PR #50, opened 2017-01-06
Store seconds as Double type in VideoTime and increase progress bar refresh rate.

### LINKED ISSUE #49: use sub-second progress bar step size
The current step size of the progress bar is one second. However, it may not be good enough in the case of short videos with a total duration of only a few seconds. In that case, the indicator jumps too much each time and it is not possible to slide the progress bar to a non-integer time point.

I am working on a fix and will send a pull request after some more testing.

Andeling:
Fixes #49.

### DIFF
--- a/iina/AppData.swift
+++ b/iina/AppData.swift
@@ -11,7 +11,7 @@ import Cocoa
 struct AppData {
 
   /** time interval to sync play pos */
-  static let getTimeInterval: Double = 0.5
+  static let getTimeInterval: Double = 0.1
 
   /** speed values when clicking left / right arrow button */
   static let availableSpeedValues: [Double] = [-32, -16, -8, -4, -2, 0, 2, 4, 8, 16, 32]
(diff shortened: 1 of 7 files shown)


lhc70000:
Switched to `develop` branch for developing. Please change the target branch.

iina/AppData.swift
```diff
@@ -11,7 +11,7 @@ import Cocoa
 struct AppData {
   
   /** time interval to sync play pos */
-  static let getTimeInterval: Double = 0.5
+  static let getTimeInterval: Double = 0.1
```
lhc70000: A shorter time interval may consume more energy.

Perhaps it's better to detect the video duration first and set the interval dynamically. What do you think?

Andeling: I assumed getting a number from mpv and asking the slider to refresh would be pretty lightweight. I performed a profiling to validate that. And it showed that `MainWindowController.updatePlayTime()` weighted 0.1% in terms of CPU time, and `PlayCore.syncUITime()` weighted 0.0%. So I guess the performance has not been affected.

The commits have been rebased to the tip of develop branch to resolve the conflicts caused by the "remove trailing spaces" commit.

PR #59, opened 2017-01-06
New Feature

droid-Q:
User can drag and drop the subtitle file to the IINA.


lhc70000:
Switched to `develop` branch for developing. Please change the target branch.

PR #64, opened 2017-01-06
Changed window dragging to use performDrag on 10.11+

DrabWeb:


lhc70000:
Thanks, but I just wonder if they have same behavior or performance, why need switching to `performDrag` with an extra version check?

DrabWeb:
Because you can feel the difference between the two, using the delta to modify position can feel laggy and off, but `performDrag` feels the same as dragging by titlebar

lhc70000:
Tested and it does have better performance. 👍

PR #77, opened 2017-01-07
New Feature And Fix bugs

droid-Q:
1.User can drag and drop the subtitle file to the IINA.
2.Fix user-defined keybindings config file does not take effect.


iina/MPVController.swift
```diff
@@ -244,6 +250,20 @@ class MPVController: NSObject {
     // get version
     mpvVersion = getString(MPVProperty.mpvVersion)
   }
+  private func getFilePath(Configs userConfigs: [String: Any]!, forConfig conf: String, showAlert: Bool = true) -> String? {
```
lhc70000: Please try to be DRY (Don't Repeat Yourself). 

We can move `getFilePath()` from `PrefKeyBindingViewController` to `Utility` as a static function.

(Also there's a typo: getFilePath(*for*Configs ...))

iina/MPVController.swift
```diff
@@ -218,10 +218,16 @@ class MPVController: NSObject {
...
-    // Receive log messages at warn level.
+    //load keybinding
+    let userConfigs = UserDefaults.standard.dictionary(forKey: PK.inputConfigs)
+    var inputConfPath =  PrefKeyBindingViewController.defaultConfigs["MPV Default"]
+    if let confFromUd = UserDefaults.standard.string(forKey: PK.currentInputConfigName) {
+      let currentConfigFilePath = getFilePath(Configs: userConfigs, forConfig: confFromUd, showAlert: false)
+      if confFromUd != "MPV Default"  &&  currentConfigFilePath != nil {
+        inputConfPath = currentConfigFilePath!
+      }
+    }
```
lhc70000: Why check `confFromUd != "MPV Default"` here? I think the following logic is enough:

```swift
if let confFromUd = UserDefaults.standard.string(forKey: PK.currentInputConfigName) {
  if let currentConfigFilePath = getFilePath(Configs: userConfigs, forConfig: confFromUd, showAlert: false) {
    inputConfPath = currentConfigFilePath
  }
}
```

PR #78, opened 2017-01-07
Improvements

alejx:
1. Handle error code when adding unsupported subtitles and audios.
2. Enable dragging subtitles in VideoView.
3. Change the original behavior in PlaylistDragDestView when dragging. It will not add non-video files into playlist now.


iina/VideoView.swift
```diff
@@ -400,5 +405,16 @@ class VideoView: NSOpenGLView {
...
   }
-
+  
+  override func draggingEntered(_ sender: NSDraggingInfo) -> NSDragOperation {
+    return .copy
+  }
+  
+  override func performDragOperation(_ sender: NSDraggingInfo) -> Bool {
+    let board = sender.draggingPasteboard()
+    let fileURL: NSURL? = NSURL(from: board)
```
lhc70000: This solution only supports single file.

Please refer to `PlaylistDragDestView.swift`.

iina/VideoView.swift
```diff
@@ -400,5 +405,16 @@ class VideoView: NSOpenGLView {
...
+    playerCore.loadExternalSubFile(fileURL as! URL)
```
lhc70000: Maybe need to check whether it's a subtitle file, if not, load the dragged file (as video).

iina/PlayerCore.swift
```diff
@@ -278,14 +278,28 @@ class PlayerCore: NSObject {
...
+  func failedLoadExternalAudioFile(_ returnCode: Int32) {
+    if returnCode >= 0 {
+      return
+    }
+    Utility.showAlert(message: "Unsupported external audio file.")
+  }
+  
   func loadExternalAudioFile(_ url: URL) {
-    mpvController.command(.audioAdd, args: [url.path])
+    mpvController.command(.audioAdd, args: [url.path], checkError: false, returnValueCallback: failedLoadExternalAudioFile)
```
lhc70000: I prefer using a block here. What do you think?

iina/PlayerCore.swift
```diff
@@ -278,14 +278,28 @@ class PlayerCore: NSObject {
...
+  func failedLoadExternalSubFile(_ returnCode: Int32) {
+    if returnCode >= 0 {
+      return
+    }
+    Utility.showAlert(message: "Unsupported external subtitle.")
+  }
+  
   func loadExternalSubFile(_ url: URL) {
-    mpvController.command(.subAdd, args: [url.path])
+    mpvController.command(.subAdd, args: [url.path], checkError: false, returnValueCallback: failedLoadExternalSubFile)
```
lhc70000: same as above

iina/PlayerCore.swift
alejx: Sure. Revise later.

iina/VideoView.swift
alejx: Do you have a list of the support format of subtitles, audio tracks and videos? I think it helps to handle.

PR #87, opened 2017-01-07
resolve path for userConfigDir in the advanced settings

xu-cheng:
Previously, path like `~/.config/mpv` won't be recognized in the
advanced settings. This fixes such issue.


lhc70000:
Nice job!

PR #89, opened 2017-01-07
add .editorconfig

xu-cheng:
This can help to enforce consistent format among contributors.

For more document http://editorconfig.org
Related to https://github.com/lhc70000/iina/pull/88


xu-cheng:
BTW, you may also want to use tools, such as swiftformat. Because consistent code format is kinda important among collaboration on the open source project.

PR #90, opened 2017-01-07
Change control bar to be more consistent with QuickTime Player

DrabWeb:
screen shot 2017-01-07 at 11 46 41 am

* Changed all the relevant assets for the control bar to be pdf and template
* Removed the settings icon in favour of the built-in `NSActionTemplate`
* Adjusted fade time for UI hiding/showing to `0.25`
* Made labels in the control bar 50% opaque to be vibrant
* Added constraints so it's able to be resized if needed
* Adjusted corner radius on the control bar to match quicktime
* Added constraints on the window titlebar so the title is actually centered, but can also fill the whole bar if needed

Still unsure about how to deal with 
screen shot 2017-01-07 at 11 49 13 am
as it just doesn't fit anymore, putting it on the bottom feels off and making the control bar taller doesn't look very good

`speed-dark` and `speedl-dark` are left as is because of their usage in preferences

Still more to be done(volume icons, time bar position indicator adjustments, control bar position constraints fixing) but wanted to start the PR to get some thoughts and opinions


lhc70000:
Thanks for your effort.

For changing UI layout, I think it's better to open an issue first and see if most users are OK with it.

By the way, next time when starting a new PR, please raise an issue before so we can discuss in advance, and avoid working on the same feature.

DrabWeb:
Alright, so should I keep this open and open an issue on the changes? Or just close and re-open when it's approved

lhc70000:
Let's keep it open. Thanks.

DrabWeb:
screen shot 2017-01-09 at 2 08 27 pm

DrabWeb:
screen shot 2017-01-09 at 2 43 53 pm

lhc70000:
Don't you think settings and playlist button are maybe too small?

Also for the floating indicator label,  maybe we don't need too much padding on the bottom of OSC to match QuickTime's design exactly, and actually I think QuickTime's OSC is not the best design.

DrabWeb:
>Don't you think settings and playlist button are maybe too small?

I think they look fine and fit in with the system, though maybe I should resize the actual button so clicking it is easier(the icon would still be the same size)

>maybe we don't need too much padding on the bottom of OSC

Yeah I could try shaving off a couple pixels for the inner padding and see how it looks

DrabWeb:
screen shot 2017-01-11 at 10 18 12 pm

Not 100% sure about this, also going to try removing some padding

DrabWeb:
screen shot 2017-01-11 at 11 06 24 pm

DrabWeb:
screen shot 2017-01-11 at 11 18 40 pm

Here's the adjusted design that fits everything, playlist/settings buttons are also 24px but the icon is only 14px

LeoNatan:
I like this design very much! I think the volume icon should also be smaller to fit the 14pt?

DrabWeb:
screen shot 2017-01-12 at 4 05 48 pm

@LeoNatan Like this?

DrabWeb:
Also noticed that the volume slider is 1px too tall to be centered relative to the mute button, not sure what should be done about that

LeoNatan:
Looks good.

LeoNatan:
QuickTime "solves" this issue by not having a volume slider at all. 😄

DrabWeb:
screen shot 2017-01-12 at 4 13 17 pm

It does have one though

LeoNatan:
Ohhh I played a screen recording (the only MOV I could find in a short notice), and since it didn't have an audio track, the slider was not displayed. Nice touch by Apple.

DrabWeb:
0e0bc37a-d8e2-11e6-8029-5355502e6d0d

Seems they just don't center it

lhc70000:
@DrabWeb 

**Thanks a lot for your effort and keeping it updated for such a long time.** I think we can get it merged before 0.0.3.

For the position, l prefer moving the slider down by 1px (not sure, may have a try tomorrow).

@LeoNatan

> since it didn't have an audio track, the slider was not displayed

Do you think this design should be adopted by IINA? 😏

DrabWeb:
screen shot 2017-01-12 at 4 23 11 pm

Here's what it would look like with the volume slider down 1px

lhc70000:
@DrabWeb Yeah I think this one looks way better (personally).

LeoNatan:
I like it. If I play something with no audio track, I don't get confused why changing the volume has no effect. Apple knows their UX.

Just a warning about the "px" talk. Remember, we are living in a retina world. So by "1px" do you mean 1pt (2px actually), or 1px, as in 0.5pt?

lhc70000:
@LeoNatan 

Yes you are right, I mean 1pt. Thanks for correcting.

Opened an issue #152 for this.

LeoNatan:
I love this player. In 0.0.2 and already my favorite! 😁

DrabWeb:
screen shot 2017-01-12 at 4 56 45 pm
screen shot 2017-01-12 at 4 56 31 pm

Just another small thing that I didn't think needed it's own PR

LeoNatan:
Can you also see why titlebar buttons disappear when entering split screen? (#149)

DrabWeb:
screen shot 2017-01-12 at 5 00 56 pm

They don't for me

LeoNatan:
Did you follow the steps in the issue? They only disappear when doing it in a specific order.

DrabWeb:
Ah, sorry, didn't notice you referenced that. I recreated it but I don't know if it should be fixed in this PR or not. I'm fine with fixing it but need to see if @lhc70000 wants it done here

LeoNatan:
Only if one line fix. If not, better to fix on its own.

DrabWeb:
@LeoNatan Mind trying this out? Works for me now

LeoNatan:
Yes, I will try in a few minutes.

LeoNatan:
When building and running your branch, I get "Fatal error" -7 "Error setting option".

DrabWeb:
I believe that's #119, don't think it's from my changes

DrabWeb:
Does this happen every time? I've gotten this error multiple times myself but never fatal

LeoNatan:
It is your branch. I think it's because it's not synced with master. Merge `master` in it, and it should work.

LeoNatan:
It looks good, and seems the buttons are fixed. 👍

Miigon:
I love this small setting button design. 
Setting button is unlike play button or forward button.Users only click setting button when they need to change some settings and this is not too frequent.So if it's too big,it may divert attention.

saagarjha:
@DrabWeb Would you mind fixing the conflicts? I'd really like to see this merged.

DrabWeb:
Oh, didn't even realize there were any. Will get right on that

saagarjha:
@lhc70000 Could we get this merged?

iina/ControlBarView.swift
```diff
@@ -15,7 +15,8 @@ class ControlBarView: NSVisualEffectView {
   var isDragging: Bool = false
 
   override func awakeFromNib() {
-    self.layer?.cornerRadius = 6
+    self.layer?.cornerRadius = 4
```
lhc70000: I personally think 6pt is better, and actually IINA prefers larger corner radius not only for OSC during early development.

lhc70000 (commented):
Great job.

Now volume and playlist icon seems not suitable at small size, I'll revamp them after merged.

iina/MainWindowController.swift
```diff
@@ -927,27 +943,9 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
...
-        // light image but with light theme: add "-dark"
-        if !isDarkTheme {
-          btn?.image = NSImage(named: currImageName + "-dark")
-          if let currAltImageName = btn?.alternateImage?.name() {
-            btn?.alternateImage = NSImage(named: currAltImageName + "-dark")
-          }
-        }
-      }
+    
+    if(isInFullScreen) {
```
alejx: I think it should not be parentheses right?

iina/ControlBarView.swift
```diff
@@ -15,7 +15,8 @@ class ControlBarView: NSVisualEffectView {
...
+    self.translatesAutoresizingMaskIntoConstraints = true;
```
saagarjha: XIBs don't make this property available directly; this would require either using `@IBInspectable` or key path runtime attributes.

iina/MainWindowController.swift
saagarjha: Yes, the parentheses are superfluous in this case.

iina/ControlBarView.swift
DrabWeb: Personally I prefer 4pt, or maybe 5, but I'll just do as requested

PR #91, opened 2017-01-07
rewrite change_lib_dependencies.rb

xu-cheng:
* It's fully automated. No manual command is required.
* No need to run with sudo.

### DIFF
--- a/other/change_lib_dependencies.rb
+++ b/other/change_lib_dependencies.rb
@@ -1,56 +1,101 @@
 #!/usr/bin/env ruby
 
-require 'open3'
+require "fileutils"
+require "open3"
+require "shellwords"
 
-LIB_NAME = "libmpv.1.dylib"
+include FileUtils::Verbose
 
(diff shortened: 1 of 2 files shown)


other/change_lib_dependencies.rb
```diff
@@ -1,56 +1,101 @@
...
-  else
-    puts 'Error!'
-    puts stderr
   end
-
 end
 
-puts "Totol #{fix_count}"
\ No newline at end of file
+puts "Totol #{fix_count}"
```
lhc70000: Oops, it's my typo (`Total`)

lhc70000:
By the way, next time when starting a new PR (especially on player related features), please raise an issue before so we can discuss in advance, and avoid working on the same feature.

other/change_lib_dependencies.rb
xu-cheng: Fixed

xu-cheng:
> next time when starting a new PR (especially on player related features), please raise an issue before so we can discuss in advance, and avoid working on the same feature.

Noted.

Issue #62, opened 2017-01-06
Behavior after finishing playing the entire video


csujedihy:
After finish playing the entire video, I cannot re-play it by clicking the play button and the window size remains the same size. TouchBar cannot control the video at this moment as well.

Andeling:
It seems that mpv closed the video after it is finished... Maybe `keep-open=yes` is necessary to pause the video at the last frame. My concern is how to replay from the beginning when space bar is pressed at the end (`eof-reached`).

lhc70000:
This will be fixed in #112 using @Andeling 's suggestion. (Wow, didn't aware that you mentioned `eof-reached` before......)

PR #112, opened 2017-01-08
Correctly handle EOF event on last media in playlist

HarukaMa:
...with a new default-on option to control the behavior

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -96,14 +96,14 @@ class PlayerCore: NSObject {
       if setPause {
         setSpeed(0)
       } else {
-        if info.videoPosition == info.videoDuration {
+        if mpvController.getFlag(MPVProperty.eofReached) {
           seek(absoluteSecond: 0)
         }
       }
       mpvController.setFlag(MPVOption.PlaybackControl.pause, setPause)
     } else {
       if (info.isPaused) {
-        if info.videoPosition == info.videoDuration {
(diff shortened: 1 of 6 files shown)


iina/PlayerCore.swift
```diff
@@ -93,12 +93,19 @@ class PlayerCore: NSObject {
   /** Pause / resume. Reset speed to 0 when pause. */
   func togglePause(_ set: Bool?) {
     if let setPause = set {
-      mpvController.setFlag(MPVOption.PlaybackControl.pause, setPause)
       if setPause {
         setSpeed(0)
+      } else {
+        if mpvController.getFlag(MPVProperty.eofReached) {
```
HarukaMa: Many thanks to @lhc70000 for the discovery of this flag

Andeling:
I have tried something similar before leaving the comments in #62, but one tricky thing I noticed was  I could not replay the video by pressing the spacebar at the end of the video. It seemed that the keybinding was processed by `mpv`, but I didn't dig into the code to figure out what really happened.

lhc70000:
@Andeling you are right, currently IINA [passes key events directly to mpv](https://github.com/lhc70000/iina/blob/master/iina/MainWindowController.swift#L299) to support mpv's `input.conf`.

However I think it's not a big problem since even in mpv, if you press space when finished playing with `--keep-open`, it will not replay the video either.

I'm almost sure that we are going to interpret key commands before passing them to mpv in the future, like to detect whether a property change is triggered by user. So maybe we can work on this problem at that time.

PR #111, opened 2017-01-08
Grammatical fixes, remap speed, and Force Touch for fast-forward/rewinding

saagarjha:
There's quite a bit of unrelated things in this pull request, so I'll go through it one by one:

1. Indentation enforcement to 2 spaces via project file: I turned this on to assist those who aren't accustomed to 2 space indents (like me!)
2. Rewording some settings, README.md, and CONTRIBUTING.md to sound better in my (very often flawed) opinion and to better match macOS's wording
3. Force Touch support for the arrow keys, made to match QuickTime Player as closely as possible. One exception is that clicking multiple times goes from
```
1x → 2x → … → 32x → 1x → 2x → …
```
instead of QuickTime Player's
```
1x → 2x → … → 32x → 2x → …
```
This was to be consistent with previous versions of IINA.
4. I have remapped the available playback speeds from (-∞, -1] ∪ [1, ∞) to (0, ∞). I've put a lot of thought into this, and here's a short list of reasons why I switched:

* Consistency with almost every other video player (e.g. VLC, MPV, YouTube), which use negative speed values for reverse playback and values in the range (0, 1) to indicate "slowed down" playback. This improves user experience by reducing the learning curve associated.
* Leaving the door open for reverse playback, if it is added to MPV (see MPV issue [#4000](https://github.com/mpv-player/mpv/issues/4000)) or even for us to add it downstream to IINA
* Using the real numbers fully, instead of leaving a "dead zone" in (-1, 1) of invalid values

5. In accordance to # 4, I've rewritten the playback speed slider to go from 0.25x to 16x, as well as added code to keep the slider and text field next to it in sync.

PR #116, opened 2017-01-09
Update MainWindowController.xib to support Force Touch

saagarjha:
This should have been part of #111.


saagarjha:
I've pushed a slightly better fix that should resolve the issue.

PR #136, opened 2017-01-11
Fix the href link 404 bug on the top-left logo. 

goodbest:
This should fix issue https://github.com/lhc70000/iina/issues/135


goodbest:
I also suggest re-format the HTML codes into blocks rather than a single line.

goodbest:
@lhc70000 Fine. It will be great if you decide to upload the website source files.

lhc70000:
感谢理解 m(_ _)m

PR #139, opened 2017-01-11
Update MPVController.swift

alejx:
There is a very small chance to get -7 even though it has been `set` twice.


lhc70000:
It may be too risky to do so. Let's have a complete test first.

PR #156, opened 2017-01-13
allow users to custom log message level

xu-cheng:
Also fix a confusing code comment.

### DIFF
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -198,6 +198,9 @@ class MPVController: NSObject {
       return Preference.RTSPTransportation(rawValue: v)!.string
     }
 
+    // Receive log messages at warn level.
+    chkErr(mpv_request_log_messages(mpv, "warn"))
+
     // Set user defined conf dir.
     if ud.bool(forKey: PK.useUserDefinedConfDir) {
       if var userConfDir = ud.string(forKey: PK.userDefinedConfDir) {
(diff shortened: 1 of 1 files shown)


xu-cheng:
There seems to be a problem, as my msg-level setting does not take effect. So probably not merge this PR for the moment.

lhc70000:
This option only controls what mpv output internally (event type `MPV_EVENT_LOG_MESSAGE`).

It was handled by `handleEvent` and output to stdout. So basically it's only for debugging use by our developers. I usually set it to `v` when debugging.

For the log file output for user, I use `--log-file`, which according to documentation,

> Opens the given path for writing, and print log messages to it. Existing files will be truncated. The log level always corresponds to -v, regardless of terminal verbosity levels.

is always verbose.

xu-cheng:
I just notice this. Also I find below document for `mpv_request_log_messages`:
```
 An exception is the value "terminal-default", which uses the
  log level as set by the "--msg-level" option. This works
even if the terminal is disabled. (Since API version 1.19.)
```

I'm testing if it is possible for user to custom the log level using the above options.

xu-cheng:
Oh, I understood it now. 

@lhc70000 Do you think it is a good idea to build IINA's own log file output function instead of using `--log-file` from mpv? Therefore, user can control the log level.

lhc70000:
Let's put it on the back burner for now. `--log-file` is also a common usage in mpv community. If there's no user requesting this feature, it should be fine.

(p.s. May be we can have an offline meeting as I noticed you are in BU :P)

xu-cheng:
> Let's put it on the back burner for now. --log-file is also a common usage in mpv community. If there's no user requesting this feature, it should be fine.

OK

> May be we can have an offline meeting as I noticed you are in BU :P

I will be quite busy in the coming weeks, and leave Hong Kong during the Chinese new year. So let's talk about this maybe in next month.

Issue #155, opened 2017-01-13
control+shift+Power  error


SHIELD-SKY:
There is an error when I use control+shift+Power。

> Fatal error: MPV API error: "error running command", Return value: -12. 
The application will exit now

lhc70000:
Close due to fixed in [0.0.3](https://github.com/lhc70000/iina/releases/tag/v0.0.3).

PR #157, opened 2017-01-13
Handle power key correctly

HarukaMa:
Prevent crash on certain key combinations such as control+shift+power.


HarukaMa:
Should "fix" #155

lhc70000:
Are you sure this key code is consistent on all devices we support?

HarukaMa:
@lhc70000 ...I'm not even sure the power button should be 127, but currently the log says so...

HarukaMa:
fd10402b-2379-405b-a589-2c0c6d1d03a2

0x7F should be okay I think.

PR #159, opened 2017-01-13
only exit application if mainWindow loaded

### LINKED ISSUE #158: malfunctioned advanced option will prevent app starting up
If you have a malfunctioned option defined in "Additional MPV options", the app will report an alert message following with exiting immediately.

This unfortunately causes the app unusable, as there is no way you can remove the malfunctioned option without opening app. The only choice is to manually delete the preference file in the disk.

xu-cheng:
Fixes #158

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -65,7 +65,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     }
     // show alpha in color panels
     NSColorPanel.shared().showsAlpha = true
-    
+
     // check update
     UpdateChecker.checkUpdate(alertIfOfflineOrNoUpdate: false)
   }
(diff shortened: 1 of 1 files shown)


lhc70000:
This also break some other behavior, like IINA will not quit after closing "open file" window or preference window.

Perhaps we can add variable like `someWindowOpened` and set it to true in `openFile` and PrefWindowController.showWindow`.

xu-cheng:
This is probably a matter of design. For example, for me, I don't think we should quit program because I open and close preference window before opening any file. (Noted that, if I close preference window after opening any file, it will quit the program).

I suggest to merge this as it is(because it fixes a crucial bug described in #158), and asking other users whether they want to quit the app even if they don't open any file.

lhc70000:
>  I don't think we should quit program because I open and close preference window before opening any file.
QuickTime does this though.

> I suggest to merge this as it is(because it fixes a crucial bug described in #158)
Totally agree.

Issue #128, opened 2017-01-10
Save playing position


JohnSinclair111:
After watching a video and reopened it later, it will starts at the beginning. A preference to watch videos from the last position (like MPlayerX) would be nice.

lhc70000:
Duplicate with #10.

We'll support it soon.

lhc70000:
Close due to duplication.

PR #160, opened 2017-01-13
support save-position-on-quit

### LINKED ISSUE #10: Support watch later | 退出后继续进度
mpv配置文件里面设置了选项
`save-position-on-quit=yes`

使用mpv命令行本身，继续播放这个功能是可用的。

然后使用IINA默认无法实现这个功能。

尝试使用高级功能里面的额外mpv选项，手动添加参数，没有效果

指定使用配置目录，也没有效果。

xu-cheng:
Fixes #10, #128

Based on the official example from https://github.com/mpv-player/mpv-examples/blob/master/libmpv/cocoa-openglcb/cocoa-openglcb.m

### DIFF
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -398,15 +398,13 @@ class MPVController: NSObject {
 
     switch eventId {
     case MPV_EVENT_SHUTDOWN:
-      if playerCore.isMpvTerminated {
-        // quit from IINA, e.g. Cmd+Q
-        mpv_detach_destroy(mpv)
-        mpv = nil
-      } else {
-        // quit by mpv, e.g. press 'q'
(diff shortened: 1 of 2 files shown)


lhc70000:
Works like a charm 👍

silic0ns0ul:
I just read the mpv manual and this "save-position-on-quit" is an option that is disabled by default for mpv, yet behavior is the opposite on iina which is a bit confusing. Would be nice to have this as a toggle option in Preferences, so user can set behavior as desired.

lhc70000:
@silic0ns0ul `Preference > General > Behavior > Resume last playback position` should be what you want. It will enable/disable `save-position-on-quit` as well as `resume-playback`.

PR #170, opened 2017-01-14
Changed timer behavior, fix #141

### LINKED ISSUE #141: On screen controller does not disappear while fullscreen if mouse does not move.
Reproduce: Use Ctrl+Cmd+F to get into fullscreen and keep the mouse still, then the on screen controller will not disappear.

alejx:
Also fix a typo


lhc70000:
Maybe this is not the best solution... since it affects all mouse enter events.

lhc70000:
See [ab1d80a](https://github.com/lhc70000/iina/commit/ab1d80ace47cefbd0fc57378e03ef8cbf1e8dfc9)

PR #173, opened 2017-01-14
Sync pause in meun with info.isPaused. Fix #70

### LINKED ISSUE #70: Update Playback->Play/Pause menu item according to playback status.
The Play/Pause item in Playback menu does not update when the playback status was changed elsewhere, such as through pressing play button or keybinding. It would be nice if it could be updated based on the actual playback status.

uiryuu:
Also delete an unused variable.

### DIFF
--- a/iina/MainWindowMenuActions.swift
+++ b/iina/MainWindowMenuActions.swift
@@ -15,10 +15,8 @@ extension MainWindowController {
   @IBAction func menuTogglePause(_ sender: NSMenuItem) {
     if sender.title == "Play" {
       playerCore.togglePause(false)
-      sender.title = "Pause"
     } else {
       playerCore.togglePause(true)
-      sender.title = "Play"
     }
   }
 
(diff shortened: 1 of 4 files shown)


xu-cheng:
One long-term suggestion: for many properties like pause, it would be better to read info directly from mpv(though `mpv_observe/get/set_property`) instead of maintaining a separate copy in iina. The point is to reduce the extra work for sync information and also to better support mpv lua scripts for advanced users.

lhc70000:
There's no need to post notifications only for menu. I'll fix this by menu delegate.

@xu-cheng pause is now synced by `observe_property`. I just don't want all classes calling mpv API directly (especially most of them requires arguments in C style) and frequently. But it's a nice suggestion, I'm thing about doing similar stuff in `PlayerCore` in a future refactor.

lhc70000:
@alejx See [2cb3653](https://github.com/lhc70000/iina/commit/2cb36533b579c48e88301eab0ca541d88b0851c6)

PR #185, opened 2017-01-15
README typo fix

nyx:


alejx:
Thank you for your investigation, but we do not accept non-functional pull requests.

nyx:
The pull request was a single letter typo fix, not sure what "non-functional" is intended to mean here.

PR #200, opened 2017-01-18
Add auto layout

lhc70000:
Please have a check on:

- Preference panel:
  - Has a minimum width of 480pt
  - Auto resizes when content is wider than 480pt
  - Window resizes smoothly when switching tabs
  - All settings still work well
- Quick settings panel:
  - All settings still work well


alejx:
Looks good

PR #206, opened 2017-01-19
Add checking when parsing keybinding files

inflation:
If iina get the wrong number of split, it will crash without a fallback.

This PR will add an alert and fallback to IINA Default. Maybe later we can store a cache to the last working config and fallback to that.

### DIFF
--- a/iina/PrefKeyBindingViewController.swift
+++ b/iina/PrefKeyBindingViewController.swift
@@ -301,6 +301,17 @@ class PrefKeyBindingViewController: NSViewController, MASPreferencesViewControll
       }
       // split
       let splitted = line.characters.split(separator: " ", maxSplits: 1)
+      if splitted.count < 2 {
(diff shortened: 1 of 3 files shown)


lhc70000:
Maybe just skip the line with problem will be better?

inflation:
I think we want users to know their config file has problems. Maybe we can provide the line number?

iina/PrefKeyBindingViewController.swift
```diff
@@ -301,6 +301,17 @@ class PrefKeyBindingViewController: NSViewController, MASPreferencesViewControll
       }
       // split
       let splitted = line.characters.split(separator: " ", maxSplits: 1)
+      if splitted.count < 2 {
+        let alert = String(format: NSLocalizedString("alert.keybinding_config_error", comment: "Error setting keybinding configuration"), currentConfName)
+        Utility.showAlert(message: alert)
```
lhc70000: Seems that this `showAlert` will also crash...

I'll add a workaround and merge.

Issue #66, opened 2017-01-06
Resizing window with pinch gesture


oskarkrawczyk:
Movist does this and it's a really awesome feature, no need to look for the drag handle at the corner, just hover over the window and pinch in or out for resizing.

lhc70000:
Seems duplicate with #23.

Some users also want entering fullscreen by pinch gesture. Maybe need to let them choose the behavior.

alejx:
Close due to duplication.

PR #210, opened 2017-01-19
Two fingers pinching resize v2

### LINKED ISSUE #198: Adjust window size via two figures' pinching while retaining the aspect ratio
It's quite like a page zoom in/out feature. I think it's more convenient to adjust the window size via figures pinching other than A precisely corner dragging etc.

neesonqk:
Enhanced version.

Fixes #198, #66, #23.


neesonqk:
I wonder how shall we manipulate the control bar during the resizing, shall we also resize the control bar with the same ratio as the main window does? OR shall we keep the size but maintain the same distance to the bottom (or any other boundaries) of the main window?

I saw QuickTime uses the second scenario.

saagarjha:
I'd think that keeping the control bar the same size would be best, since that's the way it's done currently.

neesonqk:
but if let's say I resized to a small width/height then the control bar would become too small to click, compare to other mainstream players, most of them keep the size but stick to main window's bottom bound.

lhc70000:
Could you rebase your branch and test if it works on latest `develop`?

neesonqk:
Yes will do, I've tested yesterday against to the newer rendering branch, it didn't generate any conflicts, suppose should be working with the latest develop branch.

I test later when I reach home.

Neeson.Z

neesonqk:
I've tested the latest develop branch, it still works.

neesonqk:
@saagarjha @lhc70000 Tested several times, but I still recommend the control bar should retain the distance to the bottom during the resizing, else it looks quite lousy in a sense.

Feel free to taste it out if you have the same opinions.

lhc70000:
I'll merge it into another branch and continue your work.

lhc70000:
Some problems I found, for your reference:

- You shouldn't define screen rect as a constant, since user may switch to other screens during playback.
- You shouldn't just check whether width exceeding screen frame, since the video can be in any aspect ratio.
- Window aspect ratio is not correct after resized.
- Generally `;` should not be used at line end in Swift.

PR #229, opened 2017-01-21
Add option to change scroll actions.

### LINKED ISSUE #150: Allow disabling action for horizontal and / or vertical scroll
Right now, it is only possible to change the seek type for horizontal scroll. However, I often find it preferable to disable actions for horizontal / vertical scrolls in order to prevent accidental scrolls on the touchpad.

LeoNatan:
Defaults remain unchanged - horizontal scroll seeks; vertical scroll changes volume.
Closes #150

Screenshot:

screen shot 2017-01-21 at 21 13 24


lhc70000:
- I don't think it's a good idea to have a separate seek mode option, since it only works for seek by scrolling (i.e. seek by adjusting progress bar or by "Playback-Jump To" etc. are not controlled by this option).
- And it's the same for "amount to seek". It only controls horizontal scrolling.
- If we aren't going to add more actions for scroll (it's not in plan), it's better to use checkboxes rather than a popup button.
- Newly added layout constraints are inconsistent with existing ones (all elements should be 66pt leading to container, and have equal width with the first element).

Sorry that I went to sleep last night before you asked me in the issue. Please make sure we have a short discussion before working on some bigger changes.

Would you mind me adding this option on my own?

LeoNatan:
I don't mind.

BTW, any reason why you are not using a stack view for the settings?
https://developer.apple.com/reference/appkit/nsstackview
It would be a lot easier to add new settings with a stack view.

PR #248, opened 2017-01-23
fix translation bug and some minor changes

darkbrow:
fix translation bug and some minor changes


shilch:
Nice work! This is the 4th language now. Maybe we should translate gh-pages too?

lhc70000:
@shilch Let's do it when we have a stable version.

Issue #252, opened 2017-01-24
快进放慢问题 (normal speed skipped)


SaneBow:
快进放慢之间没有正常速度，不科学。
Between speed up (x2) and low down (x0.5) there should be a normal speed (x1). Currently we can only pause then resume to get to normal speed.

SaneBow:
This is how QuickTime behaves. Not a bug.

PR #165, opened 2017-01-14
Picture in Picture support, AutoLayout improvements

saagarjha:
This pull request adds support for Picture in Picture. I would advise against merging this as-is, since it still needs a bit of work. I just wanted to get your feedback on it.

A couple of things that need addressing:

* Crash when switching between Picture in Picture and "normal" too quickly using the keyboard shortcut (Command+Shift+P)
* Some way of alerting the user on the main window that the video is playing as Picture in Picture
* You tell me

PR #255, opened 2017-01-24
fix bug: 1x speed skipped

SaneBow:
A fix for issue #252

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -1146,14 +1146,15 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
   @IBAction func leftButtonAction(_ sender: NSButton) {
     if arrowBtnFunction == .speed {
       let speeds = AppData.availableSpeedValues.count
-      // If fast forwarding change speed to 1x
-      if speedValueIndex > speeds / 2 {
-        speedValueIndex = speeds / 2
-      }
       
       if sender.intValue == 0 { // Released
         if maxPressure == 1 { // Single click ended
-          speedValueIndex = oldIndex - 1
+          // If fast forwarding change speed to 1x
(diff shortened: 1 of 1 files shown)


saagarjha:
@SaneBow Would you mind explaining in detail what exactly this pull request does? Also, would you mind holding off on this until #165 is merged, since it has a couple changes to the rewind/fast forward mechanism.

SaneBow:
This fix the issue described in #252. In the current code
```swift
   @IBAction func leftButtonAction(_ sender: NSButton) {
      if arrowBtnFunction == .speed {
        let speeds = AppData.availableSpeedValues.count
        // If fast forwarding change speed to 1x
        if speedValueIndex > speeds / 2 {
          speedValueIndex = speeds / 2
        }
        
        if sender.intValue == 0 { // Released
          if maxPressure == 1 { // Single click ended
            speedValueIndex = oldIndex - 1
 ```

The `If fast forwarding change speed to 1x` part of code tried to reset speed to 1x when click on left button (for speed down) while it's playing in fast speed (e.g. 2x). But then the `single click` part again set speed to slower index. So duplicated action happened. This pull request is all about this bug.

Thank you for referring #165, I have checked but it didn't fix this issue.

PS: I think the function of the left button is slow down rather than rewind, maybe someone should fix the typo in comment XD

saagarjha:
This was intended behavior, as that's how QuickTime does it. The left button is intended to be for rewinding, but it's being used to slow down because MPV/IINA currently doesn't support it: https://github.com/lhc70000/iina/blob/master/iina/AppData.swift#L19

SaneBow:
I see. This is a little bit inconvenient in my opinion, but as apple does that, we may better just follow the style.
Thank you for your time. I will close this pull request.

PR #261, opened 2017-01-24
Fix the wrong ObjectID

inflation:
- ObjectID of “Download subtitles from:” is wrong in the localization
  file.

Also a friendly ping to @shilch, it seems you missed this label 😃


shilch:
Thank you! I did a PR

PR #264, opened 2017-01-25
Add function to save playlist.

### LINKED ISSUE #263: Feature Request: Save playlist
We would like to have a function to save the playlist when playing.

inflation:
In order to fix #263.

image

### DIFF
--- a/iina/MainWindowMenuActions.swift
+++ b/iina/MainWindowMenuActions.swift
@@ -340,5 +340,23 @@ extension MainWindowController {
     inspector.showWindow(self)
     inspector.updateInfo()
   }
+  
+  @IBAction func menuSavePlaylist(_ sender: NSMenuItem) {
+    let _ = Utility.quickSavePanel(title: "Save to playlist", types: ["m3u8"]) {
+      (url) in if url.isFileURL {
+        var playlist = ""
+        for item in playerCore.info.playlist {
+          playlist.append((item.filename + "\n"))
+        }
+        
+        do {
+          try playlist.write(to: url, atomically: true, encoding: String.Encoding.utf8)
+        } catch let error as NSError {
+          print("error writing to url \(url)")
+          print(error.localizedDescription)
+        }
+      }
+    }
+  }
 
 }
--- a/iina/MainWindowMenuActions.swift
+++ b/iina/MainWindowMenuActions.swift
@@ -352,8 +352,7 @@ extension MainWindowController {
         do {
           try playlist.write(to: url, atomically: true, encoding: String.Encoding.utf8)
         } catch let error as NSError {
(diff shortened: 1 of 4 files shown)


iina/MainWindowMenuActions.swift
```diff
@@ -302,5 +302,23 @@ extension MainWindowController {
     inspector.showWindow(self)
     inspector.updateInfo()
   }
+  
+  @IBAction func menuSavePlaylist(_ sender: NSMenuItem) {
+    let _ = Utility.quickSavePanel(title: "Save to playlist", types: ["m3u8"], ok: {
```
lhc70000: Please use trailing closure:
```
Utility.quickSavePanel(title: "Save to playlist", types: ["m3u8"]) {
  ...
}

File: iina/MainWindowMenuActions.swift [hunk 2]
```diff
@@ -332,5 +332,23 @@ extension MainWindowController {
...
+        var playlist = ""
+        for item in playerCore.info.playlist {
+          playlist.append((item.filename + "\n"))
+        }
+        
+        do {
+          try playlist.write(to: url, atomically: true, encoding: String.Encoding.utf8)
+        } catch let error as NSError {
+          print("error writing to url \(url)")
+          print(error.localizedDescription)
```
lhc70000: Please show an alert to user, or at least use `Utility.log`

iina/Utility.swift
```diff
@@ -79,6 +79,21 @@ class Utility {
...
+  static func quickSavePanel(title: String, types: [String], ok: (URL) -> Void) -> Bool {
+    let panel = NSSavePanel()
+    panel.title = title
+    panel.canCreateDirectories = true
+    panel.allowedFileTypes = types
+    if panel.runModal() == NSFileHandlingPanelOKButton {
+      if let url = panel.url {
+        ok(url)
+      }
+      return true
```
lhc70000: Better return `false` when `url` is `nil`?

lhc70000 (changes requested):

alejx:
A suggestion. What about create a new file format for IINA playlist instead of m3u8? Just in case of adding any future fancy functions into the playlist.

alejx:
Add a keyboard shortcut like `cmd+shift+s`?

iina/Utility.swift
inflation: For these panels, we don't check the return value anyway. Checking url's non-emptiness seems to be done all in the callbacks.

inflation:
@alejx I would more like to keep with the open standard. You don't want users to find the exported playlist cannot be opened elsewhere.

lhc70000 (approved):

PR #274, opened 2017-01-26
remove border for file path field in inspect window

phill84:
Inspect window is now made unresizable, and the border around file path text field is removed.
Also the end of file path will be truncated with ellipsis if it is too long.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+language: objective-c
+osx_image: xcode8.2
+xcode_workspace: iina.xcworkspace
+xcode_scheme: iina
+before_install:
+- brew update
+- brew outdated xctool || brew upgrade xctool
+- gem install xcpretty
+install:
+- pod install --project-directory=$TRAVIS_BUILD_DIR
+before_script:
+- xctool -workspace ${TRAVIS_XCODE_WORKSPACE} -scheme ${TRAVIS_XCODE_SCHEME} run-tests
+script:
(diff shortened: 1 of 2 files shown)


megadr01d:
@phill84 Just curious, why unresizable?

alejx:
@megadr01d Cause textfields and many other components in the inspect window cannot been resized currently.

phill84:
Exactly, nothing gets auto-resized currently when you change the window size.

saagarjha:
On it!

lhc70000:
**PLEASE READ [CONTRIBUTING.md](https://github.com/lhc70000/iina/blob/master/CONTRIBUTING.md) FIRST.**

> Please open an issue before actually starting doing something to make sure that there's not someone else working on the same feature.

Especially when it's UI related. IINA is now in rapid development, which means there will be a brunch of stuff to be added or removed. We need to have a **discussion** before you starting working to make sure there's no conflict with our plan. If I didn't make it clear in CONTRIBUTING.md, sorry for that and I'll update it on next release.

---

For inspector, it's even not working fully. According to existing feature requests and plan, we are going to redesign the layout (maybe use a tab view) to add more information (subtitles info, mpv status, etc.) and current one will possibly be completely replaced. So it makes no sense to improve current layout (and that's why I didn't add AutoLayout for inspector only).

The reason why file path field is boarded is that I want to let user know they can select and copy it. 

@saagarjha Please keep only one topic for each PR. See CONTRIBUTING.md:

> Please submit separate pull requests for different features.

phill84:
@lhc70000 thanks for the clarification. However I would say that changing of cursor is an indicator clear enough for the users.

saagarjha:
Heh heh, I have a tendency to add a bunch of things to my pull requests. I'll revert them.

PR #279, opened 2017-01-27
New rendering mechanism & update to mpv 0.23

lhc70000:
This branch mainly does two things:

**Update mpv to 0.23.0.**

 - Added some more libs like `vapoursynth` when compiling
 - Support macOS 10.10

**Change current rendering mechanism.**

Since some changes in mpv 0.22 made current (improper) render mechanism blocking UI (especially when resizing window), we need to find another way. It mainly contains:

- Use `CAOpenGLLayer` rather than `NSOpenGLView` or custom `NSView`
- Render video in background thread by some dirty (but necessary) hack
- No extra frame buffer and `CVDisplayLink` needed

It could solve a lot of existing problems (and even add some bonus):

- Optimized App launch time
- Solve high CPU consumption, especially when video paused
- Provide smooth window resizing experience
- Possible to avoid requiring OpenGL 3.2+, therefore support some old MacBook models
- Let mpv to control the synchronization with screen refresh rate

If everything works well, this will be included in v0.0.4.

**TODO**

- [x] Test whether it works when switching between screens with different backing scale
- [x] Report flip to mpv

---

**Testing**

If you want to have a test, just switch to this branch and build.

Alternatively you can download a binary in Telegram group.


alejx:
@lhc70000 Fantastic!

saagarjha:
> macOS 10.10

Regarding this, do we plan to support Yosemite? The xcodeproj file has the deployment target set to 10.11, and we're using some things that aren't available on 10.10.

lhc70000:
@saagarjha The minimum deployment target is always 10.10. The reason why it doesn't support 10.10 now is that some mechanism in 10.10 requires all dylibs have the same minimum deployment target as the app itself, i.e. a set of dylibs compiled on 10.10 will make it work.

I installed a Yosemite on virtual machine and compiled them a week ago, but they are based on mpv 0.23 (latest)...... so I must solve the rendering problem first. There are a few Yosemite users who tested some experimental builds during these days and seems that everything is fine. So I guess if this branch can get merged, we'll support Yosemite.

HarukaMa:
@saagarjha Regarding the 10.11 target setting on the project itself, that setting is being overriden by target-specific setting which is 10.10, so it can be safely ignored.

Issue #283, opened 2017-01-28
Display remaining time when clicking on duration label


metacosm:
Clicking on the duration label (rightLabel) in the control bar should cycle between remaining time and duration, similar to what Quicktime Player (and others) do.

metacosm:
This issue should be closed now, no?

lhc70000:
I'll close it after releasing v0.0.4, in 10 minutes.

lhc70000:
Fixed in [0.0.4](https://github.com/lhc70000/iina/releases/tag/v0.0.4).

PR #284, opened 2017-01-28
Fix for Issue 283

metacosm:
Fix for #283: Implemented cycling through duration and remaining time for right label in control bar.


Podfile.lock
```diff
@@ -12,4 +12,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 1660afcf8e29f4d915d2c02e6adc97cd9068a0dd
 
-COCOAPODS: 1.2.0.beta.3
+COCOAPODS: 1.2.0.rc.1
```
lhc70000: Please discard changes to this file.

iina/DurationDisplayView.swift
```diff
@@ -0,0 +1,50 @@
...
+//  DurationDisplayView.swift
+//  iina
+//
+//  Created by Christophe Laprun on 26/01/2017.
+//  Copyright © 2017 lhc. All rights reserved.
+//
+
+import Foundation
+
+class DurationDisplayView: NSTextField {
```
lhc70000: I prefer a name ended with `TextField`, like `DurationDisplayTextField`.

iina/MainWindowController.swift
```diff
@@ -1088,9 +1088,7 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
     let percantage = (pos.second / duration.second) * 100
     leftLabel.stringValue = pos.stringRepresentation
     touchBarCurrentPosLabel?.stringValue = pos.stringRepresentation
-    if withDuration {
-      rightLabel.stringValue = duration.stringRepresentation
-    }
+    rightLabel.updateText(with: duration, given: pos)
```
lhc70000: Why removed `if withDuration`?

Podfile.lock
metacosm: Sorry, didn't mean to include it, missed it.

iina/DurationDisplayView.swift
metacosm: Will do.

iina/MainWindowController.swift
metacosm: Seems like when the control bar is shown on mouse hover, withDuration is set to false. This means that, when the mode is set to remaining time, if we keep the check, the label will never be updated. The other alternative would be to set withDuration to true when the control bar is shown but I think that would amount to removing the check (which is what I did). I might have missed something but I actually don't quite understand why the label is only updated when withDuration is set to true.

lhc70000:
Resolved conflicts. Please update your local branch if necessary (with `git pull --rebase`).

iina/MainWindowController.swift
lhc70000: Oh that makes sense. Please stay at your current changes.

metacosm:
Changes have been made. Thank you!

lhc70000:
`Podfile.lock` is still modified...

Issue #215, opened 2017-01-19
Add Next/Previous buttons


chrikop:
Currently there is no possibility to skip to the next/previous file in the playlist other than to open the playlist and click on the file you want to play.
In VLC they look like this:
<img width="261" alt="vlc" src="https://cloud.githubusercontent.com/assets/9819843/22110838/7f96ace6-de5d-11e6-91aa-fbc726a0b254.png">

megadr01d:
I prefer to use hotkeys for everything but, *only as a suggestion*, to keep the controller simple as it is and using the OS X way of providing alternative commands without clutter, holding the <kbd>⌥</kbd> (Option) key could turn the rewind/forward buttons into previous/next buttons.

neesonqk:
@chrikop There actually is such a feature to config the left right buttons into previous/next media.

![image](https://cloud.githubusercontent.com/assets/7279067/22131802/aeffb930-def0-11e6-8c27-530cdc7970d5.png)


@lhc70000 What if we change the left-right buttons' icon accordingly to its configuration? which will make less confusion I suppose...

neesonqk:
Adding two more extra icons doesn't seem to be the best choice, makes the control bar redundant.

lhc70000:
@megadr01d @neesonqk good suggestion! I'll consider adding it in 0.0.4.

neesonqk:
@DrabWeb 

I wrote an implementation of this feature, however, the icons for Previous/Next Video looks a bit blurry and incorrect size, if you could help to doctor a proper version will be a great plus since I have no idea about how to produce icon at all XD.

![image](https://cloud.githubusercontent.com/assets/7279067/22491437/5dc1537c-e85e-11e6-8473-67bca2c5c80c.png)

saagarjha:
I believe @DrabWeb is using PDF vector icons. Maybe he'd be able to put them here?

DrabWeb:
The icon PDFs are all in https://github.com/lhc70000/iina/tree/master/iina/Assets.xcassets
@neesonqk I could put some icons together tomorrow if you want

neesonqk:
@DrabWeb Yes please, I realized there are actually many icons in Assets already, but the two icons I want are absent XD, like the above screenshot I took, I'd like a pair of mirror icon which indicates Next/Previous video. 

Just right like this: 
![image](https://cloud.githubusercontent.com/assets/7279067/22493985/99a4111c-e86e-11e6-82a4-6b67673d0854.png)

[...]

PR #305, opened 2017-01-31
Implementation of changing the left-right buttons' icon automatically.

neesonqk:
Implementation #215.

Icons are provided by @DrabWeb


lhc70000:
What I want most is changing the icon when pressing option key though...

And what's the difference between `▶︎▶︎|` and `▶︎|`? (If no difference I prefer the former one)

For UI changes, especially icon design, please do let me have a check in advance.

---

**Before working on a PR, please read [CONTRIBUTING.md](https://github.com/lhc70000/iina/blob/master/CONTRIBUTING.md).**
TL;DR:
- Open an issue if there isn't one
- Ask for the author's permission in the issue

neesonqk:
Agree, the first icon is more meaningful.

"Changing when pressing option key" is more an add-on rather than a replacement of this implementation I suppose, if we change the icon only while we are holding opt key, then what if a user permanently set the two buttons into Next/Perv video from Preference, which doesn't make sense to keep the same icons what?

neesonqk:
So what's the conclusion here? am I going to continue the implementation of opt key?

lhc70000:
Let me make an icon first, and then you can start a new PR (I personally think it's not a good practice to have useless binary files in git log).

Issue #133, opened 2017-01-11
The video plays at the top, not at all times


faying:
The video plays at the top, not at all times, when I pause the video, it should be cancel the top.
And sometimes the CMD+W close windows does not work

faying:
Another problem, there is no need always use dGPU to codec, otherwise the battery will lose quickly. Maybe a option is need on the settings

lhc70000:
> there is no need always use dGPU to codec
Duplicate with #18, fixed in`master` branch. Please wait for next release.

> when I pause the video, it should be cancel the top. And sometimes the CMD+W close windows does not work
Got it.

neesonqk:
@faying I cannot reproduce the incidental closure issue of pressing CMD+W, I suspect that you floated the player which made it seemed to be in focus, while it wasn't the first responder as CMD+W was hit by the time.

lhc70000:
@neesonqk It does happen occasionally and I believe it's the same reason of #35. For some users it even happens **every time**. I'll try some fix later.

PR #308, opened 2017-02-01
Cancel floating on top during video pausing.

neesonqk:
Implementation #133

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -352,7 +352,17 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
   override func keyDown(with event: NSEvent) {
     window!.makeFirstResponder(window!.contentView)
     if !isInInteractiveMode {
-      playerCore.execKeyCode(Utility.mpvKeyCode(from: event))
+      let keyCode = Utility.mpvKeyCode(from: event);
+      playerCore.execKeyCode(keyCode)
+      //intercept SPACE keycode.
+      //MPV has --ontop flag, but iina floats top itself window, so here has to intercept SPACE key.
+      if keyCode == "SPACE" {
+        if !playerCore.info.isPaused {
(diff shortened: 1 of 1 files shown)


lhc70000:
Please don't do check in `keyDown`.

For querying status please use `playerCore.info` or `MPVController#propertyChanged` or getting mpv property at runtime, since user may pause by menu actions, lua scripts, JSON IPCs, etc.

Since it's not a normal behavior, I think a better implementation should:

- Connect `--ontop` with window floating status (like what `fullscreen` does now)
- Add a user option about whether floating on top when paused
- Set `ontop` to false when paused and above option enabled

Please continue discussion in the issue if needed.

**Before working on a PR, please read [CONTRIBUTING.md](https://github.com/lhc70000/iina/blob/master/CONTRIBUTING.md).**
TL;DR:
- Open an issue if there isn't one
- Ask for the author's permission in the issue

neesonqk:
Seems I didn't violate the rules, `fullscreen` does the similar way I suppose.

There are two implementations of pausing a video, one is through clicking pause icon on the control bar, another is through SPACE key, they seem to use different mechanisms?

By clicking pause from control bar can be implemented almost exactly the same as fullscreen, but by tapping SPACE key, if we don't intercept the `SPACE` key in `keyDown` event, I cannot image a better way to temporarily toggle off/on floating top while pausing was enabled by SPACE key.

lhc70000:
> There are two implementations of pausing a video

No, there are many other ways to change a mpv property (including pause) like other key bindings, lua script, JSON IPC or set by mpv itself, and we cannot detect these changes since they are not triggered by IINA.

So we must listen to certain property change event and synchronize with window status, take full screen as example: [c2ceedf](https://github.com/lhc70000/iina/commit/c2ceedf6118d9593dd29878130e5552f8c9bd320).

neesonqk:
Right I got your purpose.

Issue #294, opened 2017-01-30
drag'n'drop URL's on IINA in Dock and window


ww7:
d'n'd works in mpv for direct URL's and for stream from page's, provided by youtube_dl mpv hook

would be great to have this IINA feature

PR #314, opened 2017-02-02
Enable dragging URL into video view.

uiryuu:
- [x] It implements issue #294 (part 1).
- [ ] It implements issue #294 (part 2).

---

**Description:**

Did not rebase to the latest commit (06d6d9b) because it will crash after opening
files.

### DIFF
--- a/iina/Regex.swift
+++ b/iina/Regex.swift
@@ -13,6 +13,7 @@ class Regex {
   static let aspect = Regex("\\A\\d+(\\.\\d+)?:\\d+(\\.\\d+)?\\Z")
   static let httpFileName = Regex("attachment; filename=(.+?)\\Z")
   static let tagVersion = Regex("\\Av(\\d+\\.\\d+\\.\\d+)\\Z")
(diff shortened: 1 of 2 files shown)


lhc70000:
Any updates / ready for merge?

alejx:
Do you think it is necessary to treat strings that users drop into as a URL?

lhc70000:
> Do you think it is necessary to treat strings that users drop into as a URL?

interesting... let's implement it. e.g.:

1. Check for protocol like `http://` or any other that supported by mpv; load url if true
2. If no protocol presented, check whether string is a web url by a RegExp; try to load if true
3. (Check whether it's a file path?)

alejx:
> 2. If no protocol presented, check whether string is a web url by a **RegExp**;

Which RegExp would you like to use? I found there are many different RegExp to judge URLs.

> 3. (Check whether it's a file path?)

I don't think check whether string is a file path is necessary. Unlike windows users, macOS users cannot get files' path very easily.

lhc70000:
`@^(https?|ftp)://[^\s/$.?#].[^\s]*$@iS` should be OK.

PR #325, opened 2017-02-03
Fix crash in macOS 10.12.4 beta(16E144f).

alejx:
- [x] It fixes bug in #06d6d9b.

---

**Description:**
Revert some lines of code in 06d6d9b to prevent crash in 10.12.4 beta.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -9,53 +9,53 @@
 import Cocoa
 
 class MainWindowController: NSWindowController, NSWindowDelegate {
-
+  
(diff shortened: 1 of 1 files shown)


iina/MainWindowController.swift
```diff
@@ -9,53 +9,53 @@
 import Cocoa
 
 class MainWindowController: NSWindowController, NSWindowDelegate {
-
+
```
lhc70000: trailing spacessss

alejx:
@lhc70000 Done.

Issue #201, opened 2017-01-18
Add option to loop video and or playlist


st0ne:
In Quicktime it is possible, also in VLC. Would be a nice feature :-)

LeoNatan:
Yes! This is very useful.

chrikop:
Also, shuffle would be quite handy

PR #361, opened 2017-02-11
Playlist context menu, loop & shuffle

lhc70000:
- [x] It implements / fixes issue #201.

---

**Description:**

- Context menu for play list:
8dc8142e-ecc6-4365-a140-8942f88a241b

- Loop and shuffle:
8eae7698-0719-49eb-a8cd-9d98d57c7e7d

- Add menu items & shortcuts for playlist and chapters


alejx:
Several suggestions.

1. I don't think indexes that were selected should remove highlight when preforming secondary click.
2. Keep highlight after reorder the list if there was highlight before it.
3. Shuffle should not change the order of the list. Generate a random playlist in background would be fine.

Issue #315, opened 2017-02-03
[Feature Request] QuickTime-like title bar behavior


tzhao11:
- Right click on filename:
<img width="410" alt="2017-02-02 18 19 14" src="https://cloud.githubusercontent.com/assets/6540380/22574394/334aefce-e974-11e6-8d2f-6e3d5e9017d9.png">

- Click on arrow:
<img width="600" alt="2017-02-02 18 18 58" src="https://cloud.githubusercontent.com/assets/6540380/22574398/3603272c-e974-11e6-933e-bf6999d34135.png">

- File icon can be dragged.

alejx:
This feature may be related to #187 .

lhc70000:
@alejx However we're using custom view for titlebar.

saagarjha:
@lhc70000 If we need to use a custom view, we'll need to add a [`documentIconButton`](https://developer.apple.com/reference/appkit/nswindowbutton/1419136-documenticonbutton) and set [`representedURL`](https://developer.apple.com/reference/appkit/nswindow/1419066-representedurl) for the titlebar.

alejx:
See #366 .

Issue #353, opened 2017-02-10
Reduce file name when the title bar is not long enough.


alejx:
**System and IINA version:**

- macOS 10.12.4 developer beta 2
- IINA Latest develop branch build 9

**Expected behavior:**

![screen shot 2017-02-09 at 6 14 23 pm](https://cloud.githubusercontent.com/assets/20237141/22808818/9900e524-eef3-11e6-9bd8-be28c1a1938b.png)
P.S. This capture is from 0.0.4 build 7.

**Actual behavior:**

![screen shot 2017-02-09 at 6 14 52 pm](https://cloud.githubusercontent.com/assets/20237141/22808825/aacee38c-eef3-11e6-9164-a5c120fb95ab.png)

**Steps to reproduce:**

Use build 9. Play a file with relative long file name and resize the window to the smallest size. The file name overlaps window controllers (sorry I don't know how to call those three points).

**How often does this happen?**

This does not happen in 0.0.4 build 7. But happens every time in latest build 9.

lhc70000:
@saagarjha Why did you change the constraints for title bar?

saagarjha:
@lhc70000 It's a remanent from a commit I removed from #165, which had support for a Document Icon Button. I've added the support back in #366. As a side note: please let me know when you revert a commit. For a small graphical bug like this, that doesn't cause a crash or make the application unusable, with a very simple fix, is probably not a good reason to revert a commit with other things in it.

lhc70000:
@saagarjha 

I noticed that you also

- removed some height constraints. It doesn't cause error, but having height constraints will make it easier to add more controls in later development.
- used aspect constraints. Is there any special reason for doing this? I mean, they are just simple buttons so it perhaps didn't simplify a lot, and having height+width constraints is more flexible, since we may not require them to be square in other layouts (in the future).

I was not sure whether these changes breaks other UI, so I reverted it first.

I missed them when checking #165, sorry for that. But seems that you didn't mention it in the pull request (or I missed too?), so please tell me before changing something like this next time.

saagarjha:
It had a small mention in "AutoLayout improvements", but I agree I should have been a bit more clear. I've used aspect constraints because they make it easier to resize objects without losing the aspect ratio (for example, if we think the control bar is too small). If the images are not square, their aspect ratio should be used instead of the 1:1 we have right now, but either way there's fewer magic numbers in the storyboard.

Issue #253, opened 2017-01-24
Adjustable play-list panel width


SaneBow:
Currently there seems to be no way to adjust panel width of play-list. Sometimes I got long file names and couldn't manage to display full file names. This is inconvenient  e.g. when I was watching TV series, since information in filenames like episodes were hidden.

Miigon:
@lhc70000 This can be fixed if we add auto-layout for playlist panel and quicksettings panel.

Miigon:
This can be fixed if we add auto-layout for QuickSettings panel and Playlist panel
(A commit in #165 has added auto-layout for `MainWindowController.xib`)

saagarjha:
@Miigon I'm not the one who added AutoLayout; you can thank @DrabWeb with that in pull #90. I just improved the constraints (but I'd be happy to add AutoLayout to the playlist panel in a commit).

Miigon:
@saagarjha awesome!

saagarjha:
@lhc70000 Check #165, which adds AutoLayout. Dragging is coming soon, but you can change the size in `SideBarViewType` from 360 to something greater and see if it lays out the views correctly.

alejx:
Fixed in develop. See #673 .

jrz:
Please reopen, it's not fixed.

PR #366, opened 2017-02-12
Add Document Icon Button, AutoLayout Quick Settings

saagarjha:
- [X] This change has been discussed with the author (at least, it's been mentioned many times)
- [x] It implements / fixes issue #315, #353.

---

**Description:**
A proper fix for both issues outlined above.  Please test it out, especially with long filenames and the fullscreen transition.

Caveats:
* This PR uses internal APIs. Essentially, `NSWindow` keeps changing the `NSTextField` for the title whenever switching between fullscreen and windowed mode. It also doesn't make the actual text field publicly available. Thus, the two solutions to accessing it are the API I used or iterating through `window.standardWindowButtons(.documentIcon).superview.subviews` for a `NSTextField`. Your choice on which one we use.
* I removed a line in `windowWillExitFullScreen` for reverting the appearance; I don't see what its purpose is. Please take a look to see if it's important.
* There is are fundamental issues with the layering of the views in `MainWindowController` that will require a lot of work to fix. I've left it alone for now, but we will need to take a look at it sometime.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -121,17 +121,6 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
     case hidden // indicating sidebar is hidden. Should only be used by sideBarStatus
     case settings
     case playlist
-    func width() -> CGFloat {
-      switch self {
-      case .settings:
-        return 360
-      case .playlist:
-        return 240
-      default:
-        Utility.fatal("SideBarViewType.width shouldn't be called here")
-        return 0
-      }
-    }
   }
 
   var sideBarStatus: SideBarViewType = .hidden
@@ -779,6 +768,16 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
       if !isInFullScreen {
         titleTextField?.animator().alphaValue = 0
       }
+      var viewController: SidebarViewController
+      switch sideBarStatus {
+      case .playlist:
+        viewController = playlistView
+      case .settings:
+        viewController = quickSettingView
+      default:
+        return
+      }
+      viewController.downShift = 0
     }) {
       // if no interrupt then hide animation
       if self.animationState == .willHide {
@@ -806,6 +805,16 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
         if !isInFullScreen {
           titleTextField?.animator().alphaValue = 1
         }
+        var viewController: SidebarViewController
+        switch sideBarStatus {
+        case .playlist:
+          viewController = playlistView
+        case .settings:
+          viewController = quickSettingView
+        default:
+          return
+        }
+        viewController.downShift = titleBarView.frame.height
       }
     }) {
       self.animationState = .shown
@@ -857,19 +866,25 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
     }
   }
 
-  private func showSideBar(view: NSView, type: SideBarViewType) {
+  private func showSideBar(viewController: SidebarViewController, type: SideBarViewType) {
     // adjust sidebar width
-    let width = type.width()
-    sideBarWidthConstraint.constant = width
-    sideBarRightConstraint.constant = -width
+    guard let view = (viewController as? NSViewController)?.view else {
+        Utility.fatal("viewController is not a NSViewController")
+    }
+    sideBarWidthConstraint.constant = view.fittingSize.width
+    sideBarRightConstraint.constant = -view.fittingSize.width
     sideBarView.isHidden = false
     // add view and constraints
     sideBarView.addSubview(view)
-    let constraintsH = NSLayoutConstraint.constraints(withVisualFormat: "H:|-0-[v]-0-|", options: [], metrics: nil, views: ["v": view])
-    let constraintsV = NSLayoutConstraint.constraints(withVisualFormat: "V:|-0-[v]-0-|", options: [], metrics: nil, views: ["v": view])
+    let constraintsH = NSLayoutConstraint.constraints(withVisualFormat: "H:|[v]|", options: [], metrics: nil, views: ["v": view])
+    let constraintsV = NSLayoutConstraint.constraints(withVisualFormat: "V:|[v]|", options: [], metrics: nil, views: ["v": view])
     NSLayoutConstraint.activate(constraintsH)
     NSLayoutConstraint.activate(constraintsV)
     // show sidebar
+    if animationState == .shown {
+      var viewController = viewController
+      viewController.downShift = titleBarView.frame.height
+    }
     NSAnimationContext.runAnimationGroup({ (context) in
       context.duration = 0.2
       context.timingFunction = CAMediaTimingFunction(name: kCAMediaTimingFunctionEaseIn)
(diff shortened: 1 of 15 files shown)


saagarjha:
A picture is worth a thousand words, so here's 4000:

The pop-up button:
screen shot 2017-02-14 at 10 47 29 am
screen shot 2017-02-14 at 10 47 58 am

The title automatically truncates:
screen shot 2017-02-14 at 10 53 15 am
screen shot 2017-02-14 at 10 53 28 am

saagarjha:
@lhc70000 Fixed the merge conflicts.

saagarjha:
@lhc70000 Could you please take a look at this, and merge it if appropriate? I've gone through and fixed the merge conflicts.

lhc70000:
Sorry, I was busy with my coursework last week…… I'll take a look later today and promise nothing will be committed to `develop` before merging this PR.

saagarjha:
Thanks :)

iina/MainWindowController.swift
```diff
@@ -162,7 +151,21 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
   var magnificationGestureRecognizer: NSMagnificationGestureRecognizer = NSMagnificationGestureRecognizer(target: self, action: #selector(MainWindowController.handleMagnifyGesture(recognizer:)))
 
   @IBOutlet weak var titleBarView: NSVisualEffectView!
-  @IBOutlet weak var titleTextField: NSTextField!
+  var standardWindowButtons: [NSButton] {
+    get {
+      return ([.closeButton, .miniaturizeButton, .zoomButton, .documentIconButton] as [NSWindowButton]).flatMap {
```
lhc70000: Why `flatMap`? Isn't `map` enough?

saagarjha: `flatMap` strips out `nil`.

iina/QuickSettingViewController.swift
```diff
@@ -41,6 +38,9 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
   @IBOutlet weak var subTabBtn: NSButton!
   @IBOutlet weak var tabView: NSTabView!
 
+	@IBOutlet weak var buttonTopConstraint: NSLayoutConstraint!
```
saagarjha: Removed.

iina/PlaylistViewController.swift
```diff
@@ -134,7 +141,7 @@ class PlaylistViewController: NSViewController, NSTableViewDataSource, NSMenuDel
       playlistBtn.attributedTitle = NSAttributedString(string: playlistStr, attributes: Utility.tabTitleFontAttributes)
     }
   }
-
+
```
saagarjha: Removed.

saagarjha:
> QuickSettingView:
> 
> The animation is annoying to me (I set auto hide duration to 2s, so it often jumps up and down). It'll distract user's attention. I also asked some user's opinion and they don't like it a lot.

Ok, I removed it.

> There're warnings in QuickSettingView.xib: Frame for "Custom View" will be different at run time.

This appears to be a bug in Interface Builder, since the frames can't be fixed. I'll file a Radar.

> The runtime position for "1x" is also incorrect.

Fixed.

> Please don't change the layout, like sub position slider.

Reverted.

> Titlebar:
> 
> I prefer using window.standardWindowButtons(.documentIcon).superview.subviews.

Ok.

> Please set the titlebar's truncate mode to middle.

The system's behavior is to clip the end; this doesn't appear to be customizable.

> Please don't include irrelevant stuff in one PR (I've told you twice), and do discuss with others when implementing a new feature. I can't merge this PR since the QuickSettingView is totally out of the plan. My proposed solution: close this PR and open another one for titlebar (which only includes changes to MainWindowController), and discuss whether we need to change anything else.

This isn't "irrelevant stuff", and AutoLayout for the QuickSettingView had already been discussed in #253. The document icon button is "above" the QuickSettingView when it wasn't before, so I had to shift the QuickSettingView down to accommodate. This was impossible to do without fixing the layout for QuickSettingView. Unless you want the title bar and QuickSettingView to overlap, you'll need to merge these together.

lhc70000:
0

> The document icon button is "above" the QuickSettingView when it wasn't before

Sorry, I didn't even know this before. This is an important problem and definitely need discussion (you should discuss with me once found it, but not changing UI by yourself).

My thought based on current situation:
- Use fixed padding (i.e. `downShift`) for quick setting view
- Or, add back the animation (with a shorter interval)

For this PR please just revert back to the initial animation version and I can handle the remaining stuff later.

1

> This appears to be a bug in Interface Builder, since the frames can't be fixed

For slider labels and equalizer, why not putting them in a box and only add auto layout for the box? I think it won't affect the overall layout.
The messed up xib makes it hard to do further adjustment on the controls. Is the custom view approach a commonly used trick? 

2

> The system's behavior is to clip the end; this doesn't appear to be customizable.

Now that we got the `NSTextField` itself, there's nothing we can't do on it 😉. Confirmed the following code works:
```swift
if let attrTitle = titleTextField?.attributedStringValue.mutableCopy() as? NSMutableAttributedString {
    let p = attrTitle.attribute(NSParagraphStyleAttributeName, at: 0, effectiveRange: nil) as! NSMutableParagraphStyle
    p.lineBreakMode = .byTruncatingMiddle
    attrTitle.addAttribute(NSParagraphStyleAttributeName, value: p, range: NSRange(location: 0, length: attrTitle.length))
}
```

3

> > Don't change the layout

It also include "Load External Audio" and other buttons.

4

> This isn't "irrelevant stuff"

You should told me detailed info like said above. But besides this, there're more "irrelevant stuff" like fixing font weight warning, `Utility.fatal` etc., they are actually pretty good, but made it difficult to review and I could easily . Separate PR next time, please.

---

Do you have an IM account like telegram, or perhaps a Gitter or Slack chatroom is better? So that we can discuss immediately once something happened, if you plan to make further contribution on this project.

saagarjha:
> 0
> 
> The document icon button is "above" the QuickSettingView when it wasn't before
> Sorry, I didn't even know this before. This is an important problem and definitely need discussion (you should discuss with me once found it, but not changing UI by yourself).
> 
> My thought based on current situation:
> 
> Use fixed padding (i.e. downShift) for quick setting view
> Or, add back the animation (with a shorter interval)
> For this PR please just revert back to the initial animation version and I can handle the remaining stuff later.
> 

Ok, I've put the animation back. We can work on improving it later.

> 1
> 
> This appears to be a bug in Interface Builder, since the frames can't be fixed
> For slider labels and equalizer, why not putting them in a box and only add auto layout for the box? I think it won't affect the overall layout.
> The messed up xib makes it hard to do further adjustment on the controls. Is the custom view approach a commonly used trick?
> 

The "normal" method in this case is to let the Stack View handle the layout; however, this method doesn't work on OS X Yosemite. We need to use views in the middle and let them have equal widths to space them out.

> 2
> 
> The system's behavior is to clip the end; this doesn't appear to be customizable.
> Now that we got the NSTextField itself, there's nothing we can't do on it 😉. Confirmed the following code works:
> 
> if let attrTitle = titleTextField?.attributedStringValue.mutableCopy() as? NSMutableAttributedString {
>     let p = attrTitle.attribute(NSParagraphStyleAttributeName, at: 0, effectiveRange: nil) as! NSMutableParagraphStyle
>     p.lineBreakMode = .byTruncatingMiddle
>     attrTitle.addAttribute(NSParagraphStyleAttributeName, value: p, range: NSRange(location: 0, length: attrTitle.length))
> }

Are you sure? This will have to be done repeatedly, whenever the titleTextField changes (which it does often).

> 3
> 
> Don't change the layout
> It also include "Load External Audio" and other buttons.
> 

Reverted.

> 4
> 
> This isn't "irrelevant stuff"
> You should told me detailed info like said above. But besides this, there're more "irrelevant stuff" like fixing font weight warning, Utility.fatal etc., they are actually pretty good, but made it difficult to review and I could easily . Separate PR next time, please.

The changes to `Utility.fatal` came out of the new use of `fatal` from a background thread, which would crash the thread but not the whole app.

> 
> Do you have an IM account like telegram, or perhaps a Gitter or Slack chatroom is better? So that we can discuss immediately once something happened, if you plan to make further contribution on this project.

Does IINA have a Slack or Gitter? I could join that.

lhc70000:
> Are you sure? This will have to be done repeatedly, whenever the titleTextField changes (which it does often).

Seems that we only need to set it once (Cocoa will update the string value only).

> Does IINA have a Slack or Gitter? I could join that.

Created one and invited you 🎉 https://gitter.im/iina-dev/Lobby

PR #389, opened 2017-02-14
Fixing ellipsis

cgand:
Substitute all … (3 dots) with … (ellipsis) globally with: perl -i -0pe
's/\.\.\./…/mg' */*.*


saagarjha:
While your intentions were fine, it appears that you've replaced too much; for example, Array subscripting should not use ellipses. We'll need to go through with a finer tool than find/replace.

cgand:
Ooppss yes, I'd should apply replace only to *.xib and *.strings.
Perhaps it left only 'MainWindowController.swift', 'OSDMessage.swift' that can be done manually.
(Not sure about 'other/parse_doc.rb').

cgand:
I did a better one:
https://github.com/lhc70000/iina/pull/394

Issue #372, opened 2017-02-13
Show (V-A-S/P-C) panels toggle it instead of switch from one to other


cgand:
**System and IINA version:**
- macOS 10.12.2
- IINA 0.0.4

**Expected behavior:**
Cmd+A -> Show audio pannel
Cmd+V -> Show/switch to video panel

**Actual behavior:**
Cmd+A -> Show audio pannel
Cmd+V -> Hide audio panel (Instead of switch to video panel)

**Steps to reproduce:**
Press Cmd+A and then press Cmd+V

Same with all other panels Cmd+S or between Cmd+P /Cmd+C for playlist and chapter panel

**How often does this happen?**
Ever

alejx:
Duplicated with #362

alejx:
Close due to the duplication.

PR #407, opened 2017-02-17
Improve shortcut. Fix #362, Fix 372.

### LINKED ISSUE #362: Enhancement on shortcut in Chapter tab etc.
**System and IINA version:**

- macOS 10.12.4 developer beta 2
- IINA 0.0.4 build 9

**Actual behavior:**

When user is on Chapter tab, press `cmd + P` will collapse the side view. So does in Video, Subtitle and Audio view.

**Expected behavior:**

Change to the tab that user wants without collapse the side view.

**Steps to reproduce:**

Open a file, then use `cmd + C` to show Chapter tab. Then press `cmd + P`. The side view is collapsed.

Every time.

alejx:
- [x] It implements / fixes issue #362 , #372 .

---

**Description:**

Prevent collapsing when change to a different tab.

In addition, it memorizes which tab did user switch to last time.

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -98,7 +98,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
   }
 
   func applicationShouldTerminateAfterLastWindowClosed(_ sender: NSApplication) -> Bool {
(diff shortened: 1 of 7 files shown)


alejx:
See #439.

PR #430, opened 2017-02-22
Synchronize QuickSettingsView text fields and slider and localize, fix #425, #492

### LINKED ISSUE #425: Play speed input validation
**System and IINA version:**

- macOS 10.12.3
- IINA 0.0.5

**Expected behavior:**

User wants to adjust custom play speed. In Edit field is already written "1.00" so user expect to write "1.33" and that value to be accepted and applied.

**Actual behavior:**

If user tries to write "1.33" or any other decimal value using decimal point (".") it doesn't work. It is necessary to write value using comma "," like "1,33" value to be accepted, despite default value displayed with decimal point. Any future value will be displayed with decimal point as well but newly written value need to be entered using comma instead of point.

Crash report:
There is no crash.

**Steps to reproduce:**
Already described in expected/actual behaviour

Always

saagarjha:
- [X] It implements / fixes issue #425, #495

---

**Description:**
Fixes issue #425, #495.

### DIFF
--- a/iina/Base.lproj/QuickSettingViewController.xib
+++ b/iina/Base.lproj/QuickSettingViewController.xib
@@ -2,6 +2,7 @@
 <document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="12117" systemVersion="16E189a" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
     <dependencies>
         <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="12117"/>
+        <capability name="Alignment constraints with different attributes" minToolsVersion="5.1"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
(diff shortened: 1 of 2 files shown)


lhc70000:
Do other text fields have the same potential problem?

saagarjha:
It appears that the audio delay and subtitle delay text fields should have the same issue. Would you like me to add those to this pull request?

lhc70000:
Please add them, thanks.

saagarjha:
@lhc70000 Added the other text fields.

lhc70000:
- If enter a value lower than min value like 0.20, the indicator will only display "0.25" without "x"
- Indicator's position is incorrect: the first tick is not aligned with the left frame border, and the same with the last tick.
1df5f7d2-ab00-4f46-8efb-aee41b7e19cd


I'll fix them later after merged.

lhc70000:
See cdf5b7f.

Also, please don't clamp values from textfields. They're designed for inputting values that the slider can't represent.

saagarjha:
@lhc70000 I pretty sure we don't want a negative speed…

PR #475, opened 2017-03-09
Fix #471, placeholder issue

### LINKED ISSUE #471: suggestion on progress bar initial position
When IINA is launched and before playback starts, progress bar is centered and total length is a placeholder with the value 9:99:99.
It'd be more intuitive to place the progress bar at the beginning (left most position) and maybe show -:--:-- as video length.

alejx:
Change the default value of left and right labels and the initial value
for progress bar.

- [x] It implements / fixes issue #471 .

---

**Description:**
The default value of left and right labels are "-:--".


lhc70000:
Please change to something like `-:--:--` to remain the same text width

PR #476, opened 2017-03-09
Accept dragged strings and URLs on icon, fix #294 (part 2), fix #374

### LINKED ISSUE #294: drag'n'drop URL's on IINA in Dock and window

### LINKED ISSUE #374: Should not open initial open penal when open with browser extension.
**System and IINA version:**

- macOS 10.12.4 developer beta 2
- IINA 0.0.5 build 11 & Safari extension version 1

**Expected hehavior:**

Does not show initial open penal.

**Actual behavior:**

It shows.

**Steps to reproduce:**

Open from web via extension.

Every time.

alejx:
Accept dropped strings and URLs on IINA icon.

- [x] It implements issue #294 (part 2) & #374 .

---

**Description:**
Idea by http://stackoverflow.com/questions/27420016/osx-drag-and-drop-text-not-file-to-app-dock-icon .

I used a quite 'dirty' way to prevent the open panel. I call a new method instead of `openFile(_:)` after 0.5s to judge if I have to open the panel. Because system call `applicationWillFinishLaunching(_:)` first. At that time, I have no idea about how to know whether the app opened by the accept dropping service or not.

This pull request also prevents the open panel from showing up when the app is called by the browser extension.

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -14,6 +14,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
 
   var isReady: Bool = false
   var handledDroppedText: Bool = false
+  var handledURLEvent: Bool = false
 
   var pendingURL: String?
 
@@ -144,7 +145,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
   }
 
   func checkServiceStartup() {
-    if !handledDroppedText {
+    if !handledDroppedText && !handledURLEvent {
       openFile(self)
     }
   }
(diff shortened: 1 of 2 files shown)


iina/AppDelegate.swift
```diff
@@ -13,6 +13,7 @@ import MASPreferences
 class AppDelegate: NSObject, NSApplicationDelegate {
 
   var isReady: Bool = false
+  var isServiceStartup: Bool = false
```
lhc70000: I think a name like `handledDroppedText` will be better.

PR #487, opened 2017-03-14
Quick fixes for recently introduced regressions

saagarjha:
* Fix a warning for printing an String?
* Remove unnecessary "-0-" constraints
* Xcode automatically sets allowsCharacterPickerTouchBarItem="YES" for text fields

---

**Description:**


lhc70000:
Please reset to width="373" in `QuickSettingsView` (do you know why? should be related to the scroll bar……)

saagarjha:
Didn't you want the width to be 360? Anyways, the reason for the resize is that there's a constraint on the top-level view that's set to 360.

lhc70000:
Now my Xcode sets width to 360 now (but it sometimes resize the width to 373, weird).

But it still removes `allowsCharacterPickerTouchBarItem="YES"` automatically. Should it be a version problem?

saagarjha:
Which version of Xcode are you using (I'm on 8W143q)? It appears that Apple changed the behavior of allowsCharacterPickerTouchBarItem's default recently.

lhc70000:
8C1002. (I'm still on stable versions of macOS and Xcode)

saagarjha:
Alright, feel free to use whatever changes your Xcode uses–I'm not quite sure what allowsCharacterPickerTouchBarItem actually does.

PR #489, opened 2017-03-15
Add restricted number formatter class.

alejx:
Accept only numbers and dot (in case of decimal style formatter)

---

**Description:**

Accept only numbers and dot. When user input an unrelated character it will call `NSBeep()`.

### DIFF
--- a/iina/Formatters.swift
+++ b/iina/Formatters.swift
@@ -10,7 +10,7 @@ import Cocoa
 
 class RestrictedNumberFormatter : NumberFormatter {
 
-  init(_ min: Float?, max: Float?, isDecimal: Bool) {
+  init(_ min: Double?, max: Double?, isDecimal: Bool) {
     super.init()
     if isDecimal {
       self.numberStyle = .decimal
--- a/iina/Formatters.swift
+++ b/iina/Formatters.swift
@@ -8,22 +8,6 @@
 
 import Cocoa
 
-class DecimalFormatter : NumberFormatter {
-
-  override init() {
-    super.init()
-    self.numberStyle = .decimal
-  }
-
-  required init?(coder aDecoder: NSCoder) {
-    fatalError("init(coder:) has not been implemented")
-  }
-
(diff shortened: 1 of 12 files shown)


iina/Formatters.swift
```diff
@@ -23,3 +23,45 @@ class DecimalFormatter : NumberFormatter {
...
+
+class RestrictedNumberFormatter : NumberFormatter {
+
+  init(_ min: Float?, max: Float?, isDecimal: Bool) {
+    super.init()
+    if isDecimal {
+      self.numberStyle = .decimal
+    }
+    if let minValue = min {
+      minimum = minValue as NSNumber?
```
alejx: > Also, we should consider using Double for min/max instead of Float.
That's wired. Both my local version and 'File Changed' on GitHub shows that I used `Double` instead of `Float`.

iina/Formatters.swift
```diff
@@ -23,3 +23,45 @@ class DecimalFormatter : NumberFormatter {
     return true
   }
 }
+
+class RestrictedNumberFormatter : NumberFormatter {
+
+  init(_ min: Float?, max: Float?, isDecimal: Bool) {
```
alejx: Good idea!

iina/Formatters.swift
alejx: > Swift should bridge min to NSNumber automatically, right?

Swift will NOT convert it automatically. Xcode arise a error says
> Cannot assign value of type 'Double?' to type 'NSNumber?'

And it suggest me to add ` as NSNumber?`

But I found that I do not have to unwrap it explicitly. So I will change this piece of code into
`minimum = min as NSNumber?`
`maximum = max as NSNumber?`

iina/Formatters.swift
```diff
@@ -23,3 +23,45 @@ class DecimalFormatter : NumberFormatter {
...
+      maximum = maxValue as NSNumber?
+    }
+  }
+
+  override init() {
+    super.init()
+  }
+  
+  required init?(coder aDecoder: NSCoder) {
+    fatalError("init(coder:) has not been implemented")
```
alejx: This piece of code is inherited from @lhc70000 's previous code.

iina/Formatters.swift
saagarjha: Sorry, my mistake; I didn't realize that this wouldn't be bridged.

saagarjha:
LGTM.

PR #496, opened 2017-03-16
Connect more mpv window options

lhc70000:
`ontop`

- window's floating on top status should be syn ced with mpv's `ontop` property.
- able to set key bindings for `ontop` (e.g. `cycle ontop` `set ontop yes`).
- options like `ontop=yes` should work initially

`geometry`

- see [mpv's doc](https://mpv.io/manual/stable/#options-geometry) and test whether it works normally.
- make sure it doesn't break any existing function.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -77,6 +77,14 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
       playerCore.mpvController.setFlag(MPVOption.Window.fullscreen, isInFullScreen)
     }
   }
+
+  var isOntop: Bool = false {
+    didSet {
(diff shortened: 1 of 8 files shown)


iina/MainWindowController.swift
```diff
@@ -77,6 +79,14 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
       playerCore.mpvController.setFlag(MPVOption.Window.fullscreen, isInFullScreen)
     }
   }
+
+  var isOntop: Bool = false {
+    didSet {
+      playerCore.mpvController.setFlag(MPVOption.Window.ontop, isOntop)
+
```
alejx: Is the empty line necessary?

lhc70000:
updated, please merge if you everything looks good to you.

PR #500, opened 2017-03-17
More readable keybindings

lhc70000:
8bc1de26-d347-42a2-8e49-a874d8ebfb2d

7c408ac5-98ac-4cfe-a509-49f54c8473e6

### DIFF
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -153,16 +153,6 @@ extension NSPoint {
   }
 }
 
-extension Dictionary {
-  init<S: Sequence>
-    (_ seq: S) where S.Iterator.Element == Element {
-    self.init()
-    for (k, v) in seq {
-      self[k] = v
-    }
-  }
-}
-
 extension Array {
   func at(_ pos: Int) -> Element? {
     if pos < count {
@@ -180,12 +170,6 @@ extension NSMenu {
     menuItem.representedObject = obj
     menuItem.state = stateOn ? NSOnState : NSOffState
     self.addItem(menuItem)
-  }
-}
-
-extension String {
-  func splitted(by separator: Character) -> [String] {
-    return self.characters.split(separator: separator).map { String($0) }
   }
 }
 
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -153,6 +153,16 @@ extension NSPoint {
   }
 }
 
+extension Dictionary {
+  init<S: Sequence>
+    (_ seq: S) where S.Iterator.Element == Element {
+    self.init()
+    for (k, v) in seq {
+      self[k] = v
+    }
+  }
+}
+
 extension Array {
   func at(_ pos: Int) -> Element? {
     if pos < count {
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -183,6 +183,12 @@ extension NSMenu {
   }
 }
 
+extension String {
+  func splitted(by separator: Character) -> [String] {
+    return self.characters.split(separator: separator).map { String($0) }
+  }
+}
+
 extension Int {
   func toStr() -> String {
     return "\(self)"
--- a/iina/KeyBindingTranslator.swift
+++ b/iina/KeyBindingTranslator.swift
@@ -11,6 +11,12 @@ import Mustache
 
 class KeyBindingTranslator {
 
+  static let l10nDic: [String: String] = {
+    let filePath = Bundle.main.path(forResource: "KeyBinding", ofType: "strings")!
+    let dic = NSDictionary(contentsOfFile: filePath) as! [String : String]
+    return dic
+  }()
+
   static let mpvCmdFormat: [String: String] = {
     let filePath = Bundle.main.path(forResource: "MPVCommandFormat", ofType: "strings")!
     let dic = NSDictionary(contentsOfFile: filePath) as! [String : String]
@@ -28,7 +34,7 @@ class KeyBindingTranslator {
 
     // If translated
     if let mpvFormat = KeyBindingTranslator.mpvCmdFormat[cmd],
-      let cmdTranslation = KeyBindingItem.l10nDic["read." + cmd],
+      let cmdTranslation = l10nDic["read." + cmd],
       let tmpl = try? Template(string: cmdTranslation) {
       // parse command
       var data: [String: String] = [:]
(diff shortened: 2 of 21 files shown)


saagarjha:
If you're looking to improve the keybindings even more, macOS has standard icons for many keys: https://support.apple.com/kb/PH25100?locale=en_US

lhc70000:
Completed. @saagarjha would you like to review it?

Expected behavior:

- By default it displays key bindings in a more readable format
  - Table cells editing by double clicking will be disabled, but instead, double clicking will show a more user friendly panel.
- An option is provided for switching back to raw key names and commands
  - If checked, users should be able to edit table cells directly as before
- As `NSView` can't capture media keys, I added a dropdown list to enable users inputting a media key.

saagarjha:
Haha, I was actually just going through it!

iina/Extensions.swift
```diff
@@ -153,6 +153,16 @@ extension NSPoint {
   }
 }
 
+extension Dictionary {
```
saagarjha: Unnecessary, see below

iina/Extensions.swift
```diff
@@ -173,6 +183,12 @@ extension NSMenu {
   }
 }
 
+extension String {
```
saagarjha: Could you explain what this does exactly? From what it looks like you're trying to `filter` the `String`–am I correct?

iina/KeyBindingTranslator.swift
```diff
@@ -7,9 +7,106 @@
 //
 
 import Foundation
+import Mustache
 
 class KeyBindingTranslator {
 
+  static let l10nDic: [String: String] = {
+    let filePath = Bundle.main.path(forResource: "KeyBinding", ofType: "strings")!
```
saagarjha: I tend to shy away from force unwrapping; this will be the first of these comments

Use a `guard` here

iina/KeyBindingTranslator.swift
```diff
@@ -7,9 +7,106 @@
...
+    let dic = NSDictionary(contentsOfFile: filePath) as! [String : String]
```
saagarjha: Instead of force unwrapping fall back to `?? [:]`

iina/KeyBindingTranslator.swift
```diff
@@ -7,9 +7,106 @@
...
+    return dic
+  }()
+
+  static let mpvCmdFormat: [String: String] = {
+    let filePath = Bundle.main.path(forResource: "MPVCommandFormat", ofType: "strings")!
```
saagarjha: Same as above

iina/KeyBindingTranslator.swift
```diff
@@ -7,9 +7,106 @@
...
+    let dic = NSDictionary(contentsOfFile: filePath) as! [String : String]
+    return dic
+  }()
+
+  static private let UnsupportedCmdPrefix: [String] = [
```
saagarjha: Type is unnecessary here, it will be inferred

iina/KeyMapping.swift
```diff
@@ -10,6 +10,30 @@ import Foundation
 
 class KeyMapping {
 
+  static let prettyKeySymbol: [String: String] = [
```
saagarjha: Again, type is unnecessary

iina/PrefKeyBindingViewController.swift
```diff
@@ -370,4 +384,25 @@ extension PrefKeyBindingViewController: NSTableViewDelegate, NSTableViewDataSour
...
+  func editRow() {
+    guard shouldEnableEdit else { return }
+    let selectedData = currentMapping[kbTableView.selectedRow]
+    keyRecordViewController.keyCode = selectedData.key
+    keyRecordViewController.action = selectedData.readableAction
+    showKeyBindingPanel {
+      let key = keyRecordViewController.keyCode
+      let action = keyRecordViewController.action
+      guard !key.isEmpty && !action.isEmpty else { return }
+      let splitted = action.characters.split(separator: " ").map { String($0) }
```
saagarjha: See comments on `splitted` above

iina/PrefKeyBindingViewController.swift
```diff
@@ -108,6 +110,20 @@ class PrefKeyBindingViewController: NSViewController, MASPreferencesViewControll
...
 
+  func showKeyBindingPanel(ok: () -> Void) {
+    let panel = NSAlert()
+    panel.messageText = NSLocalizedString("keymapping.title", comment: "Key Mapping")
+    panel.informativeText = NSLocalizedString("keymapping.message", comment: "Press any key to record.")
+    panel.accessoryView = keyRecordViewController.view
+    panel.window.initialFirstResponder = keyRecordViewController.keyRecordView
+    panel.addButton(withTitle: "OK")
+    panel.addButton(withTitle: "Cancel")
+    let response = panel.runModal()
```
saagarjha: `if panel.runModal() == NSAlertFirstButtonReturn`

iina/KeyCodeHelper.swift
```diff
@@ -128,6 +128,10 @@ class KeyCodeHelper {
     0x7F: ("POWER", nil) // This should be KeyCode::PC_POWER.
   ]
 
+  static var reversedKeyMapForShift = Dictionary(keyMap.flatMap { (k, v) in
```
saagarjha: Untested, but try this:

```swift
static var reversedKeyMapForShift = keyMap.reduce([:]) { partial, keyMap in
	var partial = partial
	if let value = keyMap.value.1 {
		partial[value] = keyMap.value.0
	}
	return partial
}
```

iina/KeyBindingTranslator.swift
```diff
@@ -7,9 +7,106 @@
...
+    "no-osd", "osd-auto", "osd-bar", "osd-msg-bar", "raw", "repeatable", "expand-properties"
+  ]
+
+  static func readable(fromCommand action: [String]) -> String {
```
saagarjha: I'd recommend renaming this function to something that gives a bit more insight about its parameters and return value. If I'm not mistaken, it converts an action into a readable command, so something like `readableCommand(from action: [String])` might be a better choice.

iina/Extensions.swift
lhc70000: No, it just splits the string (should be the same as `components(separatedBy:)`?)

Or I can just use `components(separatedBy:)`.

saagarjha: Yes, then you're looking for `components(seperatedBy:)`.

lhc70000:
Updated, but there's still a problem:

- Edit any entry in key bindings (double click, shows panel, change the value and OK)
- Edit another one. Double click and panel shows.
- The action text field in the panel will **display text only when focused**.

lhc70000:
`NSViews`'s `keyDown` doesn't receive media key events (they must be handled in lower level), so users must select those keys manually. I'll try to add some hints.

PR #521, opened 2017-03-26
Improve "open url" dialog

lhc70000:
**Description:**

- Add GUI for HTTP authentication.
- Add links to browser extensions.
- Better handles user's input.

e742949d-b4ca-4194-85df-d2f1d0215e2a

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -181,8 +181,17 @@ class AppDelegate: NSObject, NSApplicationDelegate {
   }
 
   @IBAction func openURL(_ sender: NSMenuItem) {
-    let _ = Utility.quickPromptPanel(messageText: "Open URL", informativeText: "Please enter the url:") { str in
-      playerCore.openURLString(str)
+    let panel = NSAlert()
+    panel.messageText = "Open URL"
+    panel.informativeText = "Please enter the url:"
+    let inputViewController = OpenURLAccessoryViewController()
+    panel.accessoryView = inputViewController.view
+    panel.addButton(withTitle: "OK")
+    panel.addButton(withTitle: "Cancel")
+    panel.window.initialFirstResponder = inputViewController.urlField
+    let response = panel.runModal()
+    if response == NSAlertFirstButtonReturn {
+      playerCore.openURLString(inputViewController.urlField.stringValue)
     }
   }
 
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -183,7 +183,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
   @IBAction func openURL(_ sender: NSMenuItem) {
     let panel = NSAlert()
     panel.messageText = "Open URL"
-    panel.informativeText = "Please enter the url:"
+    panel.informativeText = "Please enter the URL:"
     let inputViewController = OpenURLAccessoryViewController()
     panel.accessoryView = inputViewController.view
     panel.addButton(withTitle: "OK")
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -191,7 +191,11 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     panel.window.initialFirstResponder = inputViewController.urlField
     let response = panel.runModal()
     if response == NSAlertFirstButtonReturn {
-      playerCore.openURLString(inputViewController.urlField.stringValue)
+      if let url = inputViewController.url {
+        playerCore.openURL(url)
+      } else {
+        Utility.showAlert("wrong_url_format")
+      }
     }
   }
 
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -316,3 +316,13 @@ extension Data {
     return fileUrl
   }
 }
+
+
+extension CharacterSet {
+  static let urlAllowed = CharacterSet.urlHostAllowed
+    .union(.urlUserAllowed)
+    .union(.urlPasswordAllowed)
+    .union(.urlPathAllowed)
+    .union(.urlQueryAllowed)
+    .union(.urlFragmentAllowed)
+}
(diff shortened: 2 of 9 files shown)


alejx:
I suggest to hide HTTP Authentication as well as the browser extension indicators by default and use a disclosure triangle button to show them on the same panel.

Besides, using NSBox to wrap both HTTP Authentication and browser extension indicators seems not good. Actually URL and HTTP Authentication are more related. It currently seems like the two bottom things are related.

Use "URL" instead of "url".

saagarjha:
The alert that IINA shows when the URL is invalid appears to lack localization:

screen shot 2017-03-26 at 14 39 43

alejx:
@saagarjha Maybe forgot to call NSLocalizedString?

saagarjha:
@alejx I haven't looked at the code yet, but that's probably the issue.

iina/AppDelegate.swift
```diff
@@ -181,8 +181,21 @@ class AppDelegate: NSObject, NSApplicationDelegate {
   }
 
   @IBAction func openURL(_ sender: NSMenuItem) {
-    let _ = Utility.quickPromptPanel(messageText: "Open URL", informativeText: "Please enter the url:") { str in
-      playerCore.openURLString(str)
+    let panel = NSAlert()
+    panel.messageText = "Open URL"
+    panel.informativeText = "Please enter the url:"
```
saagarjha: Capitalize URL

iina/OpenURLAccessoryViewController.swift
```diff
@@ -0,0 +1,58 @@
...
+      let username = usernameField.stringValue
+      let password = passwordField.stringValue
+      guard var urlValue = urlField.stringValue.addingPercentEncoding(withAllowedCharacters: .urlAllowed) else {
+        return nil
+      }
+      if URL(string: urlValue)?.host == nil {
+        urlValue = "http://" + urlValue
+      }
+      guard let urlComponents = NSURLComponents(string: urlValue) else { return nil }
+      if !username.isEmpty && !password.isEmpty {
```
saagarjha: Is this necessary? I expect that some websites don't have a password but require a valid username.

iina/OpenURLAccessoryViewController.swift
```diff
@@ -0,0 +1,58 @@
...
+      }
+      return urlComponents.url
+    }
+  }
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+
+    [safariLinkBtn, chromeLinkBtn].forEach {
+      $0!.image = NSImage(named: NSImageNameFollowLinkFreestandingTemplate)
```
saagarjha: Please don't force unwrap

iina/PlayerCore.swift
```diff
@@ -47,14 +47,14 @@ class PlayerCore: NSObject {
     openMainWindow(path: path, url: url!, isNetwork: false)
   }
 
+  func openURL(_ url: URL) {
+    let path = url.absoluteString
+    print(path)
```
saagarjha: Remove

iina/Extensions.swift
```diff
@@ -316,3 +316,13 @@ extension Data {
     return fileUrl
   }
 }
+
+
+extension CharacterSet {
+  static let urlAllowed = CharacterSet.urlHostAllowed
```
saagarjha: This is *technically* wrong but probably the easiest way to do it. I'd add a comment here for posterity explaining that we should be splitting up the URL and encoding each part separately.

iina/Preference.swift
```diff
@@ -346,7 +346,7 @@ struct Preference {
     Key.controlBarPositionHorizontal: Float(0.5),
     Key.controlBarPositionVertical: Float(0.1),
     Key.controlBarStickToCenter: true,
-    Key.controlBarAutoHideTimeout: Float(5),
+    Key.controlBarAutoHideTimeout: Float(2.5),
```
saagarjha: Any reason for this change?

iina/OpenURLAccessoryViewController.swift
alejx: Shouldn't we force unwrap `NSButton`s?

saagarjha: They should always be non-`nil`, but I'd think it's better behavior to just not set the image than to crash in this case.

iina/AppDelegate.swift
```diff
@@ -181,8 +181,21 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
+    panel.accessoryView = inputViewController.view
+    panel.addButton(withTitle: "OK")
+    panel.addButton(withTitle: "Cancel")
+    panel.window.initialFirstResponder = inputViewController.urlField
+    let response = panel.runModal()
+    if response == NSAlertFirstButtonReturn {
+      if let url = inputViewController.url {
+        playerCore.openURL(url)
+      } else {
+        Utility.showAlert("wrong_url_format")
```
alejx: Just a reminder.

iina/OpenURLAccessoryViewController.swift
lhc70000: They are `NSButton!` so it's equivalent to call `safariLinkBtn.image = ...` directly. Using `?` helps nothing, except we using `?` for all other outlets. I think force unwrapping is designed to be used, not to be avoided.

iina/Preference.swift
lhc70000: It's an unpushed commit on my `develop`.

iina/Extensions.swift
lhc70000: Yes that would be the best solution, but what user want initially is like "percent encoding the spaces" in URL. `NSURLComponent` won't recognize the url if it contains spaces, so if we don't add percent encoding at first, we must split it manually. Plus mpv is doing the same thing, so I think It's OK to stay consistent with mpv.

lhc70000:
> the text fields and their describing labels don't have the be the same width:
It's more convenient. Even if the translation in a language is longer than the text field, they won't overlap at least.

lhc70000:
> IMO we don't even need a box for the browser extensions, I'd put a "hint" suggesting they install the browser extensions

Let's improve it in next release (I don't know how hard it is to embed links in an alert).

saagarjha:
@lhc70000 I'd change the constraint to be less than or equal to, then.

PR #531, opened 2017-03-28
Issue 530

### LINKED ISSUE #530: Minor issues about textfield and slider in quick setting views
**System and IINA version:**

- macOS 10.12.5 beta
- IINA 0.0.7

**Expected behavior:**

1. When started, Both textfields and sliders should be assigned properly by the number remembered by mpv.
2. When user input nothing in textfields and hit enter, the indicator should fall back and display '0s'.

**Actual behavior:**

1. Only textfields are assigned properly. Sliders are always at 0 point.
2. Indicators display 's' only.

**Steps to reproduce:**

1. 
- Open a video, change the subtitle delay to 1s.
- Exit IINA.
- Open the same video and open quick setting view you will see the issue.
2.
- Open a video and open the quick setting view.
- Empty the textfield and hit enter.

Every time.

alejx:
- [x] It implements / fixes issue #530.

---

**Description:**
Fix #530 .


alejx:
@lhc70000 Does the `DecimalFormatter` class necessary? It only applies on `customSpeedTextField` and does nothing different from a normal `NumberFormatter` in xib.

In addition, `customSpeedTextField` should reject 0 value.

iina/QuickSettingViewController.swift
```diff
@@ -394,23 +404,19 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
     }
     let sliderValue = sender.doubleValue
     let value = AppData.minSpeed * pow((AppData.maxSpeed / AppData.minSpeed), sliderValue / sliderSteps)
```
saagarjha: Looks like I left a spare set of parentheses here, would you mind removing them?

iina/QuickSettingViewController.swift
```diff
@@ -505,6 +509,9 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
 
   @IBAction func customAudioDelayEditFinishedAction(_ sender: NSTextField) {
     let value = sender.doubleValue
+    if value == 0 {
```
saagarjha: At first glance it looks like this code doesn't do anything, so I'd prefer `sender.stringValue.isEmpty`, since it makes the meaning explicit.

iina/QuickSettingViewController.swift
```diff
@@ -566,6 +571,9 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
 
   @IBAction func customSubDelayEditFinishedAction(_ sender: NSTextField) {
     let value = sender.doubleValue
+    if value == 0 {
```
saagarjha: Ditto

iina/QuickSettingViewController.swift
```diff
@@ -394,23 +404,19 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
...
-    let formattedValue = (customSpeedTextField.formatter as? NumberFormatter)?.string(from: NSNumber(value: value)) ?? ""
-    speedSliderIndicator.stringValue = "\(formattedValue)x"
-    customSpeedTextField.stringValue = formattedValue
+    customSpeedTextField.doubleValue = value
     playerCore.setSpeed(value)
-    redraw(indicator: speedSliderIndicator, constraint: speedSliderConstraint, slider: speedSlider, value: "\(formattedValue)x")
+    redraw(indicator: speedSliderIndicator, constraint: speedSliderConstraint, slider: speedSlider, value: "\(customSpeedTextField.stringValue)x")
   }
 
   @IBAction func customSpeedEditFinishedAction(_ sender: NSTextField) {
```
saagarjha: The other text fields default to "0" when nothing is entered; I believe that we should make a similar argument for this text field but have it default to "1" (normal speed) instead.

iina/QuickSettingViewController.swift
```diff
@@ -394,23 +404,19 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
     }
```
saagarjha: We might want to implement "snapping" instead of `allowsTickMarkValuesOnly`–but that would be a different pull request.

iina/QuickSettingViewController.swift
alejx: Included in the update commit.

Issue #138, opened 2017-01-11
Black out other monitors in full screen


stevewillows:
The option to go full screen without blacking out other monitors would be ideal for those of us who do not have independent spaces. 

VLC has this by default.

NexusSix:
Going to Fullscreen still blacks out the other monitor even if preferences option is unchecked.  

Sierra 10.12.3
iina build 29
MacBook Pro (15-inch, Mid 2010) 
NVIDIA GeForce GT 330M 512 MB
Intel HD Graphics 288 MB
mini-display to hdmi

PR #562, opened 2017-04-05
Add an option to black out other screens in fullscreen mode

alejx:
- [x] It implements / fixes issue #138  .

---

**Description:**

Black out other screens in fullscreen. Did not test under macOS 10.11 or below, but I think I did not use any incompatible API.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -72,6 +72,9 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
   var hideControlTimer: Timer?
 
   var hideOSDTimer: Timer?
+  
+  var screens: [NSScreen] = []
+  var blackWindowControllers: [NSWindowController] = []
 
   // MARK: - Status
 
@@ -198,7 +201,8 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
     PK.doubleClickAction,
     PK.rightClickAction,
     PK.pinchAction,
-    PK.showRemainingTime
+    PK.showRemainingTime,
+    PK.blackOutMonitor
   ]
 
   // MARK: - Outlets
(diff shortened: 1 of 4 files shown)


iina/MainWindowController.swift
```diff
@@ -386,8 +393,30 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
...
+        if NSScreen.screens()?.count ?? 0 != self.totalScreens {
+          self.removeBlackWindow()
+          self.blackOutOtherMonitors()
+        }
+      }
+    }
+    let resignActiveObserver = NotificationCenter.default.addObserver(forName: NSNotification.Name.NSApplicationDidResignActive, object: nil, queue: .main) { [unowned self] _ in
+      // This observer handles a situation that the user drag the fullscreen video to another screen
+      if self.isInFullScreen && self.ud.bool(forKey: PK.blackOutMonitor) {
+        if self.window?.screen != self.lastScreen {
```
lhc70000: why not putting all of them in one `if`?

iina/MainWindowController.swift
```diff
@@ -64,7 +64,12 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
   var hideControlTimer: Timer?
 
   var hideOSDTimer: Timer?
+  
+  var screens: [NSScreen] = []
+  var totalScreens = 0;
```
lhc70000: remove `;`
and something like `cachedScreenCount` should be a better name

iina/MainWindowController.swift
```diff
@@ -918,6 +957,9 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
     if isOntop {
       setWindowFloatingOnTop(true)
     }
```
lhc70000: These lines doesn't exist on `develop`

PR #571, opened 2017-04-11
Add a function to show playlist when the user dragging files into the window

alejx:

---

**Description:**

Automatically show playlist when dragging after a specific trigger time.


lhc70000:
When user dragged an audio file, let them choose whether playing the file or adding as audio track

Issue #569, opened 2017-04-11
Better handling OSD


lhc70000:
There's an old problem existing from 0.0.1: when to display OSD. As currently an OSD will be displayed once the corresponding mpv property changed, IINA actually displays many unneeded OSD messages, like:

A. **"Resume"** on video start
B. **"Contrast: 0"** etc. on video start (the property actually didn't change)
C. **"Pause"** when started seek by scrolling, and **"Resume"** when finished
D. **"Sub switched"** immediately after "Sub downloaded" (and will overwrite "Sub downloaded")

---

The ideal solution should only display "useful" OSD messages. So here's my proposed solution:

1. **Parse user's `input.conf` and handle it manually.** Fixes situation A and C. Now IINA just _sends mpv key events_ to mpv and let mpv handle the key bindings, which leads to at least following problems:

    - IINA will capture all key events including caps lock, so users may fail to switch input method (#477, #132, etc.)
    - There's no way to check whether user run some commands like screenshot, therefore no OSD for them.
    - User must restart the App to make new key bindings work.
    - There's no way to set key bindings for IINA's native features, like entering PIP.

    These can be solved by checking the _corresponding mpv commands_ in advance, handle them and send them directly to mpv. Additionally by extending mpv's syntax, we may support binding keys to IINA's features like
    ```
    #@iina CTRL+META+p set pip yes
    ```

    Shortcomings: 
    - Unable to handle key sequences (like `a-b-c show-text`). Workarounds are possible though.
    - Untidy solution.

2. **Cache more mpv properties, check whether they actually changed on notification**. Fixes situation B. As far as I know mpv doesn't provide old values in the event data, so we must record them.

3. For situation D (and probably C), I can't think of any elegant solution. What I'm planning to do is:
    - When there's no need to display an OSD, record the OSD type or other info and add it into a "temporally muted OSD" queue
    - Before displaying OSD, check whether it's in the queue; if so, doesn't display the OSD and remove the it from the queue
    - If unused, the added info will be cleared automatically after like 100ms.

@saagarjha @HarukaMa @alejx @inflation please leave your options. Any other suggestions are also welcomed.

alejx:
A question about #477 , why IINA captures caps lock even if it is not the key window?

saagarjha:
@lhc70000 Another possible approach is offering to "import" a user's MPV configuation during onboarding.

PR #572, opened 2017-04-14
Better handling key bindings

lhc70000:
Implementation of #569.


iina/Extensions.swift
```diff
@@ -330,3 +330,8 @@ extension CharacterSet {
     .union(.urlQueryAllowed)
     .union(.urlFragmentAllowed)
 }
+
+
+extension NSMenuItem {
+  static let dummy = NSMenuItem(title: "Dummy", action: nil, keyEquivalent: "")
```
saagarjha: This is a hack IHMO. Just change the method signatures to accept `Any` and pass in `nil`.

iina/KeyBindingDataLoader.swift
```diff
@@ -96,7 +116,7 @@ class KeyBindingDataLoader {
   }
 
   static private func propertiesForMultiply() -> [KeyBindingItem] {
-    return propertyList.filter { $0.1 != .bool }.map { (str, type) -> KeyBindingItem in
+    return propertyList.filter { $0.1 != .bool && $0.1 != .string }.map { (str, type) -> KeyBindingItem in
```
saagarjha: Just use `flatMap` and return `nil` when it's a `.bool` or a `.string`

iina/KeyBindingTranslator.swift
```diff
@@ -114,12 +119,16 @@ class KeyBindingTranslator {
   static func string(fromCriterions criterions: [Criterion]) -> String {
     var mapped = criterions.filter { !$0.isPlaceholder }.map { $0.mpvCommandValue }
 
-    let firstName = (criterions[0] as! TextCriterion).name
+    let firstCriterion = criterions[0] as! TextCriterion
```
saagarjha: `criterions.first` is more ergonomic IMO

iina/KeyBindingTranslator.swift
```diff
@@ -114,12 +119,16 @@ class KeyBindingTranslator {
   static func string(fromCriterions criterions: [Criterion]) -> String {
     var mapped = criterions.filter { !$0.isPlaceholder }.map { $0.mpvCommandValue }
```
saagarjha: The plural of "criterion" is "criteria", FYI

iina/KeyBindingTranslator.swift
```diff
@@ -114,12 +119,16 @@ class KeyBindingTranslator {
...
+    let firstCriterion = criterions[0] as! TextCriterion
+
+    if firstCriterion.isIINACommand {
+      mapped.insert("@iina", at: 0)
+    }
 
     // special cases
 
     /// [add property add|minus value] (length: 4)s
-    if firstName == "add" {
```
saagarjha: I'd use a `switch` here

iina/KeyMapping.swift
```diff
@@ -34,15 +34,37 @@ class KeyMapping {
...
   var key: String
 
   var action: [String]
 
+  private var privateRawAction: String
+
+  var rawAction: String {
+    set {
+      if newValue.hasPrefix("@iina") {
+        privateRawAction = newValue.substring(from: newValue.index(newValue.startIndex, offsetBy: 5)).trimmingCharacters(in: .whitespaces)
```
saagarjha: `5` looks like a magic number here; I'd use `"@iina".count` here to make the relationship explicit

iina/PlayerCore.swift
```diff
@@ -612,6 +632,16 @@ class PlayerCore: NSObject {
     }
   }
 
+  func playbackRestarted() {
+    DispatchQueue.main.async {
+      Timer.scheduledTimer(timeInterval: TimeInterval(0.2), target: self, selector: #selector(self.reEnableOSDAfterFileLoading), userInfo: nil, repeats: false)
```
saagarjha: This looks like a hack; why is it necessary?

iina/PlayerCore.swift
```diff
@@ -698,6 +728,15 @@ class PlayerCore: NSObject {
     // querying `mainWindow.isWindowLoaded` will initialize mainWindow unexpectly
     guard let mw = mainWindow, mw.isWindowLoaded else { return }
 
+    if info.disableOSDForFileLoading {
+      switch osd {
```
saagarjha: What is going on here?

iina/PrefKeyBindingViewController.swift
```diff
@@ -135,8 +135,15 @@ class PrefKeyBindingViewController: NSViewController, MASPreferencesViewControll
   @IBAction func addKeyMappingBtnAction(_ sender: AnyObject) {
     showKeyBindingPanel { key, action in
       guard !key.isEmpty && !action.isEmpty else { return }
-      let splitted = action.characters.split(separator: " ").map { String($0) }
-      currentMapping.append(KeyMapping(key: key, action: splitted))
+      if action.hasPrefix("@iina") {
+        let trimmedAction = action.substring(from: action.index(action.startIndex, offsetBy: 5)).trimmingCharacters(in: .whitespaces)
```
saagarjha: See above

iina/KeyMapping.swift
```diff
@@ -64,19 +86,56 @@ class KeyMapping {
...
+  init(key: String, rawAction: String, isIINACommand: Bool = false, comment: String? = nil) {
     self.key = key
-    self.action = action
+    self.privateRawAction = rawAction
+    self.action = rawAction.components(separatedBy: " ")
+    self.isIINACommand = isIINACommand
     self.comment = comment
   }
 
+  static func parseInputConf(_ path: String) -> [KeyMapping]? {
```
saagarjha: I'd prefer `parseInputConf(atPath path: String)`

iina/MPVController.swift
```diff
@@ -347,6 +338,10 @@ class MPVController: NSObject {
     }
   }
 
+  func command(raw rawString: String) -> Int32 {
```
saagarjha: I'd prefer `command(fromRawString:)`

iina/MainWindowController.swift
```diff
@@ -1848,6 +1879,41 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
...
+    switch cmd {
+    case .openFile:
+      (NSApp.delegate! as! AppDelegate).openFile(self)
+    case .openURL:
+      (NSApp.delegate! as! AppDelegate).openURL(self)
+    case .togglePIP:
+      if #available(OSX 10.12, *) {
+        self.menuTogglePIP(.dummy)
+      }
+    case .videoPanel:
```
saagarjha: See above for discussion of `.dummy`

iina/PlayerCore.swift
lhc70000: mpv will send track switched event and property changed event for all observed properties when started file. There's no way to identify them as far as I know.

iina/Extensions.swift
lhc70000: There would be lots of changes since many of them read data from the `NSMenuItem`. If we want to add support for those methods, we may still have to pass dummy `NSMenuItem`s with `tag` or `representedObject` attached.

iina/KeyBindingDataLoader.swift
lhc70000: I think filtering is not the intended usage of flat map.

iina/KeyBindingTranslator.swift
lhc70000: 🤣

iina/KeyBindingTranslator.swift
lhc70000: I think `first` is designed to be used when we don't know whether the array is empty. Here it must have a first element; and if not, although I can't image such scenario, crash will be a better way to track what happened.

iina/PlayerCore.swift
lhc70000: only allow OSD for "file started" in the first 0.2s.

iina/KeyBindingDataLoader.swift
saagarjha: Well, no, but `flatMap` only requires a single pass.

iina/KeyBindingTranslator.swift
saagarjha: If this is so, then an `assert` at the top of the function would give a better error message than "Index out of range".

iina/PlayerCore.swift
saagarjha: No, I mean that you have a `switch` that `break`s and a `default` that `returns`. Why not use
```swift
guard osd == .fileStart else {
	return
}
```

lhc70000: Because Swift can't handle such enum cases with associated values.

saagarjha: Ahh, didn't see that. How about this:

```swift
guard case .fileStart = osd else {
	return
}
```

lhc70000: I didn't even know this syntax before……

saagarjha: No worries, it's relatively new 😄

PR #590, opened 2017-04-24
Use DAR on file load

HarukaMa:

---

**Description:**
A quick bug fix, reported from telegram group: Videos with anamorphic (non-square) pixels would be played using SAR(Sample Aspect Ratio) on file load instead of DAR(Display Aspect Ratio).

I'm not able to always reproduce it: sometimes it stuck at SAR, sometimes it would jump from SAR to DAR. This commit let the program use DAR on file load.

Even with the commit, player would still use SAR on load for a moment, then resize to DAR size. Breakpoint on file load shows that mpv wouldn't report correct display width on the beginning. Maybe this commit fixes _nothing_. More tests are needed.

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -588,7 +588,7 @@ class PlayerCore: NSObject {
     guard let mw = mainWindow else {
       Utility.fatal("Window is nil at fileLoaded")
     }
-    guard let vwidth = info.videoWidth, let vheight = info.videoHeight else {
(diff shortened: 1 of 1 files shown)


HarukaMa:
Another report about player window size bug with square pixel videos reveals that this part may have deeper problems other that current commit. This part should be checked thoroughly...

lhc70000:
This should work, at least better than current version.

PR #598, opened 2017-04-28
Fixed at exit crashes during playback

vit9696:
**Description:**
You have mistakes in handling thread synchronisation allowing OpenGL context to be destroyed in the middle of the draw procedure. As a result closing the window during video playback will cause crashes.

This change reworks the way uninitLock is used and eliminates the race condition.

### DIFF
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -453,9 +453,6 @@ class MPVController: NSObject {
     case MPV_EVENT_SHUTDOWN:
       let quitByMPV = !playerCore.isMpvTerminated
       if quitByMPV {
-        playerCore.terminateMPV(sendQuit: false)
-        mpv_detach_destroy(mpv)
-        mpv = nil
         NSApp.terminate(nil)
       } else {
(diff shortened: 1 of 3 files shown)


lhc70000:
I'm on vacation and will take a look a few days later. Thanks in advance 👍

alejx:
@vit9696 If I open an audio file, then press q to exit, it will still crash. This can exist before and is not imported by this pr, but do you find any clue to cause this issue?

vit9696:
@alejx ugh, this is a separate issue, but I fixed it anyway.

**Description:**
mpv_opengl_cb_uninit_gl calls OpenGL functions (obviously) and you cannot call OpenGL functions in a thread different from the thread you created your context in. For this reason handleEvent is obliged to only post an event to the main thread to let it quit mpv. 

You are already doing this in applicationShouldTerminate and NSApp.terminate already does what I explained, so the deleted code is just invalid.

alejx:
@vit9696 Thank you very much!

Reference: #2 .

PR #600, opened 2017-04-28
Fix #599

### LINKED ISSUE #599: click to drag does not work
Click to drag only works when the player window already has focus, otherwise it just gets focus.
Whereas MPV does not need focus to be dragged.

**System and IINA version:**

- macOS 10.12.4
- IINA 0.0.7 and current CVS

**Steps to reproduce:**

always, it kind of negates on-top functionality.

saagarjha:
- [X] It implements / fixes issue #599.

---

**Description:**
A quick fix for #599 that makes the first click be usable for dragging the window.

### DIFF
--- a/iina/VideoView.swift
+++ b/iina/VideoView.swift
@@ -83,6 +83,10 @@ class VideoView: NSView {
   override func draw(_ dirtyRect: NSRect) {
     // do nothing
   }
+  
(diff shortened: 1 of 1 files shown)


lhc70000:
You can push branches directly to the main repo.

PR #620, opened 2017-05-04
Full localization

lhc70000:
**Description:**

This PR aims to provide full localization for all UI elements. Important changes:
- Translations for `alert.mpv_error` in all languages need to be updated. Please pay attention to the order of `%@`s!

A translation checker is also added to do automatic checking for missing translations or wrong format strings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,8 +75,6 @@ Pods/
 
 # Carthage
 #
-
-
 # Add this line if you want to avoid checking in source code from Carthage dependencies.
 # Carthage/Checkouts
 
--- a/.gitignore
+++ b/.gitignore
@@ -75,6 +75,8 @@ Pods/
 
 # Carthage
 #
+
+
 # Add this line if you want to avoid checking in source code from Carthage dependencies.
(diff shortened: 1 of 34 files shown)


lhc70000:
de: @shorshe
fr: @superzazu
it: @cgand 
ja: @matsune  
ko: @darkbrow 
pl: -
zh-Hant: @LEOYoon-Tsaw 

Here's the checking result.
[translation_checker_result.txt](https://github.com/lhc70000/iina/files/977154/translation_checker_result.txt)


Please search your language in this file and check for those missing keys.

- For inspector, German and Italian translation seems to be obsoleted (before 0.0.4). Please _delete the string file and create a new one in Xcode_ to start.
- Please pay attention to "Wrong format string" alerts especially. These may lead to crash at runtime!
- For some placeholders like "Title" and "Multiline Label", it's OK to ignore them.
- Some keys were newly added in this PR, so please wait until merged. For up-to-date translations like Japanese and Korean, seems that you only need to add those new keys.

Thanks for your effort and contribution!

PR #654, opened 2017-05-16
Touchbar enhancement

lhc70000:
**Description:**

0. Move touch bar related code to another file
1. Draw custom slider
2. Fix play/pause button not updated
3. Support show remaining time for time label


iina/MainWindowTouchBarSupport.swift
```diff
@@ -0,0 +1,252 @@
...
+      self.touchBarCurrentPosLabel = label
+      item.view = label
+      item.customizationLabel = "Time Position"
+      return item
+
+    case NSTouchBarItemIdentifier.ahead15Sec,
+         NSTouchBarItemIdentifier.back15Sec,
+         NSTouchBarItemIdentifier.ahead30Sec,
+         NSTouchBarItemIdentifier.back30Sec:
+      guard let data = touchBarItemBinding[identifier] else { return nil }
```
saagarjha:
```swift
return touchBarItemBinding[identifier].flatMap { data in
	buttonTouchBarItem(withIdentifier: identifier, imageName: data.0, tag: data.1, customLabel: data.2, action: #selector(touchBarSeekAction))
}
```

lhc70000: I think it adds extra complexity when `guard` is simple and clear enough. You can replace almost every `guard` with a `flatMap` but where's the point?

saagarjha: In this case you're returning `nil` if the value you're testing is `nil`. That's basically what `flatMap` does; if you didn't have to return `nil` or even return at all I'd agree with you. Plus `flatMap` is a line shorter.

lhc70000: `guard` is designed for guarding and nil test. Like the sample in documentation, the most valuable usage of `flatMap` on `Any?` should be producing optional values though complicated logic in order to be embedded in optional chain.

```
let possibleNumber: Int? = Int("42")
let nonOverflowingSquare = possibleNumber.flatMap { x -> Int? in
    let (result, overflowed) = Int.multiplyWithOverflow(x, x)
    return overflowed ? nil : result
}
```

It's totally my personal opinion, but I still prefer `guard` for logical simplicity.

PR #659, opened 2017-05-17
Update version and build number.

### LINKED ISSUE #645: Version number not updated.
The 0.0.9 release still shows version number as 0.0.8. This causes redundant update messages on starting IINA.

kkanungo17:
Fixes #645

- [x] It implements / fixes issue #645 .

---

**Description:**
Update the values in iina/Info.plist

### DIFF
--- a/iina/Info.plist
+++ b/iina/Info.plist
@@ -361,7 +361,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>0.0.8</string>
+	<string>0.0.9</string>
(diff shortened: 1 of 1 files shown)


HarukaMa:
`develop` branch is solely for **development**. Version strings there are not considered relevant at all. All versions are only checked before every *release*.

0.0.9 was pushed in hurry in order to fix an important bug, so we may have missed that in the dev branch. Even though, users who use dev branch product should be able to fix it themselves as the buildbot is down at the moment and the only way is to compile it.

kkanungo17:
I see, not updating the build number was causing unnecessary update checks, so I thought this might help. I'll remove it then.

Issue #672, opened 2017-05-21
[BUG] "Add files in same folder automatically" feature duplicates playlist items


alayliercan:
**System and IINA version:**

- macOS 10.12.5
- IINA 0.0.9

**Expected behavior:**

*Add files in same folder automatically* feature should only add video files into the playlist once unless it's a new video file that doesn't already exist on the playlist.

**Actual behavior:**

Re-ordering items on it duplicates video files on playlist. 

**Steps to reproduce:**

1) Enable *Add files in same folder automatically* feature
2) Open up a video file from a directory that's got more than one video file in it
3) Open up the playlist and move currently playing video file to somewhere else in the list.
4) Start playing another video file by clicking on it on playlist.
5) Now you will notice that the video file that you had moved was added into playlist again.

**How often does this happen?**

Every time you re-order playlist items.

megadr01d:
Maybe related to https://github.com/lhc70000/iina/issues/455.

alayliercan:
I don't use shuffle function though. Have you tried to reproduce it?

lhc70000:
It's going to be fixed in next version. See `auto-load` branch and #673.

alayliercan:
Thanks for letting me know. Cheers.

alejx:
Fixed in develop. See #673 .

Issue #455, opened 2017-03-03
Duplicated items in playlist after pressing shuffle button.


alejx:
**System and IINA version:**

- macOS 10.12.4 beta
- IINA 0.0.6

**Actual behavior:**

See below.

- Crash report:

- MPV log:

**Steps to reproduce:**

Open IINA with some items in the playlist, press shuffle button, then randomly change to other videos in the playlist. In some cases there emerge some duplicated items.

- [x] MPV does not have this problem.

**How often does this happen?**

Every time.

alejx:
After investigating a lot, I found it's not anything about our code, but the behavior of shuffle in the playlist of mpv is very strange. At least I did not find any pattern in it.

I suggest to avoid to use the shuffle in mpv and maintain a 'shuffled' playlist. I'm currently working on it, and will include this change in my next pr.

Issue #140, opened 2017-01-11
Refined Playlist


alejx:
Currently the playlist in IINA is a little bit simple and less developed. I have some ideas to improve it.

- ~~Can drag medias to where it released, but not simply append them to the end of  the playlist.~~
- ~~Can drag medias that are already in the playlist to change their order.~~
- ~~Is able to add subtitles into specific media in the playlist. When a subtitle file is detected to be dragged on the file name in the playlist, turn a highlight on the file name. There should be indicators to show whether a media matched subtitles (including matched online).~~
- ~~A toggle to change the behavior of autoplay next item in the playlist~~.
- ~~A toggle to change the behavior of repeating the playlist~~.
- ~~Save the playlist into a file (with related subtitles)~~.
- An indicator to show whether a file has played before (highlight if hasn't played).

Point out if you think any of my idea is redundant.

PR #673, opened 2017-05-22
Auto load and playlist enhancement

lhc70000:
**Description:**

- Removed `autoload.lua` and use our logic instead
- Match subtitles automatically for videos (more intelligent, search sub directories)
- Add or remove matched subtitles for each playlist item
- Collapse common filename prefixes
- Adjustable playlist/chapter panel width
- Set initial directories when loading audio tracks or subtitles


alejx:
This pull request will fix: #672 , #455 , #140 , #253 .

iina/Extensions.swift
```diff
@@ -163,6 +163,16 @@ extension Array {
   }
 }
 
+extension Dictionary {
+  mutating func safeAppend<T: Equatable>(_ value: T, forKey key: Key) where Value == Array<T> {
+    if self[key] == nil {
+      self[key] = Array<T>()
```
saagarjha: With more thought, I think we should just provide an overloaded subscript function that automatically initializes any key's value when queried:
```swift
extension Dictionary where Value: ExpressibleByArrayLiteral {
	subscript(autogenerating key: Key) -> Value {
		mutating get {
			if let value = self[key] {
				return value
			} else {
				self[key] = []
				return self[autogenerating: key]
			}
		}
	}
}
```
Unfortunately this is blocked by [SR-3547](https://bugs.swift.org/browse/SR-3547), so the above solution is the best we can do for now.

iina/Extensions.swift
```diff
@@ -327,6 +337,20 @@ extension String {
       return self.data(using: .utf8)!.md5
     }
   }
+
+  mutating func deleteLast(_ num: Int) {
+    guard num <= characters.count else { self = ""; return }
+    self = self.substring(to: self.index(endIndex, offsetBy: -num))
+  }
+
+  public func occurancesOf(_ str: String, inRange r: Range<Index>?) -> Int {
```
saagarjha: I'd suggest `countOccurances(of string: String, in range: Range<Index>)`.

iina/FileGroup.swift
```diff
@@ -0,0 +1,167 @@
...
+  var priorityStringOccurances = 0
+
+  var prefix: String {  // prefix detected
+    didSet {
+      suffix = filename.substring(from: filename.index(filename.startIndex, offsetBy: prefix.characters.count))
+    }
+  }
+  var suffix: String  // filename - prefix
+
+  init(_ url: URL) {
```
saagarjha: This is a normal initializer; I don't see any need to elide the label.

iina/PlayerCore.swift
```diff
@@ -642,10 +651,169 @@ class PlayerCore: NSObject {
...
+        subtitles.append(contentsOf: contents.filter { subExts.contains($0.pathExtension) }.map { FileInfo($0) })
+      }
+    }
+
+    // grop video files
+    let series = FileGroup.group(files: groups[.video]!)
+    info.commonPrefixes = series.flatten()
+
+    // group sub files
+    let _ = FileGroup.group(files: subtitles)
```
saagarjha: I believe the `let` is superfluous here

iina/Utility.swift
```diff
@@ -13,6 +13,14 @@ class Utility {
   static let tabTitleFontAttributes = FontAttributes(font: .system, size: .system, align: .center).value
   static let tabTitleActiveFontAttributes = FontAttributes(font: .systemBold, size: .system, align: .center).value
 
+  static let supportedFileExt: [MPVTrack.TrackType: [String]] = {
+    var result: [MPVTrack.TrackType: [String]] = [:]
```
saagarjha: You can do this as a literal in one go; did you mean do it this way?

iina/FileGroup.swift
```diff
@@ -0,0 +1,167 @@
...
+    self.path = url.path
+    self.ext = url.pathExtension
+    self.filename = url.deletingPathExtension().lastPathComponent
+    self.characters = [Character](self.filename.characters)
+    self.prefix = ""
+    self.suffix = self.filename
+    self.getSegments()
+  }
+
+  private func getSegments() {
```
saagarjha: Sorry, but what does this function do?

iina/FileGroup.swift
```diff
@@ -0,0 +1,167 @@
...
+      for finfo in contents {
+        if i >= finfo.characters.count {
+          shouldContinue = false
+          continue
+        }
+        let c = finfo.characters[i]
+        var p = prefix
+        p.append(c)
+        lastPrefix = p
+        if tempGroup[p] == nil {
```
saagarjha: Why not use `safeAppend` here?

iina/ObjcUtils.m
```diff
@@ -9,26 +9,70 @@
...
+  @try {
+    tryBlock();
+    return YES;
+  }
+  @catch (NSException *exception) {
+    // do nothing
   }
+}
+
++ (NSUInteger)levDistance:(NSString *)str0 and:(NSString *)str1 {
```
saagarjha: Any reason why this is written in Objective-C?

saagarjha:
There appears to be some issues with localization:

screen shot 2017-05-29 at 12 56 30

I'm not sure if this was introduced in this pull request, but it's something to keep in mind.

iina/FileGroup.swift
```diff
@@ -0,0 +1,167 @@
...
+  var path: String
+  var filename: String
+  var ext: String
+  var characters: [Character]
+  var dist: [FileInfo: UInt] = [:]
+  var minDist: [FileInfo] = []
+  var segments: [String] = []
+  var priorityStringOccurances = 0
+
+  var prefix: String {  // prefix detected
```
saagarjha: I'd move the comment down to the `didSet` line to make it more clear

iina/PlayerCore.swift
```diff
@@ -642,10 +651,169 @@ class PlayerCore: NSObject {
...
+    var groups: [MPVTrack.TrackType: [FileInfo]] = [.video: [], .audio: [], .sub: []]
+    let allTypes: [MPVTrack.TrackType] = [.video, .audio, .sub]
+    for file in files {
+      let fileInfo = FileInfo(file)
+      guard let mediaType = allTypes.first(where: { Utility.supportedFileExt[$0]!.contains(fileInfo.ext) }) else { continue }
+      groups[mediaType]!.append(fileInfo)
+    }
+
+    // natural sort
+    groups[.video]!.sort { file1, file2 -> Bool in
```
saagarjha: I don't think the return type is required here

iina/PlayerCore.swift
```diff
@@ -642,10 +651,169 @@ class PlayerCore: NSObject {
...
+      return file1.segments.lexicographicallyPrecedes(file2.segments) { a, b in
```
saagarjha: Looks like a job for `localizedCompare(_:)`.

iina/PlayerCore.swift
```diff
@@ -642,10 +651,169 @@ class PlayerCore: NSObject {
...
+          return a < b
+        }
+      }
+    }
+
+    // get all possible sub files
+    var subtitles = groups[.sub]!
+    for subDir in subDirs {
+      if let contents = try? fm.contentsOfDirectory(at: subDir, includingPropertiesForKeys: nil, options: searchOptions) {
+        subtitles.append(contentsOf: contents.filter { subExts.contains($0.pathExtension) }.map { FileInfo($0) })
```
saagarjha: `flatMap` will do this in a single passthrough, instead of the two that `filter` and `map` require.

iina/PlayerCore.swift
```diff
@@ -642,10 +651,169 @@ class PlayerCore: NSObject {
...
+        let minDistToSub = video.dist.reduce(UInt.max, { min($0.0, $0.1.value) })
+        subtitles
+          .filter { video.dist[$0]! == minDistToSub && $0.minDist.contains(video) }
+          .forEach { info.matchedSubs.safeAppend($0.url, forKey: video.path); matchedSubs.insert($0) }
+      }
+      // add subs that contains video name
+      if subAutoLoadOption.shouldLoadSubsContainingVideoName() {
+        subtitles
+          .filter { $0.filename.contains(video.filename) }
+          .forEach { info.matchedSubs.safeAppend($0.url, forKey: video.path); matchedSubs.insert($0) }
```
saagarjha: Put this on two lines, please

iina/PlayerCore.swift
```diff
@@ -642,10 +651,169 @@ class PlayerCore: NSObject {
     }
   }
 
-  @objc private func reEnableOSDAfterFileLoading() {
+  @objc
+  private func reEnableOSDAfterFileLoading() {
     info.disableOSDForFileLoading = false
   }
 
+  /// Add files in the same folder to playlist.
```
saagarjha: Just a question: have you tried running this on a large folder? This function looks like it might not be fast enough if there are too many videos and subtitles.

iina/FileGroup.swift
lhc70000: We also want to append to `currChars` when `p` is a new key.

iina/ObjcUtils.m
lhc70000: Objective-C (C) is faster, and more convenient when initializing and manipulating fixed 2D arrays like in this algorithm.

lhc70000:
> There appears to be some issues with localization:

I can't reproduce in current `develop` branch. Which localization are you using?

saagarjha:
English?

iina/PlayerCore.swift
lhc70000: `localizedCompare(_:)` can't handle numbers "naturally", e.g.

```swift
"abc100de".localizedCompare("abc88de")
// =>  .orderedAscending
```

saagarjha: Oops, I meant `localizedStandardCompare(_:)`.

lhc70000: Looks good; didn't discover that 😂

iina/FileGroup.swift
saagarjha: Ahh, got it.

iina/PlayerCore.swift
saagarjha: Oops, I meant `localizedStandardCompare(_:)`.

iina/Extensions.swift
lhc70000: I'm not sure whether it would work. `var v = self[key]` will copy `self[key]` to a new array.

```swift
var dict: [String: [String]] = ["foo": ["a"]]

if var v = dict["foo"] {
  v.append("b")
  print(v.count)
  print(dict["foo"]!.count)
}

print(dict)

// 2
// 1
// ["foo": ["a"]]
```

iina/Utility.swift
lhc70000: Swift used to have performance issue when compiling complicated dict literals if I remember correctly. I'll switch to literal since it's not a problem nowadays.

iina/Extensions.swift
saagarjha: Ah, I knew I was forgetting something. Here's some slightly uglier code:
```swift
if self[key] != nil {
	self[key]?.append(value)
} else {
	self[key] = [value]
}
```

iina/MainWindowController.swift
```diff
@@ -17,6 +17,8 @@ fileprivate let TitleBarHeightWithOSCInFullScreen: CGFloat = 24 + 10
 fileprivate let OSCTopMainViewMarginTop: CGFloat = 26
 fileprivate let OSCTopMainViewMarginTopInFullScreen: CGFloat = 6
 
+fileprivate let PlaylistMinWidth: CGFloat = 240
+fileprivate let PlaylistMaxWidth: CGFloat = 400
```
stevetranby, line 21: I'm not sure why the max value is set to 400? Is there a concern users will make it too wide, and if so what is the issue if say the maximum width were set to 80% of window width?

Ironically, if I shrink the window the max width allows me to show the playlist covering up to 90% of the window.
(note: my default viewing size is roughly 1600x800 within a 2560x1440)

Is there a reason the max width is kept to such a small amount?

I'll probably just fork the project and make use my own build version with a larger max width.

PR #696, opened 2017-06-01
Auto load 2

lhc70000:
**Description:**

Improved auto loading and subtitle matching algorithm. Includes updates after @saagarjha's review on last PR.

### DIFF
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -164,7 +164,7 @@ extension Array {
 }
 
 extension Dictionary {
-  mutating func safeAppend<T: Equatable>(_ value: T, forKey key: Key) where Value == Array<T> {
+  mutating func safeAppend<T: Equatable>(_ value: T, for key: Key) where Value == Array<T> {
     if self[key] == nil {
       self[key] = Array<T>()
     }
@@ -343,10 +343,10 @@ extension String {
     self = self.substring(to: self.index(endIndex, offsetBy: -num))
   }
 
-  public func occurancesOf(_ str: String, inRange r: Range<Index>?) -> Int {
+  public func countOccurances(of str: String, inRange r: Range<Index>?) -> Int {
     if let range = range(of: str, options: [], range: r, locale: nil) {
       let nextRange = range.upperBound..<self.endIndex
-      return 1 + occurancesOf(str, inRange: nextRange)
+      return 1 + countOccurances(of: str, inRange: nextRange)
     } else {
       return 0
     }
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -343,10 +343,10 @@ extension String {
     self = self.substring(to: self.index(endIndex, offsetBy: -num))
   }
 
-  public func countOccurances(of str: String, inRange r: Range<Index>?) -> Int {
-    if let range = range(of: str, options: [], range: r, locale: nil) {
-      let nextRange = range.upperBound..<self.endIndex
-      return 1 + countOccurances(of: str, inRange: nextRange)
+  func countOccurances(of str: String, in range: Range<Index>?) -> Int {
+    if let firstRange = self.range(of: str, options: [], range: range, locale: nil) {
+      let nextRange = firstRange.upperBound..<self.endIndex
+      return 1 + countOccurances(of: str, in: nextRange)
     } else {
       return 0
     }
--- a/iina/FileGroup.swift
+++ b/iina/FileGroup.swift
@@ -131,7 +131,7 @@ class FileGroup {
     if stopGrouping(currChars) || maxSubGroupCount < 3 {
       contents.forEach { $0.prefix = self.prefix }
     } else {
-      groups = tempGroup.flatMap { FileGroup(prefix: $0, contents: $1) }
+      groups = tempGroup.map { FileGroup(prefix: $0, contents: $1) }
       // continue
       for g in groups {
         g.tryGroupFiles()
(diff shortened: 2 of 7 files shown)


iina/Extensions.swift
```diff
@@ -343,10 +343,10 @@ extension String {
     self = self.substring(to: self.index(endIndex, offsetBy: -num))
   }
 
-  public func occurancesOf(_ str: String, inRange r: Range<Index>?) -> Int {
+  public func countOccurances(of str: String, inRange r: Range<Index>?) -> Int {
```
saagarjha: I see you've improved the signature, but again, I'd prefer `countOccurances(of string: String, in range: Range<Index>?)`. There are a couple reasons:

1. Parameter names are "public-facing"; they should be reasonably "nice". I'd avoid abbreviations and truncations since they serve can serve as pseudo-documentation. In this case, where you're already using `range`, I'd prefer that you used shorter names *inside* the function rather than in the signature.
2. We can drop range from `inRange` since type information is strong and readily available. Besides, there's not a whole lot of things that can go after an `in:`.

iina/FileGroup.swift
```diff
@@ -11,21 +11,31 @@ import Foundation
...
+  var isMatched = false
 
-  var prefix: String {  // prefix detected
+  var prefix: String {  // prefix detected by FileGroup
     didSet {
-      suffix = filename.substring(from: filename.index(filename.startIndex, offsetBy: prefix.characters.count))
+      if prefix.characters.count < self.characters.count {
+        suffix = filename.substring(from: filename.index(filename.startIndex, offsetBy: prefix.characters.count))
+        getNameInSeries()
+      } else {
```
saagarjha: A `prefix` longer than the file's actual name is a programmer error, right? I'd add an assertion here as well.

iina/FileGroup.swift
```diff
@@ -122,19 +127,20 @@ class FileGroup {
       i += 1
     }
 
-    if stopGrouping(currChars) {
+    let maxSubGroupCount = tempGroup.reduce(0, { max($0, $1.value.count) })
+    if stopGrouping(currChars) || maxSubGroupCount < 3 {
       contents.forEach { $0.prefix = self.prefix }
     } else {
-      groups = tempGroup.map { FileGroup(prefix: $0, contents: $1) }
+      groups = tempGroup.flatMap { FileGroup(prefix: $0, contents: $1) }
```
saagarjha: 😕 Any particular reason to switch to `flatMap` here? I don't see any `Optional`s.

iina/PlayerCore.swift
```diff
@@ -579,17 +580,19 @@ class PlayerCore: NSObject {
...
     info.currentURL = URL(fileURLWithPath: path)
-    // add files in same folder
-    if ud.bool(forKey: Preference.Key.playlistAutoAdd) {
-      autoLoadFilesInCurrentFolder()
-    }
-    // auto load matched subtitles
-    if let matchedSubs = info.matchedSubs[path] {
-      for sub in matchedSubs {
-        loadExternalSubFile(sub)
+    backgroundQueue.async {
```
saagarjha: Well, that's one way to do it…

iina/FileGroup.swift
lhc70000: Should be a typo.

iina/PlayerCore.swift
lhc70000: Any other suggestion? Actually with the improved algorithm it's fast enough for normal situations. Except for some extreme cases, like there're 1000 videos with completely different name and 1000 subs whose names don't contain any video name, (although I think it won't happen,) the algorithm takes about 15s.

iina/FileGroup.swift
lhc70000: The current algorithm is still unstable; I'll remove this after enough test.

iina/PlayerCore.swift
saagarjha: There probably is a better way to do it, but then I'd have to actually think about writing efficient code 😅 . If it's possible, we could try to have the results show up as they're found, if it's easy to do; otherwise just leave it. We'll wait and see how this responds in real-world scenarios.

saagarjha: So I had a chance to try this algorithm, and it's falling apart in my real-world usage. Basically, I have a video in my downloads folder, which has slightly over 1500 items. If I play it, IINA will idle at 100% CPU as it goes through this method. I can attach a stackshot if you'd like.

lhc70000: How many subtitles are there in the folder? It will only run brute force match on videos and subs if they can't be matched by (detecting and matching series) and (sub's filename containing video's).

saagarjha: I have no idea; nothing is *explicitly* a subtitle. All I know is that moving it to my desktop, which has around 10 items, "fixes" the issue.

lhc70000: That's weird. It shouldn't even run this algorithm if there's no subtitles. By default it will search all sub folders, if you can confirm that there's still no subtitle, I'll double check the algorithm.

saagarjha: I don't think there are any. However, I guess a recursive search takes time.

lhc70000: It only takes considerable time when calculating edit distance between filenames. The complexity is like `O(mn)`, so e.g. if you have 1000 videos and 1 sub, it will calculate 1000 times, but if there are 1000 videos and subs respectively, it will run 1,000,000 times (which takes 15s on my mbp). I'll check and see if I can reproduce this problem.

PR #711, opened 2017-06-05
Auto load 3

lhc70000:
**Description:**

- Fixed several crashes introduced by auto load feature
- Improved thread safety
- Improved precision

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -18,6 +18,8 @@ class AppDelegate: NSObject, NSApplicationDelegate {
 
   var pendingURL: String?
 
+  private var lastOpenFileTimestamp: Double?
+
   lazy var playerCore: PlayerCore = PlayerCore.shared
 
   lazy var aboutWindow: AboutWindowController = AboutWindowController()
@@ -135,6 +137,13 @@ class AppDelegate: NSObject, NSApplicationDelegate {
   }
 
   func application(_ sender: NSApplication, openFile filename: String) -> Bool {
+    // When dragging multiple files to IINA icon, cocoa will simply call this method repeatedly.
+    // IINA (mpv) can't handle opening multiple files correctly, so I have to guard it here.
+    // It's a temperory solution, and the min time interval 0.3 might also be too arbitrary.
+    let c = CFAbsoluteTimeGetCurrent()
+    if let t = lastOpenFileTimestamp, c - t < 0.3 { return false }
+    lastOpenFileTimestamp = c
+
     if !isReady {
       UserDefaults.standard.register(defaults: Preference.defaultPreference)
       playerCore.startMPV()
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -13,7 +13,21 @@ class PlayerCore: NSObject {
   static let shared = PlayerCore()
 
   unowned let ud: UserDefaults = UserDefaults.standard
+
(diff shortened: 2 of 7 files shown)


iina/AppDelegate.swift
```diff
@@ -135,6 +137,13 @@ class AppDelegate: NSObject, NSApplicationDelegate {
   }
 
   func application(_ sender: NSApplication, openFile filename: String) -> Bool {
+    // When dragging multiple files to IINA icon, cocoa will simply call this method repeatedly.
+    // IINA (mpv) can't handle opening multiple files correctly, so I have to guard it here.
+    // It's a temperory solution, and the min time interval 0.3 might also be too arbitrary.
+    let c = CFAbsoluteTimeGetCurrent()
```
saagarjha: I'm not sure if we should be dipping down into CoreFoundation; is `Date` not enough?

iina/PlayerCore.swift
```diff
@@ -582,17 +598,22 @@ class PlayerCore: NSObject {
     info.disableOSDForFileLoading = true
     guard let path = mpvController.getString(MPVProperty.path) else { return }
     info.currentURL = URL(fileURLWithPath: path)
+    backgroundQueueTicket += 1
```
saagarjha: Ticket incrementing and checking should be done on the same queue

iina/PlayerCore.swift
```diff
@@ -661,13 +682,25 @@ class PlayerCore: NSObject {
...
+   - For unmatched videos and subs, perform fuzzy (but slow, O(n^2)) match for them.
+
+   **Remark**: 
+   
+   This method is expected to be executed in `backgroundQueue` (see `backgroundQueueTicket`).
+   Therefore accesses to `self.info` and mpv playlist must be guarded.
+   */
+  private func autoLoadFilesInCurrentFolder(ticket: Int) {
+    let info: (() -> PlaybackInfo?) =  {
+      guard ticket == self.backgroundQueueTicket else { return nil }
```
saagarjha: `self.backgroundQueueTicket != nil ? self.info : nil`?

iina/PlayerCore.swift
```diff
@@ -727,26 +760,35 @@ class PlayerCore: NSObject {
...
+        guard ticket == self.backgroundQueueTicket else { return }
         addToPlaylist(video.path)
       } else {
         let count = mpvController.getInt(MPVProperty.playlistCount)
         let current = mpvController.getInt(MPVProperty.playlistPos)
+        guard ticket == self.backgroundQueueTicket else { return }
         addToPlaylist(video.path)
-        playlistMove(count, to: current)
+        mpvController.command(.playlistMove, args: ["\(count)", "\(current)"], checkError: false) { err in
+          if err == -12 { needQuit = true }
```
saagarjha: What is error -12? Please make an enum or something for this.

iina/AppDelegate.swift
lhc70000: I just want a quick way to get a timestamp, rather than creating a powerful but useless `Date` object.

iina/PlayerCore.swift
lhc70000: Why? It's designed to interrupt a running block. If we post this to the same serialized queue, it can't affect any running tasks. Accesses to this property should be synchronized though.

lhc70000:
Merged for 0.0.11 release. I'll push to `develop` for any further update.

PR #713, opened 2017-06-05
Better compatibility for video filters and hardware decoding

lhc70000:
**Description:**

- Add HwDec `auto-copy` option.
- Ask user to switch to `auto-copy` or `no` before applying video filters.

### DIFF
--- a/iina/Preference.swift
+++ b/iina/Preference.swift
@@ -96,11 +96,9 @@ struct Preference {
     // Codec
 
     static let videoThreads = "videoThreads"
-
-    static let useHardwareDecoding = "useHardwareDecoding"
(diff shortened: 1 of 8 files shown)


iina/Preference.swift
```diff
@@ -372,6 +370,24 @@ struct Preference {
     }
   }
 
+  enum HardwareDecoderOption: Int {
```
saagarjha: Why not just make the enum backed by `String`? I don't see any place that the actual integer value is used.

lhc70000: It's bound to `selectedTag` of the pop up button. The same for all similar enums in `Preference`.

Issue #721, opened 2017-06-07
Adding multiple files to playlist


anonmonogatari:
**System and IINA version:**

- macOS 10.12.5
- IINA 0.0.11

**Expected behavior:**
After pressed '+'  button to add files to playlist, we should be able to add multiple files to play list.

**Actual behavior:**
You can't choose multiple files. Selection window only allows choosing one 1 file.

PR #739, opened 2017-06-10
Added option to add multiple files to playlist

dejanskledar:
- [x] It implements / fixes issue #721.

---

**Description:**
Added option to add multiple files to playlist

### DIFF
--- a/iina/PlaylistViewController.swift
+++ b/iina/PlaylistViewController.swift
@@ -307,15 +307,16 @@ class PlaylistViewController: NSViewController, NSTableViewDataSource, NSTableVi
   // MARK: - IBActions
 
   @IBAction func addToPlaylistBtnAction(_ sender: AnyObject) {
-    Utility.quickOpenPanel(title: "Add to playlist", isDir: false) { (url) in
-      if url.isFileURL {
-        self.playerCore.addToPlaylist(url.path)
-        self.reloadData(playlist: true, chapters: false)
-        self.mainWindow.displayOSD(.addToPlaylist(1))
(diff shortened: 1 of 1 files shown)


iina/PlaylistViewController.swift
```diff
@@ -307,16 +307,17 @@ class PlaylistViewController: NSViewController, NSTableViewDataSource, NSTableVi
...
-    Utility.quickOpenPanel(title: "Add to playlist", isDir: false) { (url) in
-      if url.isFileURL {
-        self.playerCore.addToPlaylist(url.path)
-        self.reloadData(playlist: true, chapters: false)
-        self.mainWindow.displayOSD(.addToPlaylist(1))
+    Utility.quickMultipleOpenPanel(title: "Add to playlist") { urls in
+      for url in urls {
+        if url.isFileURL {
+          self.playerCore.addToPlaylist(url.path)
+          self.mainWindow.displayOSD(.addToPlaylist(1))
```
lhc70000: Please use `PlayerCore#sendOSD` to post OSD safely onto main thread. 

Also, why show OSD for each added file rather than for all of them? 

```swift

for url in urls {
  ...
}
playerCore.sendOSD(.addToPlaylist(urls.count))

```

alejx: - Use `playerCore.sendOSD` instead.
- Count valid url in urls, not "1".

dejanskledar: I see. I had trouble understanding what the `displayOSD` is for and the meaning of the `Int` parameter in the `addToPlaylist` enum.

PR #752, opened 2017-06-14
Animation fix

alejx:

---

**Description:**

This pull request refactors `hideUI()` and `showUI()` in MainWindowController. In principle, I made those two functions symmetrical.

```diff
     fadeableViews.forEach { (v) in
-      v.alphaValue = 1
       v.isHidden = false
     }
```
`alphaValue` is managed by the animator. Change it manually is useless and somewhat invalid.
```diff
+    standardWindowButtons.forEach { $0.isEnabled = false }
```
Should be disabled.
```diff
-      self.animationState = .shown
+      // if no interrupt then hide animation
+      if self.animationState == .willShow {
+        self.animationState = .shown
+      }
```
Like `hideUI()`, `showUI()` should also be able to be interrupted by `hideUI()`.

```diff
-    if isMouseInWindow && animationState == .hidden {
+    if isMouseInWindow {
       showUI()
```
Whatever `animationState` is, we should call `showUI()` directly. `showUI()` can interrupt `hideUI()` effectively.

```diff
+    if mouseExitEnterCount >= 2 && action == .hideOSC {
+      // the counter being greater than or equal to 2 means that the mouse re-entered the window
+      // showUI() must be called due to the movement in the window, thus hideOSC action should be cancelled
+      return
+    }
```
These code fixed a bug. You can reproduce it by steps below:
- Make sure `NSEvent.doubleClickInterval()` have some value.
- Make single click action to be "Hide OSC".
- Click once in the window.
- Move the mouse out and in the window then stop there (quickly!).
- OSC should be shown due to the mouse movement in the window, however "Hide OSC" is activated after show OSC. OSC is hidden.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -144,6 +144,8 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
   }
 
   var wasPlayingWhenSeekBegan: Bool?
+  
+  var mouseExitEnterCount = 0
 
   // MARK: - Enums
 
@@ -770,6 +772,7 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
           } else {
             // else start a timer
             singleClickTimer = Timer.scheduledTimer(timeInterval: NSEvent.doubleClickInterval(), target: self, selector: #selector(self.performMouseActionLater(_:)), userInfo: singleClickAction, repeats: false)
+            mouseExitEnterCount = 0
           }
         } else if event.clickCount == 2 {
           // double click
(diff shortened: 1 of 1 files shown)


iina/MainWindowController.swift
```diff
@@ -20,6 +20,11 @@ fileprivate let OSCTopMainViewMarginTopInFullScreen: CGFloat = 6
 fileprivate let PlaylistMinWidth: CGFloat = 240
 fileprivate let PlaylistMaxWidth: CGFloat = 400
 
+fileprivate let UIAnimationDuration: TimeInterval = 0.25
```
saagarjha: `TimeInterval` is a typealias for `Double`, so you don't need to explicitly specify the type.

saagarjha:
Here's a bug:

1. Click and hold a button.
2. Move mouse off of the controller.
3. Let go of the mouse.
4. Controller fails to hide.

alejx:
@saagarjha Currently we update the timer for showing OSC only in overrode `mouseMoved(with event: NSEvent)`, however clicking on a button and dragging out do not call that function. Any good idea?

PR #753, opened 2017-06-14
Thumbnail preview

lhc70000:
**Description:**

screen shot 2017-06-14 at 11 15 09 pm

#### How do we get thumbnails?

mpv itself doesn't provide practical way to get thumbnails (unless launching another mpv instance in background, seek to different time points and draw to some pixel buffer, etc).

However since mpv is based on ffmpeg, we have all ffmpeg libraries available in iina (libav*, libav for short), and we can link to them simply by adding libav's headers, and use ffmpeg directly bypassing mpv.

`FFmpegController` is written in Objective-C to deal with libav's C API which involves too much pointer manipulations. 

#### What's changed?

120+/158 files in the diff are libav headers. We should include them in the repository because they are specific to the ffmpeg version that libmpv linked to and are essential for building.

I also renamed `libmpv/` to `deps/` so that it doesn't only include libmpv, plus github won't count `deps/` folder in language statistics.

#### How does this system work?

On `file-start`, `playerCore` will check whether there're cached thumbnails.

1.  Cache exists:
  - Load thumbnails from cache in its `backgroundQueue`
  - Set `info.thumbnailsReady` to true
2. Cache doesn't exist:
  - Request thumbnails from `FFmpegController`
  - `FFmpegController` will add the following operation into a `NSOperationQueue`, and cancel all waiting tasks (of course, the running task will keep running, if it exists)
    - `FFmpegController` get the duration of the video stream
    - Set an interval of `duration/100`, and seek to those 101 time points (including start and end), get the closest keyframe. The number 100 can be configured by changing `FFmpegController`'s `thumbnailCount`.
    - Since keyframes are not accurate, also get the timestamp of this frame. Save them both to a `FFthumbnail`.
    - If there's already a frame saved with same timestamp, skip (e.g. the video has < 100 key frames)
    - If there's any error during thumbnail generation, do nothing
  - Notify `playerCore` about the result and thumbnail list
  - `playerCore` set `info.thumbnailsReady` to true
  - `playerCore` save thumbnails in its `backgroundQueue`

In the UI, every time when the `timeLabel` is needed to be updated, also check `info.thumbnailsReady`. If ready, display the closest thumbnail before the current time (pointed by pointer).

#### How does it cache the thumbnails?

All thumbnails for a video, regardless of the actual count, are saved in a single binary file, by `ThumbnailCache`. Here's the structure of this binary file:

|Offset|Length|Content|
|---|---|---|
|0|8|Version number or other info (not used for now)|
|9|~EOF|Thumbnail data blocks|

For each data block:

|Offset|Length|Content|
|---|---|---|
|0|8|Length of the block (`len`)|
|9|8|Timestamp of the thumbnail, in double|
|17|`len`-8|JPEG data of the thumbnail|

The filename of the cache file is the same as mpv's watch_later file, i.e. basically a md5 of the video path.


lhc70000:
For anyone, please review if you'd like.

iina/FFmpegController.h
```diff
@@ -0,0 +1,37 @@
...
+
+@end
+
+
+@protocol FFmpegControllerDelegate <NSObject>
+
+/** 
+ Did generated thumbnails for the video.
+ */
+- (void)didGeneratedThumbnails:(NSArray<FFThumbnail *> *)thumbnails withSuccess:(BOOL)success;
```
saagarjha: `didGenerateThumbnails:succeeded:`

iina/FFmpegController.m
```diff
@@ -0,0 +1,266 @@
...
+
+
+@interface FFmpegController () {
+  NSMutableArray<FFThumbnail *> *_thumbnails;
+  NSMutableSet *_addedTimestamps;
+  NSOperationQueue *_queue;
+}
+
+- (int)getPeeksForFile:(NSString *)file;
+- (void)sameThumbnail:(AVFrame *)pFrame width:(int)width height:(int)height index:(int)index realTime:(int)second;
```
saagarjha: Did you mean "save"?

iina/FFmpegController.m
```diff
@@ -0,0 +1,266 @@
...
+    if (self.delegate) {
+      [self.delegate didGeneratedThumbnails:[NSArray arrayWithArray:_thumbnails] withSuccess:(success < 0 ? NO : YES)];
+    }
+  }];
+}
+
+
+- (int)getPeeksForFile:(NSString *)file
+{
+  int i, ret;
```
saagarjha: You don't really need `ret`.

iina/FFmpegController.m
```diff
@@ -0,0 +1,266 @@
...
+        ret = avcodec_receive_frame(pCodecCtx, pFrame);
+        if (ret < 0) {  // something happened
+          if (ret == AVERROR(EAGAIN))  // input not ready, retry
+            continue;
+          else
+            break;
+        }
+
+        // Check if duplicated
+        NSNumber *currentTimeStamp = [[NSNumber alloc] initWithLongLong:pFrame->best_effort_timestamp];
```
saagarjha: `@(pFrame->best_effort_timestamp)` is shorter

iina/FFmpegController.m
```diff
@@ -0,0 +1,266 @@
...
+                        pFrameRGB->data,
+                        pFrameRGB->linesize);
+        CHECK_SUCCESS(ret, @"Cannot convert frame")
+
+        // Save the frame to disk
+        [self sameThumbnail:pFrameRGB
+                      width:pFrameRGB->width
+                     height:pFrameRGB->height
+                      index:i
+                   realTime:(pFrame->best_effort_timestamp*timebaseDouble)];
```
saagarjha: Add spacing around the operator here please.

iina/MainWindowController.swift
```diff
@@ -1460,16 +1467,39 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
   }
 
   /** Display time label when mouse over slider */
-  private func updateTimeLabel(_ mouseXPos: CGFloat) {
+  private func updateTimeLabel(_ mouseXPos: CGFloat, originalPos: NSPoint) {
     let timeLabelXPos = playSlider.frame.origin.y + 15
-    timePreviewWhenSeek.frame.origin = CGPoint(x: round(mouseXPos + playSlider.frame.origin.x - timePreviewWhenSeek.frame.width / 2), y: timeLabelXPos + 1)
```
saagarjha: `NSPoint` is equivalent to `CGPoint`, is it not?

iina/ThumbnailCache.swift
```diff
@@ -0,0 +1,112 @@
...
+//  Created by lhc on 14/6/2017.
+//  Copyright © 2017 lhc. All rights reserved.
+//
+
+import Cocoa
+
+class ThumbnailCache {
+
+  static private let sizeofDouble = MemoryLayout<Double>.size
+  static private let sizeofInt64 = MemoryLayout<Int64>.size
```
saagarjha: Why `Int64` specifically? Why not just `Int`?

iina/ThumbnailCache.swift
```diff
@@ -0,0 +1,112 @@
...
+    // version
+    var version = Int64(1)
+    let versionData = Data(bytes: &version, count: sizeofInt64)
+    file.write(versionData)
+
+    // data blocks
+    for tb in thumbnails {
+      let timestampData = Data(bytes: &tb.realTime, count: sizeofDouble)
+      guard let tiffData = tb.image.tiffRepresentation else {
+        Utility.log("Cannot geenerate tiff data.")
```
saagarjha: generate

iina/ThumbnailCache.swift
```diff
@@ -0,0 +1,112 @@
...
+        return
+      }
+      guard let jpegData = NSBitmapImageRep(data: tiffData)?.representation(using: .JPEG, properties: imageProperties) else {
+        Utility.log("Cannot geenerate jpeg data.")
```
saagarjha: ditto

iina/ThumbnailCache.swift
```diff
@@ -0,0 +1,112 @@
...
+      file.write(timestampData)
+      file.write(jpegData)
+    }
+
+    Utility.log("Finished writing thumbnail cache.")
+  }
+
+  /// Read thumbnail cache to file.
+  /// This method is expected to be called when the file exists.
+  static func read(from name: String) -> [FFThumbnail]? {
```
saagarjha: `readFromFile(atPath path: String)`

iina/ThumbnailCache.swift
```diff
@@ -0,0 +1,112 @@
...
+    NSImageCompressionFactor: 0.75
+  ]
+
+  static func exists(name: String) -> Bool {
+    return FileManager.default.fileExists(atPath: urlFor(name).path)
+  }
+
+  /// Write thumbnail cache to file. 
+  /// This method is expected to be called when the file doesn't exist.
+  static func write(thumbnails: [FFThumbnail], to name: String) {
```
saagarjha: `write(_ thumbnails: [FFThumbnail], toFileAtPath path: String)`

iina/ThumbnailCache.swift
```diff
@@ -0,0 +1,112 @@
...
+      tb.realTime = timestamp
+      tb.image = image
+      result.append(tb)
+    }
+
+    Utility.log("Finished reading thumbnail cache...")
+    return result
+  }
+
+  static private func urlFor(_ name: String) -> URL {
```
saagarjha: `urlForFile(atPath path: String)`

iina/ThumbnailCache.swift
```diff
@@ -0,0 +1,112 @@
...
+class ThumbnailCache {
+
+  static private let sizeofDouble = MemoryLayout<Double>.size
+  static private let sizeofInt64 = MemoryLayout<Int64>.size
+
+  static private let imageProperties: [String: Any] = [
+    NSImageCompressionFactor: 0.75
+  ]
+
+  static func exists(name: String) -> Bool {
```
saagarjha: `fileExists(atPath path: String)`

saagarjha:
Could you get this to work while dragging the scrubber as well?

lhc70000:
@inflation said he will work on Touch Bar later.

saagarjha:
Sure, no problem. That would be part of a separate pull request anyways.

iina/Utility.swift
```diff
@@ -343,6 +343,12 @@ class Utility {
     return url
   }()
 
+  static let thumbnailCacheURL: URL = {
+    let url = Utility.appSupportDirUrl.appendingPathComponent(AppData.thumbnailCacheFolder, isDirectory: true)
```
lhc70000: Generating thumbnails can be costly, like 10-20s for large videos, so we may want to save them permanently. Of course I think we should also provide a button in Preferences to clear all caches.

iina/FFmpegController.m
lhc70000: Why?

iina/MainWindowController.swift
lhc70000: I just want to stay consistent to `NSPoint` when nothing is related to Core Graphics and more `CG` operations.

iina/ThumbnailCache.swift
lhc70000: Technically `Int` could be `Int32`(although it's not possible in IINA) or `Int64`, so just to make sure make it's exactly 4 bytes.

iina/ThumbnailCache.swift
lhc70000: Actually it accepts a name (md5 value), not a path. I'll improve them according to your suggestions though.

iina/Utility.swift
saagarjha: Alright, sounds good.

iina/ThumbnailCache.swift
saagarjha: Yes, but why does it matter? Are we doing anything that requires 4 bytes?

iina/ThumbnailCache.swift
saagarjha: Oh, right. Got it.

iina/FFmpegController.m
saagarjha: Can't you just pass the result into `CHECK_SUCCESS`?

```objective-c
CHECK_SUCCESS(avformat_open_input(&pFormatCtx, cFilename, NULL, NULL), @"Cannot open video")
```

lhc70000: Oh, I just don't want to write them together :stuck_out_tongue: (especially for function calls that spans multiple lines).

iina/ThumbnailCache.swift
lhc70000: When designing a binary format, we always want a fixed file structure with each byte of data predictable, for maximum compatibility.

saagarjha: Got it.

PR #775, opened 2017-06-26
Multiple windows support

lhc70000:
### Description

A list of `PlayerCore`s are stored to support multiple window.

```
        playerCores
      +--------------+
[MPV]-+-[PlayerCore]-+-[MainWindow]
[MPV]-+-[PlayerCore]-+-[MainWindow]
[MPV]-+-[PlayerCore]-+-[MainWindow]
[MPV]-+-[PlayerCore]-+-[MainWindow]
      +      ...     +
      +--------------+
```

Once created, a `PlayerCore` will not be removed from the list. When IINA needs a `PlayerCore`, it will always try to find an idle one from the list and create a new instance only when the search fails. (I found it quite painful to create and destroy `PlayerCore`s dynamically, e.g. thread safety issues. If you can get it work, please feel free to push) 

On termination of the app, IINA will destroy all `PlayerCore`s.

### Features

**Show an initial window on launch **

Configurable by user. The initial window is actually the `MainWindow` itself, but with OSC/OSD disabled. User can still open sidebars and set some config before opening media.

**Open media in new windows**

Configurable by user. If enabled, IINA will always try to create new windows for playback.

**Create a new window**

User can create a new initial window manually.

**When user pressed `q`**

By default `q` sends `quit` command to mpv. The current `PlayerCore` will be shut down and need to be destroyed ASAP; however as said before, it's maybe too tedious for now. Currently when user press `q` the **whole App will quit**.


alejx:
I will list some opinions I came up with.

### Opinions that support the initial window
#### 1. Make user easier to get access to the locations.
We can list all location that users want in the initial window. Without the initial widow, users can only use modifier keys (e.g. <kbd>⌥</kbd>) to open a media in a new window.
#### 2. Functionalities from the window itself.
> User can still open sidebars and set some config before opening media.

### Opinions that against the initial window
#### 1. Conciseness.
Most of macOS applications do not have such an initial window. Xcode, however, does have the initial window, because there are different and complex ways to create and open documents. Other typical apps like QuickTime, iWork suite (contrast to Office suite) do not have initial windows.
#### 2. Chaos Logic
The initial window acts like a normal window, so it can access almost all the functionalities. However the media has not been added. We will have to sort out the logic and disable the invalid actions. Namely, the initial window might be buggy.

saagarjha:
Nice work! I haven't had a chance to dive into the code yet, but I have a couple of comments:

1. I think opening files in new windows should be the default; all the video players I know do this.We should rename the menu items from "Open in New Window…" to "Open…" and "Open…" to "Open in This Window…" to reflect this behavior.
2. The (custom) initial window certainly does look nice, but I'm not sure if it's necessary. It's only really useful for reopening files. In Xcode, for example, this can be useful, since returning to the same project after launching Xcode is a common occurrence. For a video player, however, people are much less likely to open previously watched videos, so this isn't as useful. IMO we should just go the iWork/QuickTime route of opening a file chooser if IINA is launched without a file. BTW, if we do decide to have an initial window, this should be available to users so that they can open it at any time under the Window menu. Also, you might consider adding a checkbox that allows users to hide the window as Xcode does.

saagarjha:
@lhc70000 Regarding the `PlayerCore` issue…could you give me a bit more detail as to why destroying them is annoying? Are they performing any cleanup, or does anything have to be done on a specific thread?

lhc70000:
> 1

It's controlled by "Always open media in new window" in preferences and can be turned on by default. 

> 2

There's also an option for it…… and user can choose from showing the initial window, file chooser or do nothing. I personally prefer an initial window which provides a general entrance for all ways to open media, like opening URL or choosing directly from recent files (When I watch a series, I'm used to select a recent file and find the desired episode in playlist, rather than digging from folder hierarchies in open panel).

I'll collect more opinions on default values of the above two settings.

> PlayerCore issue

Like, some code in Touch Bar will crash when user press `q` (`mpv` destroyed at first). I must `guard` everywhere and handle the order of shutting down / destroying `mpv`, unloading `MainWindow` and releasing `PlayerCore` carefully. mpv's `SHUTDOWN` event also seems unreliable. But whatever, I just gave up and focused on other functions and didn't do much research. There should be a solution and we can work it out.

saagarjha:
> It's controlled by "Always open media in new window" in preferences and can be turned on by default.

Yes, I know that. I'm saying that this should be the default option.

As for `PlayerCore`s, I see that there is `terminateMPV(sendQuit:)` that seems to be called for deinitialization. Why not have a `deinit` block which is automatically executed when its reference count drops to zero? If this needs to run on the main thread, wrap it in a `DispatchQueue.main.async`.

lhc70000:
This is how quitting works previously.

379ce83d-9d62-4884-9fe0-65be3d76d14e

I still can't make all of them work, as mpv can quit before _or_ after closing window, and called by _or_ not by IINA.

PR #810, opened 2017-07-07
Preference -> Preferences (Preferences window)

northamerican:

The title of the Preferences window should read 'Preferences'.


alejx:
We are not using `en.lproj` any more. And actually, we have already change `Preference` to `Preferences` in `Base.lproj`. `en.lproj` will be removed soon.

saagarjha:
@northamerican Thanks for the effort, though!

PR #823, opened 2017-07-11
Fixed some memory leaking

skyline75489:
- [X] It implements / fixes issue - general memory issues.

---

**Description:**

### DIFF
--- a/iina/AutoFileMatcher.swift
+++ b/iina/AutoFileMatcher.swift
@@ -46,11 +46,9 @@ class AutoFileMatcher {
 
     // group by extension
     for file in files {
-      autoreleasepool{
-        let fileInfo = FileInfo(file)
-        if let mediaType = Utility.mediaType(forExtension: fileInfo.ext) {
-          filesGroupedByMediaType[mediaType]!.append(fileInfo)
-        }
+      let fileInfo = FileInfo(file)
+      if let mediaType = Utility.mediaType(forExtension: fileInfo.ext) {
+        filesGroupedByMediaType[mediaType]!.append(fileInfo)
       }
     }
 
--- a/iina/AutoFileMatcher.swift
+++ b/iina/AutoFileMatcher.swift
@@ -14,7 +14,8 @@ class AutoFileMatcher {
     case ticketExpired
   }
 
-  weak private var player: PlayerCore!
+  private unowned var player: PlayerCore
+
   var ticket: Int
 
   private let fm = FileManager.default
@@ -45,9 +46,11 @@ class AutoFileMatcher {
 
     // group by extension
     for file in files {
-      let fileInfo = FileInfo(file)
-      if let mediaType = Utility.mediaType(forExtension: fileInfo.ext) {
-        filesGroupedByMediaType[mediaType]!.append(fileInfo)
+      autoreleasepool{
+        let fileInfo = FileInfo(file)
+        if let mediaType = Utility.mediaType(forExtension: fileInfo.ext) {
+          filesGroupedByMediaType[mediaType]!.append(fileInfo)
+        }
       }
     }
 
--- a/iina/FileGroup.swift
+++ b/iina/FileGroup.swift
@@ -17,7 +17,6 @@ class FileInfo: Hashable {
   var filename: String
   var ext: String
   var nameInSeries: String?
-  var characters: [Character]
   var dist: [FileInfo: UInt] = [:]
   var minDist: [FileInfo] = []
   var relatedSubs: [FileInfo] = []
@@ -26,7 +25,7 @@ class FileInfo: Hashable {
 
   var prefix: String {  // prefix detected by FileGroup
     didSet {
-      if prefix.characters.count < self.characters.count {
+      if prefix.characters.count < self.filename.characters.count {
         suffix = filename.substring(from: filename.index(filename.startIndex, offsetBy: prefix.characters.count))
         getNameInSeries()
       } else {
@@ -42,7 +41,6 @@ class FileInfo: Hashable {
     self.path = url.path
     self.ext = url.pathExtension
     self.filename = url.deletingPathExtension().lastPathComponent
-    self.characters = [Character](self.filename.characters)
     self.prefix = ""
     self.suffix = self.filename
   }
@@ -106,12 +104,13 @@ class FileGroup {
       var lastPrefix = ""
       for finfo in contents {
         // if reached string end
-        if i >= finfo.characters.count {
+        if i >= finfo.filename.characters.count {
           tempGroup.safeAppend(finfo, for: prefix)
           currChars.append(("/", prefix))
           continue
         }
-        let c = finfo.characters[i]
+        let index = finfo.filename.index(finfo.filename.startIndex, offsetBy: i)
+        let c = finfo.filename.characters[index]
         var p = prefix
         p.append(c)
         lastPrefix = p
@@ -144,8 +143,7 @@ class FileGroup {
 
   func flatten() -> [String: [FileInfo]] {
     var result: [String: [FileInfo]] = [:]
-    var search: ((FileGroup) -> Void)!
-    search = { group in
+    func search(_ group: FileGroup) {
       if group.groups.count > 0 {
         for g in group.groups {
           search(g)
--- a/iina/FilterWindowController.swift
+++ b/iina/FilterWindowController.swift
@@ -47,7 +47,9 @@ class FilterWindowController: NSWindowController {
   }
 
   deinit {
-    NotificationCenter.default.removeObserver(self)
+    ObjcUtils.silenced {
+      NotificationCenter.default.removeObserver(self)
+    }
   }
 
   // MARK: - IBAction
--- a/iina/FilterWindowController.swift
+++ b/iina/FilterWindowController.swift
@@ -47,8 +47,11 @@ class FilterWindowController: NSWindowController {
   }
 
   deinit {
-    ObjcUtils.silenced {
-      NotificationCenter.default.removeObserver(self)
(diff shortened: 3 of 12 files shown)


iina/FileGroup.swift
```diff
@@ -17,7 +17,6 @@ class FileInfo: Hashable {
   var filename: String
   var ext: String
   var nameInSeries: String?
-  var characters: [Character]
```
lhc70000: Are you sure there's no performance issue caused by accessing `characters` and constructing `Index` in each iteration?

iina/HistoryWindowController.swift
```diff
@@ -77,6 +77,10 @@ class HistoryWindowController: NSWindowController, NSOutlineViewDelegate, NSOutl
     outlineView.expandItem(nil, expandChildren: true)
   }
 
+  deinit {
+    NotificationCenter.default.removeObserver(self)
+  }
```
lhc70000: Please wrap it with `ObjcUtils.silenced`.

It seems that Cocoa tends to release (leak) the whole memory block after quit and let system GC to handle it, so `deinit` is not being called at a fixed time. In early versions IINA crashes when trying to remove nonexistent observers. I'm not sure it's still needed but it's better to add them for safety reasons.

iina/FileGroup.swift
skyline75489: I don't see any obvious performance issue at this point. The amount of data isn't large enough, after all.

iina/MPVController.swift
```diff
@@ -342,8 +342,13 @@ class MPVController: NSObject {
     strArgs.insert(command.rawValue, at: 0)
     strArgs.append(nil)
     var cargs = strArgs.map { $0.flatMap { UnsafePointer<Int8>(strdup($0)) } }
+    defer {
```
saagarjha: Why is a `defer` necessary?

iina/QuickSettingViewController.swift
```diff
@@ -147,8 +147,10 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
   }
 
   deinit {
-    observers.forEach {
-      NotificationCenter.default.removeObserver($0)
+    ObjcUtils.silenced {
```
saagarjha:
```swift
guard #available(OSX 10.11, *) else {
  ObjcUtils.silenced { [unowned self] in
    self.observers.forEach(NotificationCenter.default.removeObserver)
  }
  return
}
```

iina/PlaylistViewController.swift
```diff
@@ -102,7 +102,9 @@ class PlaylistViewController: NSViewController, NSTableViewDataSource, NSTableVi
   }
 
   deinit {
-    NotificationCenter.default.removeObserver(self.playlistChangeObserver!)
+    ObjcUtils.silenced {
```
saagarjha:
```swift
guard #available(OSX 10.11, *) else {
  ObjcUtils.silenced { [unowned self] in
    self.playlistChangeObserver.flatMap { NotificationCenter.default.removeObserver($0) }
  }
  return
}
```

iina/AutoFileMatcher.swift
```diff
@@ -45,9 +46,11 @@ class AutoFileMatcher {
 
     // group by extension
     for file in files {
-      let fileInfo = FileInfo(file)
-      if let mediaType = Utility.mediaType(forExtension: fileInfo.ext) {
-        filesGroupedByMediaType[mediaType]!.append(fileInfo)
+      autoreleasepool{
```
saagarjha: Why is this necessary?

iina/FilterWindowController.swift
```diff
@@ -47,7 +47,9 @@ class FilterWindowController: NSWindowController {
   }
 
   deinit {
-    NotificationCenter.default.removeObserver(self)
+    ObjcUtils.silenced {
```
saagarjha:
```swift
guard #available(OSX 10.11, *) else {
  ObjcUtils.silenced { [unowned self] in
    NotificationCenter.default.removeObserver(self)
  }
  return
}
```

iina/HistoryWindowController.swift
```diff
@@ -77,6 +77,12 @@ class HistoryWindowController: NSWindowController, NSOutlineViewDelegate, NSOutl
     outlineView.expandItem(nil, expandChildren: true)
   }
 
+  deinit {
+    ObjcUtils.silenced {
```
saagarjha:
```swift
guard #available(OSX 10.11, *) else {
  ObjcUtils.silenced { [unowned self] in
    NotificationCenter.default.removeObserver(self)
  }
  return
}
```

iina/KeyRecordViewController.swift
```diff
@@ -69,6 +69,12 @@ class KeyRecordViewController: NSViewController, KeyRecordViewDelegate, NSRuleEd
     }
   }
 
+  deinit {
+    ObjcUtils.silenced {
```
saagarjha:
```swift
guard #available(OSX 10.11, *) else {
  ObjcUtils.silenced { [unowned self] in
    NotificationCenter.default.removeObserver(self)
  }
  return
}
```

iina/FileGroup.swift
```diff
@@ -144,8 +143,7 @@ class FileGroup {
 
   func flatten() -> [String: [FileInfo]] {
     var result: [String: [FileInfo]] = [:]
-    var search: ((FileGroup) -> Void)!
-    search = { group in
+    func search(_ group: FileGroup) {
```
skyline75489: There seems to be a bug inside Swift itself which causes memory leak when there's recursive call of closure. Using inner function prevents this kind of leak.

saagarjha: > There seems to be a bug inside Swift itself which causes memory leak when there's recursive call of closure.

Is there? Give me a few more details and I'll file it on the [Swift JIRA](http://bugs.swift.org/).

iina/MPVController.swift
skyline75489: AFAIK, `defer` is a good practice for releasing unsafe resources. Because you can use it right after the alloctaion code, it's more noticeable & error-resistant.

iina/AutoFileMatcher.swift
skyline75489: Putting `autoreleasepool` inside loop is also a common practice to avoid memory spike, if that's what you asking.

iina/MPVController.swift
saagarjha: Yes, I know that. The resources are released two lines down, though.

iina/FileGroup.swift
skyline75489: Have a look at instruments & you'll find out there's leak around here.

iina/FileGroup.swift
skyline75489: This change is mainly because `self.characters = [Character](self.filename.characters)` is causing too much memory copying (and somehow shows leaking in Instruments). If you guys think performance has higher priority, I'm OK with reverting this.

iina/AutoFileMatcher.swift
saagarjha: `autoreleasepool` is useful when you're working with `autorelease`d types. You're working with Swift `struct`s here.

iina/MPVController.swift
skyline75489: I'd like a consistent coding style with this kind of allocation. So I changed this on and the one in md5.

iina/AutoFileMatcher.swift
skyline75489: Which `struct` are you talking about? `FileInfo` is a class here.

iina/FileGroup.swift
saagarjha: Arrays are COW. Since no modifications are taking place no copying should be done.

skyline75489: That's my first thought but Instruments suggest otherwise. You know what, I'm beginning to doubt whether Instruments is pointing me to the right direction.

iina/FileGroup.swift
skyline75489: At first I thought maybe `[weak search]`? But Swift complains about "'weak' may only be applied to class and class-bound protocol types, not '(FileGroup) -> Void'".

iina/AutoFileMatcher.swift
saagarjha: Oops, I had the wrong `FileInfo`. However, my point still stands–`FileInfo` should not have to be `autorelease`d.

skyline75489: If the performance thing about character stands, it means there will be a lot of `FileInfo`. So...why not `autorelease` it?

saagarjha: IIRC Chris Lattner had said somewhere that the compiler will detect this and return an object with a +0 retain count in this case.

iina/MPVController.swift
```diff
@@ -342,8 +342,13 @@ class MPVController: NSObject {
     strArgs.insert(command.rawValue, at: 0)
     strArgs.append(nil)
     var cargs = strArgs.map { $0.flatMap { UnsafePointer<Int8>(strdup($0)) } }
```
skyline75489: This actually causes bad memory access for some reason.

saagarjha: Oh wait, never mind. Leave it as is.

iina/AutoFileMatcher.swift
skyline75489: Huh, in that case this isn't necessary. Is this thing applied to all loops? Forgive me cuz I'm mostly a objc guy.

lhc70000:
All issues in this PR seems to be already resolved in current codebase. Closing.

PR #831, opened 2017-07-12
Ukrainian translation

sevenfourk:

---

@lhc70000, firstly, thanks for IINA. I wanted something like QuickTime/mpv hybrid on my Mac. UI is important indeed, thought functionality goes right along too.

Added Ukrainian translation.  Ran check translation script.  Was showing 0 issues. But got this one only:

```
# InfoPlist.strings #
  [x] Cannot read file "./iina/Base.lproj/InfoPlist.strings"
```


lhc70000:
Great job and thanks a lot! I'll review it tomorrow.

lhc70000:
Everything is perfect except some `read.` keys in `KeyBinding.strings`.

For example, in

```
"read.seek" = "Пропустити {{#fwd}}forward{{/fwd}}{{#bwd}}backward{{/bwd}}{{#abs}}до{{/abs}} {{value}}{{#sec}}с{{/sec}}{{#per}}%{{/per}} {{#exact}}(exact){{/exact}}{{#kf}}(keyframe){{/kf}}";
```

"forward", "backward", etc. should be in your language.

sevenfourk:
@lhc70000, thanks, pushed fix for your review. I see that in other languages these two strings: "read.seek", "read.sub-seek" not fully translated, like in my version.

lhc70000:
@sevenfourk Thanks, I'll check them later.

PR #862, opened 2017-07-16
Use monospaced numbers

J-rg:


iina/MainWindowTouchBarSupport.swift
```diff
@@ -97,6 +97,7 @@ extension MainWindowController: NSTouchBarDelegate {
     case NSTouchBarItemIdentifier.time:
       let item = NSCustomTouchBarItem(identifier: identifier)
       let label = DurationDisplayTextField(labelWithString: "00:00")
+      label.font = NSFont.monospacedDigitSystemFont(ofSize: 14, weight: NSFontWeightMedium)
```
saagarjha: A better solution might be to put this code in `DurationDisplayTextField`.

J-rg: Not entirely sure if I get what you're suggesting, but as that class is also used elsewhere and I don't see a way of using the monospaced font without also specifying size and weight, it seems odd to me to put Touch Bar specific code into that class. Or am I missing something here?

lhc70000:
Why closed this one?

Issue #735, opened 2017-06-10
PIP crashes if activated when another application is already using it


saagarjha:
PIP will crash if another application has already activated PIP.

**System and IINA version:**

- macOS 10.13
- IINA 0.0.11

**Expected behavior:**

No crash; IINA should dismiss any other PIP windows before trying to present itself in PIP.

**Actual behavior:**

Crash

- Crash report:
```
Process:               IINA [10350]
Path:                  /Applications/IINA.app/Contents/MacOS/IINA
Identifier:            com.colliderli.iina
Version:               0.0.11 (29)
Code Type:             X86-64 (Native)
Parent Process:        ??? [1]
Responsible:           IINA [10350]
User ID:               501

Date/Time:             2017-06-09 20:19:32.435 -0700
OS Version:            Mac OS X 10.13 (17A264c)
Report Version:        12
Anonymous UUID:        BB0A8A4A-E7AF-10A7-C32D-317E0F40DD6C

Sleep/Wake UUID:       0F629E4A-F27F-4E96-ABC6-48FA5005C340

Time Awake Since Boot: 18000 seconds
Time Since Wake:       13000 seconds

System Integrity Protection: disabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_INSTRUCTION (SIGILL)
Exception Codes:       0x0000000000000001, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Illegal instruction: 4
Termination Reason:    Namespace SIGNAL, Code 0x4
Terminating Process:   exc handler [0]

Application Specific Information:
Crashing on exception: presentViewController:animator:: <NSViewController: 0x6000004cdba0> has already been presented by <PIPViewController: 0x7fde98fda8b0>

Application Specific Backtrace 1:
0   CoreFoundation                      0x00007fffac14c28b __exceptionPreprocess + 171
1   libobjc.A.dylib                     0x00007fffd24fcd06 objc_exception_throw + 48
2   CoreFoundation                      0x00007fffac151ec2 +[NSException raise:format:arguments:] + 98
3   Foundation                          0x00007fffae202f05 -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 193
4   AppKit                              0x00007fffa9c0fa16 -[NSViewController presentViewController:animator:] + 342
[...]

Issue #511, opened 2017-03-23
Changed aspect ratio does not apply to PiP


blaichch:
**System and IINA version:**

- macOS 10.12.4 Beta (16E192b)
- IINA 0.0.6 Build 14

**Expected behavior:**
When changing the aspect ratio from a video in IINA it should also apply for PiP.
.
**Actual behavior:**
PiP uses the original video aspect ratio instead of the aspect ratio that has been set in IINA.

**Steps to reproduce:**
1. Open a video
2. Change aspect ratio in IINA
3. Open PiP

**How often does this happen?**
Always

![screen shot 2017-03-23 at 23 46 07](https://cloud.githubusercontent.com/assets/12624897/24273725/3ab4f3fa-1024-11e7-95fa-f630004f21f2.png)

lhc70000:
@saagarjha Any progress?

saagarjha:
I think I have a branch locally that has some sort of fix for this…let me see if I can clean it up. I'm not sure if I'm getting the scale correctly, though; how can I query for the value?

lhc70000:
@saagarjha Should be `playerCore.info.displayWidth`.

saagarjha:
I mean getting the scale.

lhc70000:
`playerCore.info.displayWidth` / `playerCore.info.displayHeight`, if you mean the aspect ratio……

Issue #879, opened 2017-07-18
Bring IINA to front when exiting PIP


oskarkrawczyk:
**System and IINA version:**

- macOS 10.12.5
- IINA 0.0.12

**Expected behavior:**

When exiting PIP, the app should bring itself to the front – I personally keep have to find INAA after exiting PIP in order to put it in the front.

The only other app I know that uses PIP is the unofficial WWDC.app, and it works as expected: https://v.usetapes.com/jzF6GmpSaf

**Actual behavior:**

Exiting PIP, keeps the app below the currently active window.

saagarjha:
@oskarkrawczyk I'm working on fixing PIP and this is one of the things on my checklist. Stay tuned!

lhc70000:
Fixed/added in 0.0.13.

Issue #380, opened 2017-02-14
Option to hide window in Picture in Picture mode


matthewdias:
Will need to restore the window when the restore button is clicked, or the shortcut is activated.
![](https://i.imgur.com/kV2I8mA.png)

saagarjha:
As a workaround, you can use the hide command (⌘-H) while in Picture in Picture to hide the main window. Stopping PIP, however, will not raise the window automatically (though I can probably get it to do that, if desired).

matthewdias:
I'm using that for now. I think it'd be fine to just tell the window to hide when entering pip, and unhide when leaving. This setting should be opt-in of course.

uiryuu:
You can configure IINA to minimize when entering PIP

PR #880, opened 2017-07-19
Improve PIP

saagarjha:
- [X] It implements / fixes issue #735, #879.

---

I've gone through and fixed some PIP bugs as well as simplified the related code.

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -77,7 +77,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     NSColorPanel.shared().showsAlpha = true
 
     // other
-    if #available(OSX 10.12.2, *) {
+    if #available(macOS 10.12.2, *) {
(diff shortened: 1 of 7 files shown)


saagarjha:
@lhc70000 If you're running macOS Sierra would you mind testing this out?

lhc70000:
@saagarjha 👌

lhc70000:
LGTM. Could you also fix #511?

alejx:
screen shot 2017-07-23 at 12 37 02 pm
Thumbnail preview should be disabled in picture in picture.

alejx:
screen shot 2017-07-23 at 12 40 03 pm
It would be better if we can handle this strange box for OSDs in pip.

lhc70000:
@alejx Let's discuss in #350 or #380.

Perhaps after adding the "Music mode", we can switch to that mode when entered PIP.

saagarjha:
@lhc70000 Done.
@alejx This might be a bug with `NSVisualEffectView`–if so, there's not much we can do other than file a radar.

PR #905, opened 2017-07-23
Use launch arguments for localization

skyline75489:

---

**Description:**

Launch arguments offer more flexibility over schema. This PR adds several launch arguments for localization usage.


lhc70000:
I prefer schemes for quick switching among them……

Issue #931, opened 2017-07-30
Wrong URL Format


MuntashirAkon:
**System and IINA version:**

- macOS 10.13 public beta 3
- IINA 0.0.12 buld 37

**Expected behaviour:**

IINA should open the link (below) and play it

**Actual behaviour:**

IINA shows me an alert saying `Wrong URL format`

Here's the link:
```
http://172.127.100.100/download/new2/2017/Pirates%20of%20the%20Caribbean%3A%20Dead%20Men%20Tell%20No%20Tales%20(2017)/Pirates%20Of%20The%20Caribbean%20Dead%20Men%20Tell%20No%20Tales%20(2017)%20720p%20HD-TS%20x264%20AAC%20-%20[MSP].mkv
```

**Steps to reproduce:**
N/A

**How often does this happen?**
Everytime

**PS:** I've tried playing the same URL using VLC where it worked fine.

bmelts:
This is a fun RFC mismatch. The provided URL is valid under RFC 1738, which NSURL and most things that interact with URLs use; however, it's invalid under RFC 3986, which NSURLComponents uses (in RFC 3986, square brackets are considered reserved characters within the path, and must be percent-encoded).

The easy solution is to make an NSURL from the string first, then make the NSURLComponents from the NSURL, since there's some magic in there to handle the RFC differences. The harder solution is to ensure the string is RFC 3986-compliant, but that would require rewriting the URL escaping algorithm, since it's a lot more complicated to percent encode different parts with different allowed characters.

MuntashirAkon:
@bmelts In the [OpenURLAccessoryViewController.swift](https://github.com/lhc70000/iina/blob/02cc0962950207d85b5b15c32614ebdf8b1e118a/iina/OpenURLAccessoryViewController.swift#L34), I found this:
```swift
guard let urlComponents = NSURLComponents(string: urlValue) else { return nil }
```
So, this should be something like this:
```swift
let nsurl = NSURL(string: urlValue)!
guard let urlComponents = NSURLComponents(url: nsurl, resolvingAgainstBaseURL: false) else { return nil }
```

lhc70000:
@MuntashirAkon If you found the solution, could you submit a pull request?

MuntashirAkon:
Ok, [here](https://github.com/lhc70000/iina/pull/933) it is.

ErikMinekus:
Since the above PR has been merged, this can be closed?

PR #933, opened 2017-07-31
Fix "Wrong URL Format" error for correct URL

MuntashirAkon:
- [x] It implements / fixes issue #931.

---

**Description:**
Due to RFC mismatch, a valid RFC 1738 URL is shown as invalid (saying `"Wrong URL Format"`). I attempted to solve the problem by making an NSURL from the string and passing it to NSURLCompnents.


lhc70000:
Merged f6a5682 via ff12cb8.

For 56cc613, please discuss with us before proposing UI changes.

MuntashirAkon:
Ok. I will.

PR #1016, opened 2017-09-06
Solve a minor issue at QuickSettingViewController.xib

MuntashirAkon:
- [x] This change has been discussed with the author (who told me to explain it further).
- [x] Fixed the width and arrangement of the custom field for `Aspect Ratio`
- [x] Fixed the arrangement of the `Custom...` button of `Crop`

---

**Description:**
For the sake of clarity, let's look at the changes by comparing two images:
Before:
screen shot 2017-09-07 at 2 01 48 am
After:
screen shot 2017-09-07 at 2 02 35 am

(Screenshots were taken on macOS 10.13. First shot was from iina v0.0.12 build 37 and the later was the result after the compilation)


MuntashirAkon:
As you said earlier, merge only the [`56cc613`](https://github.com/lhc70000/iina/pull/1016/commits/56cc613fe5447f39a8f1f4cc49876ed0368ca61d).

lhc70000:
I'm still not sure if this is the best solution - we should always avoid cropping text.

How about adding a wrapping scroll view for the segment control?

MuntashirAkon:
Would this do any justice to the UI? I think it will look even worse.

MuntashirAkon:
I've reverted the change and closing this. Maybe I can do better :)

Issue #819, opened 2017-07-09
Migration to Swift 4


saagarjha:
Just wanted to open an issue to track the migration to Swift 4. Currently, IINA fails to build even in "Swift 3" mode in Xcode 9 due to its dependencies (will list below as progress continues); I think we'll have to go nag them to update to Swift 4.

saagarjha:
GzipSwift has a bad modulemap, resolved by using the swift4 branch.

saagarjha:
MASPreferences is having issues with `NSUserInterfaceItemIdentifier`: https://github.com/shpakovski/MASPreferences/issues/54
Currently falling back to fork: https://github.com/saagarjha/MASPreferences

saagarjha:
GRMustache needs updating to Swift 4.
Currently falling back to fork: https://github.com/saagarjha/GRMustache.swift

- [X] Pull request filed: https://github.com/groue/GRMustache.swift/pull/46
- [ ] Pull request merged
- [ ] Pod updated

saagarjha:
Just needs updating to Swift 4.
Currently falling back to fork: https://github.com/saagarjha/Just

- [X] Pull request filed: https://github.com/JustHTTP/Just/pull/83
- [ ] Pull request merged
- [ ] Pod updated

skyline75489:
@saagarjha With the help of Swift 4, maybe we can just drop usage of libraries like Just?

lhc70000:
@skyline75489 Maybe AEXML, but not Just please, it's lightweight enough, and we need a networking library for both current and potential need in the future.

skyline75489:
@lhc70000 I should've made my point clear. With the help of Swift 4 and its new `Codable` API, JSON parsing would be easier. `Just` is basically as easy-to-use as an `NSURLSession + NSJSONSerialization` wrapper when it comes to simple HTTP GET. On the other hand, if you want further possibility, `Just` isn't as popular, fully-functional and actively maintained as `Alamofire`. So I wouldn't say using it is a good choice.

saagarjha:
PromiseKit had some compile errors in Swift 3, resolved by using the swift4-beta1 branch.

saagarjha:
IINA compiles in Swift 4! (I'd put some confetti emoji here, but it causes [Safari to crash](https://bugs.webkit.org/show_bug.cgi?id=174404) so I'd thought I'd spare you guys).

saagarjha:
[...]

PR #1033, opened 2017-09-13
Swift 4 support

saagarjha:
- [X] It implements / fixes issue #819.

---

**Description:**
With Xcode 9 in GM, I think it's time to begin migrating our Swift 4 code to the develop branch. Please take a look at this in Xcode 9, make sure it compiles, and try out core features. If you see an opportunity to use a new Swift 4 feature that I missed, please don't hesitate to call it out.


lhc70000:
Thanks, I'm going to review and merge it by weekend.

iina/Extensions.swift
```diff
@@ -482,3 +482,9 @@ extension URL {
     return (try? self.resourceValues(forKeys: [.isDirectoryKey]))?.isDirectory ?? false
   }
 }
+
+extension NSPasteboard.PasteboardType {
+  static let nsURL = NSPasteboard.PasteboardType("NSURL")
```
saagarjha: I'd argue that this should be `static let URL = NSPasteboard.PasteboardType("NSURL")`. `NSString` has already been migrated by Apple as `string` so I think dropping the prefix would be more consistent.

iina/Extensions.swift
```diff
@@ -482,3 +482,9 @@ extension URL {
...
+  static let nsFilenames = NSPasteboard.PasteboardType("NSFilenamesPboardType")
```
saagarjha: Same as above

iina/FileGroup.swift
```diff
@@ -27,7 +27,7 @@ class FileInfo: Hashable {
   var prefix: String {  // prefix detected by FileGroup
     didSet {
       if prefix.characters.count < self.characters.count {
-        suffix = filename.substring(from: filename.index(filename.startIndex, offsetBy: prefix.characters.count))
+        suffix = String(filename[filename.index(filename.startIndex, offsetBy: prefix.characters.count)...])
```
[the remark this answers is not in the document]
saagarjha: You'd need to put a cast anyways, since `suffix(from:)` returns a `Substring` too:
```swift
suffix = String(filename.suffix(from: filename.index(filename.startIndex, offsetBy: prefix.characters.count)))
```
This makes it a bit longer than the slicing it with a subscript.

saagarjha:
Thanks for adding your commits; it looks a lot better now. One thing to note: this pull request is merely makes IINA *compile*. I still have to make another pass and modernize the code to take advantage of the new features in Swift 4. I'm a bit busy with other work currently so I'll hold off on it for now, but we should revisit it sometime.

iina/Extensions.swift
lhc70000: `NSPasteboard.PasteboardType` already has a member `url` for "public.url" (and `filenames` for "public.filenames"), so `nsURL` would be appropriate if we want to refer to "NSURL" clearly, as well as to avoid name collision.

lhc70000:
@saagarjha Sure :D

lhc70000:
2-4 fields for each controller. It's an Obj-C library so I think using `@objcMembers` would be a more safe (and lazy) way.

Do you think we should eliminate Obj-C lib usages? I think we'd better implement the pref panel by ourselves or write a Swift version of this library.

Issue #451, opened 2017-03-01
Touch Bar Persistence Missing


StarlightDesign:
I'm not sure if this is a missing feature or an already implemented feature that's bugged, but I'm hoping for persistent Touch Bar controls in the Control Strip once you leave the IINA full screen space. Kind of like what iTunes or Safari does with songs or HTML video, and how you can control them while working in a different app screen.

A great use case and my example of what I'm looking for is seeking while using Picture-in-Picture and working in a different app space.

Is this something that can be implemented or fixed?

inflation:
With a simple search, it supposed to be a private API. Do we want to implement it? @lhc70000

lhc70000:
I think it's OK, anyway IINA has already used private APIs.

saagarjha:
@StarlightDesign Please send screenshots/more detail, as I can't get the controls to show up in the background on my computer for Safari or iTunes.

StarlightDesign:
<img width="1085" alt="touch bar shot 2017-03-22 at 10 15 13 pm" src="https://cloud.githubusercontent.com/assets/8271041/24233084/28de944c-0f4d-11e7-9bab-779f62a8dea2.png">
<img width="1085" alt="touch bar shot 2017-03-22 at 10 15 24 pm" src="https://cloud.githubusercontent.com/assets/8271041/24233083/28ddd980-0f4d-11e7-8916-53ab88f29029.png">

This is how the Touch Bar media persistence shows up on my laptop. Just to elaborate, this is for the new MacBook Pro computers for the Touch Bar specifically. Picture in Picture on Mac doesn't have any controls available in the window, and you have to return to the actual player's space to control it. It would be really handy to access seek controls from the Touch Bar while using a different app altogether.

saagarjha:
@StarlightDesign Thanks for the screenshots. I'm still trying to get it to show up on my computer (which doesn't have a physical Touch Bar); once I can get it working I'll look into how it's implemented.

StarlightDesign:
@saagarjha Many thanks, appreciate the help.

alejx:
@saagarjha @inflation @lhc70000 As far as I'm concerned, this feature is not related with the touch bar. If an application can be controlled by the system media keys, then it will be displayed in the touch bar.

[...]

Issue #846, opened 2017-07-15
Capslock to switch Korean IME isn't working while IINA is running. 


goodhyun:
**System and IINA version:**

- macOS 10.12.5
- IINA 0.0.11

**Expected behavior:**

As of macOS 10.12, capslock key is used to switch Korean input method.

**Actual behavior:**

However, this functionality isn't working while IINA is running. 

- Crash report:
```

```

- MPV log:
```

```

**Steps to reproduce:**

- [x] MPV does not have this problem.

**How often does this happen?**

lhc70000:
Please turn off media key as a workaround. It'll be solved in 10.13.

Dupe of #477

Issue #799, opened 2017-07-06
AirPods should pause the playback when taken off


fabiocody:
**Feature Request**

- macOS 10.12.5
- IINA 0.0.11

My AirPods don't pause the playback when I take them off, and I know this is possible since other apps do it, so I was just asking if this could be added/fixed in some later version. Thanks

saagarjha:
@fabiocody Could you name an app that does this? It will help us figure out how best to add this feature.

saagarjha:
Might also be related to #791.

fabiocody:
Of course iTunes does it, and also Safari when using HTML5 player. At the moment, IINA is the only non-Apple player I have, so I will download and test some other players and see if they do it. I'll let you know

PhilSwn:
I think [MPRemoteCommandCenter](https://developer.apple.com/library/content/samplecode/MPRemoteCommandSample/Listings/README_md.html) is relevant. 

> The ability for media applications to support the automatic pause and play functionality with the AirPods when taking them in and out of the ear.

 
And it's supported in:

> macOS 10.12 SDK or later
> Swift 3.0

Issue #791, opened 2017-07-03
Play/Pause signal from earphones are not captured by IINA


swj1442291549:
**System and IINA version:**

- macOS 10.12.5
- IINA 0.0.11

**Expected behavior:**

When IINA is running, the play/pause button from the **earphones** should be able to perform their functions.

**Actual behavior:**

When only IINA is running, the play button would open iTunes.
When iTunes is also in the background, the button on the earphone will control the iTunes.

- Crash report:
```

```

- MPV log:
```

```

**Steps to reproduce:**

- [x] MPV does not have this problem.

**How often does this happen?**

Everytime

alejx:
It's a known issue. Stay tuned!

PR #1063, opened 2017-10-02
MediaPlayer framework

alejx:
- [x] It implements / fixes issue #451, #477, #846, #778, #799, #791. (macOS 10.13+ only)

---

**Description:**

This pull request integrates MediaPlayer framework and enables multiple system media controls. 

Even though the document says MediaPlayer framework is available on macOS 10.12.1+, based on by test, it only works on 10.13. So that I kept macOS 10.13- using the legacy mpv media key functionality and enables MediaPlayer framework on macOS 10.13+. The label `Use media keys` in Preferences > Keybindings will be displayed as `Use system media control` in macOS 10.13+.

It is supposed to
- make IINA stay in the touch bar "media center" system (#451)
- not capture keys(by mpv) anymore (i.e. the caps lock key) (#477, #846, #778)
- make IINA to be controlled by system (i.e. bluetooth devices, AirPods, media keys) (#799, #791)
- shows IINA in the "Now Playing" widget

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -451,55 +451,55 @@ class RemoteCommandController {
   static var useSystemMediaControl: Bool = false
 
   static func setup() {
-    remoteCommand.playCommand.addTarget(handler: { _ in
+    remoteCommand.playCommand.addTarget { _ in
       PlayerCore.lastActive.togglePause(false)
       return .success
-    })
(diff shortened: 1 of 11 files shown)


alejx:
@saagarjha Thanks for your review, but I haven't heavily test it on 10.13, so I may push some commits later. Sorry about that.

Issue #1062, opened 2017-10-01
Translation for 0.0.13


lhc70000:
fr @superzazu
it @cgand
de @J-rg @shorshe
zh-Hant @LEOYoon-Tsaw (I'll use OpenCC if you are busy)
ko @darkbrow
ru @euf
ja @matsune
tr @eminhaqi
nl @dionvl
pl (email)

Dear translators, please update your translation for 0.0.13 release at convenience.

---

## Changed Strings

**FilterPresets.strings**

(All)

**FreeSelectingViewController**

(All)

**Localizable.strings**

- `alert.nothing_to_open`
- `alert.no_video_track`
- `subencoding.auto`

**MainMenu.strings**

- `bGN-E4-V5A.title`
- `ArE-2s-JIV.title`

**PrefCodecViewController**

- `g4c-vG-URh.title`

**PrefControlViewController**

- `Gyd-Qt-FEo.title`
- `iFF-Gq-tVx.title`
- `uqM-UM-Xvb.title`
- `7wy-xt-c8Z.title`
- `unj-zt-RyU.title`
- `dpH-PO-WKt.title`
- `M6K-ME-cve.title`
- `Ba4-eT-rTv.title`
- `hDh-Lk-FJe.title`
- `jnR-2I-4x4.title`

**PrefGeneralViewController**

- Delete `FbF-jJ-W5a.title` and use `FbF-jJ-W5b.title` instead.
- Delete `MlG-hl-goA.title` and use `MlG-hl-goB.title` instead.
- `TvB-UM-FrS.title`
- `GNV-VA-NBW.title`
- `FbF-jJ-W5b.title`
- `MlG-hl-goB.title`

**PrefSubViewController**

- `Y75-4k-2Du.title`
- `0fB-es-DmV.title`
- `29W-lw-0pD.label`
- `R5q-Xd-PPr.label`

**PrefUIViewController**

- `t6s-yV-fdc.title`
- `kYu-0l-jQQ.title`
- `GYe-Y3-bOR.title`

dionvl:
Strings do not change in _PrefGeneralViewController_ and _PrefSubViewController_ . Is anyone else having this issue?

lhc70000:
@dionvl Firstly please make sure your branch is up to date. Xcode can't synchronize the `strings` files. You can just append the keys to the end of each file, like 

```
"t6s-yV-fdc.title" = "Your translation";
```

[...]

PR #1074, opened 2017-10-05
Japanese Translation for 0.0.13

matsune:
- [x] It implements / fixes issue #1062.

---
- Japanese translation for 0.0.13
- Fixed a few translations


alejx:
Thanks for your work!

I'm just reviewing the app in Japanese and I noticed that you translated "exact" in "control" tab as "精密" but in "Keybindings" tab it is translated as "精细" (sorry for Chinese). Is there any special concern? If not, I hope they are displayed by a same word.

matsune:
Thanks for reviewing.
There is no special concern, my mistake. I will fix it and conflict.

PR #1118, opened 2017-10-17
Update Podfile

zichenJack:
fix some problem of cocoapods config,when `pod install ` maybe notice `Could not automatically select an Xcode project` & change `GRMustache.swift` git source , notice `Failed to download 'GRMustache.swift'` 
This problem only show at first open project.

---

**Description:**

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -4,13 +4,15 @@ target 'iina' do
   
   use_frameworks!
 
+  project 'iina.xcodeproj'
+  
   # Pods for iina
   pod 'MASPreferences'
   pod 'Just', :git => 'https://github.com/JustHTTP/Just.git', :branch => 'swift-4'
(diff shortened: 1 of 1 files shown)


saagarjha:
This works on my machine; which version of CocoaPods are you using?

azag0:
Had the same problem. The cause of the issue is that cocoapods clones the target repo differently depending on whether the `.git` extension is present. With it, it clones the full repo. Without it, it does a shallow copy (`--single-branch --depth 1`). The shallow copy results in a different commit. Perhaps this changed in some recent version? The version recorded in the Podfile assumes that shallow copy is made, which is the case only when `.git` is not present with a recent cocoapods. Sounds almost like a bug in cocoapods to me, but I haven't found anything by quick search.

PR #1159, opened 2017-10-23
mini player: add touch bar support for music

doukasd:
Hey @lhc70000,

First of all thanks for IINA, I've only been using it for a few days and love it. I just started using it for music and realised that the recent Mini Player did not have touch bar support yet, so I added it.

1. Sorry for not opening an issue first, I did this feature for myself but thought I'd share afterwards.
2. The addition is very simple and self contained. The whole logic is copied from the Main Window touch bar support so all styles and design is the same.
3. We can probably share some logic for touch bar support between the 2 classes but Swift is not my main language so I started the simple way.

Cheers,
Dimitris


doukasd:
Issue created: #1160

lhc70000:
Sorry for late reply, I was busy during the last whole month and will resume the development now.

Please try not to write repeated code. It would be better to create a delegate class for handling the touch bar logic, which is shared between two windows.

lhc70000:
Since I'm working on an overall enhancement for music mode, I implemented it on my branch (8a4d3f9) and you can take a look. I'm going to to close this PR, but any others are still welcomed. Thanks for your effort.

doukasd:
Yep totally understand, as I mentioned I'm not really familiar with Swift so I had to do it the simple way. I'll look out for "official" support when it's done. Cheers.

PR #1181, opened 2017-10-28
Command Line Interface

lhc70000:
**Description:**

Basic CLI support.

- Supports all mpv options (except `--no-*` ones) with `--mpv-` prrefix.
- Supports playing from stdin if invoked with `--stdin`.
- Supports passing other Cocoa arguments together, like `-Name VALUE`.


saagarjha:
FWIW I would recommend splitting the command line interface off into a separate target, instead of having it reside in the main binary.

lhc70000:
@saagarjha 

- Separate binary may require additional installation
- Still need to pass the options to main binary via some way, and need to handle piping/stdin too

saagarjha:
The binary would be bundled as part of IINA, and would handle parsing arguments and handling input/output. Then it would call the main binary and set it up for playback.

lhc70000:
In which way it calls the main binary? Is it possible to call methods in the main binary directly?

saagarjha:
> Is it possible to call methods in the main binary directly?

Fair point; this is not easy to do and so probably not worth doing. A few things we should still look into: making IINA behave like a standard command line tool, and offering to symlink it to a directory in the user's `PATH` such as `/usr/local/bin/`.

MuntashirAkon:
Splitting CLI may not be a good idea. Many applications (although not written in Swift) support CLI and GUI at the same time, including VLC, wine, etc.

iina/AppDelegate.swift
```diff
@@ -100,8 +146,20 @@ class AppDelegate: NSObject, NSApplicationDelegate {
       parsePendingURL(url)
     }
 
-    // check whether showing the welcome window after 0.1s
-    Timer.scheduledTimer(timeInterval: TimeInterval(0.1), target: self, selector: #selector(self.checkForShowingInitialWindow), userInfo: nil, repeats: false)
+    let _ = PlayerCore.first
```
alejx: I suggest wrapping this line into a class function of PlayerCore. Kind of wired.

Et0h:
I'm really pleased to see work being done on CLI support for IINA.

PR #1196, opened 2017-11-04
Add .gitattributes to prevent `git diff` on auto-generated text files.

holymonson:

---

**Description:**

Adding .gitattributes will give us a much clearer output on `git diff`.


alejx:
Though disabling these will give us a clearer git diff, these files' diffs are sometimes needed.

PR #1220, opened 2017-11-14
Could you add double speed playback?Custom fullscreen

fff555333:

---

**Description:**


lhc70000:
If you're asking a question or requesting a feature, please open an issue for that.

PR #1270, opened 2017-12-09
Save filters & enhancement for 3D LUT files

lhc70000:
**Description:**

- Users can now save an added filter, and assign a shortcut key to it.
- Support dragging a 3D LUT file (3dl, cube, dat or m3d) to main window to apply it directly.

### DIFF
--- a/iina/FilterPresets.swift
+++ b/iina/FilterPresets.swift
@@ -160,8 +160,16 @@ extension FilterPreset {
     return dic
   }()
 
+  static private let customMPVFilterPreset = FilterPreset("custom_mpv", params: ["name": PM.text(defaultValue: ""), "string": PM.text(defaultValue: "")]) { instance in
+      return MPVFilter(rawString: instance.value(for: "name").stringValue + "=" + instance.value(for: "string").stringValue)!
+  }
+  // custom ffmpeg
+  static private let customFFmpegFilterPreset = FilterPreset("custom_ffmpeg", params: [ "name": PM.text(defaultValue: ""), "string": PM.text(defaultValue: "") ]) { instance in
+    return MPVFilter(name: "lavfi", label: nil, paramString: "[\(instance.value(for: "name").stringValue)=\(instance.value(for: "string").stringValue)]")
+  }
+
   /** All filter presets. */
-  static let presets: [FilterPreset] = [
+  static let vfPresets: [FilterPreset] = [
     // crop
     FilterPreset("crop", params: [
       "x": PM.text(), "y": PM.text(),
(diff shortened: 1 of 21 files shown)


iina/FilterPresets.swift
```diff
@@ -69,38 +69,47 @@ class FilterPresetInstance {
...
-    let pm = FilterParameter(.float, defaultValue: FilterParamaterValue(float: defaultValue))
+    let pm = FilterParameter(.float, defaultValue: FilterParameterValue(float: defaultValue))
     pm.min = min
     pm.max = max
     return pm
   }
 
-  private init(_ type: ParamType, defaultValue: FilterParamaterValue) {
+  static func choose(from choices: [String], defaultChoiceIndex: Int = 0) -> FilterParameter {
+    guard !choices.isEmpty else { fatalError("FilterParameter: Choices annot be empty") }
```
saagarjha: typo

iina/FilterWindowController.swift
```diff
@@ -62,51 +111,160 @@ class FilterWindowController: NSWindowController {
...
   }
 
   func tableView(_ tableView: NSTableView, objectValueFor tableColumn: NSTableColumn?, row: Int) -> Any? {
-    guard let filter = filters.at(row) else { return nil }
-    if tableColumn?.identifier == .key {
-      return row.toStr()
-    } else if tableColumn?.identifier == .value {
-      return filter.stringFormat
+    if tableView == currentFiltersTableView {
+      guard let filter = filters.at(row) else { return nil }
```
saagarjha: You're only using this once, so…

iina/FilterWindowController.swift
```diff
@@ -62,51 +111,160 @@ class FilterWindowController: NSWindowController {
...
+      if tableColumn?.identifier == .key {
+        return row.toStr()
+      } else if tableColumn?.identifier == .value {
+        return filter.stringFormat
```
saagarjha: …`return filters.at(row)?.toStr()`?

iina/KeyCodeHelper.swift
```diff
@@ -7,7 +7,7 @@
 //
 
 import Foundation
-
+import Carbon
```
saagarjha: Ouch…is there no other way to do this? I'd really not like us to use deprecated APIs.

lhc70000: Unfortunately no…… `UCKeyTranslate` itself is not from Carbon, but in order to get the "key layout data", some HIToolbox APIs are needed.

saagarjha: I guess…

dsayed:
It would be good to add a file picker rather than have to type in the path to the file in the filter.

lhc70000:
@dsayed You can drag & drop the file directly into the player window (to apply the LUT immediately) or the input box (to fill in the path).

PR #1307, opened 2018-01-02
Add “Open Link in IINA” context menu item to Safari extension

jwells89:

---

**Description:**
This PR adds an “Open Link in IINA” item to context menus in Safari. It’s context-aware, so the item will only appear when the user is right-clicking on a link. This is useful because it allows the user to open linked content in IINA directly without having load the page first or copy the URL and paste it in IINA.


lhc70000:
Works like a charm. Thanks for your contribution!

PR #1315, opened 2018-01-05
spanish translation for 0.0.15

Aikyuichi:

---

**Description:**


alejx:
I noticed that the Spanish localization has many untranslated strings. You should use `iina/others/check_translation.swift es` to find them; or simply use [iina-translation-utility](https://github.com/iina/iina-translation-utility), it will prompt you untranslated or invalid strings.

Aikyuichi:
@alejx 
it is necessary to translate strings like: Label, Text Cell, Box, Table View Cell?

alejx:
They are not necessary to be translated.

alejx:
There are still two files to be translated: `FreeSelectingViewController.strings` and `FilterPresets.strings`.

Issue #1303, opened 2018-01-01
Translation for 0.0.15


lhc70000:
fr @superzazu
it @cgand
de @J-rg @shorshe
zh-Hant @LEOYoon-Tsaw (I'll use OpenCC if you are busy)
ko @darkbrow
ru @euf @imKota 
ja @matsune
tr @eminhaqi
nl @dionvl
pl -
es @Aikyuichi 
uk @sevenfourk 
sk @Marvin99com
da @thkir 

Dear translators, please update your translation for 0.0.15 release.

---

A lot of keys were added or deleted since the last release. **I created a utility app for checking updating translations.** If you're interested, please check out https://github.com/lhc70000/iina-translation-utility and download it from release page

If you'd like to stick with your current way and need a list of keys, please:

- Run `other/check_translation.swift <your language code>` (strongly recommended)
- Or use a local git client like Fork or Source Tree to compare current `develop` branch against tag `v0.0.14.1`.

alejx:
We have some new localizations compare to last release, please add new translators' in `Contribution.rtf` of your language. Simply copy new lines from base `Contribution.rtf` and paste in your language version, then revise the English into your language.

alejx:
We recently preformed a rename in the base localization. We renamed "Sub" to "Subtitle" (see 846fc335000cec1175053a05e31d0a5871783437). If you mistranslated them before, please update them. Sorry for any inconvenience.

imKota:
russian translation updated #1327

PR #1392, opened 2018-01-19
Revamped Trad. Chinese translation

LEOYoon-Tsaw:
- [ ] It implements / fixes issue #1303.

---

**Description:**
Revamped Trad. Chinese translation, and fixed a few untranslated strings.


lhc70000:
Please double check `iina/zh-Hant.lproj/Localizable.strings`, where all `alert.` keys were deleted. It may be a bug of your translation tool.

LEOYoon-Tsaw:
I was using the Xcode export/import localization tool to generate translation strings. It might be that the missing strings are just no longer used, but I'm not sure.
I put the missing strings back through the new commit, probably I can find out why this ever happened some time later.

lhc70000:
@LEOYoon-Tsaw Xcode just searches for all strings appeared in NSLocalizedString calls…… 

As IINA generates them dynamically, perhaps those keys should be put in another file.

PR #1436, opened 2018-01-30
Beta channel

lhc70000:


lhc70000:
@inflation 

Sparkle only detects newer versions by build number, so there should be no problem as long as we keep the build number organized (release - beta - ...  - beta - release). Users who opt out should receive update to the next public release normally.

You can try it out now (I've changed the latest build number of `appcast-beta.xml` to 999).

Issue #1450, opened 2018-02-03
Feature request: Option to show subs only while paused


devandsev:
As a user I want to be able to see subtitles only when I pause the playback, otherwise they are hidden. This is for cases when a user can understand most of the film, but sometimes misses something. With this feature enabled, user just pauses for a second, reads the thing that he has missed, and then continues to watch his film without being distracted by subtitles.

PR #1451, opened 2018-02-03
Add option to show subs only while paused

devandsev:
- [x] It implements / fixes issue #1450.

### Description:

This feature adds an option to show subs only while playback is paused. User can turn in on in the preferences.

### Integration:

It integrates with MPV by adjusting MPVOption.Subtitles.subVisibility setting.

### Screenshot:

iina


brunobely:
I like this option! +1 👍

Issue #1460, opened 2018-02-07
Feature Request: better sync tool


etienne-dldc:
**System and IINA version:**

- macOS 10.13.3
- IINA 0.0.15

**Feature request:**

Whenever I have to sync subtitle or audio I always struggle with the direction I have to delay and sometime finding the correct duration is a bit tricky too.
So here is my idea to make this easier: we would have two buttons "Mark subtitle" and "Mark video". Then you just identify something, like someone start talking on the vidéo, press "Mark video", find when the corresponding subtitle appears and click "Mark subtitle", IINA would then automatically delay the subtitles to align the two points.
Note that I'm taking subtitle as an example here but this also apply to audio/video.
These two buttons could be added under the `delay` slider (both in `AUDIO` and `SUBTITLE` settings) as a helper to configure it.

Tell me what you think of this,
Thanks for the great work 👍

inflation:
I think this would be better fit for a subtitle adjustment tool, since player won’t always remember how you tweak your subtitles. 
Current mechanism is sort of an “ad-hoc”, so I’d just open the subtitle and see the timestamp, or just find another corresponding one. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

etienne-dldc:
@inflation I don't get your point, I mean the feature already exist in IINA (subtitle delay), this is just better ways (in my opinion) to find the correct value...
And yes, it's probably a good feature for a subtitle adjustment tool but I think it's not really the question here... Being fit for something else doesn't mean it's IINAppropriate ( 😂 ) for others.

etienne-dldc:
hi 👋 

Any update on this ? Or it out of IINA scope and I should close the issue ?

PR #1467, opened 2018-02-08
Add sub/audio synchronization with video

devandsev:
- [x] It implements / fixes issue #1460.

---

### Description

This feature adds "Synchronize" buttons to the Audio and Subtitles panels. If a user clicks the button, it changes its state and title, informing the user to mark the video playback position, and after that -subtitles/audio position. When both marks are placed, calculated delay is applied to the subtitles/audio.

### Screenshots

screen shot 2018-02-08 at 21 31 19
screen shot 2018-02-08 at 21 31 44

### DIFF
--- a/iina/SyncManager.swift
+++ b/iina/SyncManager.swift
@@ -1,0 +1,88 @@
+//
+//  SyncManager.swift
+//  iina
+//
+//  Created by Andrey Sevrikov on 07/02/2018.
+//  Copyright © 2018 lhc. All rights reserved.
+//
+
+import Foundation
+
+protocol SyncManagerDelegate: class {
+  /** State of synchronization process has changed */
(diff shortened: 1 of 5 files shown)


iina/SyncManager.swift
```diff
@@ -0,0 +1,88 @@
...
+//
+
+import Foundation
+
+protocol SyncManagerDelegate: class {
+  /** State of synchronization process has changed */
+  func didChange(state: SyncManager.State, source: SyncManager.Source)
+  
+  /** Synchronization is finished with calculated *delay* */
+  func didSync(source: SyncManager.Source, delay: Double)
```
Cee: Better to use `TimeInterval` instead of `Double`, though they are the same.

PR #1469, opened 2018-02-09
Time and battery display

siamak01:

---

**Description:**
I do not know how to report for a bug and new feature. So I am here.
On Display time does not come back by hovering the mouse arrow. It reappears only after pause/play or fullscreen/maximaze/minimazie.

Requested feature which I badly want and none of other players have as far as I can say is:
Crossfade for Audio(iTunes does have) But no player has have Video crossfade.
I like to set a time in second for crossfade for audio an video play.


inflation:
Please use Issues to report bugs or feature request

PR #1484, opened 2018-02-13
Fix MPV´s CVE Security bug CVE-2018-6360 

pradorocchi:
This release Fix CVE Security bug CVE-2018-6360 (https://nvd.nist.gov/vuln/detail/CVE-2018-6360) by updating 'mpv' to version '0.27.1' (see release note: https://github.com/mpv-player/mpv/releases/tag/v0.27.1). Also ffmpeg is updated to 3.4.1_2 because is needed for mpv 0.27.1. Tested on El Capitan and High Sierra. No broken features, everything running fine and fast. May not work on Yosemite because mpv´s new version. (and I have not tested it on Yosemite).

Please make your validations and feel free to comment.

---

mpv-0 27-iina

**Description:**


lhc70000:
Thanks for the update.

However there are a few issues with this PR: 
- libmpv headers should not be updated because mpv 0.27 still uses libmpv 1.25 (and if you update the headers, you should also update `MPVProperty.swift`, `MPVOption.swift` and `MPVCommand.swift`).
- In order to make IINA work on Yosemite,  all dylibs should be built with `-mmacosx-version-min=10.10` flag.
- mpv just released 0.27.2 and 0.28.2 🤣.

I'd prefer building a new set of dylib by myself, but I'm afraid that I won't have enough time recently (we also need to do some user tests before releasing it, which will take a few days). Since mpv 0.28.x via brew will be available soon (which depends on ffmpeg 3.5), it may be better to wait for another few days.

@HarukaMa @inflation @saagarjha @alejx What do you think?

saagarjha:
I'm in favor of waiting a few days. At the risk of having to eat my words later, I don't think this issue is very problematic: users can just not use youtube-dl for a couple of days.

> we also need to do some user tests before releasing it, which will take a few days

What user tests?

pradorocchi:
I agree with you and with all observations :)

xu-cheng:
@lhc70000 

I think you probably want to fix this instead of waiting for mpv 3.8 (which is unlikely to happen in homebrew due to ffmpeg issue).

FWIW, this is a **crucial security issue** rather than a bug. After all, I don't think many IINA users are aware of this security vulnerability.

Considering 1.0.0 is still a beta, you probably should release 0.0.16 in addition to a fix in the develop branch. The idea is to make sure the majority of IINA users not vulnerable.

alejx:
A new version 0.0.15.1 with mpv 0.28.2 is up cdbaf0ece443aa813d5fe3f3ee5dba8d66c7571d. Closed.

PR #1497, opened 2018-02-16
Fix counting of string types in check_translation.

MaddTheSane:

---

**Description:**

Modernize/potential bug fixes for check_translation:

Better range getter NSRegularExpressions: they expect NSRange, UTF-16 conted strings, but Swift Strings use a different counting method.


lhc70000:
Seems that in Swift 4.0, we can use

```swift
NSRange(str.startIndex..<str.endIndex, in: str)
```

PR #1507, opened 2018-02-18
Use .fileSystemRepresentation when getting file paths from NSStrings.

MaddTheSane:

---

**Description:**
`-[NSString fileSystemRepresentation]` better handles strange characters and is the recommended way of sending file paths to lower-level file APIs.

### DIFF
--- a/iina/FFmpegController.m
+++ b/iina/FFmpegController.m
@@ -88,7 +88,7 @@ return -1;\
 {
   int i, ret;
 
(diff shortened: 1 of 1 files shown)


saagarjha:
I'm surprised that nobody's merged this in yet; it seems like a no-brainer change. That being said, we should probably go through our code and try to find similar file bridging issues.

Issue #1551, opened 2018-03-13
Localization for 1.0.0 beta1


lhc70000:
fr @superzazu
it @cgand
de @J-rg @shorshe
zh-Hant @LEOYoon-Tsaw (I'll use OpenCC if you are busy)
ko @darkbrow
ru @euf @imKota
ja @matsune
tr @eminhaqi
nl @dionvl
pl -
es @Aikyuichi
uk @sevenfourk
sk @Marvin99com
da @thkir

Dear translators, please update the translation if possible. This is a beta release, so the update of translation is not mandatory.

From this version, missing strings in `Localizable.strings` will be appended to the file using default English translation, with a `FIXME` comment. Please pay attention to them. **If you are using the translation utility, please download the latest release ([0.1.3](https://github.com/iina/iina-translation-utility/releases/tag/0.1.3)) before starting.**

imKota:
@lhc70000, no problem :3

Greenek:
I can update Polish translation if Patryk is no longer available.

lhc70000:
Hi @Greenek, that would be great!

pradorocchi:
I will review and update pt-br

LEOYoon-Tsaw:
Sorry for the delay, I'm primarily using English now, and I think that using OpenCC for trad Chinese is quite viable.

PR #1558, opened 2018-03-16
Update Polish translation

Greenek:
- [x] It implements / fixes issue #1551.

```shell
➜ swift other/check_translation.swift pl
...
Finished. Issues count:
pl 0
```
---

**Description:**

The translation, except updating files, required rewriting most of it. Mostly because many labels were translated "literally", without the context, which made some options difficult to understand.

This PR does not contain all planned fixes but includes the most visible ones.


lhc70000:
Thanks!

PR #1565, opened 2018-03-19
Japanese translation for 1.0.0 beta1

matsune:
- [x] It implements / fixes issue #1551.

---

**Description:**
`check_translation.swift` reports that 10 keys don't exist, but they are not used now.


lhc70000:
Thanks!

PR #1579, opened 2018-03-25
Properly disable Initial window size controls on launch

ErikMinekus:

---

**Description:**

When ```Initial window size``` was unchecked, the related controls weren't disabled on launch, because it was disabling the controls for ```Initial window position``` instead.


alejx:
Thanks!

Issue #917, opened 2017-07-26
A-B Loop with logistical problem


tywtyw2002:
A-B Loop does not cancel when video finished.

Repeat steps:
1. Having a series of videos `s01.mp4, s02.mp4.......`
2. Click A-B Loop or `CMD+L` in 00:10:00 of the `s01.mp4`.
3. When `s01.mp4` finished, player jump to 00:10:00 of the `s01.mp4`.   <----- BUG 1
4. Close player windows, but do not close IINA.
5. Play `s02.mp4` and wait for it finish, player jump to 00:10:00 of the `s02.mp4`. <------ BUG 2.

BUG 1: It seem like, IINA not clean the A-B Loop `A` position when video is played over.
BUG 2: IINA did not clean the A-B Loop control, when player windows is closed or video changed.

PR #1604, opened 2018-04-01
Fix ab loop stays even when switching media files

braineo:
- [x] It implements / fixes issue #917 .

---

**Description:**

It seems that mpv retains the status of ab loop even when switching files in playlist or play next/prev. Added clearing ab loop when a new files is loaded.

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -447,9 +447,6 @@ class PlayerCore: NSObject {
     let a = mpv.getDouble(MPVOption.PlaybackControl.abLoopA)
     let b = mpv.getDouble(MPVOption.PlaybackControl.abLoopB)
     if a != 0 || b != 0 {
-      mpv.setDouble(MPVOption.PlaybackControl.abLoopA, 0)
-      mpv.setDouble(MPVOption.PlaybackControl.abLoopB, 0)
-      mpv.command(.abLoop)
       info.abLoopStatus = 0
     }
   }
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -444,10 +444,12 @@ class PlayerCore: NSObject {
   }
 
   func clearAbLoop() {
-    let a = mpv.getDouble(MPVOption.PlaybackControl.abLoopA)
-    let b = mpv.getDouble(MPVOption.PlaybackControl.abLoopB)
-    if a != 0 || b != 0 {
-      info.abLoopStatus = 0
+    if let resetOption = mpv.getString(MPVOption.ProgramBehavior.resetOnNextFile) {
+        if resetOption.contains(MPVOption.PlaybackControl.abLoopA) &&
+                   resetOption.contains(MPVOption.PlaybackControl.abLoopB) {
+          info.abLoopStatus = 0
+          sendOSD(.abLoop(info.abLoopStatus))
+      }
     }
   }
 
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -443,6 +443,17 @@ class PlayerCore: NSObject {
     sendOSD(.abLoop(info.abLoopStatus))
   }
 
+  func clearAbLoop() {
+    let a = mpv.getDouble(MPVOption.PlaybackControl.abLoopA)
+    let b = mpv.getDouble(MPVOption.PlaybackControl.abLoopB)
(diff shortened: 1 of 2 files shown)


braineo:
oh yea, interesting. That is much better. But still we need to clear the internal state of IINA somewhere when switching the file right?

braineo:
No problem. 😃 

So is it better to set the config when initializing IINA?

lhc70000:
Yes, I think it should be in `MPVControler#mpvInit()`, somewhere before [applying user defined options](https://github.com/lhc70000/iina/blob/develop/iina/MPVController.swift#L278) (so user can override this), maybe at line 277.

iina/PlayerCore.swift
```diff
@@ -447,9 +447,6 @@ class PlayerCore: NSObject {
     let a = mpv.getDouble(MPVOption.PlaybackControl.abLoopA)
     let b = mpv.getDouble(MPVOption.PlaybackControl.abLoopB)
     if a != 0 || b != 0 {
```
lhc70000: Is it still necessary to check these values? Can we just reset `info.abLoopStatus` to 0 without checking?

braineo: I'm looking for how to get the option string of IINA, because we do not know if user have overwritten the config value that whether to reset AB loop when opening next.

braineo: So yes I don't think we need to check the ab loop time value, and it is necessary to check the option value

lhc70000: That's a good point; if you want to cover all possible cases, it's possible that mpv loads ab loop status from watch_later files, or user controls ab loop via a lua script or a local socket.

So a perfect solution should listen to `ab-loop-a` and `ab-loop-b` changes and update corresponding status (probably send OSD as well), just like how IINA deals with a lot of other properties.

However I think there's very little chance for a user to override this option, so it's OK to update it only when file loaded.

I just found if you get `ab-loop-a` or `b` using `getFlag()`, it will return `false` when the loop point is not set, which I think is better than checking if it's zero; and then you can set `info.abLoopStatus` to 0, 1 or 2 depends on the two flags.

braineo:
Instead of checking current state with `getFlag` I used 'getString' to get the current setting of `mpv` so that it should behavior correctly.

iina/PlayerCore.swift
```diff
@@ -443,6 +443,16 @@ class PlayerCore: NSObject {
     sendOSD(.abLoop(info.abLoopStatus))
   }
 
+  func clearAbLoop() {
+    if let resetOption = mpv.getString(MPVOption.ProgramBehavior.resetOnNextFile) {
+        if resetOption.contains(MPVOption.PlaybackControl.abLoopA) &&
+                   resetOption.contains(MPVOption.PlaybackControl.abLoopB) {
+          info.abLoopStatus = 0
+          sendOSD(.abLoop(info.abLoopStatus))
```
lhc70000: Maybe I didn't make it clear. You don't need to send notification unless you listen to mpv's `ab-loop` change event and update IINA's state at runtime.

iina/PlayerCore.swift
```diff
@@ -443,6 +443,16 @@ class PlayerCore: NSObject {
     sendOSD(.abLoop(info.abLoopStatus))
   }
 
+  func clearAbLoop() {
+    if let resetOption = mpv.getString(MPVOption.ProgramBehavior.resetOnNextFile) {
+        if resetOption.contains(MPVOption.PlaybackControl.abLoopA) &&
+                   resetOption.contains(MPVOption.PlaybackControl.abLoopB) {
```
lhc70000: There's no problem with this approach, but I still prefer checking whether the two loop points are set (and set info.abLoopStatus to 0, 1, or 2 accordingly), because it also handles the case where A-B loop status is loaded from watch_later file by mpv.

iina/PlayerCore.swift
braineo: I took a look at the callback handle. in `MPVController::handleEvent`, seems when I set AB loop there is no event caught by readEvent

iina/PlayerCore.swift
braineo: Sure.

myxal:
Any idea when a build with this fix might come out? (Or are there any nightly builds? I'm having a different issue with looping in a released build, but suspect it might be fixed in latest code.)

PR #1610, opened 2018-04-04
Add argument for stdin piping to iina-cli help

snipem:

---

The piping capability was only described but the actual arguments where not shown. I only found out about this by looking at the code.

### DIFF
--- a/iina-cli/main.swift
+++ b/iina-cli/main.swift
@@ -54,12 +54,12 @@ if userArgs.contains(where: { $0 == "--help" || $0 == "-h" }) {
             Open all files in separate windows.
     --help | -h:
             Print this message.
-
+    --stdin:
+            You may also pipe to stdin directly.
+    
     MPV Option:
     Raw mpv options without --mpv- prefix. All mpv options are supported here.
     Example: --volume=20 --no-resume-playback
-
-    You may also pipe to stdin directly.
     """)
   exit(0)
 }


lhc70000:
No, you don't need to add `--stdin` to pipe. Try something like `cat /path/to/file | iina` and it should work.

snipem:
@lhc70000 I'm afraid it does only work when using the `--stdin` argument:

```
~  $ iina --help
IINA 0.0.15 Build 68
This binary is not intended for being used as a command line tool. Please use the bundled iina-cli.
Please ignore this message if you are running in a debug environment.
^C
~  $ curl --silent http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4 | iina-cli
^C
~  $ curl --silent http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4 | iina-cli --stdin
```

The second to last command does nothing besides opening the IINA file chooser. The last command with the `--stdin` argument does play the video piped over curl.

snipem:
I read through your response and saw that you mentioned `iina` instead of `iina-cli`. The result is the same:

```
~  $ curl --silent http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4 | iina  
^C
```

Only the IINA file chooser is opening.

ErikMinekus:
```cat /path/to/file | iina-cli``` works with a 12MB MP3 file, but not with a 15MB MP3 file. So it seems that if the file is larger than some unknown buffer size, or if the file size is unknown, the file is not fully passed to IINA.

lhc70000:
@snipem IINA cannot detect whether stdin has data when you pipe from an online resource, so yes, this is where `--stdin` should be used (and I forgot it). 

Could you update the help message accordingly?

PR #1640, opened 2018-04-21
Make `FontPickerWindowController` shows up quicker.

alejx:
**Description:**
When opening font picker window, there is a noticeable delay (~0.7s, ~260 fonts) until it shows up. Based on my investigation, this maybe caused by `NSFontManager.localizedName(forFamily: String, face: String?)`. Seems to be inevitable, so that in order to prevent that delay, I move the code block where calls this method to `PrefSubViewController` and call it in an asynchronous way.


lhc70000:
I don't think putting the logic in `PrefSubViewController` is a good choice. In most cases loading the font list is not necessary until user actually clicked the "select font" button; plus, the font picker is also used somewhere else like the sidebar so you'll need to change them all.

PR #1656, opened 2018-04-29
Add function to select None subtitle in menu, fix #1653

### LINKED ISSUE #1653: Selecting “<None>” in OSC’s subtitles menu has no effect
I have configured my OSC’s right buttons to include the “Choose subtitle track” button. That button pops a menu (as opposed to the right pane). Selecting a subtitle there works, however selecting “None” as the subtitle has no effect.

screen shot 2018-04-28 at 12 18 13

alejx:
- [x] It implements / fixes issue #1653 .

---

**Description:**
Add function to select None subtitle in menu; use "No subtitle loaded" and "No external subtitle found" instead of "\<None\>" when no suitable subtitle.

Before: screen shot 2018-04-28 at 12 18 13

After: image
Click "\<None\>" here now switch to \<None\>.

In case of there's no loaded subtitle and no external found:
image

### DIFF
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -60,7 +60,7 @@
 "osd.video_eq.saturation" = "Saturation: %i";
 "osd.video_eq.brightness" = "Brightness: %i";
 "osd.find_online_sub" = "Finding online subtitles...";
-"osd.sub_not_found" = "No subtitle found";
+"osd.sub_not_found" = "No subtitles found";
 "osd.sub_found" = "%d subtitle(s) found. Downloading...";
 "osd.sub_downloaded" = "Subtitle downloaded";
 "osd.sub_saved" = "Subtitle saved";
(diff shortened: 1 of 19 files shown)


alejx:
CC @LeoNatan

iina/Base.lproj/Localizable.strings
```diff
@@ -109,6 +109,9 @@
 
 "track.none" = "<None>";
 
+"subtrack.no_loaded" = "No subtitle loaded";
```
saagarjha: "Subtitles" is generally plural when used in this sense.

alejx: Thanks! Just curious, since I was taught usually No + noun singular form, why we should use plural here?

saagarjha: It's just convention; I'd assume it's because you never really just have one subtitle. Kind of like train tracks? See dict://subtitle. By the way, don't you need to update every instance of this, rather than just the ones you've done here?

alejx: So it's because "subtitle" itself is usually in plural form, right?
>  By the way, don't you need to update every instance of this, rather than just the ones you've done here?

Those ones were generated by script, so I forgot them, sorry!

LeoNatan:
https://developer.apple.com/macos/human-interface-guidelines/menus/menu-anatomy/

> Use title-style capitalization. This style is used consistently for menus and menu item titles throughout the system.

LeoNatan:
Thank you!

PR #1678, opened 2018-05-06
Search subtitles online automatically (#639)

alejx:
Due to there will be a UI change in Preferences panel, no GUI work in this PR.


alejx:
@lhc70000 Updated.

PR #1691, opened 2018-05-11
Minor refactorings

adamnemecek:
Just minor refactorings.

### DIFF
--- a/iina/AppData.swift
+++ b/iina/AppData.swift
@@ -88,14 +88,14 @@ struct Constants {
 }
 
 extension Notification.Name {
-  static let iinaMainWindowChanged = Notification.Name("IINAMainWindowChanged")
-  static let iinaPlaylistChanged = Notification.Name("IINAPlaylistChanged")
-  static let iinaTracklistChanged = Notification.Name("IINATracklistChanged")
-  static let iinaMediaTitleChanged = Notification.Name("IINAMediaTitleChanged")
-  static let iinaVFChanged = Notification.Name("IINAVfChanged")
-  static let iinaAFChanged = Notification.Name("IINAAfChanged")
-  static let iinaKeyBindingInputChanged = Notification.Name("IINAkeyBindingInputChanged")
-  static let iinaFileLoaded = Notification.Name("IINAFileLoaded")
-  static let iinaHistoryUpdated = Notification.Name("IINAHistoryUpdated")
-  static let iinaLegacyFullScreen = Notification.Name("IINALegacyFullScreen")
+  static let MainWindowChanged = Notification.Name("IINAMainWindowChanged")
+  static let PlaylistChanged = Notification.Name("PlaylistChanged")
+  static let TracklistChanged = Notification.Name("TracklistChanged")
+  static let MediaTitleChanged = Notification.Name("MediaTitleChanged")
+  static let VFChanged = Notification.Name("VFChanged")
+  static let AFChanged = Notification.Name("AFChanged")
+  static let KeyBindingInputChanged = Notification.Name("KeyBindingInputChanged")
+  static let FileLoaded = Notification.Name("FileLoaded")
+  static let HistoryUpdated = Notification.Name("HistoryUpdated")
+  static let LegacyFullScreen = Notification.Name("LegacyFullScreen")
 }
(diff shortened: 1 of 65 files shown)


iina/PlaybackHistory.swift
```diff
@@ -63,7 +63,7 @@ class PlaybackHistory: NSObject, NSCoding {
   init(url: URL, duration: Double, name: String? = nil) {
     self.url = url
     self.name = name ?? url.lastPathComponent
-    self.mpvMd5 = Utility.mpvWatchLaterMd5(url.path)
```
lhc70000: mpv has a little more logic to calculate file hashes, especially:
- When the file is a network URL,
- Or user specified `--ignore-path-in-watch-later-config`, in which case the md5 is calculated from the filename only.

To stay consistent with mpv, IINA should implement all these logics. It's not implemented yet (all `Utility.mpvWatchLaterMd5` do is returning the md5 of the path), but probably will be in the future. So I suggest still using `Utility.mpvWatchLaterMd5` for better compatibility.

adamnemecek: Having a Utility type is an anti-pattern. I'm not sure I entirely follow the use case. I guess I'll setup an MD5Hash struct and use that? It will have multiple constructors which will all hash the argument appropriately.

adamnemecek: If you could merge this so that it doesn't get stale. I'll submit the hash thing very soon.

iina/AppData.swift
```diff
@@ -88,14 +88,20 @@ struct Constants {
...
-  static let iinaPlaylistChanged = Notification.Name("IINAPlaylistChanged")
-  static let iinaTracklistChanged = Notification.Name("IINATracklistChanged")
-  static let iinaMediaTitleChanged = Notification.Name("IINAMediaTitleChanged")
-  static let iinaVFChanged = Notification.Name("IINAVfChanged")
-  static let iinaAFChanged = Notification.Name("IINAAfChanged")
-  static let iinaKeyBindingInputChanged = Notification.Name("IINAkeyBindingInputChanged")
-  static let iinaFileLoaded = Notification.Name("IINAFileLoaded")
-  static let iinaHistoryUpdated = Notification.Name("IINAHistoryUpdated")
-  static let iinaLegacyFullScreen = Notification.Name("IINALegacyFullScreen")
+  static let MainWindowChanged = Notification.Name("IINAMainWindowChanged")
```
lhc70000: Seems that you forgot to change this line. Also please use camel case for static member name.

iina/DraggingDetect.swift
```diff
@@ -158,10 +158,10 @@ extension PlayerCore {
   }
 
   /**
-   Get called for all drag-and-drop enabled window/views in their `draggingEntered(_:)`.
+   Get called for all drag-and-drop enabled window/views in their `draggingEntered`.
```
saagarjha: The correct name for this method is `draggingEntered(_:)`.

iina/Extensions.swift
```diff
@@ -292,12 +284,9 @@ extension NSColor {
         self.init(red: CGFloat(splitted[0]!), green: CGFloat(splitted[1]!), blue: CGFloat(splitted[2]!), alpha: CGFloat(1))
       } else if splitted.count == 4 {  // if has alpha value
         self.init(red: CGFloat(splitted[0]!), green: CGFloat(splitted[1]!), blue: CGFloat(splitted[2]!), alpha: CGFloat(splitted[3]!))
-      } else {
-        return nil
       }
-    } else {
-      return nil
     }
+    return nil
```
saagarjha: One case where I think this change might be justified.

iina/MainWindowController.swift
```diff
@@ -683,7 +679,7 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
       button.bezelStyle = .regularSquare
       button.isBordered = false
       button.image = buttonType.image()
-      button.action = #selector(self.toolBarButtonAction(_:))
+      button.action = #selector(self.toolBarButtonAction)
```
saagarjha: To be clear: this is fine. However, the name of the selector is still `MainWindowController.toolBarButtonAction(_:)`.

iina/VideoTime.swift
```diff
@@ -16,33 +16,25 @@ class VideoTime {
...
   var stringRepresentation: String {
-    get {
-      if self == Constants.Time.infinite {
-        return "End"
-      }
-      let ms = (m < 10 ? "0\(m)" : "\(m)")
-      let ss = (s < 10 ? "0\(s)" : "\(s)")
-      let hs = (h > 0 ? "\(h):" : "")
-      return "\(hs)\(ms):\(ss)"
+    if self == Constants.Time.infinite {
```
saagarjha: While you're at it, would you mind changing the check here to a `guard`?

lhc70000:
Closing due to too many conflicts. Let's refactor after v1.2 when finished all pending features.

Issue #1695, opened 2018-05-13
Flashing Blue Screen when Playing Videos


WANGJIEKE:
**System and IINA version:**

- macOS 10.13.4
- IINA 0.0.15.1

**Expected behavior:**
Play the video normally.

**Actual behavior:**
When the setting "Preference -> Codec -> Video -> Hardware Decoder" is set to "Auto" and the file is a .mkv file with HEVC encoding, the video flashes between blue screen and the content of the video. If the hardware decoder is disabled, the problem disappears.
当播放HEVC编码的mkv文件，且硬件解码器被设置为“自动”时，画面在蓝屏和实际视频内容之间不断闪烁。如果将硬件解码器设置成“禁用”，问题消失。

**Steps to reproduce:**
1. Set the Hardware Decoder to "Auto"
2. Play a .mkv file whose encoding is HEVC

- [ ] MPV does not have this problem.

**How often does this happen?**
Always, if the file is a .mkv file with HEVC encoding

lhc70000:
It should be related to FFmpeg 4.0 / mpv 0.28.

There're similar reports in https://github.com/mpv-player/mpv/issues/5755.

jerkstore369:
Link is dead?

tmm1:
I have a hacky fix in http://ffmpeg.org/pipermail/ffmpeg-devel/2018-May/229483.html, but would be better to figure out the root cause.

saagarjha:
Since this is an upstream issue, we're unlikely to merge it in here directly. I think the best course of action is to allow FFmpeg to come to consensus on how to resolve this issue, then wait for the changes to percolate downstream.

tmm1:
Here is an updated version of the patch if anyone would like to help test it: https://github.com/tmm1/FFmpeg/commit/d5624b0401b6d65de9cdfd726f5c6e7a1496a494

If it works, I will commit to FFmpeg master.

tmm1:
The fix has been merged to FFmpeg master: https://github.com/FFmpeg/FFmpeg/commit/8f146b526ff8d63adc02e1c5db15850f4589230b

danshao:
Still getting this issue with the 0.0.15.1 Build 69. Does not occur while using mpv.

inflation:
@danshao I believe build 69 does not build from ffmpeg HEAD, try build 80.

frost134:
Getting the issue too, totally the same situation.

lhc70000:
[...]

PR #1712, opened 2018-05-18
Update mpv for beta 2 release

alejx:


lhc70000:
Let's use FFmpeg HEAD since #1695 is fixed.

PR #1729, opened 2018-05-27
Fix #1720 by rewriting FFmpegController in Swift

### LINKED ISSUE #1720: Memory leaks.
**System and IINA version:**

- macOS 10.13.4
- IINA 0.0.15.1

**Expected behavior:**

No leaks.

**Actual behavior:**

**Steps to reproduce:**

Play a video without cached thumbnails.

Every time a thumbnail is generated.

saagarjha:
- [X] It implements / fixes issue #1720.

---

This is a partial fix for #1720. I'm asking for a review because this isn't code that I've touched before so I'd like someone else to take a look at it.

I mention this being a "partial" fix because there's still a memory leak if `CHECK_SUCCESS` fails, since C does not have a defer equivalent. I'd really much rather have this rewritten in Swift, so if that's an option please let me know and I'll convert it.


lhc70000:
How about using `goto`? It's common idiom for early exit and cleanup in C.

saagarjha:
I don't think I could live with myself if I had to write a `goto`. But jokes aside, what's wrong with writing this is Swift?

lhc70000:
`goto` is not that evil as long as you use it in the correct way. 

I thought writing in Obj-C is more simple, clean and easy to read/maintain when interacting with complicated C API (at least no more `UnsafeutablePointer`s) and Swift's ability of bridging Obj-C code is something we can make use of rather than to avoid. However I never actually tried.

You can have a try if you would like, and I'll merge if it doesn't look so bad.

saagarjha:
@lhc70000 Done. I've rewritten all of FFmpegController.h/m in Swift and moved it to FFmpegController.swift. In doing so, I also did a run through and cleaned up some naming and modernized some of the external interfaces (I left most of the internal code intact, though I did find some *other* issues such as a use-after-free in `+[FFmpegController probeVideoDurationForFile:]` and a couple missing deinitialization calls or failure to check return values in `-[FFmpegController getPeeksForFile:]`. I'd suggest going line-by-line and comparing it to the old code to make sure I didn't miss anything; it should map pretty much one-to-one). Some of the code dealing with `UnsafePointer` is slightly awkward, but I feel that it isn't *too* bad for the increase in safety we get. Feel free to ask me if anything seems amiss, but fair warning: some of the seemingly unecessarily verbose code must be done a certain way as far as I'm aware to ensure that we do not do anything undefined. If we need to interact with the FFmpeg API in the future, I would suggest that we write a good, well annotated wrapper, but as for now I think just using the C API is fine.

saagarjha:
@lhc70000 were you planning on looking at this, or should I merge it in?

lhc70000:
Actually, I'm not sure.

FFmpeg (libav*) is a huge and complex library, and using it requires deep understanding of multimedia technology. I'm not interested in writing a Swift wrapper for it, or frankly speaking, that will be a huge project and is beyond my abilities. There's nothing wrong with your implementation, and I really like the use of `defer`. However it's definitely more obscured and complicated than the Objective-C version, especially at around [L192](https://github.com/lhc70000/iina/pull/1729/files#diff-30a625e609588195414a112d242a03ebR192), like writing 6 lines just to call one method.

We do need to add more code that calls FFmpeg directly (Quick Look plugin, etc) in the future, and I'm afraid that migrating to Swift could bring extra difficulty for both maintainers and potential contributors. We should concentrate on writing business logic but not converting between `UnsafePointer`s. Swift provides us the ability to bridge C/Obj-C code, but the mechanism is rather awkward, especially when dealing with pointers. IINA has a wrapper for mpv API, but again, FFmpeg is a large, way more complicated and fast-changing C library. So from my point of view, the best way is still implement intensive C API calls in Obj-C, and expose simple, `UnsafePointer` free interface to Swift. To fix this issue (#1720), there's no elegant like `defer` but some commonly used solutions do exist in C/Obj-C.

@alejx @inflation @HarukaMa What do you think?

saagarjha:
Even if you decide to throw out the Swift code, I still feel that some parts of this pull request should be looked at seperately. Namely, the additional bugs I found in memory management and refactoring/renaming of some of the thumbnail API stand on their own and are unrelated to whether this is written in Swift or Objective-C. And now on to my argument of why this should be in Swift:

> FFmpeg (libav*) is a huge and complex library, and using it requires deep understanding of multimedia technology. I'm not interested in writing a Swift wrapper for it, or frankly speaking, that will be a huge project and is beyond my abilities.

We don't have to wrap all of FFmpeg–or even wrap it at all. Even good API annotations (e.g. nullability) on the methods we use would go a long way.

> However it's definitely more obscured and complicated than the Objective-C version, especially at around L192, like writing 6 lines just to call one method.

Well, the tradeoff we have here is an improvement over memory-safety bugs that C/Objective-C has versus the verbosity that Swift brings. Personally, and as I'm sure you know, I favor the approach that uses Swift, since I think it saves us time debugging issues later. Sure, we get code written quickly _now_, but we pay for it later.

> We do need to add more code that calls FFmpeg directly (Quick Look plugin, etc) in the future, and I'm afraid that migrating to Swift could bring extra difficulty for both maintainers and potential contributors. 

My view was the opposite: if we continue to write code in Objective-C people will contribute code that may have bugs because they don't understand the subtleties of undefined behavior or memory management coming from languages like Swift. Also, as a heads up: QuickLook plugins cannot be written in Swift currently due to ABI stability requirements. This may be something that might change in the future, either as 1. Swift gains ABI stability or 2. Apple deprecates the QuickLook plugin API and moves to an app extension model.

> To fix this issue (#1720), there's no elegant like defer but some commonly used solutions do exist in C/Obj-C.

What's your opinion on C++/Objective-C++?

lhc70000:
> Namely, the additional bugs I found in memory management and refactoring/renaming of some of the thumbnail API stand on their own and are unrelated to whether this is written in Swift or Objective-C.

Please fell free to fix them when we finished this discussion.

> Well, the tradeoff we have here is an improvement over memory-safety bugs that C/Objective-C has versus the verbosity that Swift brings.

In this case, we do nothing other than calling API. The security issue happened here is that I missed to call some destructors. This can happen in any language, and it has nothing to do with which language we use. Talking about safety, I think when we start to use `UnsafePointer`, it's already unsafe and unmanaged. Without the protection of Swift, there's only very limited advantages:
- Nullability. It's not necessary, if a function returns non-zero, we know something happened. If a function returns success but some pointer is actually `NULL` (although very unlikely to happen), that's not our fault.
- `defer`. Like I said before, it's just like a syntax sugar, and that's not something we can't do in C/Obj-C.

> if we continue to write code in Objective-C people will contribute code that may have bugs because they don't understand the subtleties of undefined behavior or memory management coming from languages like Swift.

That's a different topic. We can't assume that every potential contributor of this project knows Swift well or has little knowledge about C/Obj-C. Plus, it's our responsibility to ensure the code quality before merging.

> What's your opinion on C++/Objective-C++?

No problem, as long as it compiles and works.

saagarjha:
> It's not necessary, if a function returns non-zero, we know something happened.

Not necessarily: some of the API returns `NULL` itself.

> Like I said before, it's just like a syntax sugar, and that's not something we can't do in C/Obj-C.

It would be hard to do in pure C or Objective-C.

> The security issue happened here is that I missed to call some destructors. This can happen in any language, and it has nothing to do with which language we use.

Well, some of the calls were not just *missed*: there were execution paths that didn't call the destructors, even though they were there at the bottom of the function. Without `defer` I really think that bugs like these are inevitable. I think `guard` with `Optional` is also a great construct because it forces you to check return values and encodes this in the type system.

inflation:
Swift is not a garbage collecting language. The memory management is essentially the same with Objective-C, i.e. ARC. So just to be clear, any discussion is with C not Objective-C. 

Dealing with C APIs in Swift is such a hassle. Those unnecessary boilerplate code are what Swift want to prevent at the first place, right? Memory safety is all your responsibility when you communicate with external C APIs. That doesn’t change whichever language you are using. So I’m with code simplicity and clarity here. 

As for code quality, what we lack are proper unit testings and profiling. All in Swift may ease the burden for new contributors, but that doesn’t improve code quality, especially when Swift can’t protect you.

lhc70000:
> Well, some of the calls were not just missed: there were execution paths that didn't call the destructors, even though they were there at the bottom of the function. Without defer I really think that bugs like these are inevitable. 

It shouldn't be "inevitable": if so, how can those C projects without defer survive till today?

saagarjha:
> Swift is not a garbage collecting language. The memory management is essentially the same with Objective-C, i.e. ARC. So just to be clear, any discussion is with C not Objective-C.

Sure, but Objective-C lets you opt out of ARC easily by dropping down to C, while Swift requires explicit approval.

> Dealing with C APIs in Swift is such a hassle. Those unnecessary boilerplate code are what Swift want to prevent at the first place, right? Memory safety is all your responsibility when you communicate with external C APIs. That doesn’t change whichever language you are using. So I’m with code simplicity and clarity here.

To be clear: when I was talking about memory safety here, I was talking about the programming paradigms that Swift encourages rather than the syntax of the language. I'm going to repeat the same, tired argument: Swift has `defer` and `guard`, which makes it extremely clear that you are cleaning up resources on *every* execution path. Plus, it's always easy to see when you've opted out of memory safety: you get an `UnsafePointer` immediately. What you see as verbosity, I see as a clear "warning, this code performs unsafe behavior". And again, some of the superficial verbosity can be reduced with a simple apinotes file.

> As for code quality, what we lack are proper unit testings and profiling.

This is something that I'd like to work on in the future, and it might dovetail with the discussion about a new build system since it would enable us to compile with a lot more debugging tools enabled.

> how can those C projects without defer survive till today?

Well, there's an argument that they don't: they're insecure or crash or leak memory ;)

HarukaMa:
Calling C codes from Swift IS more troublesome comparing with from Obj-C code, and maintaining the code generally (not limited to safe comcerns) is more difficult. Except a few patterns, it only incerases the code complexity. I still think that the code safety should not solely rely to some language features. It is still the code author who finally determine how safe the code is.

> Swift has defer and guard, which makes it extremely clear that you are cleaning up resources on every execution path. 

I think for people who doesn't use them well, Swift would contribute nothing to the memory safety.

PR #1764, opened 2018-06-04
Revised Logging System

lhc70000:
Updated from #1726.

### DIFF
--- a/iina/Logger.swift
+++ b/iina/Logger.swift
@@ -8,10 +8,20 @@
 
 import Foundation
 
-class Logger {
+struct Logger {
 
-  enum LogLevel: Int, Comparable, CustomStringConvertible {
-    static func < (lhs: LogLevel, rhs: LogLevel) -> Bool {
+  struct Subsystem: RawRepresentable {
+    var rawValue: String
+
+    static let general = Subsystem(rawValue: "iina")
+
+    init(rawValue: String) {
+      self.rawValue = rawValue
+    }
+  }
+
+  enum Level: Int, Comparable, CustomStringConvertible {
+    static func < (lhs: Logger.Level, rhs: Logger.Level) -> Bool {
       return lhs.rawValue < rhs.rawValue
     }
 
@@ -20,9 +30,7 @@ class Logger {
     case warning
     case error
 
-    static var preferred: LogLevel {
-      return LogLevel(rawValue: Preference.integer(for: .logLevel).clamped(to: 0...3))!
-    }
+    static var preferred: Level = Level(rawValue: Preference.integer(for: .logLevel).clamped(to: 0...3))!
 
     var description: String {
       switch self {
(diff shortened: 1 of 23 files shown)


iina/Logger.swift
```diff
@@ -8,10 +8,20 @@
 
 import Foundation
 
-class Logger {
+struct Logger {
 
-  enum LogLevel: Int, Comparable, CustomStringConvertible {
-    static func < (lhs: LogLevel, rhs: LogLevel) -> Bool {
+  struct Subsystem: RawRepresentable {
```
saagarjha: I'd prefer if this was an enum; is there any particular reason why you went with a struct? Plus, I'd add an `init(_ rawValue: String)` constructor.

iina/Logger.swift
```diff
@@ -8,10 +8,20 @@
...
+
+    static let general = Subsystem(rawValue: "iina")
+
+    init(rawValue: String) {
+      self.rawValue = rawValue
+    }
+  }
+
+  enum Level: Int, Comparable, CustomStringConvertible {
+    static func < (lhs: Logger.Level, rhs: Logger.Level) -> Bool {
```
saagarjha: Does dropping the `Logger` qualifier work here?

iina/Logger.swift
```diff
@@ -20,9 +30,7 @@ class Logger {
     case warning
     case error
 
-    static var preferred: LogLevel {
-      return LogLevel(rawValue: Preference.integer(for: .logLevel).clamped(to: 0...3))!
-    }
+    static var preferred: Level = Level(rawValue: Preference.integer(for: .logLevel).clamped(to: 0...3))!
```
saagarjha: I think it's more logical for the clamping to occur in `Level`'s constructor, so we don't have to depend on `Level`'s clients to behave correctly.

iina/Logger.swift
```diff
@@ -34,11 +42,8 @@ class Logger {
     }
   }
 
-  var label: String
-  var level: LogLevel
-  var logToConsole = false
-
   static let enabled = Preference.bool(for: .enableLogging)
```
saagarjha: I'm still not sold on this, versus a "disabled" log level. Could you explain this to me?

iina/Logger.swift
lhc70000: There's nothing to enumerate; each component should create their own subsystem and their names are not fixed. I think it's more like a `NSImage.Name` or `NSNib.Name`.

iina/Logger.swift
saagarjha: Sure, if you have a good reason for it then go ahead and keep it.

iina/Logger.swift
saagarjha: > I think it's more like a NSImage.Name or NSNib.Name.

That's the API I was aiming for. The reason why this isn't a struct as those are is because LogLevel cannot have an arbitrary value as those do.

iina/Logger.swift
saagarjha: Use Swift 4.2, which has [SE-0194](https://github.com/apple/swift-evolution/blob/master/proposals/0194-derived-collection-of-enum-cases.md).

lhc70000: Then let's put it on `mojave` branch.

PR #1767, opened 2018-06-05
Respect "Show remaining time" preference in Music Mode

ErikMinekus:

---

**Description:**
When changing the "Show remaining time" preference, the change is immediately visible in the OSC. However, after restarting IINA, Music Mode always shows the duration.


alejx:
Thanks!

ErikMinekus:
I forgot to mention that the text is cut off when the remaining time is over an hour, so maybe someone can look into the label width. Monospaced numbers in DurationDisplayTextField would also be nice.

PR #1775, opened 2018-06-09
Fix mouse tracking issue in macOS 10.14

alejx:
In mojave, titlebar seems overlap content view for some reason; If we still add tracking area on main window's content view, then the system will call mouseExit when we move the mouse to the titlebar. Adding tracking area on content view's super view can fix this.


alejx:
This pr fixes #1779.

saagarjha:
I'm not sure this is the right fix for this issue. I don't have time to look at this more in detail right now, but I'd advise against merging this until we can figure out what the underlying issue is.

Issue #1377, opened 2018-01-16
Use monospaced fonts for timepos


SLboat:
否则它会一直变动滚动条位置-Mac 10.13,中文
![image](https://user-images.githubusercontent.com/4231376/34993971-02b6c544-fb0d-11e7-8e80-95386649d892.png)

SLboat:
Arial不错

HarukaMa:
@SLboat Mac 哪来 Arial……

SLboat:
Mac OS X was the first Mac OS version to include Arial; it was not included in classic Mac OS. The operating system ships with Arial, Arial Black, Arial Narrow, and Arial Rounded MT. However, the default Mac OS X font for sans-serif/Swiss generic font family is Helvetica. The bundling of Arial with Windows and Mac OS X has contributed to it being one of the most widely distributed and used typefaces in the world.

SLboat:
osx开始有的

ErikMinekus:
This was fixed in #1777 and can be closed.

PR #1777, opened 2018-06-09
Monospaced font for time labels in OSC and Touch Bar

ErikMinekus:
- [x] It implements / fixes issue #1377.

---

**Description:**
I couldn't find a constant for the System Mini font size (9pt) for the mini player, so it's hardcoded.

### DIFF
--- a/iina/Base.lproj/MainWindowController.xib
+++ b/iina/Base.lproj/MainWindowController.xib
@@ -389,6 +389,9 @@
                         <color key="textColor" name="textColor" catalog="System" colorSpace="catalog"/>
                         <color key="backgroundColor" name="textBackgroundColor" catalog="System" colorSpace="catalog"/>
                     </textFieldCell>
+                    <connections>
+                        <binding destination="-2" name="font" keyPath="monospacedFont" id="uHM-ya-PZy"/>
+                    </connections>
                 </textField>
                 <textField horizontalHuggingPriority="251" verticalHuggingPriority="750" alphaValue="0.5" allowsCharacterPickerTouchBarItem="YES" translatesAutoresizingMaskIntoConstraints="NO" id="GgV-Cc-sk0" userLabel="Right Label" customClass="DurationDisplayTextField" customModule="IINA" customModuleProvider="target">
                     <rect key="frame" x="450" y="5" width="50" height="14"/>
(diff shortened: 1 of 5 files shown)


saagarjha:
I could have sworn we've had other pull requests to this effect…

ErikMinekus:
#862 was closed by the author and never merged. I can put the code in ```DurationDisplayTextField``` if you can tell me which function to use. It didn't work for me in the init function.

ErikMinekus:
I tried
```swift
override var font: NSFont? = .monospacedDigitSystemFont(ofSize: font!.pointSize, weight: .regular)
```
but it returned these errors:
- Cannot override with a stored property 'font'
- Variable used within its own initial value

I did however just find ```viewWillDraw```:
```swift
  override func viewWillDraw() {
    if #available(OSX 10.11, *), let currentFont = font {
      font = .monospacedDigitSystemFont(ofSize: currentFont.pointSize, weight: .regular)
    }
  }
```
Which works great for the main window and mini player, but not for the Touch Bar. That's because the font property returns the system font (San Francisco UI size 13), but the actual font used in the Touch Bar is San Francisco Condensed size 15.

saagarjha:
You need to override it with a computed property:

```swift
override var font: NSFont? {
	get {
		return .monospacedDigitSystemFont(ofSize: 0, weight: .regular)
	}
	set {
	}
}
```

ErikMinekus:
Thanks, that works for the UI. The Touch Bar still renders the text at size 13 though, instead of size 15. Even though it works fine when I remove the font override and use the same code in the ```TouchBarSupport``` class:
```swift
label.font = .monospacedDigitSystemFont(ofSize: 0, weight: .regular)
```

ErikMinekus:
That code compiles, but it still uses the font set in Interface Builder, without monospacing.

alejx:
@saagarjha Is there any benefit to override `font` in `DurationDisplayTextField`? Though `DurationDisplayTextField`s have same font, they have different font size. So that we have to set their size separately.

ErikMinekus:
@alejx I imagine the benefit is to get the font size that's set in Interface Builder, instead of hardcoding it like I have done now. But so far overriding the font in ```DurationDisplayTextField``` seems to actually cause issues with the different font sizes.

saagarjha:
> That code compiles, but it still uses the font set in Interface Builder, without monospacing. It also doesn't solve the font size issue in the Touch Bar.

Yeah, now that I think about it that won't work. I think your solution is OK.

> That's because the font property returns the system font (San Francisco UI size 13), but the actual font used in the Touch Bar is San Francisco Condensed size 15.

Wait, what? How are you checking this?

ErikMinekus:
It's mentioned in the Advanced Touch Bar session from WWDC 2017: https://developer.apple.com/videos/play/wwdc2017/222/?time=2416

After that they explain using the system font of size 0 to get the right font, and how to set the monospaced font.

alejx:
@saagarjha Can I merge this in?

PR #1838, opened 2018-07-16
Respect theme settings in additionalInfoView 

Alex-Sokolov:
All float views respect theme settings, but additionalInfoView all time light-gray. For example:

This PR fix that


alejx:
Thanks!

Issue #1852, opened 2018-07-22
IINA does not switch to music mode when playing music from a URL


ErikMinekus:
**System and IINA version:**

- macOS 10.13.6
- IINA 1.0.0-beta2

**Expected behavior:**
It would be nice if IINA switched to music mode when playing music or podcasts from a URL.

**Actual behavior:**
IINA doesn't switch to music mode, because the ```checkCurrentMediaIsAudio``` function explicitly returns ```.notAudio``` for URLs. I didn't see an explanation in the commit messages for why this is the case, and when I comment the guard it seems to work as expected.

**Steps to reproduce:**
1. Open a URL, for example http://9to5mac.com/wp-content/uploads/sites/6/2018/07/9to5Mac-Daily-07-20-2018.mp3.

**How often does this happen?**
Always

alejx:
You are welcomed to make a pull request on this issue!

PR #1857, opened 2018-07-23
Automatically switch to Music Mode for audio URLs

ErikMinekus:
- [x] It implements / fixes issue #1852.

---

**Description:**
I don't believe this guard is needed (anymore?). The rest of the function works fine for URLs.


alejx:
Seems I found why we placed that line here. For some reason, mpv will firstly get the audio track from a network resource (e.g. a YouTube video), so that the url will be regarded as audio. There should be better solution than this. Revert this PR.

Issue #1840, opened 2018-07-17
Localization for 1.0.0 beta3


alejx:
- [x] fr @superzazu
- [x] it @cgand
- [ ] de @J-rg @shorshe 
- [x] ko @darkbrow
- [ ] ru @euf @imKota @Semty
- [ ] ja @matsune
- [ ] tr @eminhaqi
- [x] nl @ErikMinekus
- [ ] pl @Greenek
- [x] es @Aikyuichi
- [ ] uk @sevenfourk
- [x] sk @Marvin99com
- [ ] da @thkir
- [ ] pt-BR @pradorocchi
- [x] en -
- [x] zh-Hans -
- [x] zh-Hant -

Dear translators, please update the translation if possible. This is a beta release, so the update of translation is not mandatory.

For this version, missing strings in `Localizable.strings` will be appended to the file using default English translation, with a `FIXME` comment. Please pay attention to them. **If you are using the translation utility, please download the latest release ([0.1.3](https://github.com/iina/iina-translation-utility/releases/tag/0.1.3)) before starting.**

alejx:
Note: `PrefUtilsViewController.xib` is newly added in this beta version, don't forget to localize it.

PR #1890, opened 2018-08-03
IT translate FIXME to italian

cgand:
Translate fixme to italian and few refinements.
Note: All FIXME already present in localizable strings.


alejx:
Ref: #1840

PR #1917, opened 2018-08-16
Support for Swift 4.2 and Mojave

lhc70000:
**Description:**

- Updated to Swift 4.2
- Not perfect but full dark mode support (actually a full "light mode" was added as well), with some UI fine-tunes (which will be enabled in 10.14 only)
- Fixed some layout issues

This branch should be merged ASAP as we are going to ship Mojave support in beta 4. All our following development should base on this branch.

We should also update build instructions to require Xcode 10 beta for building and address [the fix needed for cocoapods](https://github.com/CocoaPods/CocoaPods/pull/7720/files).


iina/MainWindowController.swift
```diff
@@ -1863,7 +1872,11 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
     // prerequisites
     guard let window = window else { return }
 
-    window.backgroundColor = NSColor(calibratedWhite: 0.1, alpha: 1)
+    if #available(macOS 10.14, *) {
+      window.backgroundColor = .windowBackgroundColor
+    } else {
+      window.backgroundColor = NSColor(white: 0.1, alpha: 1)
```
alejx: `calibratedWhite`?

PR #1922, opened 2018-08-17
Show playlist when a dragging cursor stop at right side

### LINKED ISSUE #1848: Show playlist when the user dragging files into the window

alejx:
Fix #1848

### DIFF
--- a/iina/VideoView.swift
+++ b/iina/VideoView.swift
@@ -32,6 +32,13 @@ class VideoView: NSView {
 
   var uninitLock = NSLock()
 
+  var draggingTimer: Timer?
+  var triggered: Bool = false
+
+  var lastPosition: NSPoint?
+
+  var hasPlayableFiles: Bool = false
+
   // MARK: - Attributes
 
   override var mouseDownCanMoveWindow: Bool {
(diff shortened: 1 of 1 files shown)


iina/VideoView.swift
```diff
@@ -94,11 +101,65 @@ class VideoView: NSView {
...
+      let nowIn = inTriggerArea(position)
+      let lastIn = inTriggerArea(lastPosition)
+      if nowIn && !lastIn {
+        createTimer()
+      } else if nowIn && lastIn {
+        destroyTimer()
+        createTimer()
+      } else if !nowIn && lastIn {
+        destroyTimer()
+      }
```
lhc70000:
```swift
if lastIn {
  destroyTimer()
}
if nowIn {
  createTimer()
}
```

iina/VideoView.swift
```diff
@@ -94,11 +101,65 @@ class VideoView: NSView {
...
+    }
+    lastPosition = position
```
lhc70000: You don't need this assignment if `position == lastPosition`. It should be moved to the `if` block above.

iina/VideoView.swift
```diff
@@ -94,11 +101,65 @@ class VideoView: NSView {
...
+
+    return super.draggingUpdated(sender)
```
lhc70000: Should we return `.copy` here?

PR #1941, opened 2018-08-30
[WIP] Implementing quicklook plugin

HarukaMa:

---

**Description:**
Currently, all media files which built-in quicklook could play are handled by built-in generator.

For self-handled files, this plugin currently generates one preview image extracted from video.
I think for audios we could try to encode the file and try to play that. TBD

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -17,3 +17,10 @@ target 'iina' do
 #  end
 
 end
+
+target 'iina-server' do
+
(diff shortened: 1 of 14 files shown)


AuroraWright:
If possible I'd make it optional, given there is a more full-featured alternative: https://github.com/Marginal/QLVideo

Issue #1947, opened 2018-09-01
Localization for 1.0.0 Beta 4


lhc70000:
- [x] fr @superzazu
- [x] it @cgand
- [x] de @J-rg @shorshe 
- [x] ko @darkbrow
- [ ] ru @euf @imKota @Semty
- [x] ja @matsune
- [ ] tr @eminhaqi
- [ ] nl @ErikMinekus
- [ ] pl @Greenek
- [ ] es @Aikyuichi
- [ ] uk -
- [x] sk @Marvin99com
- [ ] da @thkir
- [ ] pt-BR @pradorocchi
- [x] en -
- [x] zh-Hans -
- [x] zh-Hant -

Dear translators, please update the translation if possible. If your translation is not submitted by the next Friday (Sep 7), English translations will be used for new strings by default.

If you are no longer able to provide new translations, please also kindly leave a message here.

Thanks a lot.

**It's recommended to use the IINA translation utility. Please download the latest release ([0.1.4](https://github.com/iina/iina-translation-utility/releases/tag/0.1.4)) before starting.**

sevenfourk:
Hi @lhc70000, I'm sorry I don't have time to work on translations.  Can you please pass translation to un-assigned, someone might pick it up.  I've edited wiki to remove myself from there.  Thanks.  If anything needed further needed from my side, please let me know.

MarcSteven:
@lhc70000  I think we can help to do the Chinese translation  if you agree.

alejx:
@MarcSteven Sure thing! You are welcome to contribute zh-Hant localization. Currently we are using OpenCC to handle zh-Hant localization, which is somehow inaccurate.

imKota:
@lhc70000 Already in the process of updating the translation.

imKota:
@lhc70000 Done
https://github.com/lhc70000/iina/pull/1987

Issue #1427, opened 2018-01-27
Some German Localisation Flaws


ksaerdna:
In "Preferences/General":

- "Wiedergabe wenn ..." must be "Wiedergabe, wenn ..." (twice in the dialogue)
- "Dateien im gleichen Ordner ..." must be "Dateien im selben Ordner ..."

Note that "the same" is either "das gleiche" (a similar specimen of same type) oder "dasselbe" (the very same specimen, i.e. there is only one object) in German. Note also that in low level German many people cannot distinguish, if they e.g. have two cars of the same type or share one car with somebody else.

In main menu:

- "Nach Updates suchen" should be "Nach Aktualisierungen suchen" or better "Nach neuer Version suchen" or "Auf neue Version prüfen". In best case this should match the dialogue "Preferences/General", where the update check interval can be configured. Here the German translation „Auf Aktualisierungen prüfen“ is very nice.

Note that "Updates" is low level German and typical Microsoft slang ("download updates" becomes "Updates downloden" in Windows, which is really ugly).

lhc70000:
Thanks, you can open a pull request directly, or I can help you do the commit. But before that, perhaps you need to discuss it with German translators: @J-rg and @sgoehre.

sgoehre:
He is right and those strings are among the things I want to fix before the next release. I didn’t get around to it yet, however. I’ll add it do my to-do list. :-) Do I have a deadline?

Offtopic
=======
Just a quick note to Andreas: You are right - just your explanations aren’t entirely. 😉 

What you mean is slang or informal German. There’s no such thing as low level German except in foreign language acquisition.

The use of the loanword „Update“, however, is something entirely different. That’s called mixing and is a totally normal process (and happened before in the German language, especially with Latin and French). The word has been part in the German lexis for over 20 years now (since 1996 to be exact, according to the Duden). So it’s become an official (technical) term and is widely used. However, in the Apple world your suggested translation is more common and should as such be adopted.

J-rg:
Great job spotting those errors. One objection though: I wouldn’t consider „Update“ to be low register, it’s merely an Anglicism. Apple changed terminology a while ago and now uses „Updates“ in both App Store and System Preferences, so I think we should go with consistency here.

sgoehre:
Ughh. J-rg is right. While „Aktualisierungen“ seems to be used by most apps, Apple doesn’t anymore. My bad.

ksaerdna:
Just for fun:

 

 

English:


 

[...]

PR #1949, opened 2018-09-02
Update German localization for beta 4

J-rg:
- [x] It implements / fixes issue #1947 & #1427.


alejx:
```
# Localizable.strings #
  [!][de] Wrong format string for key alert.config.cannot_create in Localizable
        Base:        Cannot create config file!
%@
        Translation: Konfigurationsdatei konnte nicht erstellt werden.

# PreferenceWindowController.xib #
  [x][de] File "PreferenceWindowController.strings" doesn't exist
  [x] Cannot read file "./iina/de.lproj/PreferenceWindowController.strings"

# SubChooseViewController.xib #
  [x][de] File "SubChooseViewController.strings" doesn't exist
  [x] Cannot read file "./iina/de.lproj/SubChooseViewController.strings"

```

J-rg:
I fixed the string format and added the omitted files.
The missing key in PrefCodecViewController is just a placeholder if I’m not mistaken.

iina/MainWindowController.swift
```diff
@@ -2114,7 +2114,7 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
     if player.info.isNetworkResource {
       bufferIndicatorView.isHidden = false
       bufferSpin.startAnimation(nil)
-      bufferProgressLabel.stringValue = "Opening stream..."
+      bufferProgressLabel.stringValue = NSLocalizedString("main.opening_stream", comment:"Opening stream…")
```
alejx: This is kind of a breaking change to all translators. Since some of the translators have finished beta 4 work, we should avoid such braking changes. You can revert it and make it another pr. I'll consider merge after beta 4 is released.

iina/OSDMessage.swift
```diff
@@ -253,7 +253,7 @@ enum OSDMessage {
     case .startFindingSub(let source):
       return (
         NSLocalizedString("osd.find_online_sub", comment: "Finding online subtitles..."),
-        .withText("from " + source)
+        .withText(NSLocalizedString("osd.find_online_sub.source", comment: "from") + " " + source)
```
alejx: Ditto

iina/MainWindowController.swift
J-rg: Sorry, I wasn’t aware this would be an issue, I thought you’d add the English wording for missing strings before release anyway?
But if that’s a problem, feel free to merge this PR after release, there’s no hurry on my side.

alejx: I mean you could revert just these two line and move them into another PR. Other parts of this PR are good. And I want to update German localization in beta 4 since German localization has not been updated for quite a while.

J-rg: Alright then, done.

alejx:
One more minor issue before I would merge this in: `U4W-Xv-hkh.hkh.ibShadowedToolTip` is missing in `PlaylistViewController.strings`. If you are using IINA translation utility, please download the latest release [0.1.4](https://github.com/iina/iina-translation-utility/releases/tag/0.1.4).

J-rg:
Frankly I don’t know how to translate this because I don’t know what the button does exactly and I couldn’t make it show up.
What part of the filename does IINA consider to be a common prefix and under which circumstances is it truncated?
(The tooltip also shows up if there is no ellipsis btw, I doubt it’s supposed to do that.)

lhc70000:
> The tooltip also shows up if there is no ellipsis btw

It shoudn't. I'll fix it later.

> What part of the filename does IINA consider to be a common prefix and under which circumstances is it truncated?

Typically like 

```
A_long_series_name_001.mp4
A_long_series_name_002.mp4
A_long_series_name_003.mp4
```

will be displayed as

```
...001.mp4
...002.mp4
...003.mp4
```

J-rg:
Okay, that’s about what I figured, doesn’t work for me though. If I add three files with those filenames to the playlist, they show up as
A_long_serie...me_001.mp4
A_long_serie...me_002.mp4
A_long_serie...me_003.mp4

Anyway, I added the correct translation for the tooltip to commit d25e5fa.

PR #1982, opened 2018-09-14
improve FileSize.format() when size is unit size

shengdoushi:
for example, the output of 1024 will be 1KB, not 1024B

---

1KB is better than 1000B.


lhc70000:
Thanks!

PR #2000, opened 2018-09-23
Add Xunlei online subtitle vendor.

tinkernels:
As Xunlei subtitle api using clear text HTTP, more insecure exceptions appended:
- xunlei.com
- geilijiasu.com


inflation:
Thanks for the interests of contribution! 

However, there are a lot of requests about adding custom subtitle providers, and we simply don't have the effort to check them all. But we are going to introduce a plugin system for IINA, so that you can implement anything you want and maintain them separately.

Stay tuned! (And we would list your plugin in our wiki if you want then.)

tinkernels:
fair enough 😉

Issue #1988, opened 2018-09-18
"Open in IINA" is no longer support by safari 12.0


Bluetus:
**System and IINA version:**

- macOS 10.13.6
- IINA 1.0.0-beta 4 Build 92

**Expected behavior:**

**Actual behavior:**

- Crash report:
```

Apple just delete the extension automatically after update safari 12.0
Doesn't allow to reinstall either

```

- Pic:
<img width="466" alt="screen shot 2018-09-18 at 3 38 13 am" src="https://user-images.githubusercontent.com/25444910/45672264-b4ed3b00-baf5-11e8-8b2d-6236b3e4bce4.png">

LeoNatan:
@alexjs What does “not our problem” even means? Of course it’s your problem. As developers of IINA, part of the “contract” with the community is to support the latest technologies and constraints of the OS vendor.

lhc70000:
Yes, surely it's our problem. 

We only need to submit the plugin to the Safari Extension Gallery. I'll get this done ASAP.

LeoNatan:
@lhc70000 Is this the right approach? I think the gallery is being closed later this year. I think a better approach is to build an appex (extension) for Safari.

LeoNatan:
> Support for .safariextz-style Safari Extensions installed from the Safari Extensions Gallery is deprecated with Safari 12 on macOS. Submissions to the Safari Extensions Gallery will no longer be accepted after December 2018. Developers are encouraged to transition to Safari App Extensions.

AnanyaKirti:
Hope this happens!
Love IINA

Bluetus:
Thank you, everyone! I confused when it was labeled as ”not our problem” too. Thought the extension was a different project somewhere else. Happy that this will happen. Looking forward to it! :)

inflation:
@saagarjha Gladly

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

saagarjha:
After looking at #1990, I'm not quite sold on the need for a browser extension. Why can't we just vend a service so that it can be invoked from anywhere in macOS?

AnanyaKirti:
yes, but I would still prefer an extension.
I mostly use IINA to watch 4k youtube videos in IINA (since I use Safari)

lhc70000:
[...]

PR #2029, opened 2018-10-09
Add Safari app extension

saagarjha:
- [X] It implements / fixes issue #1988.

---

Converts the old Safari JavaScript extension into an app extension.

Currently, it still needs some work:
* We need a new toolbar icon. Please replace the icon with an appropriate PDF image.
* Right clicking on links doesn't quite work, for a variety of reasons. We can discuss how we want to move forward with this here.
* This can only be installed if the *entire app bundle*, including the extension, is properly signed. Ad-hoc signatures will cause it to be rejected by Safari.


lhc70000:
> Right clicking on links doesn't quite work, for a variety of reasons. We can discuss how we want to move forward with this here.

Could you elaborate more on the exact reasons? I just tested roughly and seems that each time a new `SafariExtensionHandler` is created dynamically so there's no way to store the URL received in `messageReceived` inside the handler class, but if I store it outside the class, like in a `fileprivate var`, it works.

saagarjha:
Wait, hold on: context menus can define a custom `userInfo` dictionary. Ignore what I said before.

I think the only thing this needs now is an icon file and it should be good to go–unless you see anything wrong with it?

jimblue:
Hey!

Seems to be merged but I don't see iina extensino in the appstore...

Did I miss something?

inflation:
@jimblue You don’t need to manually install the extension. Once you have the next version installed, Safari would load it automatically. 

saagarjha:
You may actually have to run IINA before it shows up, just FYI. Then you should be able to enable it from Safari's preferences.

jimblue:
@inflation @saagarjha, thanks, but what do you mean by the next version installed?
I'm actually running iina 1.0.0-beta4 (92)

inflation:
@jimblue We would be rolling out the official 1.0 in the following weeks. 

jimblue:
oh ok that's great! thank you so much guys, I love iina ❤️ !!!

pimvanriezen:
Just wanted to point out that it is in no way obvious that the extension is automatically added to Safari. I spent some time today trying to download and activate the old safariextz, and searching through the Safari extension gallery, before I figured out it was already added to my extensions, but disabled.

saagarjha:
Agree that it should be made more explicit to users how to “install” the extension. I feel like this falls under the work I outlined in #1988.

LEoREo2247:
I updated IINA to 1.0.0 through in-app updater, however the new extension was not added to Safari and I had to download then re-install it from github's release page in order to make it appear in Safari.

New extension should have bundled in app even if it's updated through updater. I wonder if there is any function to let or force Safari to re-detect bundled extensions so that we can run that on first-launch after next update. If that is possible, then we can also re-add Safari button at preferences so users can manually do that.

Issue #2054, opened 2018-11-02
Localizations for 1.0 Release


alejx:
We are delighted to announce that we are going to release IINA 1.0.0 soon. There's no accurate timeline for this release, but hopefully within a month from now.

IINA 1.0 is the first stable version, every words that users finally see should be neat, clear and without any glitches. UI is now frozen for 1.0 release. That is to say, no UI/wording update on `develop` or `release/1.0.0` branch until 1.0 release. 

For those whose language translation is up-to-date for 1.0.0 beta 4: A few UI updates are made after beta 4. Make sure all xib files have string files for your language (The IINA translation utility will not notify you if there’s a xib file without the string file for your language, so be sure to double check it). Besides, I hope you can check all the translations thoroughly within the app (not just in strings files), and fix problems. It's a little bit tricky to build IINA currently, see the current `README.md` in develop branch.

For those whose language translation is not up-to-date: Thanks for your work for IINA, but we are not going to ship incomplete localization to users in a stable release. We will drop incomplete localization if we cannot find new translators at 1.0.0 release. The strings files still remain in the repository; whenever it becomes completed, we will use it again.

For new contributors: You are welcome to add your language to IINA, or pick up one incomplete localization and help us to finish it before release. Please follow [this page](https://github.com/lhc70000/iina/wiki/Translation) to submit your translations.

alejx:
**Status**
- [ ] fr @superzazu
- [ ] it @cgand
- [ ] de @J-rg @shorshe 
- [ ] ko @darkbrow
- [ ] ru @euf @imKota @Semty
- [ ] ja @matsune, @LEoREo2247 
- [ ] tr @eminhaqi
- [ ] nl @ErikMinekus
- [ ] pl @Greenek
- [ ] es @Aikyuichi
- [ ] uk -
- [ ] sk @Marvin99com
- [ ] da @thkir
- [ ] pt-BR @pradorocchi
- [x] en -
- [ ] zh-Hans -
- [ ] zh-Hant -

alejx:
From this point, new Pull Requests should base on `release/1.0.0` branch, not `develop` branch. Please checkout `release/1.0.0` before starting your work.

imKota:
@alejx Done #2088

PR #2095, opened 2018-11-24
Add Swedish translation

anohren:
I have attempted to use best practices for:
1. Hyphen (-) and En-dash (–) respectively ("A–B-loop")
2. Separation of quantities from their units but not their sign ("+5 %", "-1 s")
3. Localized decimal separator ("1,1") — this requires a code change to make the UI consistent

...but not for replacing "x" with multiplication sign (×) since that might requrie a non-trivial change in code to catch all cases.

---

**Description:**


alejx:
Ref: #2054

alejx:
Only one translation is missing:
PlaylistViewController.xib #
[!][sk] Key "U4W-Xv-hkh.ibShadowedToolTip" doesn't exist in PlaylistViewController

alejx:
Thanks for your work! Merged locally.

PR #2121, opened 2018-12-12
Fix video and audio filters can't delete.

xjbeta:
- [*] This change has been discussed with the author.

---

**Description:**
https://mpv.io/manual/master/#command-interface-%22filter1

`Remove the given filters from the video chain. Unlike in the other cases, the second parameter is a comma separated list of filter names or integer indexes. 0 would denote the first filter. Negative indexes start from the last filter, and -1 denotes the last filter.`


xjbeta:
Fixed #2156

PR #2122, opened 2018-12-12
Fix nil unwrap on display configuration change

anohren:
When unplugging an external display while IINA is playing in fullscreen on a secondary display (not the one with Dock), two NSWindowDidChangeScreenNotification are sent in short succession. When receiving the first, CGDisplayCreateUUIDFromDisplayID will return nil for that window's screen, resulting in a crash.


alejx:
Thanks! Merged locally into release branch.

PR #2128, opened 2018-12-13
Don't exit fullscreen on entering PiP

anohren:
Entering PiP and exiting fullscreen are orthogonal concepts; PiP manages its own window, so the two aren't mutually exclusive. Requesting one of them should not imply requesting the other. 

Whatever fullscreen / split-screen setup the user has should be preserved whenever possible — not torn down. For example, I might be watching something in fullscreen but need to do something else real quick. The natural way to do this is to activate PiP, move to another space and do the task, then use the Exit PiP button in the PiP window to _restore the setup to its previous state_.

Even worse is if I've put IINA in split-screen with another app to be able to watch a video without having a PiP window obscuring any part of the other app. If I want to complete a short task on another space, then I'm currently confronted with the choice of A) pausing the video, or B) destroying my carefully laid out split-screen setup.

---

**Description:**

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -2800,11 +2800,6 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
 extension MainWindowController: PIPViewControllerDelegate {
 
   func enterPIP() {
-    // Exit fullscreen if necessary
-    if fsState.isFullscreen {
-      toggleWindowFullScreen()
-    }
-
     pipStatus = .inPIP
     showUI()
 


saagarjha:
IINA's original implementation of PIP behaved the way you expect it to, IIRC. During the rewrite of picture-in-picture, I changed the implementation to match that of Safari, which *does* exit fullscreen when entering picture-in-picture. Since there is obviously no standard human interface guidelines for private API, I usually base my implementations on system apps (in this case, this is essentially Safari). Interestingly, it seems like Safari's behavior is ad-hoc and it's possible to get it to enter picture-in-picture while in fullscreen in certain situations (as well as break it, lol), so I'm not opposed to revisiting this choice. Does anyone else have thoughts?

choco:
I'm just a IINA user lurking around, but I agree with @anohren.

I think Safari makes PiP and Fullscreen mutually exclusive because you end up with a tab telling you the video is playing in fullscreen, and in turn the fullscreen window pointing to the PiP window. I can understand why this "double" indirection wouldn't be the best solution in their eyes. IINA on the other end wouldn't incur in the same issue.

anohren:
Thanks for your inputs. I agree that it's a good idea to start with mimicing the system behavior so as to not miss any hidden benefits -- that is, until we can articulate reasons for doing it differently, like in this case. 

> Interestingly, it seems like Safari's behavior is ad-hoc and it's possible to get it to enter picture-in-picture while in fullscreen in certain situations

That explains why I was of the impression that Safari already did it this way, and when I went back to look it suddenly didn't 🙃 

I came to the same conclusion as @choco -- this makes more or less sense depending on context. I would still argue that safari and itunes does this the Wrong way because of the reasons in the OP -- entering PiP should be a zero cost operation -- but the behavior is at least borderline reasonable in the case of web browsers if they've already spawned additional windows.

lhc70000:
I agree with @anohren and @choco and let's get it merged.

PR #2161, opened 2019-01-01
Fix bitrate display in Inspector window

ErikMinekus:

---

**Description:**
- ~~The bitrate is expressed in bits per second, so the value must be divided by 1000, instead of 1024.~~
- ~~The bitrate is now always displayed in kbps, which is the most common format.~~
- Fixed the audio bitrate label being greyed out when switching to the next track.

### DIFF
--- a/iina/InspectorWindowController.swift
+++ b/iina/InspectorWindowController.swift
@@ -119,7 +119,6 @@ class InspectorWindowController: NSWindowController, NSTableViewDelegate, NSTabl
           MPVProperty.currentAo: self.aoField,
           MPVProperty.audioParamsFormat: self.aformatField,
           MPVProperty.audioParamsChannels: self.achannelsField,
-          MPVProperty.audioBitrate: self.abitrateField,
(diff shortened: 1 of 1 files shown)


saagarjha:
If #2183 is merged, you'll need to change your code slightly. Just a heads up.

alejx:
Any updates?

ErikMinekus:
@alexj I wasn't able to figure out how to always display the bitrate in kbps with @saagarjha's changes, and I don't have time currently to look into it. So the only fix left in this PR is the audio bitrate label being greyed out when switching tracks.

alejx:
@ErikMinekus Sorry for such a substantial late. I've made a PR https://github.com/iina/iina/pull/2604 to fix localization issues in the inspector window. It seems that this issue ceases to exist in my PR. Can we close this now?

ErikMinekus:
@alejx yes, this seems to be fixed in #2604. Thanks!

PR #2184, opened 2019-01-03
Minor adjustment of Swedish translation

anohren:
Correct the mention of keys to map correctly to the mention of their respective functions

---

**Description:**

left / right was incorrectly translated to the equivalent of right / left


alejx:
Thanks!

Issue #2157, opened 2019-01-01
No effect on volume on scroll in Music Mode.


bw-space:
**System and IINA version:**

- macOS - 10.14.2
- IINA 1.0.0 build 100

**Expected behaviour:**
On scrolling up or down the volume should change in the app.
Also scrolling left or right should shift the timeframe accordingly.

**Actual behavior:**
There is no change in volume or shift in timeframe on scrolling.

PR #2191, opened 2019-01-04
Added scroll controlling to mini player

LEoREo2247:
- [x] It implements / fixes issue #2157.

---

**Description:**

I did it because I personally just wanted that.
Implementation is somewhat lazy and its nearly just a copy-paste from MainWindowController.
However instead of having the dedicated OSD, this scroll controlling on mini player indicates the volume change by showing volume popover while moving trackpad/mouse.

### DIFF
--- a/iina/MiniPlayerWindowController.swift
+++ b/iina/MiniPlayerWindowController.swift
@@ -354,7 +354,12 @@ class MiniPlayerWindowController: NSWindowController, NSWindowDelegate, NSPopove
       scrollDirection = nil
     }
     
-    let scrollAction = scrollDirection == .horizontal ? horizontalScrollAction : verticalScrollAction
+    var scrollAction = scrollDirection == .horizontal ? horizontalScrollAction : verticalScrollAction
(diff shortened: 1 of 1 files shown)


anohren:
You don't explicitly say what is being controlled by scrolling, but I assume it's exclusively the volume.

That's exactly what I was missing as well. Tried scrolling while hovering over the window, the album art, the volume icon, the volume popover, the volume slider -- nothing :)

anohren:
From testing now I think it works really well. Scrolling reacts everywhere it should.

alejx:
Sorry for late reply. While I'm not oppose this idea, I see there are a lot of code share between `MainWindowController` and `MiniPlayerWindowController`. I'm considering abstract more from those two to reduce code sharing. So I'll probably close this one.

alejx:
Will be fixed by #2772

anohren:
@alejx The feature has not been fully implemented as of release 1.0.7. See https://github.com/iina/iina/issues/2999.

Issue #2147, opened 2018-12-28
Hide main window when entering PiP


CourierKyn:
**System and IINA version:**

- macOS 10.14.2 (18C54)
- IINA 1.0.0-rc

**Expected behavior:**
When entering Picture in Picture mode, the former IINA window should hide immediately, leaving only the PiP visible. That's just like how it works in iTunes. 

**Actual behavior:**
When entering Picture in Picture mode, the main IINA window remains visible, together with the PiP.

**Steps to reproduce:**
N/A

- [x] MPV does not have this problem.

**How often does this happen?**
N/A

anohren:
Highly disagree. For those interested, the reasons for this can be found in this discussion: https://github.com/iina/iina/pull/2128

CourierKyn:
I appreciate your support very much.

dhoepfl:
I agree with the request.

Having a normal window telling me that there is a PIP window is like having a normal window telling me that the content is presented in fullscreen. Who would do that?

LEoREo2247:
I personally also appreciate this.
Sometimes I just wanted to use PiP to simply reduce all the scattered windows, and this feature can make that possible. However, I'm pretty sure some may not appreciate this change as well, therefore this can be optional.
Anyway, I'll try to look into it to see if it's something that I can do it.

saagarjha:
I think we've fixed this with #2192.

PR #2192, opened 2019-01-04
Hide main window when entering Picture in Picture

LEoREo2247:
- [x] It implements / fixes issue #2147.

---

**Description:**

This PR will add ability to hide main window when entering Picture in Picture if window is not in fullscreen mode.
I looked in to it and I noticed that implementation is quite simple.
It doesn't require specification to re-show window when going back from PiP as PiP seems to be doing that already.
I also added option to enable/disable this behavior, but this may not needed. It depends on how you all thinks.

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -249,6 +249,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
   }
 
   func applicationShouldTerminate(_ sender: NSApplication) -> NSApplication.TerminateReply {
+    guard !PlayerCore.active.mainWindow.isWindowHidden else { return .terminateCancel }
     Logger.log("App should terminate")
     for pc in PlayerCore.playerCores {
      pc.terminateMPV()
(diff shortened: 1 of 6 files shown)


anohren:
So, should the discussion be held here? If yes: what were the reasonings for minimizing the window, raising it or whatever? I can see the point of minimizing the window since that's effectively hiding the window, but that's pretty much it. 

If those suggestions are alternatives to this one, then they fit quite naturally in here.

Sounds kind of weird to move core functionality like PiP to a plugin though.

saagarjha:
The core PIP functionality (e.g. actually entering/exiting PIP) likely cannot be a plugin, and will remain as-is. The question is allowing users to customize "reactions" to these events, so they they can e.g. minimize the main window when entering picture-in-picture, or close it, etc. So my question is 1. whether we should have this customization point, 2. whether we should say have a preference that lists some commonly requested actions and performs that when PIP is invoked, or 3. not have a preference at all, but allow plugins to customize this behavior.

anohren:
I suspected that that was your intention. However, since, as you say, PiP will remain a core functionality it needs to have a sensible default behavior regardless of whether this will be made a customization point and regardless of whether any such plugins are installed.

If I'm not mistaken, this PR is mostly about implementing the most sensible default behavior. Perhaps it should be split up so that the discussion about if/how to make this customizable can continue independently.


Unless of course that your suggestion is to implement default behavior like this as an official plugin. Sounds a bit overkill, but perhaps that's been the plan all along.

LEoREo2247:
Alright, I think that is pretty neat. I can try implementing that if anyone else wouldn't.

LEoREo2247:
So I've added the options that saagarjha listed. I need someone to check if there is any problem or a flaw.

saagarjha:
Great job, @LEoREo2247! There's still some jankiness with the transition, but that's my problem, not yours; I'll look into whether this can be fixed later.

PR #2207, opened 2019-01-08
Fix nil unwrap in SubPopoverViewController.playerCore.getter

anohren:
Threw a couple question marks at a crash report, since the code was already prepared for nil everywhere. Recreated this PR since apparently it's impossible to change source branch.

Exception Type:        EXC_BAD_INSTRUCTION (SIGILL)
Exception Codes:       0x0000000000000001, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Illegal instruction: 4
Termination Reason:    Namespace SIGNAL, Code 0x4
Terminating Process:   exc handler [26767]

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   com.colliderli.iina           	0x000000010664ce50 closure #1 in SubPopoverViewController.playerCore.getter + 208
1   com.colliderli.iina           	0x0000000106655255 specialized SubPopoverViewController.numberOfRows(in:) + 53
2   com.colliderli.iina           	0x000000010664c376 @objc SubPopoverViewController.numberOfRows(in:) + 38

### DIFF
--- a/iina/PlaylistViewController.swift
+++ b/iina/PlaylistViewController.swift
@@ -728,9 +728,9 @@ class SubPopoverViewController: NSViewController, NSTableViewDelegate, NSTableVi
   @IBOutlet weak var playlistTableView: NSTableView!
   @IBOutlet weak var heightConstraint: NSLayoutConstraint!
 
-  lazy var playerCore: PlayerCore = {
-    let windowController = self.playlistTableView.window!.windowController
-    return (windowController as? MainWindowController)?.player ?? (windowController as! MiniPlayerWindowController).player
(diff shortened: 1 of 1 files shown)


alejx:
I've suggested another fix #2231. I think `player` should not be optional, and directly setting `player` in the view controller when the playlist view is loaded is better than searching the `player` reversely when it's used.

anohren:
That's good, whatever works. No idea why it's written like that. 

Although your changes make it an optional as well. But it looks like you know more about the lifetime of that object.

PR #2231, opened 2019-01-18
Fix sometimes player cannot be determined and yield nil unwrap

alejx:
Fix the issue metioned in #2207


saagarjha:
Do you know the underlying reason why this happens? Otherwise, this seems like a bit of a kludge…

alejx:
I've no idea, but directly setting `player` in the view controller when the playlist view is loaded is better than searching the `player` reversely when it's used IMO.

lhc70000:
While it's not easy to reproduce this issue, I think the crash occurred because `playlistView.window` was `nil`, i.e. the view is not attached to a window in the mean time (such as it was in the middle of switching to music mode).

Therefore IMO this approach is at least better than the original one since it set up the `player` at the beginning, so it won't need to look up for it in the window hierarchy each time.

saagarjha:
With that in mind, I'm satisfied with these changes.

PR #2306, opened 2019-02-15
Fix memory leaks in FFmpegController

stps:
- [x] It implements / fixes issue #1720.

---

**Description:**

I was getting some pretty substantial memory leaks from FFmpeg generating the thumbnails when opening a new file (Instruments reported 250mb in av_malloc leaks each time I opened a 1min 4K video). This change fixes the FFmpeg-related leaks from all the files I've tested.

### DIFF
--- a/iina/FFmpegController.m
+++ b/iina/FFmpegController.m
@@ -220,9 +220,11 @@ return -1;\
 
         ret = avcodec_receive_frame(pCodecCtx, pFrame);
         if (ret < 0) {  // something happened
-          if (ret == AVERROR(EAGAIN))  // input not ready, retry
+          if (ret == AVERROR(EAGAIN)) { // input not ready, free packet and frame and retry
+            av_frame_unref(pFrame);
+            av_packet_unref(&packet);
             continue;
-          else
+          } else
             break;
         }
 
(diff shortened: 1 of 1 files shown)


saagarjha:
@lhc70000 @HarukaMa I think this is the part where I say “I told you so”. We should rewrite FFmpegController, in Swift or Objective-C++ or whatever that makes bugs like these less likely to occur. 

(For those out of the loop, #1729.)

saagarjha:
@stps PromiseKit has been removed from CocoaPods, so don’t worry about the failing test; I’ll fix that at some point.

HarukaMa:
I stand that swift is not the best way to write that interface due to the extra UnsafePointer layer.

CI is fixed.

saagarjha:
Objective-C isn't right either. I think this leaves Objective-C++?

saagarjha (changes requested):
Working on an improvement in this area; so I'm going to block merging of this for now (sorry, @stps). I'll reopen this if that doesn't end up working out.

bdolewski:
@HarukaMa 
> I stand that swift is not the best way to write that interface due to the extra UnsafePointer layer.

Could you elaborate more about that? UnsafePointer is the way to deal with C code (like using CommonCrypto in Swift). I don't think Objective-C is safer by any means, you still deal with pointers and memory ownership is on your side. At least Swift layer is far more approachable by developers than Objective-C :)

PR #2311, opened 2019-02-16
Rewrite FFmpegController in Objective-C++

saagarjha:
- [X] It implements / fixes issue #1720.

---

I'll rip out the confusing Swift parts and replace it with Objective-C++, I promise. Just give me a bit of time.


saagarjha:
Ok, I've done a first pass at writing this up. I still cannot run IINA, so the only testing I have put into it is to run a build; however, I think the only bugs we should have is if I missed a line when translating or some of the slightly more obtuse runtime coercion fails in an unexpected way. Please try running this and seeing if the thumbnails seem to work.

@lhc70000, @HarukaMa, would you mind looking at this? I think this addresses the issues you had with the previous iteration of this pull request, since `std::unique_ptr`'s usage is significantly less unwieldy than Swift's `Unsafe*Pointer` APIs; however, the "nice" parts of the API can remain in Swift. One thing I did want to discuss was the complexity of `-[FFmpegController synchronouslyGenerateThumbnailsForFileAtPath:]`: I have done a straight translation best I could, but we might want to consider adding comments or something similar to explain what is going on because I frankly don't have much clue what each thing is doing (which makes it easy for me to make mistakes).

@stps, if you're up to in running an Instruments pass on this to make sure I didn't miss anything that would be great; if you find something feel free to let me know and I'll merge in your commits fixing it.

inflation:
I’m not having a very strong background in cpp, and so does the the rest of the team. To me, there are a lot of STL boilerplates to just make things worse. 

But if you think you can maintain it in the future, all lights green.

iina/FFmpegController.mm
```diff
@@ -0,0 +1,220 @@
...
+  }
+  formatContext.reset(raw_formatContext);
+
+  if (avformat_find_stream_info(formatContext.get(), nullptr) < 0) {
+    return NO;
+  }
+  auto videoStreamIndex = std::find_if(formatContext->streams, formatContext->streams + formatContext->nb_streams, [](const AVStream *stream) {
+    return stream->codecpar->codec_type == AVMEDIA_TYPE_VIDEO;
+  });
+  if (videoStreamIndex - formatContext->streams < formatContext->nb_streams) {
```
stps: This condition seems to always trigger so it exits here. Though videoStreamIndex looks to be the correct `AVStream**`.

iina/FFmpegController.mm
```diff
@@ -0,0 +1,220 @@
...
+  frameRGB->width = DEFAULT_THUMBNAIL_WIDTH;
+  frameRGB->height = DEFAULT_THUMBNAIL_WIDTH * codecContext->height / codecContext->width;
+  frameRGB->format = AV_PIX_FMT_RGBA;
+
+  auto size = av_image_get_buffer_size(
+      AV_PIX_FMT_RGBA,
+      frameRGB->width,
+      frameRGB->height,
+      1);
+  if (0 < size) {
```
stps: Seems like this should be `size < 0` right?

saagarjha: Right you are. Thanks!

iina/FFmpegController.mm
saagarjha: I've changed the condition to `videoStreamIndex < formatContext->streams + formatContext->nb_streams`.

saagarjha:
I'm considering merging this at some point soon; does anybody have any objections to this? @inflation, I can maintain this, but I'll look into whether it's possible to remove some of the STL boilerplate if it helps.

iina/FFmpegController.mm
stps: Looking again, since `find_if` returns `last` if it doesn't find anything I believe it should actually be `videoStreamIndex == formatContext->streams + formatContext->nb_streams` to catch the case of it not finding. What's there right now still exits early (and my bad in comment above).

stps:
It doesn't seem to be working on the branch currently (even after I make the change in my comment above). Can't get into debugging it right now but looks like it's generating hundreds of corrupted images and not counting the progress correctly.

lhc70000:
Sorry for my recent inactivity. I might still be busy in the next few months.

First of all I'm neutral about this.

I do have a few concerns:
- Is it really necessary? We're going to introduce a new **language** (which is extremely complicated and powerful) just in order to use `unique_ptr` in one single file.
- There's only 200 lines of code in this file and it's unlikely to be extended in the future. Although I may forgot to free something somewhere, it's still in control and we can just fix that.
- A lot of projects using the libav API, including mpv, are written in C, and so far they are happy with allocating and releasing resources.
- Is there any overhead introduced by using Objective-C++, in both compile time and runtime?

And if you insist to use Objective-C++:
- The deleter structs are awkward. Could you at least use lambdas for deleters to make the code more readable?
- Is it better to write a macro similar to `at_exit` using RAII?

saagarjha:
(Just to clarify my earlier comment: I will not be merging this until the rest of the core team agrees to this and I can actually test this properly/make sure it works. I was just trying to solicit feedback so we could resolve this issue.)

> Sorry for my recent inactivity. I might still be busy in the next few months.

Sure, no problem. Thanks for letting us know.

> Is it really necessary? We're going to introduce a new **language** (which is extremely complicated and powerful) just in order to use `unique_ptr` in one single file.

> There's only 200 lines of code in this file and it's unlikely to be extended in the future. Although I may forgot to free something somewhere, it's still in control and we can just fix that.

I really think so, yes. The current code (even with @stps's fixes) can leak resources if any method call fails and we return an error, because we bail out before closing/freeing resources that we acquire. The only way to fix this in C that I know of is a careful set of `goto`s littered everywhere, which is quite difficult to reason about in this case because we have multiple nested control flows that make it quite hard to ensure that resources are freed. I have tried my best here to limit the use of Objective-C++ to the bare minimum that is required (the two methods that we care about)–everything else is standard Swift, and I chose to organize it this way to separate the two as much as possible while still making usage reasonable.

> A lot of projects using the libav API, including mpv, are written in C, and so far they are happy with allocating and releasing resources.

Their API is actually a bit awkward to use from anything that isn't C, hence the boilerplate in Swift and a the smaller amount in Objective-C++. I think there are some proposals for something called `std::inout_ptr` that will help in this case, but they're still in their planning phases. However, from our side I don't think it is easy to write clear, correct code in anything that doesn't have some sort of RAII/`defer` pattern, for the reason I mentioned above: our usage in particular is complex and there are a lot of code paths to consider.

> Is there any overhead introduced by using Objective-C++, in both compile time and runtime?

Compared to Objective-C? I don't think there should be any measurable performance impact. C++ is supposed to be a "zero-cost" abstraction, and we should be getting something close to that–and we might be faster in some cases (some of the Swift code might be optimized better, and the C++ standard library is generally faster than Foundation). I can profile this if you're curious, but I am pretty certain we won't see anything meaningful here.

> The deleter structs are awkward. Could you at least use lambdas for deleters to make the code more readable?

We could, but it would be difficult. We'd lose access to the zero-parameter constructor, which would make creating the `std::unique_ptr` somewhat annoying because we don't control allocation of libav's objects. Custom structs have a call operator that "works" in this situation.

> Is it better to write a macro similar to `at_exit` using RAII?

I am considering something like this. I think it might be possible to write Objective-C++ that looks similar to Objective-C, with the boilerplate hidden behind a `DEFER` macro and the rest "translated" back to C (essentially: drop `std::unique_ptr` for a custom class, don't use `auto`, and replace `std::find_if` with a loop). If this would help I can look into making this work.

saagarjha:
Oh, and thanks for looking at this @stps. I know I kind of dragged you into this conversation, so I appreciate your testing my broken code that I can't run myself ;) I'll look into the corrupted images issue in a bit.

saagarjha:
@stps I revisited this and was able to test the code, and I think it works now. I still haven't rewritten the code to integrate feedback from @lhc70000, though; I'll get around to doing that soon.

iina/FFmpegController.mm
saagarjha: The correct condition is `videoStreamIndex >= formatContext->streams + formatContext->nb_streams`.

saagarjha:
@lhc70000 I have replaced the use of `std::unique_ptr` with a RAII `DEFER` macro based on `std::shared_ptr` (which has a small memory overhead as compared to `std::unique_ptr`, unfortunately, but it's not noticeable). Let me know if you'd like me to change anything else.

@stps I ran this code through Instruments and I couldn't find any leaks from this; could you confirm this if you have the time?

stps:
Looks good for leaks but it seems there’s some bug where if you open multiple videos it will show the thumbnails belonging to a different video.

saagarjha:
Huh, that's interesting. I feel like my changes affected code "below" the layer where that should be handling this; is this something that you're seeing in general (i.e. with the IINA we ship currently) or does this only show up with my modifications?

iina/FFmpegController.swift
```diff
@@ -0,0 +1,80 @@
...
+  private var thumbnails = [FFmpegThumbnail]()
+  private var timestamps = Set<Int64>()
+  private var lastDelegateCallbackTime = 0 as CFTimeInterval
+
+  public var delegate: FFmpegControllerDelegate? = nil
+  public var thumbnailCount: Int {
+    return thumbnails.count
+  }
+
+  public func generateThumbnailsForFile(atPath path: String) {
```
stps: I think the issue is the thumbnails array and timestamps set need to be emptied here.

saagarjha: Oh, right, forgot to copy that over. Thanks for spotting that; I've updated the code.

iina/FFmpegController.mm
```diff
@@ -0,0 +1,205 @@
...
+#define __CONCATENATE(a, b) __ACTUALLY_CONCATENATE(a, b)
+#define DEFER(cleanup) std::shared_ptr<void> __CONCATENATE(__defer_, __LINE__)(nullptr, cleanup)
+
+@implementation FFmpegController (Bridge)
+
+- (BOOL)synchronouslyGenerateThumbnailsForFileAtPath:(NSString *)path {
+  AVFormatContext *formatContext = NULL;
+  if (avformat_open_input(&formatContext, path.fileSystemRepresentation, nullptr, nullptr) < 0) {
+    return NO;
+  }
```
alejx: I personally prefer now to write optional parentheses in C&C++. We do have some if statements that simply return `NO`, so I think remove those parentheses might be better in terms of readability.

iina/FFmpegController.mm
```diff
@@ -0,0 +1,205 @@
...
+//
+//  Created by Saagar Jha on 2/16/19.
+//  Copyright © 2019 lhc. All rights reserved.
+//
+
+#import "FFmpegController.h"
+
+// Perhaps this will work someday? Until then, we need to keep around our dummy
+// interface in the header file.
+//#import "IINA-Swift.h"
```
alejx: Does it work now?😊

iina/FFmpegController.mm
```diff
@@ -0,0 +1,205 @@
...
+#define DEFER(cleanup) std::shared_ptr<void> __CONCATENATE(__defer_, __LINE__)(nullptr, cleanup)
+
+@implementation FFmpegController (Bridge)
+
+- (BOOL)synchronouslyGenerateThumbnailsForFileAtPath:(NSString *)path {
+  AVFormatContext *formatContext = NULL;
+  if (avformat_open_input(&formatContext, path.fileSystemRepresentation, nullptr, nullptr) < 0) {
+    return NO;
+  }
+  DEFER([&](...) {
```
alejx: What about putting `[&](...)` into the macro?

uiryuu (changes requested):

iina/FFmpegController.mm
saagarjha: Are you talking about the brackets?

iina/FFmpegController.mm
saagarjha: I've currently chosen to leave it out to prevent putting too much "magic" into the macro (so it works like an actual "defer" that takes a full closure). But I'd be fine with hoisting the capture and parameter lists into the macro–I'd assume you're looking for something like this at call site?

```objective-c++
DEFER({
	// code
});
```
Another choice is to use Objective-C blocks, which are layout-compatible with C++ lambdas and would save two characters ;)

```
DEFER(^(...) {
	// code
});
```

iina/FFmpegController.mm
alejx: Yeah, sorry about that.

iina/FFmpegController.mm
alejx:
```C++
DEFER({
	// code
});
```
I prefer this one, however it's on your choice.

saagarjha: Since it's only being used in this file, I'll hoist it up into the macro. If we ever need this anywhere else I think we should go for the more general solution.

iina/FFmpegController.mm
saagarjha: I put a bit of effort into this and I just don't think it was worth it. A lot of things break and need to be subtly tweaked so the Swift importer is happy and I just gave up after disabling modules, forward-declaring a bunch of private APIs and moving a bunch of code around didn't fix the issue.

iina/FFmpegController.mm
saagarjha: Actually, now that we have a reliable cleanup mechanism I'm thinking that we should bring back the macros to handle these.

saagarjha:
Updated macros are live.

PR #2313, opened 2019-02-19
Correct label/tooltip for Safari toolbar button

anohren:


alejx:
Do you know what does it effect?

anohren:
@alejx Yes, the tooltip. And the label, when customizing the toolbar.

alejx:
Then thanks for you fix!

saagarjha:
This seems reasonable then. Let me merge it.

saagarjha:
…I guess I'm too slow?

PR #2346, opened 2019-03-03
README: update mpv install instructions (no flag)

vitorgalvao:

---

**Description:**

Homebrew maintainer here. [Options are no longer supported](https://brew.sh/2019/02/02/homebrew-2.0.0/) in the main [`homebrew-core`](https://github.com/Homebrew/homebrew-core) formulae repo.

Running the current command would give `Error: invalid option: --with-uchardet`. It is [already included in the formula](https://github.com/Homebrew/homebrew-core/blob/a506f8508e40d47040698349ba95d7df06f34d75/Formula/mpv.rb#L27).

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,9 +58,8 @@ $ brew install cocoapods
 1. Install mpv:
 
 ```console
-$ brew install mpv --with-uchardet
+$ brew install mpv
 ```
-Feel free to include your own copies of the other libraries if you'd like.
 
 2. Copy latest [header files](https://github.com/mpv-player/mpv/tree/master/libmpv) into `libmpv/include/mpv/`.
 


saagarjha:
Thanks for submitting this! I *think* that we can merge this, but we should probably wait a little bit and see what the other maintainers have to say (@alejx?). The issue is a bit complicated, and while we're well aware of Homebrew removing support for options in homebrew/core, we have held off on updating the README because there have been some recent conflicts with Homebrew's version of mpv and we are still working on figuring out how we want to build mpv. It's nice to see that some of the options that we use have been added as dependencies, but I think we still need to check whether Homebrew's formula is using correct set of flags that IINA needs. In any case we should also add instructions on how to build mpv from source as well, as well as documentation as to how to integrate it (but I think that's a separate discussion).

vitorgalvao:
Note that options themselves are not deprecated. If Homebrew’s current formula doesn’t suffice, you can always get the previous version with options from git history, and maintain it yourselves with a [tap](https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md).

saagarjha:
Yup, know about that too. We have some sort of tap [here](https://github.com/iina/homebrew-mpv-iina) which @alejx was working on.

lhc70000:
Thanks, let's get it merged.

saagarjha:
I'll open a new pull request once I get a stable way to get these integrated with MacPorts as well–but that'll require changes to change_lib_dependencies.rb as well so this could take a bit. Stay tuned!

Issue #2347, opened 2019-03-04
Context menu on right clicking video


anohren:
**Actual behavior:**
I couldn't find an existing discussion about this. 

When right clicking the video area you can currently either Play/Pause, Hide the on screen controls, or do nothing, but I think that what's generally even more expected is a context menu. That's what you get from a right click practically everywhere else in the OS, so it is a fairly strong expectation.

For most users the first experience with PiP on macOS will probably be in Safari where you enter it by right clicking a video and select it. It's very natural to look for PiP in that same place when using IINA for the first time (assuming that you noticed that it advertises PiP when you downloaded it).

I also think it would fit nicely in a context menu for existing users: Now that PiP can be entered from a fullscreen window without tearing it down, it can be a lot more convenient to just right click the video and select PiP from a menu than to move the cursor across the entire screen to get to the corresponding toolbar item (which might not even exist depending on your customization). 

I think it's a good idea to appeal to people who are used to watching video in Chrome/Safari, especially since IINA is becoming a very good replacement for that with the support of youtube-dl. A context menu would go a long way I think. I know it was the first thing I did after installing and wanting to try out PiP, at least.

I expect a good default context menu would not need more than two or three entries, plus maybe a temporary "Customize..." item at the bottom which would help with discovery of a potential settings panel for customizing the menu.

anohren:
A more convenient way for existing users interested in PiP: https://github.com/iina/iina/pull/2350

lhc70000:
This is actually what I'm considering recently. Please stay tuned :)

anohren:
I took the matter into my own hands and I'm perfectly happy with the PR I referenced :) but it's still a good feature for new users. Context Menu is a nice default mouse action in Control preferences.

PR #2350, opened 2019-03-05
Add Toggle PiP to mouse control actions

anohren:
A quick convenient way to enter picture in picture. Especially in fullscreen.

---

**Description:**
Corresponding items have been added to the menus in preferences for right click, double click and middle click.


alejx:
I don't think this is a good idea, but I'd like to discuss your idea in #2347 more. Currently what I and @lhc70000 think is to combine Video, Audio and Subtitles menu and make them show in the main window. Both right click and middle click can be configured to trigger the menus.

anohren:
I've used this personally since I created the PR and I've found that I use it _all the time_. If you have any reasons why this is a worse idea than binding a right click to e.g. toggle fullscreen or hide the OSD -- which is currently already possible -- I would be interested to hear it. Until then I will remain convinced that this is an awesome idea, from personal experience :)

anohren:
> Currently what I and @lhc70000 think is to combine Video, Audio and Subtitles menu and make them show in the main window. Both right click and middle click can be configured to trigger the menus.

That's interesting. I'm mostly concerned about having a huge context menu that quickly becomes overwhelming. You subconsciously avoid it if there are too many items. 

The value of a context menu in macOS isn't really that the menu bar items become accessible in fullscreen, unlike e.g. VLC on Windows, since the menu bar is always easily accessible in macOS. Instead the value of the context menu would be quick access to your most used actions. A lot of submenus would not be very inviting in that regard. That's my take on it.

iina/MainWindowController.swift
```diff
@@ -1056,6 +1056,11 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
       player.togglePause(nil)
     case .hideOSC:
       hideUI()
+    case .togglePIP:
+      if #available(OSX 10.12, *) {
```
alejx: OSX -> macOS

iina/MainWindowController.swift
```diff
@@ -1056,6 +1056,11 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
...
+        togglePIP()
```
alejx: You should directly use `menuTogglePIP` and pass a `.dummy`.

iina/MainWindowController.swift
alejx: OSX -> macOS

iina/MainWindowController.swift
alejx: You should directly use `menuTogglePIP` and pass a `.dummy`.

anohren:
> I don't think this is a good idea

Is this now considered for merging? If so I can take a look.

alejx:
Yes I'll merge this

iina/MainWindowController.swift
anohren: I'm sorry, I couldn't follow your reasoning here. Could you expand on it?

Is it the refactoring of common functionality from `menuTogglePIP` into `togglePIP` that you object to? It's done to preserve clear semantics of symbol names, since toggling pip isn't done exclusively from menus. I find it makes intention clearer than passing superfluous dummy objects.

anohren:
Well, I'm sorry @alejx but I cannot make changes in good conscience that I can't see the reasoning behind. You are free to merge and then apply additional changes, or close the PR, at your discretion.

iina/MainWindowController.swift
alejx: Sorry I didn't see your refactoring when I reviewed your pr. Your refactoring is great, but we are currently using menu actions with dummy menu item everywhere. So I believe we will take this into account when we do the next major refactoring.

alejx:
My bad I totally forgot this one. Don't worry, I'll merge this soon.

alejx:
Merged locally by db014075f0c93c177f0691cb47055c69a6eb9521

Issue #2221, opened 2019-01-15
application not responding open file in fullscreen


frimin:
- MacBook (Retina, 12-inch, 2017) macOS MoJave 10.14
- IINA 1.0.1 Build 103

In this case, IINA is playing a video in fullscreen, then open same video (or different) can show this problem. (IINA not crash)

![dongtu](https://user-images.githubusercontent.com/6148936/51189505-26f65500-191b-11e9-90ed-0354c2fea9fa.gif)

Steps:
1. open any video
2. double click to full screen
3. change to main desktop (control + <-)
4. open same video
5. desktop switch to IINA fullscreen (automatically)
6. **IINA not responding** 

------

在全屏模式下打开 IINA 时，滑切到另外的桌面点开其它文件时（不管是不是同一个文件） IINA会无响应（但并未崩溃）

lhc70000:
Sorry but I can't reproduce this issue. Could you provide detailed steps if you can reproduce it stably?

frimin:
> Sorry but I can't reproduce this issue. Could you provide detailed steps if you can reproduce it stably?

This problem is only happend in my MacBook(12-inch), working properly in MacBook Pro (2015; 12-inch; macOS High Sierra 10.13.6).

frimin:
I got a crash message when I kill program:
[iina.txt](https://github.com/iina/iina/files/2760416/iina.txt)

owenzhao:
The same here. It is easy to reproduce by:

1. in finder, double click to open a video
2. press f to full screen
3. press q to quit
4. open the same video

iina will open the video with full screen mode at the last played time, however everything hangs. 

Here is my screenshot when crashing. 

[![kCOIPg.md.png](https://s2.ax1x.com/2019/01/20/kCOIPg.md.png)](https://imgchr.com/i/kCOIPg)

version 1.0.1 Build 103

xjbeta:
This bug is caused by this commit.
https://github.com/iina/iina/commit/c25ef55781676d0407ea0a754dc16432859c9cc8

Noctem:
Disabling "Resume last playback position" in Preferences prevents these crashes for me.

dalDevelo:
I'm also having application hangs when using IINA in fullscreen and switching between desktop spaces. However, I am unable to reliably replicate the hang. I'm also unable to replicate using the steps mentioned above. I will post an update here as and when I can reliably replicate on my system.

[...]

PR #2378, opened 2019-03-26
Update for Swift 5

saagarjha:

---

Since Xcode 10.2 dropped yesterday, I think we should merge this in as soon as we get this looked over again. I think there are some `UnsafePointer` APIs left that I should look at; other than that it's mostly boilerplate.


saagarjha:
(We will also need a new libmpv compiled without Swift, FWIW)

Noctem:
Seems to be working smoothly with the same build tweaks that I normally make (0fe559ef119dfc128f4aeab4270f6d8ceb624d55) and these formulae for [ffmpeg](https://github.com/Noctem/homebrew-custom/blob/master/ffmpeg-iina.rb) and [mpv](https://github.com/Noctem/homebrew-custom/blob/master/mpv-iina.rb).

I still intermittently experience hangs/crashes when a video is launched directly to fullscreen (probably #2221), but I've been experiencing that for months. The conditions for triggering it don't seem to be entirely consistent however.

Podfile
```diff
@@ -5,10 +5,10 @@ target 'iina' do
   use_frameworks!
 
   # Pods for iina
-  pod 'Just', :git => 'https://github.com/JustHTTP/Just.git'
-  pod 'PromiseKit'
+  pod 'Just', :git => 'https://github.com/iina/Just', :branch => 'swift-5'
+  pod 'PromiseKit', :git => 'https://github.com/iina/PromiseKit', :branch => 'xcode-10.2-beta1'
```
Noctem: Upstream added Swift 5 support and seems to work for me.

saagarjha: Thanks for letting me know they've added this! I'll update this to point back at the upstream repository.

saagarjha:
Figured out how to build the libraries myself and double-checked that IINA at least runs. Moving out of draft status and requesting review.

PR #2384, opened 2019-03-30
Improve library management stuff

saagarjha:
- [X] It implements / fixes issue #2346.

---

Updates to the build process to accommodate alternative methods of building dependencies. The specifics of the changes are up for discussion (I don't actually know Ruby, lol), but I hope that the intent is clear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,18 +55,32 @@ $ brew install cocoapods
 
 ### Building mpv manually
 
-1. Install mpv:
+1. Build your own copy of mpv. If you're using a package manager to manage dependencies, the steps below outline the process.
+
+#### Homebrew
+
+Use our tap, since passes the correct flags during configuration:
 
 ```console
-$ brew install mpv
+$ brew tap iina/homebrew-mpv-iina
+$ brew install mpv-iina
+```
+
+#### MacPorts
+
+Pass in the flags when installing:
+
+```console
+# port install mpv +uchardet -bundle -rubberband configure.args="--enable-libmpv-shared --enable-lua --enable-libarchive --enable-libbluray --disable-swift --disable-rubberband" 
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -55,18 +55,32 @@ $ brew install cocoapods
...
-1. Install mpv:
+1. Build your own copy of mpv. If you're using a package manager to manage dependencies, the steps below outline the process.
+
+#### Homebrew
+
+Use our tap, since passes the correct flags during configuration:
+
+```console
+$ brew tap iina/homebrew-mpv-iina
+$ brew install mpv-iina --with-uchardet
```
alejx: `uchardet` is a dependency even if you do not specific it (both by [homebrew](https://github.com/Homebrew/homebrew-core/blob/6abd14c467c2bde2c6d857f7c4c3b4336541e4d3/Formula/mpv.rb#L27) and [us](https://github.com/iina/homebrew-mpv-iina/blob/3d3cef96cfacd847e6433df13a76e149c0ed5477/mpv-iina.rb#L27)). Besides, AFAIK, options has removed entirely removed by brew, which means we cannot use options even with our own taps.

saagarjha: > uchardet is a dependency even if you do not specific it (both by homebrew and us).

Good point, I forgot to remove that.

> Besides, AFAIK, options has removed entirely removed by brew, which means we cannot use options even with our own taps.

They've only been dropped for homebrew/core; Homebrew still supports (and in fact the "workaround" for the removal of options is to maintain them in personal taps). But as we are building for IINA only I agree with you that we shouldn't deal with options here.

saagarjha:
@alejx if you have the time, would you mind checking if I've fixed what you were asking for?

alejx:
@saagarjha Sure, I’ll probably check it tomorrow.

PR #2395, opened 2019-04-01
Fix stretched screen when entering and exiting fullscreen

alejx:

### DIFF
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -1080,6 +1080,9 @@ fileprivate func mpvGetOpenGLFunc(_ ctx: UnsafeMutableRawPointer?, _ name: Unsaf
 
 fileprivate func mpvUpdateCallback(_ ctx: UnsafeMutableRawPointer?) {
   let layer = unsafeBitCast(ctx, to: ViewLayer.self)
+  if layer.blocked {
+    return
+  }
   layer.mpvGLQueue.async {
     layer.draw()
   }
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -1510,11 +1510,6 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
       ])
     }
 
-    // is paused or very low fps (assume audio file), draw new frame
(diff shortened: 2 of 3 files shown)


iina/MPVController.swift
```diff
@@ -1080,6 +1080,9 @@ fileprivate func mpvGetOpenGLFunc(_ ctx: UnsafeMutableRawPointer?, _ name: Unsaf
 
 fileprivate func mpvUpdateCallback(_ ctx: UnsafeMutableRawPointer?) {
   let layer = unsafeBitCast(ctx, to: ViewLayer.self)
```
saagarjha: This isn't part of your changes, but I feel like `unsafeBitCast(_:to:)` is incorrect here. Perhaps this should use `UnsafeMutableRawPointer.assumingMemoryBound(to:)`?

iina/MPVController.swift
```diff
@@ -1080,6 +1080,9 @@ fileprivate func mpvGetOpenGLFunc(_ ctx: UnsafeMutableRawPointer?, _ name: Unsaf
...
+  if layer.blocked {
```
saagarjha: I'd use a guard here.

iina/MainWindowController.swift
```diff
@@ -1440,31 +1432,24 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
...
-      self.player.mpv.setFlag(MPVOption.Window.keepaspect, false)
-      DispatchQueue.main.sync {
-        for (_, constraint) in self.videoViewConstraints {
-          constraint.constant = 0
-        }
-        self.videoView.needsLayout = true
-        self.videoView.layoutSubtreeIfNeeded()
-        self.videoView.videoLayer.draw()
-      }
+    for (_, constraint) in self.videoViewConstraints {
```
saagarjha: `for constraint in self.videoViewConstraints.values`

lhc70000:
@saagarjha 

Fixed in dcc346f. An unmanaged reference of the VideoLayer was passed to mpv as the context, so here we should also use `Unmanaged` to load the pointer.

PR #2482, opened 2019-06-03
Fixed indentation in mpv build instructions

anohren:
Convert headings into subheadings to a particular step in the described build process

---

**Description:**
The non-indented headings threw me off and made me skip steps that weren't intended to be skipped.


saagarjha:
Thanks, I didn't know you could indent sections :)

Issue #2472, opened 2019-05-27
Can't open multiple YouTube videos via Safari extension


JohnClay:
As of 1.0.4, the Safari extension will only open a single video/stream. If you use the extension on a different page, it will replace the previously opened stream.

If you open the URL via the "Open URL in new window" option, that works as expected. The Safari extension now plays the video/stream over THAT window.

**System and IINA version:**

- macOS 10.14.5
- IINA 1.0.4 (107)

**Expected behavior:**

Each video/stream should open in a new window.

**Actual behavior:**

Each video/stream is forced into the same window as the previous file, interrupting playback.

**Steps to reproduce:**

Open video/stream via the Safari extension.

Open a second video/stream via the Safari extension.

All the time.

orion1vi:
`let url = URL(string: "iina://weblink?url=\(escapedURL)")` 
should be changed to 
`let url = URL(string: "iina://weblink?url=\(escapedURL)&new_window=1")`
in `SafariExtensionHandler.swift`, then it would work as expected.

anohren:
Just to be clear, this happens even when you have "always open media in new window" enabled?

JohnClay:
> Just to be clear, this happens even when you have "always open media in new window" enabled?

Yes, that's correct. It's a new behaviour in 1.0.4 w/o any change in setting. Dropping back to 1.0.3 fixes it.

orion1vi:
> "always open media in new window" enabled
When parsing urls, user defaults aren't checked, only the arguments that are parsed together. 
Personally, I don't see the use case of opening videos over existing stream from Safari, it seems that it would only cause annoyance, e.g. music playlist is playing in iina and then video opens over it.

anohren:
Regardless of its actual behavior, that setting is worded in no uncertain terms. It should probably either be honored or rephrased. AFAIR the menu options actually change into "Open URL in new window..." depending on that setting, so expecting the same from the browser doesn't feel far fetched.

Reusing a music mode window for videos doesn't sound like a good strategy, but on the other hand earmarking a window as most recently used by a browser extension and only then reuse it could save the annoyance of having 15 simultaneous video windows open by the browser.

JohnClay:
This is still an issue in 1.0.6 (113).

anohren:
Maybe because the extension technically has the option to “open in new iina window”

JohnClay:
> Maybe because the extension technically has the option to “open in new iina window”

[...]

PR #2483, opened 2019-06-04
Fixes videos opening over existing stream when launching from Safari

orion1vi:
- [x] It implements / fixes issue #2472.

---

**Description:**
Opens videos in new window when launched from Safari, instead of opening them over existing stream.


fenhl:
Why was this closed?

PR #2494, opened 2019-06-12
Correct directory names in readme

anohren:

---

**Description:**
I couldn't find `deps/libs` so I'm assuming `deps/lib`.


saagarjha:
Thanks!

PR #2496, opened 2019-06-13
Properly escape whitespace in change_lib_dependencies.rb

anohren:
Handle whitespace in path by surrounding it with double quotes

---

**Description:**
It's possible there's more than one place that need double quotes, I didn't check the whole document


alejx:
Thanks!

PR #2531, opened 2019-07-12
Fix infinite loops

lejuyan:
- [x] It implements / fixes issue #.

---

**Description:**
This commit intends to fix a kind of infinite loops:

If there are files like:
/-test.mp4
|-dir0/a.sub
|-dir1/a.sub
|-dir2/a.sub
|-dir3/a.sub

The tryGroupFiles method would find it self infinitely looping i++ when dealing with these "a.sub" files. And the user may find iina using 100% CPU at this time.

### DIFF
--- a/iina/FileGroup.swift
+++ b/iina/FileGroup.swift
@@ -89,13 +89,13 @@ class FileGroup {
     group.tryGroupFiles()
     return group
   }
-  
+
   init(prefix: String, contents: [FileInfo] = []) {
     self.prefix = prefix
     self.contents = contents
     self.groups = []
   }
-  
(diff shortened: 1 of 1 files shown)


iina/FileGroup.swift
```diff
@@ -133,6 +135,10 @@ class FileGroup {
         currChars.removeAll()
       }
       i += 1
+      // if all items have the same name
+      if false == anyProcessed {
```
alejx: `if !anyProcessed {`

lejuyan: Done.

PR #2598, opened 2019-10-02
Have the info window hide when the app isn't the top-most app.

MaddTheSane:
This fix makes it so that the info window doesn't hover over all other windows if IINA is not the front-most app.


alejx:
I'm not sure whether I like this idea. CC @lhc70000

Issue #2597, opened 2019-10-01
Localise the decimal separator


BasO:
**System and IINA version:**

- macOS 10.15 beta 10
- IINA 1.0.4 build 107

**Expected behavior:**

The decimal signifier is either a decimal or a dot, depending on where you live. If you use a localised version of IINA, you'd expect IINA to apply the correct sign in dynamically set values (`%.1`, `%.2`) e.g. the audio delay in the Menu and OSC. For Dutch, you'd expect to find a comma, e.g. `Audiovertraging: 0,00s` 

**Actual behavior:**

These numbers currently show up with the incorrect (American) dot separator for strings with dynamic numbers.

- [ ] MPV does not have this problem (not sure).

**How often does this happen?**
Always for Dutch. Other localisations of IINA that may likewise be affected are:

- Danish
- French
- German
- Italian
- Polish
- Russian
- Spanish
- Turkish
- Ukrainian
- Slovak
- Czech

anohren:
And Swedish.

Issue already exists: https://github.com/iina/iina/issues/2090

alejx:
Are there other conventions about numbers in your languages? If so, then I guess it might be necessary to use Number Formatter and set the locale to the current locale before we display any number on the user interface.

anohren:
Number formatter is the way to go. I think I confused that with string with format in the existing issue.

BasO:
I agree, and @anohren I imagine it's a common source of confusion when the function called localizedStringWIthFormat doesn't localize the format contents (: . The decimal signifier is what people are more sensitive to (even Spotlight calculator is strict about using the correct one for your locale) but the 'thousands operator' also differs per locale. [Examples](https://en.wikipedia.org/wiki/Decimal_separator#Examples_of_use).

PR #2604, opened 2019-10-06
Localize decimal numbers in OSD and main menu

alejx:
Ref: #2597

All decimals appear in OSD should be localized properly. However, some default values in mainMenu.xib should be localized in the corresponding `.string` files (`mainMenu.strings`).

CC @anohren @BasO Please take a look. Sorry about other junks in this PR.


BasO:
Verified. Good job, seems fine to me (using the Dutch locale). Just one section where I missed the thousands separator:

Schermafbeelding 2019-10-06 om 15 47 04

anohren:
I'll check it out.

Not on topic, but the units do not look right in @BasO's picture. The K in KB shouldn't be capitalized. Looks like integrating the Units framework is a future project for someone feeling bored :)

anohren:
First of all, the number formatting looks good pretty much everywhere. It's not a big deal if you didn't get every single one.

There's one place though where it's not used, which is in the Inspector. That window is just taking raw mpv output and displaying it. The proper way to do it would probably be to run it through the number formatter twice. 

I would have recommended to leave it for later since there are so many things in there that aren't localized, but the weird thing is that some of the values in the inspector are localized now (bitrate numbers) while most aren't. That inconsistency is not an improvement compared to leaving the entire view unlocalized.

Don't know if it's worth mentioning the keyboard shortcuts section at all, since that code isn't exactly the most...welcoming to localization. 

> CC @anohren @BasO Please take a look. Sorry about other junks in this PR.

 I'm not sure if you want me to ignore all the other changes or not, but if not: you seem to have intended to replace all ellipses with a single ligature, but it doesn't look like you used the right one. It seems to align to the median instead of the baseline.

alejx:
> you seem to have intended to replace all ellipses with a single ligature, but it doesn't look like you used the right one. It seems to align to the median instead of the baseline.

@anohren Thanks! Hopefully a simple replace would fix that.

> I would have recommended to leave it for later since there are so many things in there that aren't localized, but the weird thing is that some of the values in the inspector are localized now (bitrate numbers) while most aren't. That inconsistency is not an improvement compared to leaving the entire view unlocalized.

> Don't know if it's worth mentioning the keyboard shortcuts section at all, since that code isn't exactly the most...welcoming to localization.

I'll try to fix those as well.

> Just one section where I missed the thousands separator:

@BasO In fact, I only localized decimal numbers in this PR, will push more commit to handle thousand separtors. But this one should also be localized in the `.strings` files.

alejx:
@BasO I just merged your #2596 into develop branch and fixed the conflicts in this PR.
image
For this conflict, I have added '…' after "Zoek". Is that correct?

ErikMinekus:
@alejx macOS does not use an ellipsis for the placeholder in search fields, I guess that's why @BasO removed it.

BasO:
> @alejx macOS does not use an ellipsis for the placeholder in search fields, I guess that's why @BasO removed it.

Right. The search fields in macOS's built-in apps function as a filter (i.e. same as that function in IINA) but their placeholder is 'Zoek' ('Search'? 'Find'?), so I opted for the macOS default instead of a literal translation.

alejx:
I didn't check the context of this label. And because it does not trigger a new window, there should not be an ellipsis and we should change it for all languages.

Issue #2558, opened 2019-08-06
Browser addon: Make extension icon's menu available in context menu


anohren:
**Actual behavior:**

It would be convenient to have access to all the options currently only available in the extension's toolbar item (add to playlist, etc.) also in the context menu, so that it's not required to open 20 tabs to enqueue 20 items.

Since this feature sacrifices the convenience of single clicking in the context menu to open a link in favor of an extension specific submenu, it should probably be a configurable option for those who find that tradeoff acceptable.

Somewhat related:  #2537

PR #2607, opened 2019-10-07
Add more options to browser addon context menu

anohren:
Mimicks the toolbar button menu. No need to open a tab for each link just to enqueue a bunch of URLs.

Future improvement could include a setting to only show a single item in the context menu in order to avoid the submenu.

- [x] It implements / fixes issue #2558

---

### DIFF
--- a/browser/Chrome_Open_In_IINA/background.js
+++ b/browser/Chrome_Open_In_IINA/background.js
@@ -11,4 +11,40 @@ updateBrowserAction();
             openInIINA(tab.id, info[linkType]);
         },
     });
+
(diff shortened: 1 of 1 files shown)


anohren:
Opening a youtube channel's video page doesn't seem to work from the context menu, but works from the toolbar menu.

anohren:
I only use it on Firefox. It was three months since I suggested the change, but looking at the browser/Firefox folder it only contains a manifest file, so they should share all other resources. 

I use it as an unsigned extension. It’s just a zoo with a different extension.

Issue #2326, opened 2019-02-24
Detect desktop switching and have IINA switch to PiP?


anohren:
**Actual behavior:**

I started wondering whether it's possible to detect when the window is on a desktop/space (fullscreen app) that is no longer visible on the screen. If so that would be a nice point to automatically switch to PiP if there's a video playing. 

This way, when you have a fullscreen video playing and swipe to another desktop the video automatically "follows" you. Having a video play audio while not being visible seems like a corner case to me, so this would automate the general case, although it would of course not be the default behavior.

Since there's already a section in settings for PiP, it would fit naturally in there as an option.

Would anyone other than me use it?

Update:
[This looks relevant](https://developer.apple.com/documentation/appkit/nsworkspace/1527073-activespacedidchangenotification) although its description is pretty lacking.

PR #2632, opened 2019-11-03
WIP: Enter PIP when Inactive

dsdshcym:

---

**Description:**
Hi, there! 

This PR adds enterPIPWhenInactive feature to IINA, which I think would be really useful for multitasking.
iina

A few todos:
1. [ ] add preference toggle
2. [ ] call enterPIP & exitPIP less aggressively (now, exitPIP would be called when dragging the PIP window)

Let me know what you think about this feature and what I need to do next, thank you!

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -1582,6 +1582,10 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
       player.togglePause(false)
       isPausedDueToInactive = false
     }
+    // TODO: add Preference check
+    if #available(macOS 10.12, *) {
+      exitPIP()
+    }
   }
 
   func windowDidResignKey(_ notification: Notification) {
@@ -1593,6 +1597,13 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
       if Preference.bool(for: .pauseWhenInactive), !player.info.isPaused {
         player.togglePause(true)
         isPausedDueToInactive = true
+      }
+    }
+
+    // TODO: add Preference check
+    if NSApp.keyWindow == nil {
+      if #available(macOS 10.12, *) {
+        enterPIP()
       }
     }
   }


anohren:
Can you list the benefits explicitly?

Personally I think a subset of this feature would be useful, which I describe in https://github.com/iina/iina/issues/2326. Though, to switch to PiP _any_ time the window loses focus seems a bit too coarse a behavior to me, and I imagine I would be taken by surprise by the feature from time to time.

Sometimes it's necessary to react to unexpected events in other apps without necessarily wanting to. If that in turn entails side effects where windows fly across the screen it could ironically become a damper for interacting briefly with other apps, instead of encouraging and empowering multitasking. 

One problematic use case would be if we're in a split-screen configuration. As soon as you focus on the other app you'll find yourself in split screen with an empty window. Probably not what was intended.

I haven't tried it though. Perhaps it is useful to people who prefer all their windows on a single desktop.

https://github.com/iina/iina/pull/2350 is the solution I'm currently using in order to quickly and purposefully switch to/from PiP, without gotchas. I find that it brings both a sense of speed and control.

dsdshcym:
@anohren 

Yes, I totally agree that to switch to PiP any time the window loses focus is too coarse. That's why I listed the two todos above (preference toggle & switch to PiP less aggressively)

To make the goal more explicit here: I think of this feature as `pauseWhenInactive`, which is a preference IINA already provides. It should be an opt-in feature, so people who don't need this feature won't be surprised, and people who need this feature can turn it on.

To answer your first question, I think the main benefit is to be able to multitasking with a single display setup. The workflow would be: watching a video in fullscreen, switching to a message app, and the video would automatically be in PiP so I can still see the video.

Thanks for you reply! And if you can give it a try, that would be great. :)

anohren:
> That's why I listed the two todos above (preference toggle & switch to PiP less aggressively)

I can only speak for myself, but I don't consider the possibility to enable a feature as it being more fine-grained. It just makes it...usable. 

The second item, "exitPIP would be called when dragging the PIP window", simply sounds like a bug that needs fixing -- also not something I would call making the feature more fine-grained. Or did you have more improvements in mind here that you didn't mention?

By fine-grained I mean how sensitive the feature is to the context in which it is intended to activate automatically. 

> I think the main benefit is to be able to multitasking with a single display setup. The workflow would be: watching a video in fullscreen, switching to a message app, and the video would automatically be in PiP so I can still see the video.

Ok, I see. Qualifying the use case with a single monitor setup definitely simplifies things. I have some followup questions:

1. Most of the benefit can currently already be had by explicitly enabling PiP. How is this feature so much better that it weighs up the possible cons (which haven't been listed)?

2. For instance, is there any possibility that someone who has enabled the feature will forget that they enabled it? If you forgot about it, how frustrated would you become if the app did things seemingly on its own without being asked? This might not be a problem, I don't know. It's just a general danger when apps do things automatically.

3. PiP seems to activate also when the window is not in fullscreen. Is that just as desirable as when switching from fullscreen to another space? 

4. What about interacting with another app in split screen?

Don't get me wrong. I would potentially use such a feature -- as I mentioned I've suggested something similar -- but it doesn't mean I immediately think it would be beneficial to push it out to everyone. It depends on how easy it is to shoot yourself in the foot with it, be it accidentally or through forgetfulness. 

Right now I can imagine some hidden frustration even for those who made the conscious choice to activate this feature, since it fires in so many situations that might not be ideal (I've mentioned two). When they get bitten -- not if but _when_ -- will they continue to use it?

I might try it next time I do a build.

PR #2649, opened 2019-11-18
Add seekForwardCommand, seekBackwardCommand.

xjbeta:

---

**Description:**
Add more remote control command objects.
Press and hold the button in "Now Playing" to test the effect.
This will seek forward or backward 5s per second.

image


alejx:
Can you mimic the behavior of Music.app? Something that I noticed:
- Update the current time of MediaPlayer, otherwise it will not be updated until next UI sync, and will cause lag.
- The forward speed increases as the user holds the button longer. I didn't find the exact parttern, but it follows something like this:
5
7
7
8
10
11
13
15
18
21
21
So I think we can increase the step by 2s every 1s. Note that the fast rewind speed remains constant as the user holds the button.

PR #2678, opened 2019-12-30
Fix logger date format

al3xtjames:

---

**Description:**

Currently, a log directory for 2020-12-29 is created for the current date (December 29, 2019), which is incorrect. Use yyyy instead of YYYY to get the actual calendar year (instead of the week-based year).


alejx:
Thanks!

PR #2706, opened 2020-01-10
Double-click online subtitle to download it

andypotion:
Allows user to double-click on an online subtitle to download it immediately rather than having to select it then move the mouse to click the download button.


alejx:
Didn't try but LGTM.

PR #2712, opened 2020-01-12
Fix subtitle border width option in quick settings

stps:

---

Trying to set subtitle border width in the quick settings gets it stuck at 1, this hooks it up to use the correct selected value.

### DIFF
--- a/iina/QuickSettingViewController.swift
+++ b/iina/QuickSettingViewController.swift
@@ -707,10 +707,8 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
   }
 
   @IBAction func subTextSizeAction(_ sender: AnyObject) {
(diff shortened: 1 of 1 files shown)


stps:
> LGTM, but one question: why exactly is this failing? Is `subTextBorderWidthPopUp.stringValue` just wrong or `nil`? Why does it get stuck at 1 and not 3 (as that's the initial value of the popup)?

It seems `stringValue` is an NSControl thing, that doesn’t work correctly with NSPopUpButton. It was always returning `"1"` regardless of what’s selected.

saagarjha:
A bit of digging shows that `-[NSPopupButton stringValue]` delegates to `@(self.state).stringValue`, which means it's reading `NSControlStateValueOn` from its superclass, `NSButton`.

Issue #2729, opened 2020-01-17
Unable to toggle saved audio filters


ekstasis:
**System and IINA version:**

- macOS:  Mojave 10.14.6
- IINA:  1.0.6 Build 113 (develop HEAD)

**Behavior:**
Clicking the checkbox next to saved audio filters doesn't remove/add the checkmark, and the audio filter doesn't activate/deactivate.

Will provide mpv log on request.
<br>

**Steps to reproduce:**
1. Start a video so that you can hear the audio track.
2. Bring up Audio Filter dialog (Cmd-Shft-G) and create a new audio filter, e.g., Custom(mpv):
 Filter Name ="volume", Filter Value="volume=0.1"
3.  Hit save button and give it any name
4. Click checkbox next to this new saved filter
5. Checkmark should have changed and audio should have become barely audible

- I don't know if MPV has this problem.

**Solution:**
The issue is the `toggleSavedFilterAction` IBAction near/on line 168  in `FilterWindowController.swift`.  It calls `playercore.addVideoFilter` or `removeVideoFilter` even for audio filters.  I have rewritten the function -- let me know if I should go ahead and create a pull request.

**How often does this happen?**
Always.

saagarjha:
> I have rewritten the function -- let me know if I should go ahead and create a pull request.

That would be great!

PR #2731, opened 2020-01-18
Fix toggling saved audio filters (#2729)

ekstasis:
Rewrote "toggleSavedFilterAction" IBAction in FilterWindowController.swift.
* It was calling playercore.addVideoFilter or removeVideoFilter for
audio filters as well as video filters.

- [ x] It implements / fixes issue #2729 .

---

**Description:**

### DIFF
--- a/iina/FilterWindowController.swift
+++ b/iina/FilterWindowController.swift
@@ -164,19 +164,34 @@ class FilterWindowController: NSWindowController, NSWindowDelegate {
     saveFilter(filters[row])
   }
 
+  /// User activates or deactivates previously saved audio or video filter
+  /// - Parameter sender: A checkbox in lower portion of filter window
   @IBAction func toggleSavedFilterAction(_ sender: NSButton) {
     let row = savedFiltersTableView.row(for: sender)
-    let filter = savedFilters[row]
+    let savedFilter = savedFilters[row]
     let pc = PlayerCore.active
-    if sender.state == .on {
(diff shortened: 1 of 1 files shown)


iina/FilterWindowController.swift
```diff
@@ -164,19 +164,34 @@ class FilterWindowController: NSWindowController, NSWindowDelegate {
...
+
+    // choose approriate add/remove functions for .af/.vf
+    var addFilterFunction: (MPVFilter) -> Bool
+    var removeFilterFunction: (MPVFilter) -> Bool
+    if filterType == MPVProperty.vf {
+      addFilterFunction = pc.addVideoFilter
+      removeFilterFunction = pc.removeVideoFilter
     } else {
-      if pc.removeVideoFilter(MPVFilter(rawString: filter.filterString)!) {
+      addFilterFunction = pc.addAudioFilter(_:)
```
alejx: Be consistent with other functions references, remove the parameter list

alejx:
Merged locally. Thanks for your contribution.

iina/FilterWindowController.swift
ekstasis: You mean it should be "addFilterFunction = pc.addAudioFilter", right?  You've merged/closed, so it's just an FYI / "for future reference", right?  Thanks for the feedback.  Should I hit Resolve Conversation?  [sorry, i'm a noob]

saagarjha: Yes. There’s nothing you need to do; just keep it in mind for the future.

PR #2754, opened 2020-02-07
Guard thread unsafe dictionary in the singleton

inflation:
- [x] It implements / fixes issue #2745.

---

**Description:**
Reading mutable thread unsafe dictionary in a singleton simultaneously causes program crash. This fix guards all those access in serial order. 

If better performance is needed, we need to find another container.


Podfile.lock
```diff
@@ -42,7 +42,7 @@ CHECKOUT OPTIONS:
 SPEC CHECKSUMS:
   GRMustache.swift: 53405b02c374520f9193cf41bfa677f03a6d70aa
   GzipSwift: 5592f4d62b641e04d06443ba471f8ed76b1363e4
-  Just: 60c0995e04718eb621a07abef45ce6883c562f2b
+  Just: 3c64f42cd75ad819a4d42a1052006f7c44054d9e
```
alejx: Is this necessary?

lhc70000:
Merged locally (after resetting Podfiles.lock).

PR #2762, opened 2020-02-13
Try to utilize the new NSSwitch on 10.15

alejx:
Very ugly code, but looks good on UI. Any feedback welcomed.

image

image


saagarjha:
Can't say I'm a fan of `NSSwitch` being used in places like these :/

alejx:
Apple also uses it in the system preferences, almost the same usage.
image

anohren:
> Very ugly code, but looks good on UI. Any feedback welcomed.
> 
> image
> 
> image

As long as the effect gets applied immediately it looks fine. One of the few places where it actually makes sense to replace a checkbox with a switch, if anywhere.

PR #2771, opened 2020-02-15
Fix various issue regarding window size in music mode

lhc70000:
**Description:**

Currently, if there's no video track, mpv won't send the `video reconfig` event and thus windows won't change their sizes accordingly. Info such as aspect ratio and video size will remain to be the last video track. I'm not sure whether it worked well before and it's an internal change in mpv. Now the player remains functional due to an awkward fallback mechanism, and here are some obvious glitches:

- Disable auto music mode switching and open an audio file without album art. The window aspect ratio will be the initial one (16:10) at first, but suddenly jump to 1:1 on resize. This also happens when switching back from music mode.
- Open a landscape or portrait video, switch to music mode, and open another audio without album art. The album art part of the mini player will have a wrong aspect ratio.

In this PR, `notifyMainWindowVideoSizeChanged` is also called on `playback restart` if there's no video track to make sure windows receive the resize event. `player.originalVideoSize` is used to check if there's no video in the mini player. The default window size has been changed to 640x400.

Before merging, extensive tests are required to ensure it didn't break anything.

### DIFF
--- a/iina/MiniPlayerWindowController.swift
+++ b/iina/MiniPlayerWindowController.swift
@@ -395,8 +395,8 @@ class MiniPlayerWindowController: NSWindowController, NSWindowDelegate, NSPopove
   func updateVideoSize() {
     guard let window = window else { return }
     let videoView = player.mainWindow.videoView
-    let (width, height) = player.videoSizeForDisplay
-    let aspect = CGFloat(width) / CGFloat(height)
+    let (width, height) = player.originalVideoSize
+    let aspect = width == 0 || height == 0 ? 1 : CGFloat(width) / CGFloat(height)
     let currentHeight = videoView.frame.height
     let newHeight = videoView.frame.width / aspect
     updateVideoViewAspectConstraint(withAspect: aspect)
--- a/iina/MiniPlayerWindowController.swift
+++ b/iina/MiniPlayerWindowController.swift
@@ -396,7 +396,7 @@ class MiniPlayerWindowController: NSWindowController, NSWindowDelegate, NSPopove
     guard let window = window else { return }
     let videoView = player.mainWindow.videoView
     let (width, height) = player.originalVideoSize
-    let aspect = width == 0 || height == 0 ? 1 : CGFloat(width) / CGFloat(height)
+    let aspect = (width == 0 || height == 0) ? 1 : CGFloat(width) / CGFloat(height)
     let currentHeight = videoView.frame.height
     let newHeight = videoView.frame.width / aspect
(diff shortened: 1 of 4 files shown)


iina/MiniPlayerWindowController.swift
```diff
@@ -395,8 +395,8 @@ class MiniPlayerWindowController: NSWindowController, NSWindowDelegate, NSPopove
   func updateVideoSize() {
     guard let window = window else { return }
     let videoView = player.mainWindow.videoView
-    let (width, height) = player.videoSizeForDisplay
-    let aspect = CGFloat(width) / CGFloat(height)
+    let (width, height) = player.originalVideoSize
+    let aspect = width == 0 || height == 0 ? 1 : CGFloat(width) / CGFloat(height)
```
alejx: Very poor readability, consider adding parentheses around `width == 0 || height == 0`. Or write a helper function (extension?) since I see you use it mutiple times

iina/PlayerCore.swift
```diff
@@ -1089,6 +1086,11 @@ class PlayerCore: NSObject {
   func playbackRestarted() {
     Logger.log("Playback restarted", subsystem: subsystem)
     reloadSavedIINAfilters()
+
+    if info.vid == 0 {
```
alejx: When deciding whether we will show the default album art, `self.info.videoTracks.isEmpty` is used. Not sure which one is the better one, or should we make it a function in `PlaybackInfo`?

iina/PlayerCore.swift
```diff
@@ -1089,6 +1086,11 @@ class PlayerCore: NSObject {
...
+      notifyMainWindowVideoSizeChanged()
```
alejx: Also, should we move this to `PlayerCore.fileLoaded`?

```diff
diff --git a/iina/PlayerCore.swift b/iina/PlayerCore.swift
index dcbb67ae..1f6b6dcb 100644
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -1063,6 +1063,7 @@ class PlayerCore: NSObject {
 
       if self.isInMiniPlayer {
         miniPlayer.defaultAlbumArt.isHidden = !self.info.videoTracks.isEmpty
+        self.miniPlayer.updateVideoSize()
       }
     }
     // set initial properties for the first file
```
Because currently, since hiding & un-hiding the default art and `updateVideoSize` for mini player are not executed consecutively, we can see the default art shows in the original aspect ratio and then the size of the mini player got updated when we change from an audio file without art to a landscape video in music mode.

PR #2772, opened 2020-02-16
WIP: Refactor MainWindowController & MiniPlayerWindowController

alejx:
Make `MainWindowController` and `MiniPlayerWindowController` inherent from `PlayerWindowController`.

`PlayerWindowController` consists same or similar code from `MainWindowController`
and `MiniPlayerWindowController`. There are some functions in `MiniPlayerWindowControler`
directly copy-pasted from `MainWindowController`.

Besides, we can implement more features in the main window but not in the miniplayer without copy-pasting code (e.g. https://github.com/iina/iina/pull/2191.)

Ref: #2157

### DIFF
--- a/iina/Base.lproj/MiniPlayerWindowController.xib
+++ b/iina/Base.lproj/MiniPlayerWindowController.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="15702" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="15705" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="15702"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="15705"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <objects>
@@ -45,7 +45,7 @@
             <windowCollectionBehavior key="collectionBehavior" fullScreenNone="YES"/>
             <windowPositionMask key="initialPositionMask" leftStrut="YES" rightStrut="YES" topStrut="YES" bottomStrut="YES"/>
             <rect key="contentRect" x="196" y="240" width="300" height="72"/>
-            <rect key="screenRect" x="0.0" y="0.0" width="3840" height="1057"/>
+            <rect key="screenRect" x="0.0" y="0.0" width="2560" height="1057"/>
             <value key="minSize" type="size" width="300" height="72"/>
             <view key="contentView" id="se5-gp-TjO">
                 <rect key="frame" x="0.0" y="0.0" width="300" height="72"/>
@@ -112,7 +112,7 @@
                                     <textField horizontalHuggingPriority="251" verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="h25-bp-EvL" customClass="ScrollingTextField" customModule="IINA" customModuleProvider="target">
                                         <rect key="frame" x="133" y="26" width="35" height="16"/>
                                         <textFieldCell key="cell" lineBreakMode="truncatingMiddle" sendsActionOnEndEditing="YES" alignment="center" title="Title" id="Rs9-6l-6vp">
-                                            <font key="font" metaFont="system"/>
+                                            <font key="font" metaFont="label" size="13"/>
                                             <color key="textColor" name="labelColor" catalog="System" colorSpace="catalog"/>
                                             <color key="backgroundColor" name="controlColor" catalog="System" colorSpace="catalog"/>
                                         </textFieldCell>
@@ -149,7 +149,7 @@
                                         </constraints>
                                         <buttonCell key="cell" type="square" bezelStyle="shadowlessSquare" image="play" imagePosition="only" alignment="center" alternateImage="pause" refusesFirstResponder="YES" state="on" imageScaling="proportionallyUpOrDown" inset="2" id="F9X-N7-Uv0">
                                             <behavior key="behavior" pushIn="YES" changeContents="YES" lightByContents="YES"/>
-                                            <font key="font" metaFont="system"/>
+                                            <font key="font" metaFont="label" size="13"/>
                                         </buttonCell>
                                         <connections>
                                             <action selector="playBtnAction:" target="-2" id="7Hb-fd-gCC"/>
@@ -163,7 +163,7 @@
                                         </constraints>
                                         <buttonCell key="cell" type="square" bezelStyle="shadowlessSquare" image="nextr" imagePosition="only" alignment="center" refusesFirstResponder="YES" imageScaling="proportionallyUpOrDown" inset="2" id="DPp-zB-KXR">
                                             <behavior key="behavior" pushIn="YES" lightByBackground="YES" lightByGray="YES"/>
-                                            <font key="font" metaFont="system"/>
+                                            <font key="font" metaFont="label" size="13"/>
                                         </buttonCell>
                                         <connections>
                                             <action selector="nextBtnAction:" target="-2" id="ivD-3f-1va"/>
@@ -177,7 +177,7 @@
                                         </constraints>
                                         <buttonCell key="cell" type="square" bezelStyle="shadowlessSquare" image="nextl" imagePosition="only" alignment="center" refusesFirstResponder="YES" imageScaling="proportionallyUpOrDown" inset="2" id="bti-JU-2h9">
                                             <behavior key="behavior" pushIn="YES" lightByBackground="YES" lightByGray="YES"/>
-                                            <font key="font" metaFont="system"/>
+                                            <font key="font" metaFont="label" size="13"/>
                                         </buttonCell>
                                         <connections>
                                             <action selector="prevBtnAction:" target="-2" id="BDa-Ku-caU"/>
@@ -191,7 +191,7 @@
                                         </constraints>
                                         <buttonCell key="cell" type="square" bezelStyle="shadowlessSquare" image="playlist" imagePosition="only" alignment="center" refusesFirstResponder="YES" imageScaling="proportionallyUpOrDown" inset="2" id="9be-Tw-9hy">
                                             <behavior key="behavior" pushIn="YES" lightByBackground="YES" lightByGray="YES"/>
-                                            <font key="font" metaFont="system"/>
+                                            <font key="font" metaFont="label" size="13"/>
                                         </buttonCell>
                                         <connections>
                                             <action selector="togglePlaylist:" target="-2" id="0da-p5-HBI"/>
@@ -205,7 +205,7 @@
                                         </constraints>
                                         <buttonCell key="cell" type="square" bezelStyle="shadowlessSquare" image="volume" imagePosition="only" alignment="center" alternateImage="mute" refusesFirstResponder="YES" imageScaling="proportionallyUpOrDown" inset="2" id="9bz-eY-31X">
                                             <behavior key="behavior" pushIn="YES" lightByBackground="YES" lightByGray="YES"/>
-                                            <font key="font" metaFont="system"/>
+                                            <font key="font" metaFont="label" size="13"/>
                                         </buttonCell>
                                         <connections>
                                             <action selector="volumeBtnAction:" target="-2" id="ynf-E4-jte"/>
@@ -219,7 +219,7 @@
                                         </constraints>
                                         <buttonCell key="cell" type="square" bezelStyle="shadowlessSquare" image="toggle-album-art" imagePosition="only" alignment="center" refusesFirstResponder="YES" imageScaling="proportionallyUpOrDown" inset="2" id="VB3-JD-dpR">
                                             <behavior key="behavior" pushIn="YES" lightByBackground="YES" lightByGray="YES"/>
-                                            <font key="font" metaFont="system"/>
+                                            <font key="font" metaFont="label" size="13"/>
                                         </buttonCell>
                                         <connections>
                                             <action selector="toggleVideoView:" target="-2" id="Q4b-pa-4UM"/>
@@ -290,7 +290,7 @@
                                         </constraints>
                                         <buttonCell key="cell" type="square" bezelStyle="shadowlessSquare" image="NSStopProgressFreestandingTemplate" imagePosition="only" alignment="center" refusesFirstResponder="YES" imageScaling="proportionallyUpOrDown" inset="2" id="A3f-Go-Xss">
                                             <behavior key="behavior" pushIn="YES" lightByBackground="YES" lightByGray="YES"/>
-                                            <font key="font" metaFont="system"/>
+                                            <font key="font" metaFont="label" size="13"/>
                                         </buttonCell>
                                     </button>
                                     <button toolTip="Back to video mode" translatesAutoresizingMaskIntoConstraints="NO" id="WS6-ix-uBJ">
@@ -301,7 +301,7 @@
                                         </constraints>
                                         <buttonCell key="cell" type="square" bezelStyle="shadowlessSquare" image="back" imagePosition="overlaps" alignment="center" refusesFirstResponder="YES" imageScaling="proportionallyUpOrDown" inset="2" id="FaS-tJ-QNE">
                                             <behavior key="behavior" pushIn="YES" lightByBackground="YES" lightByGray="YES"/>
-                                            <font key="font" metaFont="system"/>
+                                            <font key="font" metaFont="label" size="13"/>
                                         </buttonCell>
                                         <connections>
                                             <action selector="backBtnAction:" target="-2" id="wwC-Mg-w9P"/>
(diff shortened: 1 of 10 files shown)


iina/Base.lproj/MiniPlayerWindowController.xib
```diff
@@ -112,7 +112,7 @@
                                     <textField horizontalHuggingPriority="251" verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="h25-bp-EvL" customClass="ScrollingTextField" customModule="IINA" customModuleProvider="target">
                                         <rect key="frame" x="133" y="26" width="35" height="16"/>
                                         <textFieldCell key="cell" lineBreakMode="truncatingMiddle" sendsActionOnEndEditing="YES" alignment="center" title="Title" id="Rs9-6l-6vp">
-                                            <font key="font" metaFont="system"/>
```
saagarjha: I assume Xcode did this automatically, but I am not sure it's what we want…

iina/MainWindowController.swift
```diff
@@ -2744,13 +2648,9 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
 
   // MARK: - Utility
 
-  func handleIINACommand(_ cmd: IINACommand) {
-    let appDeletate = (NSApp.delegate! as! AppDelegate)
+  internal override func handleIINACommand(_ cmd: IINACommand) {
```
saagarjha: It's not something new, but do we need to use `self` in this function?

iina/MiniPlayerWindowController.swift
```diff
@@ -338,57 +301,31 @@ class MiniPlayerWindowController: NSWindowController, NSWindowDelegate, NSPopove
...
-    DispatchQueue.main.async {
-      let (mediaTitle, mediaAlbum, mediaArtist) = self.player.getMusicMetadata()
-      self.titleLabel.stringValue = mediaTitle
-      self.window?.title = mediaTitle
-      // hide artist & album label when info not available
-      if mediaArtist.isEmpty && mediaAlbum.isEmpty {
-        self.titleLabelTopConstraint.constant = 6 + 10
-        self.artistAlbumLabel.stringValue = ""
+  override func updateTitle() {
+    let (mediaTitle, mediaAlbum, mediaArtist) = self.player.getMusicMetadata()
```
saagarjha: I don't think we need `self` here either?

iina/PlayerWindowController.swift
```diff
@@ -0,0 +1,170 @@
...
+  init(playerCore: PlayerCore) {
+    self.player = playerCore
+    super.init(window: nil)
+  }
+
+  required init?(coder: NSCoder) {
+    fatalError("init(coder:) has not been implemented")
+  }
+  
+  internal lazy var followGlobalSeekTypeWhenAdjustSlider: Bool = Preference.bool(for: .followGlobalSeekTypeWhenAdjustSlider)
```
saagarjha: Does this need an annotation? Also, I'm not sure if we can change the name now but if we can I would consider making it shorter and more grammatical.

iina/PlayerWindowController.swift
```diff
@@ -0,0 +1,170 @@
...
+  @IBOutlet weak var volumeSlider: NSSlider!
+  @IBOutlet weak var muteButton: NSButton!
+  @IBOutlet weak var playButton: NSButton!
+  @IBOutlet weak var playSlider: NSSlider!
+  @IBOutlet weak var rightLabel: DurationDisplayTextField!
+  @IBOutlet weak var leftLabel: NSTextField!
+  
+  override func windowDidLoad() {
+    super.windowDidLoad()
+    loaded = true
```
saagarjha: Is this different than [`NSWindowController.isWindowLoaded`](https://developer.apple.com/documentation/appkit/nswindowcontroller/1527496-iswindowloaded)?

iina/PlayerWindowController.swift
```diff
@@ -0,0 +1,170 @@
...
+    
+    guard let window = window else { return }
```
saagarjha: I would argue that this should crash.

iina/PlayerWindowController.swift
```diff
@@ -0,0 +1,170 @@
...
+    setMaterial(Preference.enum(for: .themeMaterial))
+    
+    notificationCenter(.default, addObserverfor: .iinaMediaTitleChanged, object: player) { [unowned self] _ in
+        self.updateTitle()
+    }
+    
+    updateVolume()
+  }
+  
+  internal func notificationCenter(_ center: NotificationCenter, addObserverfor name: NSNotification.Name, object: Any? = nil, using block: @escaping (Notification) -> Void) {
```
saagarjha: `notificationCenter(_ center: NotificationCenter, addObserverForName name: Notification.Name, object: Any? = nil, using block: @escaping (Notification) -> Void)`

iina/PlayerWindowController.swift
```diff
@@ -0,0 +1,170 @@
...
+    let observer = center.addObserver(forName: name, object: object, queue: .main, using: block)
+    if notificationObservers[center] == nil {
```
saagarjha: `notificationObservers[center, default: []].append(observer)`

iina/PlayerWindowController.swift
```diff
@@ -0,0 +1,170 @@
...
+      notificationObservers[center] = []
+    }
+    notificationObservers[center]!.append(observer)
+  }
+
+  
+  internal func setMaterial(_ theme: Preference.Theme?) {
+    guard let window = window, let theme = theme else { return }
```
saagarjha: When would we accept a `nil` theme?

iina/PlayerWindowController.swift
```diff
@@ -0,0 +1,170 @@
...
+    guard !player.info.fileLoading else { return }
+    let percentage = 100 * sender.doubleValue / sender.maxValue
+    player.seek(percent: percentage, forceExact: !followGlobalSeekTypeWhenAdjustSlider)
+  }
+  
+  
+  /** This method will not set `isOntop`! */
+  func setWindowFloatingOnTop(_ onTop: Bool) {
+    guard let window = window else { return }
+    if onTop {
```
saagarjha: This could be a ternary.

iina/PlayerWindowController.swift
```diff
@@ -0,0 +1,170 @@
...
+      window.level = .iinaFloating
+    } else {
+      window.level = .normal
+    }
+  }
+  
+  internal func handleIINACommand(_ cmd: IINACommand) {
+    let appDeletate = (NSApp.delegate! as! AppDelegate)
```
saagarjha: `appDelegate`

iina/PlayerWindowController.swift
```diff
@@ -0,0 +1,170 @@
...
+    switch cmd {
+    case .openFile:
+      appDeletate.openFile(self)
+    case .openURL:
+      appDeletate.openURL(self)
+    case .flip:
+      self.menuActionHandler.menuToggleFlip(.dummy)
```
saagarjha: I don't think we need `self` for all of these, too.

iina/MainWindowController.swift
alejx: No.

iina/MiniPlayerWindowController.swift
alejx: Yeah, this was surrounded by DispatchQueue, forgot to remove those `self.`

iina/PlayerWindowController.swift
alejx: Feel free to suggest a new name or something. What I did was just copy-paste from the two classes.

iina/PlayerWindowController.swift
alejx: This variable is added by me recently, see f9b23d4. Should be the same, but accessing `NSWindowController.isWIndowLoaded` requires main thread, but we have too many of them that we cannot ensure they are all on the main thread. Any idea?

iina/PlayerWindowController.swift
alejx: Then just use `!`?

iina/Base.lproj/MiniPlayerWindowController.xib
alejx: Unfortunately Xcode is doing this all the time, every time when you have changes in a xib file. But I don't think it will make a difference, otherwise Xcode shouldn't do this automatically,

iina/MiniPlayerWindowController.swift
saagarjha: Since we already have a preference key with this string, I think we should leave it for now :(

iina/PlayerWindowController.swift
saagarjha: Since we already have a preference key with this string, I think we should leave it for now :(

iina/PlayerWindowController.swift
saagarjha: Hmm, I see. It's somewhat annoying to have to keep track of this ourselves, but for now I don't have any ideas.

iina/PlayerWindowController.swift
saagarjha: Yes.

iina/MainWindowController.swift
```diff
@@ -57,16 +55,11 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
...
-
-  lazy var videoView: VideoView = {
-    let view = VideoView(frame: self.window!.contentView!.bounds)
-    view.player = self.player
-    return view
-  }()
-
-  /** A responder handling general menu actions. */
-  var menuActionHandler: MainMenuActionHandler!
+  override var videoView: VideoView {
```
saagarjha: We may want to consider making this kind of thing a property wrapper at some point. I think the proposal has [a decent example](https://github.com/apple/swift-evolution/blob/master/proposals/0258-property-wrappers.md#proposed-solution) of what this can look like.

iina/MainWindowController.swift
```diff
@@ -884,45 +736,13 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
...
-        if keyBinding.action[0] == MPVCommand.screenshot.rawValue {
-          player.sendOSD(.screenshot)
-        }
-      } else {
-        Logger.log("Return value \(returnValue) when executing key command \(keyBinding.rawAction)", level: .error)
-      }
+  @discardableResult
+  override func handleKeyBinding(_ keyBinding: KeyMapping) -> Bool {
+    let success = super.handleKeyBinding(keyBinding)
+    if (success && keyBinding.action[0] == MPVCommand.screenshot.rawValue) {
```
saagarjha: `.first`?

iina/MiniPlayerWindowController.swift
```diff
@@ -216,42 +138,41 @@ class MiniPlayerWindowController: NSWindowController, NSWindowDelegate, NSPopove
...
+  override func otherMouseUp(with event: NSEvent) {
+    guard !isMouseEvent(event, inAnyOf: [backgroundView]) else { return }
+    super.otherMouseUp(with: event)
+  }
+  
+  override func scrollWheel(with event: NSEvent) {
+    if isMouseEvent(event, inAnyOf: [playSlider]) && playSlider.isEnabled {
+      seekOverride = true
+    } else if isMouseEvent(event, inAnyOf: [volumeSliderView]) && volumeSlider.isEnabled {
+    volumeOverride = true
```
saagarjha: You're missing an indent here

iina/PlayerWindowController.swift
```diff
@@ -0,0 +1,516 @@
...
+      }
+    })
+  }
+  
+  deinit {
+    ObjcUtils.silenced {
+      for key in self.observedPrefKeys {
+        UserDefaults.standard.removeObserver(self, forKeyPath: key.rawValue)
+      }
+      for (center, observers) in self.notificationObservers {
```
saagarjha: This is unnecessary on OS X 10.11+: https://developer.apple.com/documentation/foundation/notificationcenter/1407263-removeobserver

iina/PlayerWindowController.swift
```diff
@@ -0,0 +1,516 @@
...
+        for observer in observers {
+          center.removeObserver(observer)
+        }
+      }
+    }
+  }
+  
+  internal func notificationCenter(_ center: NotificationCenter, addObserverForName name: Notification.Name, object: Any? = nil, using block: @escaping (Notification) -> Void) {
```
saagarjha: After some thought, I think this name evokes a delegate method. perhaps `addObserver(to notificationCenter: NotificationCenter, forName name: Notification.Name, object: Any?, using block: @escaping (Notification) -> Void)`?

iina/PlayerWindowController.swift
```diff
@@ -0,0 +1,516 @@
...
+        return true
+      } else {
+        Logger.log("Unknown iina command \(keyBinding.rawAction)", level: .error)
+        return false
+      }
+    } else {
+      // - mpv command
+      let returnValue: Int32
+      // execute the command
+      switch keyBinding.action[0] {
```
saagarjha: Also possibly `.first`?

iina/PlayerWindowController.swift
```diff
@@ -0,0 +1,516 @@
...
+      }
+    }
+  }
+  
+  override func mouseUp(with event: NSEvent) {
+    if event.clickCount == 1 {
+      if doubleClickAction == .none {
+        performMouseAction(singleClickAction)
+      } else {
+        singleClickTimer = Timer.scheduledTimer(timeInterval: NSEvent.doubleClickInterval, target: self, selector: #selector(self.performMouseActionLater(_:)), userInfo: singleClickAction, repeats: false)
```
saagarjha: Can we get by with just `performMouseActionLater`?

iina/PlayerWindowController.swift
```diff
@@ -0,0 +1,516 @@
...
+        return true
+      } else {
+        Logger.log("Return value \(returnValue) when executing key command \(keyBinding.rawAction)", level: .error)
+        return false
+      }
+    }
+  }
+  
+  override func mouseUp(with event: NSEvent) {
+    if event.clickCount == 1 {
```
saagarjha: Yuck, please use the actual constants if possible: https://developer.apple.com/documentation/appkit/nsevent/eventtype

iina/PlayerWindowController.swift
```diff
@@ -0,0 +1,516 @@
...
+
+    if scrollAction == .seek && isTrackpadBegan {
+      // record pause status
+      wasPlayingWhenSeekBegan = player.info.isPlaying
+      if wasPlayingWhenSeekBegan! {
+        player.pause()
+      }
+    }
+
+    if isTrackpadEnd && wasPlayingWhenSeekBegan != nil {
```
saagarjha:
```swift
if isTrackPadEnd && wasPlayingWhenSeekBegan ?? false {
	player.resume()
	wasPlayingWhenSeekBegan = nil
}
```

(Why is this an `Optional` again?)

iina/PlayerWindowController.swift
alejx: Maybe not in this PR, since there are a lot not in these files as well.

iina/PlayerWindowController.swift
alejx: Do you mean to move the timer inside `performMouseActionLater`?

iina/PlayerWindowController.swift
alejx: I don't think there's an event for single click so we have to handle it manually.

iina/MainWindowController.swift
alejx: Wow, very cool stuff. We can maybe rewrite a lot of code using property wrapper.

iina/PlayerWindowController.swift
saagarjha: Sorry, I meant to put this on the code that checks `buttonNumber`.

iina/PlayerWindowController.swift
saagarjha: No, I meant:

```swift
singleClickTimer = Timer.scheduledTimer(timeInterval: NSEvent.doubleClickInterval, target: self, selector: #selector(performMouseActionLater), userInfo: singleClickAction, repeats: false)
```

iina/MainWindowController.swift
```diff
@@ -307,57 +273,30 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
...
-  private lazy var singleClickAction: Preference.MouseClickAction = Preference.enum(for: .singleClickAction)
-  private lazy var doubleClickAction: Preference.MouseClickAction = Preference.enum(for: .doubleClickAction)
   private lazy var pinchAction: Preference.PinchAction = Preference.enum(for: .pinchAction)
-  private lazy var followGlobalSeekTypeWhenAdjustSlider: Bool = Preference.bool(for: .followGlobalSeekTypeWhenAdjustSlider)
   lazy var displayTimeAndBatteryInFullScreen: Bool = Preference.bool(for: .displayTimeAndBatteryInFullScreen)
 
-  /** A list of observed preference keys. */
-  private let observedPrefKeys: [Preference.Key] = [
-    .themeMaterial,
+  private let localObservedPrefKeys: [PK] = [
```
lhc70000: `Preference.Key`

iina/MainWindowController.swift
```diff
@@ -562,53 +449,28 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
...
-    guard let w = self.window else { return }
 
-    w.initialFirstResponder = nil
-
-    // Insert `menuActionHandler` into the responder chain
-    menuActionHandler = MainMenuActionHandler(playerCore: player)
-    let responder = w.nextResponder
-    w.nextResponder = menuActionHandler
-    menuActionHandler.nextResponder = responder
+    guard let window = self.window else { return }
```
lhc70000: No need for `self`

iina/MainWindowController.swift
```diff
@@ -884,45 +736,13 @@ class MainWindowController: NSWindowController, NSWindowDelegate {
...
-        if keyBinding.action[0] == MPVCommand.screenshot.rawValue {
-          player.sendOSD(.screenshot)
-        }
-      } else {
-        Logger.log("Return value \(returnValue) when executing key command \(keyBinding.rawAction)", level: .error)
-      }
+  @discardableResult
+  override func handleKeyBinding(_ keyBinding: KeyMapping) -> Bool {
+    let success = super.handleKeyBinding(keyBinding)
+    if (success && keyBinding.action.first! == MPVCommand.screenshot.rawValue) {
```
lhc70000: No parentheses

iina/PlayerWindowController.swift
alejx: Did you mean to use `.otherMouseUp`?
- This one is only for macOS 10.12+, says the documentation (but there's no compile-time check for this)
- It seems like all the buttons on a mouse except the right button and left button are `other mouse`, so it cannot help us to identify whether or not this is a middle button.

alejx:
Merged locally.

Issue #2111, opened 2018-12-06
[Feature Request] Have welcome window show URL history


mwinkw:
macOS 10.14.1
IINA 0.0.15.1 Build 69

**Expected behavior:**

Would be nice for the welcome screen to show a history of URLs/streams that were played.

**Actual behavior:**

Welcome screen currently shows only video files I have played from my hard drive.

darkscarf1234:
Agreed, this would be a great addition!

PR #2812, opened 2020-03-08
Make initial window source playback history

allie:
- [x] It implements / fixes issue #2111.

---

**Description:**
This change is in response to both #2111 as well as my own desires.
Before, the recent files list in InitialWindowController sourced
NSDocumentController's recent files, meaning if the last file you
played in IINA was a file on a server or a livestream, it would not
show up on the initial window's recent files list, nor would there
be a resume button available.

With this commit, the initial window's recent files list data now
comes from PlaybackHistory, meaning network files and live streams
will show up in this list. Additionally, a resume button will show
when your last played file was a network resource. If the playback
history entry looks like a normal video file but is a network
resource, just the file name will be shown (ignoring the rest of
the URL), and the last playback position will be shown in the
resume control. If the entry looks like a live stream URL, the full
URL will be displayed, and the last playback position is not shown,
as it's likely irrelevant.

Debatably, it would be nice to include network resources in the
"Open recent" menu as well, though unfortunately it looks like a
more involved task, likely needing to either subclass
NSDocumentController or write a new menu altogether. For now, I
personally think it's okay to have "Open recent" only show local
files while the initial window includes network resources as well.

Example where the last played file was a `.mkv` on my VPS: 
Screen Shot 2020-03-08 at 12 40 36 PM

Example where the last played URL was a Twitch stream:
Screen Shot 2020-03-08 at 12 40 25 PM

### DIFF
--- a/iina/InitialWindowController.swift
+++ b/iina/InitialWindowController.swift
@@ -106,8 +106,8 @@ class InitialWindowController: NSWindowController {
     }
   }
 
-  lazy var recentDocuments: [URL] = {
-    NSDocumentController.shared.recentDocumentURLs.filter { $0 != lastPlaybackURL }
+  lazy var playbackHistory: [PlaybackHistory] = {
+    HistoryController.shared.history.filter { $0.url != lastPlaybackURL }
   }()
   private var lastPlaybackURL: URL?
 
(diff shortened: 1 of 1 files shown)


allie:
To clarify on the force push, the build check had errored before and I was triggering a rebuild with no changes. Sorry about that!

saagarjha (changes requested):
Hi @allie, thanks for coding this up! Since the change isn’t huge there’s not a lot to talk about for the code, but I think there are a couple of things we can discuss about how this would fit in with what we have now. In addition to some of the comments I left, I do know that some people prefer not having network streams appear in the list so I’m not sure if we will need to add a preference for this. Hopefully the other maintainers can weigh in with what they think before we can move forwards. 

(By the way, force pushing your changes is fine, we don’t mind.)

anohren:
A question, will this change the number of history entries shown in the welcome window?

Issue #2817, opened 2020-03-11
Wrong preview screenshot direction on the progress bar


Guochengjie:
**System and IINA version:**

- macOS 10.15.2
- IINA 1.0.7-beta1 Build 115

**Expected behavior:**

The preview on the progress bar with the same direction as the video's.

**Actual behavior:**

The preview on the progress bar rotates for 90 degrees.
Like this:
![Screenshot 2020-03-11 at 11.21.51 PM.png](https://i.loli.net/2020/03/11/D4AUZLYvc5bJWoi.png)

<br>

<details>
<summary><b>mpv log:</b></summary>

```
[   0.013][v][cplayer] mpv 0.32.0 Copyright © 2000-2020 mpv/MPlayer/mplayer2 projects
[   0.013][v][cplayer]  built on Tue Feb 25 21:16:35 CST 2020
[   0.013][v][cplayer] ffmpeg library versions:
[   0.013][v][cplayer]    libavutil       56.31.100
[   0.013][v][cplayer]    libavcodec      58.54.100
[   0.013][v][cplayer]    libavformat     58.29.100
[   0.013][v][cplayer]    libswscale      5.5.100
[   0.013][v][cplayer]    libavfilter     7.57.100
[   0.013][v][cplayer]    libswresample   3.5.100
[   0.013][v][cplayer] ffmpeg version: 4.2.1
[   0.013][v][cplayer] 
[   0.013][v][cplayer] Configuration: waf configure --prefix=/usr/local/Cellar/mpv-iina/0.32.0 --enable-javascript --enable-libmpv-shared --enable-lua --enable-libarchive --enable-uchardet --enable-libbluray --disable-swift --disable-debug-build --disable-macos-media-player --confdir=/usr/local/etc/mpv --datadir=/usr/local/Cellar/mpv-iina/0.32.0/share/mpv-iina --mandir=/usr/local/Cellar/mpv-iina/0.32.0/share/man --docdir=/usr/local/Cellar/mpv-iina/0.32.0/share/doc/mpv-iina
[...]

PR #2818, opened 2020-03-12
Rotate the thumbnails according to the display matrix

inflation:
- [x] It implements / fixes issue #2817.

---

**Description:**
Basically make thumbnails rotation aware.

### DIFF
--- a/iina/FFmpegController.m
+++ b/iina/FFmpegController.m
@@ -12,6 +12,7 @@
 #import <libavcodec/avcodec.h>
 #import <libavformat/avformat.h>
 #import <libswscale/swscale.h>
+#import <libavutil/display.h>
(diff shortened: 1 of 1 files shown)


lhc70000:
Can we just display the thumbnails with the current video rotation instead of writing them to the disk? This will also work if the user rotates the video manually.

```swift
let rotation = mpv.getInt(MPVProperty.videoParamsRotate) - mpv.getInt(MPVOption.Video.videoRotate)
```

inflation:
Hmm. That requires notifications when rotation changes. Let me see if I could figure it out.

inflation:
@lhc70000 Another attempt on branch [thumbnail_rotation2](https://github.com/iina/iina/tree/thumbnail_rotation2)

PR #2834, opened 2020-03-24
Custom Crop Aspect Ratio

andypotion:

---

**Description:**

Adds a text field for typing in a custom crop aspect ratio in the crop settings view.

Screen Shot 2020-03-24 at 6 50 13 PM

### DIFF
--- a/iina/Base.lproj/CropSettingsViewController.xib
+++ b/iina/Base.lproj/CropSettingsViewController.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="14460.31" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="15705" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
     <dependencies>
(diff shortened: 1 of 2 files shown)


andypotion:
Any chance that this gets looked at?

Lucas-Quinn-1163273:
Is there any progress for this? @lhc70000 @uiryuu @low-batt Thank you.

low-batt:
I am a junior newbie contributor, it is the experienced developers that make the call on what gets merged…

Although some features have sneaked in, the focus of the recent IINA releases have been on fixing defects. There are still open defects that need to be addressed. With the latest release there are a couple of regressions that warrant a quick release to fix them. After that IINA is planning a release that is focused on features.

I have added this PR to our internal list of potential work for the feature release. That release has not been planned out yet. Looks to me like some work will need to be kicked down the road to future releases as the list is huge compared to available developer resources.

I'm not a GUI designer, so I have to guess what the experienced developers might say. I would not make changes based on my feedback.

Such a change kind of makes the `Video Panel` 	inconsistent. The `Aspect ratio` has a text field for entering a custom value. If it was possible to enter a custom value for `Crop` you might expect there to be a similar field. But of course that disturbs the panel layout because where that would go is where the `Custom…` button is. I am not sure how to resolve that without ughlifying the panel. The other developers may think it is fine to just have the field for custom values on the crop view.

I don't see validation of the value entered? This also seems to be true today for `Aspect ratio`, so possibly this is considered acceptable.

The XIB has what look to be extraneous changes. Xcode does this. When I change a XIB I usually end up shutting down Xcode, running a `git diff` and editing XIB and restoring the values by hand before I commit. I'm not sure if the other developers are concerned about these Xcode tweaks.

With the build up of pull requests this commit now has merge conflicts. @andypotion Still around? Able to resolve merge conflicts?

Issue #316, opened 2017-02-03
[Feature Request] OSA script (AppleScript & JavaScript) support


tzhao11:
Both Quicktime and VLC support automation via AppleScript and JavaScript.

Henrietta1989:
+1 for AppleScript support. At least allow AppleScript to get current position of player. Returns "10" if video is playing or paused currently at 10 second.

lawder:
+1 for AppleScript support.

winray:
+1

theLMGN:
@lawder @winray 
![image](https://user-images.githubusercontent.com/12257112/59933181-d4c4dd00-9440-11e9-9ef0-eddef2b20816.png)

Click this instead of creating useless +1 comments

ideologysec:
Any updates on this? Would be great to replace VLC completely with IINA; as it is, I still keep VLC for the sole purpose of opening network streams via AppleScript from a browser.

saagarjha:
No progress, but we do have browser extensions which might do what you want.

theLMGN:
They can only play media. My usecase involves seeing what's playing

saagarjha:
Right, doing any actual AppleScript-y things is going to require OSA support.

theLMGN:
Also, @ideologysec, have you tried the following AppleScript?

`open location "iina://open?url=http://distribution.bbb3d.renderfarming.net/video/mp4/bbb_sunflower_1080p_30fps_normal.mp4"`

Wevah:
I've done some preliminary work on this at (https://github.com/Wevah/iina/tree/applescript).

lhc70000:
Hi @Wevah,

Thanks for your work! Would you mind opening a pull request? Although it's unfinished, I think we can see the diff more clearly and discuss any further problems there.

I'm working on the Javascript plugin system at https://github.com/iina/iina/tree/plugin-system. Perhaps some work in this branch can be reused.

Wevah:
Sure, I can do that. I’m not completely sure about how I separated some things out, but I guess that’s what review is for.

varenc:
@Wevah psyched to see your branch! I can't offer any serious help on the IINA implementation details, but there's anything simpler that'd still be useful for you let me know (like trying to bug test your new AppleScript interface?)

Wevah:
Ah! I put this off because I got knee deep in rewriting a Punycode/IDNA thing to be more accurate/for NNW.

shinepixel:
[...]

PR #2857, opened 2020-04-09
AppleScript support

Wevah:
WIP for issue #316.

I tried to make as few changes to the pre-existing files as possible; most stuff is in extensions in separate files. The only existing API changes were adding parent references to tracks and playlist items. Scripting needs those for returning object specifiers, and they're also used to make sure that actions like trying to set a current track to a track belonging to another player is a no-op or throws an error. (An alternative would be to iterate every player/track or player/playlist item combo, but that felt inefficient.)

`unowned` seemed appropriate since presumably if the player goes away, so do all the tracks/playlist items. Can be changed to `weak` if necessary.


Wevah:
One other thing I wanted to add: I tried to mimic the Music/iTunes sdef reasonably closely, so I used the same `FourCharCode`s for equivalent commands/properties, and followed the FCC naming prefix convention (`e` for enum, `p` for property, etc.).

Wevah:
Hit close by accident.

iina/AppDelegate+Scripting.swift
```diff
@@ -0,0 +1,34 @@
...
+//  Created by Nate Weaver on 2020-03-02.
+//  Copyright © 2020 lhc. All rights reserved.
+//
+
+import AppKit
+
+extension AppDelegate {
+
+  func application(_ sender: NSApplication, delegateHandlesKey key: String) -> Bool {
+    return Set(["orderedPlayerWindows", "orderedPlayers"]).contains(key)
```
saagarjha, line 14: For this few items, this could probably just be an explicit check–i.e. `key == "orderedPlayerWindows" || key == "orderedPlayerWindows"`. Unless you think there's a good reason to not do it this way?

iina/AppDelegate+Scripting.swift
```diff
@@ -0,0 +1,34 @@
...
+    let windows = NSApp.orderedWindows
+    var players = [PlayerCore]()
+
+    for window in windows {
+      if window.isVisible, let controller = window.delegate as? PlayerWindowController {
+        players.append(controller.player)
+      }
+    }
+
+    return players
```
saagarjha, line 27: This could all probably be done functionally, e.g.
```
Array(NSApp.orderedWindows.lazy.filter(\.isVisible).flatMap { $0.delegate as? PlayerWindowController}.map(\.player))
```

iina/FourCharCode+Extensions.swift
```diff
@@ -0,0 +1,25 @@
...
+
+import Foundation
+
+extension FourCharCode: ExpressibleByStringLiteral {
+
+  public typealias StringLiteralType = String
+
+  public init(stringLiteral: String) {
+    // Match NSHFSTypeCodeFromFileType() behavior: Return 0 for invalid codes.
+    guard stringLiteral.count == 4, let data = stringLiteral.data(using: .macOSRoman) else {
```
saagarjha, line 17: I think it might be a good idea to have initialization fail if these requirements aren't met. What do you think?

iina/MPVPlaylistItem+Scripting.swift
```diff
@@ -0,0 +1,48 @@
...
+    guard let index = scriptingIndex else { return nil }
+
+    return NSIndexSpecifier(containerClassDescription: containerClass!, containerSpecifier: containerSpecifier, key: "scriptingPlaylistItems", index: index)
+  }
+
+}
+
+extension MPVPlaylistItem {
+
+  @objc var scriptingURL: String { isNetworkResource ? filename : URL(fileURLWithPath: filename).absoluteString }
```
saagarjha, line 36: I think naming this "URL" might a bit confusing if it's not actually a URL.

saagarjha (commented):
This is great; I think this can really help us towards our goal of making IINA extensible. I'll need to do a bit of testing with the scripting stuff to see how this works, but for now here's a couple of general comments that I had while reading the code.

iina/FourCharCode+Extensions.swift
Wevah: I’m fine doing it that way; that was my initial thought, but then I decided to match the behavior of the notes function.

iina/MPVPlaylistItem+Scripting.swift
Wevah: Good call; I’ll change it to `scriptingURLString` or something.

iina/AppDelegate+Scripting.swift
Wevah: Probably just a holdover from code I had in another app with a *lot* more matches (to reduce noise).

iina/FourCharCode+Extensions.swift
Wevah: I'd have to make it throw, since a failable initializer can't satisfy the protocol.

iina/AppDelegate+Scripting.swift
Wevah: Actually, `orderedPlayerWindows` isn't even a thing currently, so I'll remove it.

Wevah:
More notes as I remember them:

- `AppDelegate` has a stub for `handlePlayCommand:`; it'll probably just forward to the frontmost player.
- `play` could take a file or URL, but using `open` might be better? Cocoa scripting conflates the direct parameter and the command handler without some AppleEvents property finagling.

Wevah:
Looks like the build failed due to using key paths as functions and Travis not having an Xcode 11.4 image yet.

Wevah:
Is there anything that looks missing from here that should be in for a first go?

hfpf:
Why leave the WIP label here?

sclsj:
@Wevah I forked your fork and merged the applescript branch. I can now open the sdef in `Script Editor`, but doing anything results in no actions (e.g. playpause, play, and pause don't work). Is this normal? My understanding is that your code works? Can you take a look if I'm doing anything wrong? https://github.com/sclsj/iina.

Wevah:
@sclsj Hm, the only things you should need are the .sdef, the files with the extensions on various classes and the minor tweaks to those classes (I think I added a referential property to at least one of them).

Wevah:
> Why leave the WIP label here?

Good call; removed.

Wevah:
Well this is annoying, the xcworkspace file opens empty in the latest Xcode I can run on this old machin.

Wevah:
Looks like I accidentally wipted out the `NSAppleScriptEnabled` Info.plist key when I rebased.

Wevah:
> @Wevah I forked your fork and merged the applescript branch. I can now open the sdef in `Script Editor`, but doing anything results in no actions (e.g. playpause, play, and pause don't work). Is this normal? My understanding is that your code works? Can you take a look if I'm doing anything wrong? https://github.com/sclsj/iina.

I think I see the issue; you need to tell an individual player to playpause, etc., since conceivably someone could have multiple players open:

```applescript
tell application "IINA"
	tell front player to playpause
end tell
```

Maybe I should add an app-wide version that automatically forwards to the first player?

Wevah:
Now I remember why it was marked as WIP still; I missed a few things.

sclsj:
@Wevah How would I pause the currently playing player and start the last playing player?

Wevah:
@sclsj You'd probably have to store a reference to the active player, or the player's id, somewhere if you've got multiple players open.

low-batt:
Some feedback from another IINA user...
Nice work!!

I successfully built the PR natively on an Apple Silicon Mac running Monterey 12.0.1 with mpv 0.34.0 and FFmpeg 4.4.1. I ran several AppleScript tests and they passed. Yeah!

I am not an AppleScript expert, so I could be wrong about some of the issues I'm about to bring up. See what others think.

This new warning is being emitted. Whatever is causing it needs to be fixed:
```text
2021-11-21 17:19:18.691940-0500 IINA[44532:1033619] .sdef warning for argument 'FileType' of command 'save' in suite 'Standard Suite': 'saveable file format' is not a valid type name.
```
IINA's support for AppleScript should follow Apple standards as much as possible. In particular we want to insure the AppleScript capabilities provided by the Cocoa Framework work as expected. This shows a problem:
```text
low-batt@gag ~$ osascript -e 'tell application "IINA"' -e 'activate' -e 'open "/Users/low-batt/Movies/issue-316.mp4"' -e 'end tell'
missing value
```
Running the same script against QuickTime shows what is wrong:
```text
low-batt@gag ~$ osascript -e 'tell application "QuickTime Player"' -e 'activate' -e 'open "/Users/low-batt/Movies/issue-316.mp4"' -e 'end tell'
document issue-316.mp4
```
The IINA.sdef contains:
```xml
<xi:include href="file:///System/Library/ScriptingDefinitions/CocoaStandard.sdef" xpointer="xpointer(/dictionary/suite)"/>
```
Which is fine if you fully follow the Cocoa conventions. These standard definitions expect that `NSDocument` is being used. QuickTime is using documents. It makes sense for QuickTime because it is not just a player. It has operations like `trim` that make modifications that require saving, whereas IINA is strictly a player. Need to consider now if IINA will in the future add features that involve modifying and saving files. If so, then some thought should be given as to whether IINA should be using `NSDocument`. It is not hard to add that support.

The Cocoa Standard definition in question is:
```xml
<command name="open" code="aevtodoc" description="Open a document.">
  <direct-parameter description="The file(s) to be opened.">
    <type type="file"/>
    <type type="file" list="yes"/>
  </direct-parameter>
  <result description="The opened document(s).">
    <type type="document"/>
    <type type="document" list="yes"/>
  </result>
</command>
```
The open command **must** return a document if that definition is used. IINA must either adopt the Cocoa QuickTime "document" pattern or address this issue in another manner.

Looking at the definitions using the Script Editor:

issue-316-quicktime

issue-316-iina

The QuickTime definitions contains an "Internet Suite" with an "Open URL" command., would be nice if IINA had this.

Would be nice for IINA to have commands for "Next Frame" and "Previous Frame".

QuickTime has "audio volume" whereas IINA has "sound volume", the term used by the Music app . Since IINA refers to this as "audio / volume" in the menus the AppleScript definitions should match up with the naming used in the menus. Would also be good to provide some sort of indication of what values are expected like the QuickTime definitions do.

Apple's AppleScript naming guidelines stress the importance of not picking a name for a property that makes it seem like a command. I expect that is why the QuickTime property is named "muted". The IINA property is named "mute", likely coming from the Music app which failed to follow the guidelines and has a property named "mute".

IINA has a "pause" command, would be nice to have a "resume" command like QuickTime.

Wevah:
> The IINA.sdef contains:
> 
> ```
> <xi:include href="file:///System/Library/ScriptingDefinitions/CocoaStandard.sdef" xpointer="xpointer(/dictionary/suite)"/>
> ```

Good call; I should probably get rid of that.

> The QuickTime definitions contains an "Internet Suite" with an "Open URL" command., would be nice if IINA had this.
> 
> Would be nice for IINA to have commands for "Next Frame" and "Previous Frame".

Reasonable!

> QuickTime has "audio volume" whereas IINA has "sound volume", the term used by the Music app . Since IINA refers to this as "audio / volume" in the menus the AppleScript definitions should match up with the naming used in the menus. Would also be good to provide some sort of indication of what values are expected like the QuickTime definitions do.
> 
> Apple's AppleScript naming guidelines stress the importance of not picking a name for a property that makes it seem like a command. I expect that is why the QuickTime property is named "muted". The IINA property is named "mute", likely coming from the Music app which failed to follow the guidelines and has a property named "mute".

Yes, I've been going off of the Music app's names and type codes. I suppose mimicking QuickTime would make more sense.

> IINA has a "pause" command, would be nice to have a "resume" command like QuickTime.

Does "resume" do something different in QuickTime than calling "play" on a paused player in IINA? I guess I should check.

Wevah:
In retrospect, I maybe shouldn't have rebased. Oh well.

Wevah:
If I'm reading QuickTime's dictionary right, `resume` is for resuming recordings, not playback.

Wevah:
I'm going to add a custom pruned copy of the core suite instead of importing it (like Xcode does).

CarterLi:
Hello! I'm interested in merging this PR into my [iina-plus](https://github.com/iina-plus/iina) repo, which tries to collect various awesome features into one build.

However I got some conflict errors when merging

```
$ git merge Wevah/applescript
Auto-merging iina.xcodeproj/project.pbxproj
CONFLICT (content): Merge conflict in iina.xcodeproj/project.pbxproj
Auto-merging iina/Base.lproj/PrefGeneralViewController.xib
CONFLICT (content): Merge conflict in iina/Base.lproj/PrefGeneralViewController.xib
Auto-merging iina/Info.plist
CONFLICT (content): Merge conflict in iina/Info.plist
Auto-merging iina/PlayerCore.swift
Auto-merging iina/PrefGeneralViewController.swift
Automatic merge failed; fix conflicts and then commit the result.
```

As it doesn't seems to be merged into the official repo soon, would you like to open a PR to my iina-plus repo?

ddddxxx:
Hi @Wevah, I want to communicate with IINA via AppleScript in my app. This pr seems very promising and I’d like to work on it right now. I’m wondering if there will be major changes in `sdef` file. Small changes like renaming a property is acceptable. Or I need to wait a bit longer for the script interface to stabilize?

Wevah:
@ddddxxx It looks like this got merged into `develop`, so there probably won't be many more changes to the functionality as it exists now unless something breaks,

low-batt:
Busy today, having to write this in a rush...

@ddddxxx Here is my own opinion on your question:

All bets are off until this feature is in an official released version of IINA. Program against the current proposed interface at your own risk.

The reason I say that is because of the issue you are raising. Unlike other IINA features where the user interface can be "tweaked" in future releases and humans can adjust to it, this feature requires that backward compatibility with programs and scripts be taken into account when making any future changes. For that reason this feature requires an intense detailed picky peer review as well as feedback from Beta test users. Since that has not been completed there is no guarantee that the interface will not change.

@Wevah has done some great work here. It is on my list to take a close look at this but I've been distracted, trying to fix other issues.

**NOTE** that this has **NOT** been merged into IINA yet. That is why this PR is still open. The merge notice seen above is because CarterLi merged it into his fork, "IINA-Plus".

Wevah:
Ahh, whoops! I must be blind.

CarterLi:
> ... this feature requires that backward compatibility with programs and scripts be taken into account when making any future changes. For that reason this feature requires an intense detailed picky peer review as well as feedback from Beta test users.

You are right. However I don't think there are many *Beta test users*. No one can tell if an API is stable without actually using it ( implementing something based on it ). It can be a good time to prove this API is stable enough to use or not.

PR #2973, opened 2020-06-10
Fix PlaylistLoop button state doesn't update with menu item

mathric:

---

**Description:**
Fixing the issue that clicking the PlaylistLoop menu item doesn't update the PlaylistLoop button state in PlaylistView when the PlaylistView already loaded.


lhc70000:
Thanks for your contribution and sorry for my late reply.

I just found that those two buttons are not synced with mpv status at all, which makes it a more complicated problem than I thought. We should also observe mpv property changes and post notifications to update the UI when needed.

If you don't mind, I'm going to continue your work to finish the implementation. I'll merge your commit into a temporary working branch and your contribution will be preserved.

lhc70000:
Pushed to 6b9bd35

mathric:
Thanks for your reply! It's fine if you're going to finish the implementation.

Issue #2930, opened 2020-05-11
Make voice icon change with volume


Petingo:
**What you want IINA to do:**
I suggest that the voice icon may be adaptive. For example,
- volume [67, 100] -> speaker + 3 lines
- volume [34, 66] -> speaker + 2 lines
- volume [1, 33] -> speaker + 1 line
- volume 0 -> same icon as mute / speaker with no line

or just make volume 0 has a different icon than others.

**What IINA does currently:**
Voice icon remains the same between [0, 100].

**Why you think this should be added:**
In music mode, there are no clues to trace the voice volume at a glance. Especially when the volume is 0, users may be struggled to look for the problem of no sound and not realize that it is because of the volume is set to 0.

I opened this issue because I couldn't find why IINA is not playing the sound and even reboot my laptop twice...

**Examples of other projects that have something similar:**
OSX

PR #2977, opened 2020-06-12
The voice icon will now change with the volume. #2930

Brothre23:

---

**Description:**
The voice icon will now change with the volume.


lhc70000:
Thanks a lot for your contribution!

It works well, but you should change all references to `volume` to `volume-2`, such as the OSC in the main window and the popup volume slider in music mode. These icons are missing now. Alternatively, you can just rename `volume-2` to `volume`.

I can also do the rename for you if you want.

Brothre23:
Thanks for the reviewing!
I'm glad if you can do the renaming stuff for me; sorry that I didn't notice the icon was being used somewhere else.

lhc70000:
Merged locally via 408a7bb.

PR #2988, opened 2020-06-15
Change font PingFang SC to PingFang TC

jrthsr700tmax:
Change font PingFang SC to PingFang TC

---

**Description:**


alejx:
Thanks!

Issue #2999, opened 2020-06-17
Scrolling to change volume does not give visual feedback in music mode


anohren:
**System and IINA version:**

- macOS 
- IINA  1.0.7

**Expected behavior:**
Scrolling to change volume shows a volume slider in a popup 

**Actual behavior:**
It gives zero visual feedback and is therefore of questionable use in music mode. See conversation in https://github.com/iina/iina/pull/2191 for already refined behavior of the volume popup.

<details>
<summary><b>Crash report:</b></summary>

```
```
</details>

<br>

<details>
<summary><b>mpv log:</b></summary>

```
```
</details>

<br>

**Steps to reproduce:**

- [ ] MPV does not have this problem.

**How often does this happen?**

PR #3002, opened 2020-06-17
Show volume popover when scrolling to adjust in miniplayer

alejx:
Ref: #2999, #2191

Co-authored-by: LEoREo2247

- [x] It implements / fixes issue #2999.

@anohren, please take a look at whether this is the correct behavior. Since I almost used the entire @LEoREo2247's code, I also gave them credits on this commit.

### DIFF
--- a/iina/MiniPlayerWindowController.swift
+++ b/iina/MiniPlayerWindowController.swift
@@ -59,6 +59,13 @@ class MiniPlayerWindowController: PlayerWindowController, NSPopoverDelegate {
 
   private var originalWindowFrame: NSRect!
 
+  lazy var hideVolumePopover: DispatchWorkItem = {
+    DispatchWorkItem {
+      self.volumePopover.animates = true
+      self.volumePopover.performClose(self)
+    }
+  }()
+
   override var mouseActionDisabledViews: [NSView?] {[backgroundView, playlistWrapperView] as [NSView?]}
 
   // MARK: - Initialization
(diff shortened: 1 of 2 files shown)


lhc70000:
Is it still needed after #2977?

Personally I think it's a little bit awkward to trigger the popover when scrolling. Although the current icon after #2977 and 0ba3798 doesn't show the precise volume, it gives sufficient visual feedback IMHO.

anohren:
I haven't tried this PR or 2977, but spontaneously: yes, this is still needed.

If you're ok with your trackpad arbitrarily alternating between pixel-perfect scrolling and behaving like a discrete step mouse wheel it might be enough with 3 steps of feedback, but intuitively I don't think anyone wants that. Especially since there are 100 steps of volume.

If people are used to looking at a slider when adjusting volume then they will be looking for a slider here too. We can't ignore that the video window conditions users to do that, and that deviating from it is most likely an unnecessary inconvenience.

The really awkward part is having volume control in a popover, making scrolling a much more inviting method for adjusting it. Not that it isn't more convenient even if the slider were visible, but I wouldn't be surprised if people acutally *discover* the slider by using scrolling.

anohren:
Ok, I've tried this PR now and it seems to work beautifully. 

Except the first time I scrolled the popover spawned outside the window, but that might just be a Cocoa quirk.

lhc70000:
Merged via 17fe851.

lhc70000:
BTW, It looks way better than I thought!

PR #3018, opened 2020-06-28
Fix pointer related warnings

alejx:


saagarjha:
I’ll see if I can take a look later today; I’m not familiar with what the lifetimes of those parameters are supposed to be so I’ll have to check. But I’m sure you’ve figured out that you can’t pin an object to an address in Swift unless you create it using on of the `Unsafe*` APIs and manage the memory yourself :)

iina/MPVController.swift
```diff
@@ -345,15 +345,17 @@ class MPVController: NSObject {
...
-      mpv_render_param(type: MPV_RENDER_PARAM_API_TYPE, data: apiType),
-      mpv_render_param(type: MPV_RENDER_PARAM_OPENGL_INIT_PARAMS, data: &openGLInitParams),
-      // mpv_render_param(type: MPV_RENDER_PARAM_ADVANCED_CONTROL, data: &advanced),
-      mpv_render_param()
-    ]
-    mpv_render_context_create(&mpvRenderContext, mpv, &params)
-    mpv_render_context_set_update_callback(mpvRenderContext!, mpvUpdateCallback, mutableRawPointerOf(obj: player.mainWindow.videoView.videoLayer))
+    withUnsafeMutablePointer(to: &openGLInitParams) { params in
+      // var advanced: CInt = 1
+      var params = [
```
saagarjha, line 350: Shadowing the parameter is a bit confusing, especially here where the lifetime of the two matters. Would you mind disambiguating them?

iina/VideoView.swift
```diff
@@ -228,25 +228,25 @@ class VideoView: NSView {
...
-
-        if current == 1, deviceID == uuid {
-          let profileURL = info["DeviceProfileURL"] as! URL
-          ptr.pointee.profileUrl = profileURL
-          return false
+    withUnsafeMutablePointer(to: &argResult) { data in
+      ColorSyncIterateDeviceProfiles({ (dict: CFDictionary?, ptr: UnsafeMutableRawPointer?) -> Bool in
+        if let info = dict as? [String: Any], let current = info["DeviceProfileIsCurrent"] as? Int {
+          let deviceID = info["DeviceID"] as! CFUUID
+          let ptr = ptr!.bindMemory(to: ProfileData.self, capacity: 1)
```
saagarjha, line 238: I think you want [`assumingMemoryBound(to:)`](https://developer.apple.com/documentation/swift/unsaferawpointer/2431721-assumingmemorybound)

alejx: I didn't change here, so I'd rather not include this change in this pr. Could you please send another pr to resolve this later?

PR #3020, opened 2020-06-28
Fix Video QuickSettings view

unique-username313:

---

**Description:**

Constraints in the Video Quicksettings view were misconfigured. The "Aspect Ratio" section (and everything below it) had their vertical positioning constrained to the "Table View" instead of the table's respective "Scroll View". This means that if a video has more than 3 video tracks or so, Aspect ratio and everything below it are pushed down to be aligned with the bottom of the video track list, no matter how long that list is and even if everything after the third track is hidden by the limit of the scroll view.

This issue is understandably hard to come by since most media files only have 1 video track (if any, in case of audio only tracks). I happen to configure youtube-dl to show all tracks available giving me the ability to choose between them at runtime, switching resolutions, codecs, and bitrates around without having to restart the video with different ytdl format settings. This is how I encountered this issue. This commit fixes it. See screenshot below.

Here is two screenshots combined in one picture. Taken while playing a video that has 20 video tracks. Left is current behavior, right is with the proposed fix.

quicksettings-fix

Notice on the left sided screen how the Y position of the Aspect Ration section is positioned right where the entire list of video tracks is supposed to end (i.e after 20 entries). Although the latter are hidden by the limits of the scroll view. One the right however, everything is lined up how it's supposed to be.


alejx:
Thank you very much!

PR #3048, opened 2020-07-21
Removed 'IINA v1.0 is coming!' from Beta Popup window

### LINKED ISSUE #3047: User prompted to participate in 1.0 beta on initial launch
On first launch of IINA by a user, they are asked to help test beta versions of IINA 1.0 (which has already been released).

**System and IINA version:**

- macOS 10.15.6 Catalina
- IINA 1.0.7

**Expected behavior:**
User is _not_ prompted to participate in beta program _or_ user is offered the option to participate in beta testing for a future version.

**Actual behavior:**
User is presented with the following dialog box:
IINA Beta Screenshot

**Steps to reproduce:**
1. Launch IINA for the first time in a user account. 

Reliably on initial launch in each user account.

saloomi2012:
- [x] It implements / fixes issue #3047 

---

**Description:**
Fixed #3047 by removing entirely the 'IINA v1.0 is coming!' part. Now it is a simple question for the user if they want to test beta versions or not and removing any reference to the upcoming versions to avoid this issue in the future. I don't have much experience with localization so I just updated the Localizable.strings (Base). I hope that is the correct way of handling such issue.

### DIFF
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -123,7 +123,7 @@
 
 // Alerts
 "alert.beta_channel.title" = "Receive Updates for Beta Versions";
-"alert.beta_channel.message" = "IINA v1.0 is coming! Would you like to help us to test out beta versions? \n\nYou can change this behavior at anytime in Preferences - General.";
+"alert.beta_channel.message" = "Would you like to help us to test out beta versions? \n\nYou can change this behavior at anytime in Preferences - General.";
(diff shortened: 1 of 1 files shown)


lhc70000:
I thought it had been removed before……

Thanks for your effort, but we need to do a little bit more work to resolve this completely, including removing this sentence from all localizations on crowdin. Therefore I'm going to close this PR and let @alejx handle it. Sorry!

PR #3049, opened 2020-07-23
Escape whitespace in Check Localizable phase

anohren:
Can't build without it


lhc70000:
Well, I think now with crowdin we can remove this phase directly.

anohren:
That's a relief, since that's what I ended up doing anyway

alejx:
The phase is deleted. Closing.

Issue #3013, opened 2020-06-25
Fatal error: video info not available


lhc70000:
`PlayerWindowController L427`

A user confirmed that this could happen for some live streams.

We should consider continuing the playback even if `videoDuration` and/or `videoPosition` cannot be obtained.

matellis:
I am getting this all the time when using a NAS. Why does it have to be a fatal error, couldn't it just refuse to open the file?

matellis:
Submitted pull request with fix

low-batt:
This issue has been reported again in #3266.

I have hit this failure multiple times. I am able to reproduce it on my 2019 MacBook Air running macOS 10.15.7. I use a bash script to start 20 separate instances of IINA all at once. Usually somewhere between 1-3 fail. Rarely do all 20 work.

The failure is a result of a race condition between threads when mpv is opening the file. Adding some additional logging before the call to `fatal` in `updatePlayTime` shows mpv has not finished loading the file and position and duration are unknown, tripping the fatal internal error check:

>18:10:20.721 [iina][d] <NSThread: 0x7fc922605d50>{number = 1, name = main} PlayerWindowController.updatePlayTime(withDuration:andProgressBar:):438 14.mp4 player.info.justStartedFile false  **player.info.fileLoading true** player.info.videoPosition:  nil player.info.videoDuration:  nil 

While mpv is in the process of opening the file it is emitting events for properties that have been updated. The function `MPVController.readEvents` is being called to process the events. That function dispatches a call to `MPVController.handleEvent` in a background thread. The `handleEvent` function calls `handlePropertyChange`. One of the properties mpv updates is `MPVProperty.chapter`. When that property is updated `handlePropertyChange` calls `player.syncUI`. Still running in that background thread the function `syncUI` stores the current values of the mpv properties `MPVProperty.timePos` and `MPVProperty.duration` in `player.info` and then queues an asynchronous call to `PlayerWindowController.updatePlayTime`. As mpv is still loading the file while the event is being processed by the background thread the position and duration properties may or may not have been set when `syncUI` captures the values of these properties. If either of the values stored in `player.info` are still `nil` when `updatePlayTime` is run then the fatal error will be triggered.

The change in the pull request updates the first `guard` statement in `updatePlayTime` to not process the update if either position or duration are unknown and removes the second `guard` statement containing the fatal error:
```swift
// IINA listens for changes to mpv properties such as chapter that can occur during file loading
// resulting in this function being called before mpv has set its position and duration
[...]

PR #3095, opened 2020-09-06
Resolve race issue with info not loaded errors

matellis:
Seems to fix issue #3013 but not sure why.
---
Original issue for me was when I moved a bunch of files I was working on from direct attached storage to a NAS, I began to get random fatal errors upon loading video files, usually about 1 in 6 to 1 in 8 times. This would be very irritating as I'd have to reload all the files already open, and re-opening would often cause the error to happen again.

After a bit of experimentation I moved the 'loaded = true' statement from the start to the end of windowDidLoad override function. This seems to have eliminated the problem with no discernible side effects. I'd like someone to check the change in case of unknown side effects, as frankly I'm not certain why this works other than some non-specific race condition. I have tested it plenty and cannot reproduce either the error nor any side effects though.

### DIFF
--- a/iina/PlayerWindowController.swift
+++ b/iina/PlayerWindowController.swift
@@ -153,7 +153,6 @@ class PlayerWindowController: NSWindowController, NSWindowDelegate {
 
   override func windowDidLoad() {
     super.windowDidLoad()
-    loaded = true
     
     guard let window = window else { return }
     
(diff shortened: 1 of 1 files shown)


matellis:
Still experiencing issues, don't know why pull failed. Closing PR.

Issue #3121, opened 2020-09-23
Rename menu item "Float On Top" to "Stay On Top" to align with System apps


albinekb:
**System and IINA version:**

- macOS 10.15.6 (19G2021)
- IINA 1.0.4 Build 107

**Expected behavior:**
In Xcode 12 and macOS 11 apple has renamed "Float On Top" to "Stay On Top", to align with system apps, I think IINA should adopt this name change.

**Actual behavior:**

In IINA the menu item is named "Float On Top"
<br>

**Steps to reproduce:**

Install macOS 11 or Xcode Version 12.0 (12A7209) and check system apps, the menu item is now named "Stay On Top" (In the places i found it)

hadi-norouzi:
I can do this.

anohren:
> In Xcode 12 and macOS 11 apple has renamed "Float On Top" to "Stay On Top", to align with system apps

Is that the case for QuickTime player as well? (If it still exists)

uiryuu:
> 

No, it's still "Float on Top" in QuickTime.

PR #3132, opened 2020-10-01
fix issue 3121

hadi-norouzi:
- [x] It implements / fixes issue #3121 .

---

**Description:**


anohren:
If this is done in the name of consistency, then won't it result in inconsistency on all earlier versions of the OS?

hadi-norouzi:
You mean it only changed on macos 11?

anohren:
That's my understanding from reading the issue description.

hadi-norouzi:
i'm in macos catalina and it's changed and work well.
Do you want it to be the same as before in above macos 11?

Issue #3219, opened 2020-11-21
Build Error: The linked library 'lib***.dylib' is missing one or more architectures required by this target: arm64.


tinkernels:
**System and IINA version:**
- macOS 
  - 10.15.7
- IINA 
  - 1.1.2
- Xcode version
  - 12.2

**Expected behavior:**
Build success.

**Actual behavior:**
Build failed with Error:
`The linked library 'lib***.dylib' is missing one or more architectures required by this target: arm64.`

<details>
<summary><b>Crash report:</b></summary>

```
```
</details>

<br>

<details>
<summary><b>mpv log:</b></summary>

```
```
</details>

<br>

**Steps to reproduce:**

- [ ] MPV does not have this problem.

**How often does this happen?**
Every build action.

PR #3220, opened 2020-11-21
Modify build settings of EXCLUED_ARCHS.

tinkernels:
fixes issue #3219 .

**Description:**
As Xcode 12.2 default build settings included arm64 architecture, iina build fails. 
A workaround is to exclude build archs with this PR.


MaddTheSane:
An ideal fix would be to build with native arm64 support.

tinkernels:
> An ideal fix would be to build with native arm64 support.

It fails because dependencies are lack of arm64 target. Let iina exclude arm64 arch before dependency problems solved is reseaonale, I think.

MaddTheSane:
Agreed. Excluding arm64 is a great stop-gap for now.

PR #3434, opened 2021-06-23
Fix fatal error, video info not available, #3013

low-batt:
Change PlayerWindowController.updatePlayTime to not report a fatal error
if the position or duration is not known as can occur when the function
is called while mpv is still loading the file.

- [ x] It implements / fixes issue #3013.

---

**Description:**

### DIFF
--- a/iina/PlayerWindowController.swift
+++ b/iina/PlayerWindowController.swift
@@ -433,10 +433,10 @@ class PlayerWindowController: NSWindowController, NSWindowDelegate {
   }
   
   func updatePlayTime(withDuration: Bool, andProgressBar: Bool) {
-    guard loaded else { return }
-    guard let duration = player.info.videoDuration, let pos = player.info.videoPosition else {
-      Logger.fatal("video info not available")
-    }
+    // IINA listens for changes to mpv properties such as chapter that can occur during file loading
+    // resulting in this function being called before mpv has set its poistion and duration
+    // properties. If position or duration are unknown assume the file is still being loaded.
+    guard loaded, let duration = player.info.videoDuration, let pos = player.info.videoPosition else { return }
     [leftLabel, rightLabel].forEach { $0.updateText(with: duration, given: pos) }
     if #available(macOS 10.12.2, *) {
       player.touchBarSupport.touchBarPosLabels.forEach { $0.updateText(with: duration, given: pos) }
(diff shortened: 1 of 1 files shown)


iina/PlayerWindowController.swift
```diff
@@ -433,10 +433,10 @@ class PlayerWindowController: NSWindowController, NSWindowDelegate {
   }
   
   func updatePlayTime(withDuration: Bool, andProgressBar: Bool) {
-    guard loaded else { return }
-    guard let duration = player.info.videoDuration, let pos = player.info.videoPosition else {
-      Logger.fatal("video info not available")
-    }
+    // IINA listens for changes to mpv properties such as chapter that can occur during file loading
+    // resulting in this function being called before mpv has set its poistion and duration
```
saagarjha, line 437:
```suggestion
    // resulting in this function being called before mpv has set its position and duration
```

iina/PlayerWindowController.swift
```diff
@@ -433,10 +433,10 @@ class PlayerWindowController: NSWindowController, NSWindowDelegate {
...
+    // properties. If position or duration are unknown assume the file is still being loaded.
+    guard loaded, let duration = player.info.videoDuration, let pos = player.info.videoPosition else { return }
```
saagarjha, line 439: Perhaps we could just downgrade the logging to `debug`?

iina/PlayerWindowController.swift
low-batt: Fixed

iina/PlayerWindowController.swift
low-batt: Good point. See what you think of the new commit I just pushed.

low-batt: According to the mpv manual, mpv does not guarantee that mpv is not always able to determine the duration. This is mentioned in several places in the manual. This is from the property definition:
> Duration of the current file in seconds. If the duration is unknown, the property is unavailable. Note that the file duration is not always exactly known, so this is an estimate.

I have not encountered a case where mpv did not provide a value for duration, but I don't do that much streaming. Probably happens with live streams? I figured it was best to adhere to the documented behavior and not assume the duration is known.

For that reason I thought it was important in the log to distinguish between which piece of video information is missing. Of course there is more than one way to do that.

PR #3441, opened 2021-06-30
Fix crash on files within specific folders, #3413

low-batt:
The code in Extensions.read was triggering a crash when readData failed
to return all the bytes requested due to trying to read past EOF. This
happened when IINA tried to read a thumbnail cache entry that was
truncated due to there being no space left on the storage device when
the cache entry was being written.

If applied this commit will:

- Add code to read to check that readData returned the bytes requested

- Change the return type of read to be optional

- Update some callers in ThumbnailCache to handle read failures

- [ x] It implements / fixes issue #3413.

---

**Description:**

### DIFF
--- a/iina/ThumbnailCache.swift
+++ b/iina/ThumbnailCache.swift
@@ -158,20 +158,25 @@ class ThumbnailCache {
     // data blocks
     while file.offsetInFile != eof {
       // length
-      let blockLength: Int64 = file.read(type: Int64.self)
+      guard let blockLength: Int64 = file.read(type: Int64.self) else {
+        Logger.log("Cannot read image length. Cache file will be deleted.", level: .warning, subsystem: subsystem)
+        file.closeFile()
+        delete(pathURL)
+        return nil
+      }
       // timestamp
-      let timestamp: Double = file.read(type: Double.self)
+      guard let timestamp: Double = file.read(type: Double.self) else {
+        Logger.log("Cannot read image timestamp. Cache file will be deleted.", level: .warning, subsystem: subsystem)
+        file.closeFile()
+        delete(pathURL)
+        return nil
+      }
       // jpeg
       let jpegData = file.readData(ofLength: Int(blockLength) - MemoryLayout.size(ofValue: timestamp))
       guard let image = NSImage(data: jpegData) else {
         Logger.log("Cannot read image. Cache file will be deleted.", level: .warning, subsystem: subsystem)
         file.closeFile()
-        // try deleting corrupted cache
-        do {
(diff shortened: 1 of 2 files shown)


iina/ThumbnailCache.swift
```diff
@@ -158,20 +158,25 @@ class ThumbnailCache {
     // data blocks
     while file.offsetInFile != eof {
       // length
-      let blockLength: Int64 = file.read(type: Int64.self)
+      guard let blockLength: Int64 = file.read(type: Int64.self) else {
+        Logger.log("Cannot read image length. Cache file will be deleted.", level: .warning, subsystem: subsystem)
+        file.closeFile()
+        delete(pathURL)
+        return nil
+      }
```
saagarjha, line 166: I feel like this doesn't need two specific checks. How about

```suggestion
      guard let blockLength = file.read(type: Int64.self),
          let timestamp = file.read(type: Double.self) else {
        Logger.log("Cannot read image header. Cache file will be deleted.", level: .warning, subsystem: subsystem)
        file.closeFile()
        delete(pathURL)
        return nil
      }
```

iina/ThumbnailCache.swift
```diff
@@ -186,6 +191,15 @@ class ThumbnailCache {
     return result
   }
 
+  static private func delete(_ pathURL: URL) {
```
saagarjha, line 194:
```suggestion
  private static func deleteCacheFile(at pathURL: URL) {
```

iina/ThumbnailCache.swift
low-batt: Fixed.

iina/ThumbnailCache.swift
low-batt: Fixed.
I changed the ordering of keywords throughout the class to be "private static", which I prefer as well. I had reversed it to be consistent with the existing code.

I actually searched around for official guidelines on preferred ordering of keywords and didn't find any for Swift. Java allows you to use the order you want, but does have an official preferred order. The Eclipse IDE source formatter could be configured to automatically put your keywords in the official order. I miss that.

saagarjha: Yeah, I don't blame you, I just think the other ordering was worse I figured we might as well switch it around for new code :)

saagarjha (approved):
lgtm

iina/ThumbnailCache.swift
low-batt: Always feel free to ask me to cleanup style issues like this. I'm happy to do it.

Issue #3445, opened 2021-07-10
Many FileInfo objects leaked


low-batt:
I noticed IINA's memory use increasing over time. Running the `leaks` command line tool against IINA reported many leaked buffers. This issue is focused on a large leak of a collection of `FileInfo` objects. The `leaks` output shows these `FileInfo` objects are related to IINA's scanning of the files in the same directory as the video being played. If there is a large number of files in that directory the amount of memory leaked can be large.

**System and IINA version:**

- macOS 11.4 
- IINA 1.2.0 

**Expected behavior:**

IINA does not leak memory.

**Actual behavior:**

IINA leaks a lot of `FileInfo` objects.

<br>

**Steps to reproduce:**

I reproduced this issue using a development build under macOS 11.4 and Xcode 12.5. I added a log message to the `init` method of the `FileInfo` class in order to be able to see when a `FileInfo` object is allocated. I also added a `deinit` method to the class with a log message to see when a `FileInfo` object is being deallocated:
```swift
deinit {
  Logger.log("deinit FileInfo \(url)")
}
```

I started up IINA. I opened a MP4 file. I then closed the window. IINA remained running. I clicked on the IINA icon in the dock and opened the same MP4 file again. I then ran `leaks` against the IINA process. The first leak reported is a collection of `FileInfo` objects representing the files in the `Movies` directory:
<details>
<summary>Output from leaks:</summary>

```text
Process:         IINA [14284]
Path:            /Users/USER/Library/Developer/Xcode/DerivedData/iina-echacgeisxgbwucwumppkhklmnkd/Build/Products/Debug/IINA.app/Contents/MacOS/IINA
Load Address:    0x109850000
Identifier:      com.colliderli.iina
Version:         1.2.0 (129)
Code Type:       X86-64
Platform:        macOS
Parent Process:  debugserver [14286]

Date/Time:       2021-07-09 16:52:00.570 -0400
Launch Time:     2021-07-09 16:49:37.756 -0400
OS Version:      macOS 11.4 (20F71)
Report Version:  7
Analysis Tool:   /usr/bin/leaks

Physical footprint:         78.1M
Physical footprint (peak):  86.5M
----

leaks Report Version: 4.0
Process 14284: 103020 nodes malloced for 59738 KB
Process 14284: 117 leaks for 10288 total leaked bytes.

    117 (10.0K) << TOTAL >>

      52 (4.95K) ROOT CYCLE: <Swift closure context 0x6000038a0540> [32]
[...]

PR #3446, opened 2021-07-10
Fix many FileInfo objects leaked, #3445

low-batt:
Fixes a retain cycle in the FileGroup.flatten method by replacing a recursive
closure with a nested function. This eliminates a memory leak consisting of a
collection of FileInfo objects representing the files contained in the directory
of the video being played.

- [x ] It implements / fixes issue #3445.

---

**Description:**


saagarjha:
Indeed, and the strange construction just hides the error that Swift provides :P

PR #3474, opened 2021-08-05
Fix toggle menus to adhere to HIG, #3116

### LINKED ISSUE #3116: Menu items buttons don't reflect the current state
There are many menu items and buttons in IINA that fall in the ["toggled menu items" category](https://developer.apple.com/design/human-interface-guidelines/macos/menus/menu-anatomy/) (and the undocumented equivalent for buttons). However, they are not implemented as such. For example: 

Screen Shot 2020-09-20 at 14 13 45 Redacted

"Show Quick Settings Panel" should be "Hide Quick Settings Panel" when the panel is already shown.

Screen Shot 2020-09-20 at 14 19 14

"Toggle Picture-in-Picture" button stay as SF Symbol `pip.enter`, when it should be `pip.exit`.

**System and IINA version:**

- macOS 10.15.6 (19G2021)
- IINA 1.1.0-beta1 Build 121

**Expected behavior:**

Menu item titles and button images should adapt to the application/window/player's current state.

**Actual behavior:**

Menu item titles and button images do not adapt to the application/window/player's current state.

---

I think I'm able to fix it, if the maintainers think this should be fixed.

low-batt:
If applied this commit will change these toggle menus to use "Hide" in their
title instead of "Show" when selecting the menu item hides the panel:
- Show Chapters Panel
- Show Playlist Panel
- Show Quick Settings Panel

This commit adds text that will require localization.

This fix only corrects one part of the reported issue. The behavior of buttons
still needs to be updated to adhere to the HIG.

There are some issues with the localization. See the discussions in the issue.

It is part of the changes required to fix #3116

---

**Description:**

There are issues with localization, such as files missing from the Xcode project.
See the discussion in issue #3116.

### DIFF
--- a/iina/AppData.swift
+++ b/iina/AppData.swift
@@ -81,6 +81,13 @@ struct Constants {
     static let subDelay = "Subtitle Delay"
     static let pip = NSLocalizedString("menu.pip", comment: "Enter Picture-in-Picture")
     static let exitPIP = NSLocalizedString("menu.exit_pip", comment: "Exit Picture-in-Picture")
+    static let chaptersPanel = NSLocalizedString("menu.chapters", comment: "Show Chapters Panel")
+    static let hideChaptersPanel = NSLocalizedString("menu.hide_chapters", comment: "Hide Chapters Panel")
+    static let playlistPanel = NSLocalizedString("menu.playlist", comment: "Show Playlist Panel")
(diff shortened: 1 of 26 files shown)


low-batt:
Absolutely fine to postpone, definitely not critical. I did include the translations since it was pretty simple. But I did find some issues when doing so.

Also this PR does not fully address the issue. The HIG "toggle" rules apply to buttons as well. Best to postpone and completely address the issue.

PR #3477, opened 2021-08-13
Fix constant CPU usage by Playlist Panel, #3162

low-batt:
This commit will change the method PlaylistViewController.tableView to
check if PlayerCore.refreshCachedVideoInfo was able to populate
PlaybackInfo.cachedVideoDurationAndProgress and not submit a reload of
the view if the cache is missing data.

This commit also updates FFmpegController.probeVideoInfoForFile to log
errors when libavformat methods return an error code.

This fixes an infinite table view reload loop that currently occurs when
refreshCachedVideoInfo is unable to populate the cache.

The infinite reload loop may be the cause of this issue:
Excessive CPU consumption when opening the Playlist Panel, #3162

- [x ] It implements / fixes issue #3162.

---

**Description:**

This commit fixes a serious Playlist Panel issue.
Not certain it is the same problem reported in #3162.
See the replies to the issue for details.

### DIFF
--- a/iina/PlaylistViewController.swift
+++ b/iina/PlaylistViewController.swift
@@ -532,9 +532,16 @@ class PlaylistViewController: NSViewController, NSTableViewDataSource, NSTableVi
             // get related data and schedule a reload
             if Preference.bool(for: .prefetchPlaylistVideoDuration) {
               self.player.refreshCachedVideoInfo(forVideoPath: item.filename)
-              self.refreshTotalLength()
-              DispatchQueue.main.async {
-                self.playlistTableView.reloadData(forRowIndexes: IndexSet(integer: row), columnIndexes: IndexSet(integersIn: 0...1))
+              // Only schedule a reload if data was obtained and cached to avoid looping
+              if let cached = self.player.info.cachedVideoDurationAndProgress[item.filename],
+                 let duration = cached.duration {
+                if duration > 0 {
+                  // if FFmpeg got the duration succcessfully
+                  self.refreshTotalLength()
+                  DispatchQueue.main.async {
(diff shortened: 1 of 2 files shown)


alejx (approved):
LGTM

iina/PlaylistViewController.swift
```diff
@@ -532,9 +532,16 @@ class PlaylistViewController: NSViewController, NSTableViewDataSource, NSTableVi
             // get related data and schedule a reload
             if Preference.bool(for: .prefetchPlaylistVideoDuration) {
               self.player.refreshCachedVideoInfo(forVideoPath: item.filename)
-              self.refreshTotalLength()
-              DispatchQueue.main.async {
-                self.playlistTableView.reloadData(forRowIndexes: IndexSet(integer: row), columnIndexes: IndexSet(integersIn: 0...1))
+              // Only schedule a reload if data was obtained and cached to avoid looping
+              if let cached = self.player.info.cachedVideoDurationAndProgress[item.filename],
+                 let duration = cached.duration {
+                if duration > 0 {
```
saagarjha, line 538: This code is already heavily nested; we might as well get rid of one level.

```suggestion
              if let cached = self.player.info.cachedVideoDurationAndProgress[item.filename],
                 let duration = cached.duration,
                 duration > 0 {
```

saagarjha (commented):
lgtm to me as well, aside from one stylistic comment.

iina/PlaylistViewController.swift
low-batt: Fixed. Wish we had some sort of automated source code formatter. With other languages I got used to letting the IDE format and cleanup my messy code.

saagarjha: This is something I would love as well, but we probably need to plan how we want to do this first :)

PR #3480, opened 2021-08-16
Fix IINA 1.2.0 crashes instantly (Big Sur), #3478

low-batt:
This commit will change the method `SleepPreventer.preventSleep`
to submit a task to the main DispatchQueue to display an alert if macOS
returns an error when IINA asks it to not put the display to sleep.

This fixes a crash due to trying to display an alert using a thread other than
the main thread.

Logging has been added to log the failure macOS is returning.

- [x ] It implements / fixes issue #3478.

---

**Description:**

This also fixes issue #3361.

### DIFF
--- a/iina/SleepPreventer.swift
+++ b/iina/SleepPreventer.swift
@@ -17,6 +17,14 @@ class SleepPreventer: NSObject {
 
   static private var preventedSleep = false
 
+  private static func errorToString(_ code: IOReturn) -> String {
+    if let error = mach_error_string(code) {
+      return String(cString: error)
+    } else {
+      return ""
+    }
+  }
+
   static func preventSleep() {
     if preventedSleep {
       return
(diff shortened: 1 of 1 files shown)


iina/SleepPreventer.swift
```diff
@@ -29,7 +37,10 @@ class SleepPreventer: NSObject {
     if success == kIOReturnSuccess {
       preventedSleep = true
     } else {
-      Utility.showAlert("sleep")
+      Logger.log("Cannot prevent display sleep: \(errorToString(success)) (\(success))", level: .error)
```
saagarjha, line 40: Honestly we should just force unwrap this. If it fails, then something has gone really wrong and we should know about it.

```suggestion
      Logger.log("Cannot prevent display sleep: \(String(cString: mach_error_string(success))) (\(success))", level: .error)
```

low-batt: Fixed.

saagarjha (approved):

Issue #3494, opened 2021-08-30
Script parse_doc.rb, MPVProperty fails to compile


low-batt:
Currently, running `other/parse_doc.rb` generates a `MPVProperty.swift` file containing this invalid code:
```swift
  /** current-tracks/... */
  static let currentTracks... = "current-tracks/..."
```

**System and IINA version:**

- macOS 11.4
- IINA 1.2.0
**Expected behavior:**

The script `parse_doc.rb` generates Swift code that successfully compiles.

**Actual behavior:**

IINA fails to build reporting:
```text
/Users/low-batt/Documents/builds/iina/iina/iina/MPVProperty.swift:453:27: error: consecutive declarations on a line must be separated by ';'
  static let currentTracks... = "current-tracks/..."
                          ^
                          ;
/Users/low-batt/Documents/builds/iina/iina/iina/MPVProperty.swift:453:27: error: expected declaration
  static let currentTracks... = "current-tracks/..."
                          ^
/Users/low-batt/Documents/builds/iina/iina/iina/MPVProperty.swift:3:8: note: in declaration of 'MPVProperty'
struct MPVProperty {
       ^
/Users/low-batt/Documents/builds/iina/iina/iina/MPVProperty.swift:453:14: error: type annotation missing in pattern
  static let currentTracks... = "current-tracks/..."
```

**Steps to reproduce:**

Just run `parse_doc.rb` and try and compile the generated sources.

- [ x] MPV does not have this problem.

**How often does this happen?**

Everytime.

I will submit a pull request with a fix to the script shortly.

low-batt:
The failure is due to mpv adding this option:
```text
current-tracks/...

    This gives access to currently selected tracks. It redirects to the correct entry in track-list.

    The following sub-entries are defined: video, audio, sub, sub2

    For example, current-tracks/audio/lang returns the current audio track's language field (the same value as track-list/N/lang).

    A sub-entry is accessible only if a track of that type is actually selected. Tracks selected via --lavfi-complex never appear under this property. current-tracks and current-tracks/ are currently not accessible, and will not return anything.

    Scripts etc. should not use this. They should use track-list, loop over all tracks, and inspect the selected field to test whether a track is selected (or compare the id field to the video / audio etc. options).
```

The documentation for `current-tracks` deviates from how other mpv path style properties are documented, confusing the script and causing it to generate invalid code.

[...]

PR #3495, opened 2021-08-30
Fix script parse_doc.rb, MPVProperty fails to compile, #3494

low-batt:
This commit will change the script parse_doc.rb to treat the property
"current-tracks" as a special case and only add a comment to
MPVProperty that indicates why a constant wasn't generated for it.

- [ x] It implements / fixes issue #3494.

---

**Description:**

### DIFF
--- a/other/parse_doc.rb
+++ b/other/parse_doc.rb
@@ -26,7 +26,12 @@ def write_prop(file, node)
   end
   $prop_set[camel_name] += 1
 
-  if name.include? '/'
+  if name.include? 'current-tracks'
+    # According to the mpv documentation this property is not supposed to be
+    # used programmatically. We still output a comment for the property so
+    # people can figure out why the Swift constant is missing.
+    file.write "  /** #{name}  As per mpv doc, scripts etc. should not use this. */\n"
+  elsif name.include? '/'
     if name.include? 'N'
       func_name = name.to_camel
       return_str = name.gsub('N', '\(n)')
--- a/other/parse_doc.rb
+++ b/other/parse_doc.rb
@@ -30,7 +30,7 @@ def write_prop(file, node)
     # According to the mpv documentation this property is not supposed to be
     # used programmatically. We still output a comment for the property so
     # people can figure out why the Swift constant is missing.
-    file.write "  /** #{name}  As per mpv doc, scripts etc. should not use this. */\n"
+    file.write "  /** #{name}  As per mpv docs, scripts etc. should not use this. */\n"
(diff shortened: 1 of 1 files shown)


other/parse_doc.rb
```diff
@@ -26,7 +26,12 @@ def write_prop(file, node)
   end
   $prop_set[camel_name] += 1
 
-  if name.include? '/'
+  if name.include? 'current-tracks'
+    # According to the mpv documentation this property is not supposed to be
+    # used programmatically. We still output a comment for the property so
+    # people can figure out why the Swift constant is missing.
+    file.write "  /** #{name}  As per mpv doc, scripts etc. should not use this. */\n"
```
saagarjha, line 33: Nit, I think this reads better:

```suggestion
    file.write "  /** #{name}  As per mpv docs, scripts etc. should not use this. */\n"
```

saagarjha (commented):
Looks good, aside from one comment.

other/parse_doc.rb
low-batt: Fixed. I confirmed the issue still exists with parsing the current mpv documentation and a change to the script is still required to avoid compilation errors in the generated code.

I was unsure of the proper workflow for updating a pull request. I'm new to GitHub. I'm used to having to rebase and squash, I read that GitHub has the ability to do that when merging a pull request. Let me know if I need to avoid having a PR with two commits as this one now has.

It is sometimes hard to get a read on people when just communicating through text, so to be real clear I very much appreciate people taking the time to provide a detailed review of anything I'm submitting. Nits and stylistic criticism are very welcome and I am happy to address them. Thank you for all the reviews you are doing.

saagarjha: I can't speak to the "right" way to use GitHub, but the way I do it (and we happen to usually work in IINA) is that you keep your branch tidy and we merge that directly, keeping all the commits. What this means in practice is that I will usually make PRs with one or a few commits, each of which that compiles independently; things like "add UI constants" or "implement menu handling". The intent is that each of these can be applied cleanly on top of develop, as if it was incrementally worked on, but it'll come in via a merge.
 
To update my code: if there's a conflict or I've fallen behind develop (but have no changes of my own) I rebase and push. For responses to reviews comments like typos and small suggestions that are obvious I generally just amend and don't bother with a new commit. If the review comments require significant changes and I feel the need to keep around the old implementation for some reason, I'll keep the old commit and push a new one on top.

saagarjha (approved):
lgtm.

Issue #3431, opened 2021-06-18
Subtitles won't show from any source


nammminn2:
Youtube and Dailynotion are a couple that the subtitles don't show for.

low-batt:
### Analysis

After enhancing the logging in this area of IINA, the log shows the following failure when searching for a subtitle while watching a stream from YouTube:
```text
16:58:21.938 [onlinesub][d] Search subtitle from opensubtitles.org...
16:58:24.893 [opensub][d] OpenSub: logged in as user 
16:58:24.895 [onlinesub][e] Failed to obtain subtitles for https://www.youtube.com/watch?v=TEWZpUhDu8g from opensubtitles.org. Cannot get file handle. Error Domain=NSCocoaErrorDomain Code=4 "The file “watch” doesn’t exist." UserInfo={NSFilePath=/watch, NSUnderlyingError=0x60000172e820 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}
```
IINA is attempting to calculate the special hash code supported by OpenSubtitles.org that can be used to locate subtitles for a movie. The IINA implementation is based on the sample source code provided by OpenSubtitles.org: [Hash Source Codes](https://trac.opensubtitles.org/projects/opensubtitles/wiki/HashSourceCodes)

If no subtitles are found using the hash code then IINA will search OpenSubtitles.org again, this time using the filename of the  file containing the movie.

Lookup by hash code is the best way to find the appropriate subtitles, but only if all the bytes of the video file are locally available to allow the hash code to be calculated. When streaming a movie IINA needs to fall back to search by text using the title of the movie.

### Fixing

I have heard that there are plans to replace the current subtitle related code with a plugin based system. I understand that these changes I'm proposing may not be needed if the new code is ready to be merged. Feel free to discard my changes if they are not needed. Do make sure the issues being fixed are not present in the plugin system.

[...]

PR #3499, opened 2021-09-12
Fix subtitles won't show, any source, #3431

low-batt:
This commit will:

- Change OpenSubSubtitle.hash to throw noResult if the video is being
    streamed
- Change OpenSubSubtitle.requestIMDB to use the media title as the text
    for the lookup if the video is being streamed
- Change MainMenuActions.saveDownloadedSub to copy downloaded subtitles
    from streamed files to "~/Movies"
- Change OnlineSubtitle.getSubtitle to correct a failure to display the
    OSD when using shooter.cn
- Change OpenSubSubtitle.hash and ShooterSubtitle.hash to ensure file
    is closed
- Improve logging when obtaining subtitles

With these changes subtitles can be downloaded and viewed for movies
streamed from YouTube and from Dailymotion when using OpenSubtitles.org.

- [x ] It implements / fixes issue #3431.

---

**Description:**


iina/MainMenuActions.swift
```diff
@@ -367,9 +367,15 @@ extension MainMenuActionHandler {
     }
     let subURL = URL(fileURLWithPath: path)
     let subFileName = subURL.lastPathComponent
-    let destURL = currURL.deletingLastPathComponent().appendingPathComponent(subFileName, isDirectory: false)
     do {
+      // When streaming save the subtitles file to the user's ~/Movies directory, otherwise
+      // save the file to the same directory the video is in.
+      let destDirURL = player.info.isNetworkResource ? try FileManager.default.url(
```
saagarjha, line 373: I feel like people who are streaming would rather not having subtitles randomly show up in ~/Movies. Can we use a temporary directory instead?

iina/OpenSubSubtitle.swift
```diff
@@ -186,13 +186,21 @@ class OpenSubSupport {
     }
   }
 
-  func hash(_ url: URL) -> Promise<FileInfo> {
+  func hash(_ url: URL, _ playerCore: PlayerCore) -> Promise<FileInfo> {
```
saagarjha, line 189: This seems like the wrong point to bring in an entire PlayerCore, which handles video playback. Can the caller do this check instead?

iina/OpenSubSubtitle.swift
```diff
@@ -212,28 +220,29 @@ class OpenSubSupport {
...
 
-      file.closeFile()
-
       resolver.fulfill(FileInfo(hashValue: String(format: "%016qx", hash), fileSize: fileSize))
     }
   }
 
-  func requestByName(_ fileURL: URL) -> Promise<[OpenSubSubtitle]> {
-    return requestIMDB(fileURL).then { imdb -> Promise<[OpenSubSubtitle]> in
+  func requestByName(_ fileURL: URL, _ playerCore: PlayerCore) -> Promise<[OpenSubSubtitle]> {
```
saagarjha, line 227: Same here.

iina/OpenSubSubtitle.swift
```diff
@@ -212,28 +220,29 @@ class OpenSubSupport {
...
+    return requestIMDB(fileURL, playerCore).then { imdb -> Promise<[OpenSubSubtitle]> in
       let info = ["imdbid": imdb]
       return self.request(info)
     }
   }
 
-  func requestIMDB(_ fileURL: URL) -> Promise<String> {
+  func requestIMDB(_ fileURL: URL, _ playerCore: PlayerCore) -> Promise<String> {
```
saagarjha, line 234: I think a better choice would be to ask the caller to provide a fallback title for the URL.

iina/ShooterSubtitle.swift
```diff
@@ -10,6 +10,8 @@ import Foundation
 import Just
 import PromiseKit
 
+fileprivate let subsystem = Logger.Subsystem(rawValue: "shooter")
```
saagarjha, line 13: I know this is how other code does it, but we should really move to making these extensions on `Logger` (preferably, one that is namespaced) rather than just a global at the top of the file. For example:

```swift
extension Logger.Sub {
	static let subsystem = Logger.Subsytem(rawValue: "sub.shooter")
}
```

saagarjha (commented):
Just gave this a quick read-through; interested to hear what you think.

iina/MainMenuActions.swift
low-batt: For many of my submissions I've intentionally been in "only fix it" mode where I resist the temptation cleanup and refactor code and instead try and make the smallest change to just fix the issue. Always feel free to ask me to do more. This case is slightly different. I remember thinking the correct solution is to enhance the "Save Downloaded Subtitle" menu action to bring up a Save panel. The save in "Movies" was to avoid doing that. So rather than come up with some sort of temporary directory I'm thinking the correct solution is to prompt the user. Do you want me to try and make that change?

iina/OpenSubSubtitle.swift
low-batt: From what I recall I believe I decided it would be ok to pass in PlayerCore because the parent class OnlineSubtitle was already interacting with PlayerCore. I'm used to dependency injection and strict layering where lower level code has no knowledge of higher level code, so easy to sell me on the idea that these low level methods should not be directly accessing PlayerCore. I will refactor all these methods to remove PlayerCore and make the caller pass in the required information.

iina/ShooterSubtitle.swift
low-batt: Will fix.

low-batt: Need some help with this one. I'm running into "invalid redeclaration" errors if I use this pattern in more than one place.

iina/MainMenuActions.swift
low-batt: I'm working on changing this to use a Save panel. Let me know if you think that is the wrong way to handle this.

saagarjha: Ah, I just noticed that this was the codepath for saving subtitles, rather than just grabbing them and showing them when streaming. In that case, yes, please present a save dialog for these files.

iina/OpenSubSubtitle.swift
saagarjha: A lot of IINA code is a bit more tangled than it should be, unfortunately. Sorry for pushing the cleanup on you but as you can probably guess it's easier to fix these incrementally as people add new code, rather than having to come back and clean it up later.

iina/MainMenuActions.swift
low-batt: I got that limping yesterday. I will get that pushed today.

iina/OpenSubSubtitle.swift
low-batt: Not a problem asking me to do it. I am happy to do cleanup.  :-)

iina/ShooterSubtitle.swift
low-batt: This is what worked.

In OnlineSubtitle, the parent class, I'm defining the namespace and the subsystem for that class:
```swift
extension Logger {
  struct Sub {
    static let onlinesub = Subsystem(rawValue: "onlinesub")
  }
}
```

In OpenSubSubtitle I can then code it like shown above:
```swift
extension Logger.Sub {
  static let opensub = Logger.Subsystem(rawValue: "sub.shooter")
}
```

And of course all references must then be fully qualified:
```swift
Logger.log("OpenSub: heartbeat ok", subsystem: Logger.Sub.opensub)
```

iina/MainMenuActions.swift
low-batt: The "Save Downloaded Subtitle" menu item has been changed to bring up a Save panel. I was surprised to not find good examples on the net for doing that and handling saving if there is an existing file. Definitely look over how I'm handling that. I updated the menu text to include an ellipsis. I copied the existing menu text into the corresponding Localization file so the title of the Save panel is localized. Bummer the text is duplicated. If there is a better way to do this, don't hesitate to make me change this.

iina/OpenSubSubtitle.swift
low-batt: Fixed.

iina/OpenSubSubtitle.swift
low-batt: Fixed.

iina/ShooterSubtitle.swift
low-batt: Fixed.

iina/MainMenuActions.swift
low-batt: I just noticed that PRs that modify anything other than the english translations are rejected to avoid conflicts with the Crowdin process. Please let me know if I need to remove the changes to the main menu and localization files I added with the latest commit.

lhc70000:
Unfortunately, I couldn't fix the merge conflict for the subtitle part due to significant API changes in the `plugin-system` branch. Here is a brief description of the new structure:

Each subtitle provider should define these two classes:
- `Subtitle`, representing a subtitle object found by the provider, with the following methods implemented:
  - `getDescription()`, returns the description of itself; it will be displayed in the UI so the user can choose the desired one.
  - `download()`, returns an array of `URL`s, representing the network location of the corresponding files. It will be called when user choose to download this subtitle.
- `Fetcher`, with an associated `Subtitle` type, implementing
  - `fetch(from url:, withProviderID:, playerCore:) -> Promise<[Subtitle]>`, searches for subtitles using `url` and returns the subtitle objects found.

A subtitle provider is therefore defined as `Provider<Fetcher>`.

You can check the diff here to see the actual implementation:

https://github.com/iina/iina/compare/develop...plugin-system#diff-ad90502f472115e1a63c490754fd65fa3912473fc74270b97b67eee8a467914b

---

#### So what to do now

  - I hope I could include this in 1.3.0, but seems the best way is to implement this fix based on `plugin-system`. If we really want this in 1.3.0, we can merge it right now, but reset it later before merging `plugin-system`, and then write a new one
  - We can also merge this branch right after the 1.3.0 release (if the system is still unstable, I can add a flag to disable the plugin system temporarily, so the plugin system won't be included in 1.3.1)
  - Or, postpone to 1.4.0

What do you think? Since this problem is caused by my work, I'm totally fine if you want me to re-implement this fix based on the new API design.

low-batt:
I am good with whatever course of action you would like to take. If you want you can task me with making this fix again once the plugin system has been merged. We can toss this PR in the trash. Or merge this now and then pull the commits before merging the plugin system. Either works for me.

lhc70000:
This PR will be merged into `release/1.3.0` while the `develop` branch keeps the previous code structure.

Issue #3436, opened 2021-06-25
UI: Appearance setting is ignored, always set to Match System Appearance


ADemkin:
Hello,

i found that UI -> Appearance: -> Theme: setting is completely ignored. Whatever setting is selected (Light, Dark) it is always work like "Match System Appearance". If i set IINA ui to "Dark", but macOS ui to Light, then IINA set Light theme. If i set Light in IINA, but Dark in macOS, then IINA goes Light. So it looks like this setting is totally ignored.

**System and IINA version:**

- macOS Catalina 10.15.7
- IINA 1.1.2 Build 128

**Expected behavior:**
IINA keep UI theme based on selected theme it its preferences, even if it does not match macOS theme.

**Actual behavior:**

IINA always match UI theme to current macOS theme.

**Steps to reproduce:**
Set macOS theme to Dark. Set IINA theme to Light. IINA will be Dark.
Set macOS theme to Light. Set IINA theme to Dark. IINA will be Light.

**How often does this happen?**
Always.

low-batt:
This setting is working for me, **but** it is only applied to the player windows, UI elements such as the preferences window, history window, etc., continue to reflect the macOS systemwide appearance setting. I'm guessing it is this way because you can't apply an appearance at the application level until macOS 10.14 (Mojave). Earlier macOS versions require more code to support this. I'm guessing the implementation prioritized getting this to work for the player windows and keeping the behavior consistent across all macOS versions as a first phase, planning to eventually implement support for other UI elements. But that is just me guessing at the history of this feature. 

I'm proposing that for this feature we allow IINA behavior to differ across macOS versions and fix this for macOS 10.14 and above by setting the appearance at the application level.

The fix in the pull request will change AppDelegate.applicationDidFinishLoading to set the appearance at the application level based on the IINA preference and establish an observer to watch the preference for changes and update the appearance as needed. The appearance is only set when running under macOS 10.14 and above.

To minimize changes for this fix the existing code that sets the appearance at the windows level was not updated to only set the appearance when running under macOS 10.13 and below.

Thoughts? Critical reviews including criticism of style and naming are welcome.

If reviewers want the existing window level code to be updated to only be used for macOS 10.13 just let me know and I will add that cleanup to the commit.

saagarjha:
[...]

PR #3501, opened 2021-09-14
Fix  UI: Appearance setting is ignored, #3436

low-batt:
If applied this commit will change AppDelegate.applicationDidFinishLoading
to set the appearance at the application level based on the IINA
preference and establish an observer to watch the preference for changes
and update the appearance as needed. The appearance can only be set at
the application level when running under macOS 10.14 and above. When
running under earlier versions of macOS the IINA preference will
continue to just be applied to the player windows.

To minimize changes for this fix the existing code that sets the
appearance at the windows level was not updated to only set the
appearance when running under macOS 10.13 and below.

- [ x] It implements / fixes issue #3436.

---

**Description:**

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -150,6 +150,14 @@ class AppDelegate: NSObject, NSApplicationDelegate {
   func applicationDidFinishLaunching(_ aNotification: Notification) {
     Logger.log("App launched")
 
+    // IINA supports a preference that allows the user to override the systemwide user
+    // interface appearance configured in macOS System Preferences - General. Set the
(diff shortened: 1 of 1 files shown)


alejx:
I believe that ignoring the appearance preference for the settings window **was** on purpose. Don't merge this for now.

low-batt:
Understood. Do not merge. Glad you caught the issue. Thanks for review!

saagarjha:
Correct, this behavior was intentional. We can revisit this if changes are desired, but that discussion should happen elsewhere, like the original issue for this bug.

Issue #3414, opened 2021-05-28
Support automatic screen refresh rate to match video in fullscreen mode à la Kodi


DanielSmedegaardBuus:
**What you want IINA to do:**
Switch the display resolution automatically to match the framerate of the video playing back, like Kodi does.

**What IINA does currently:**

**Why you think this should be added:**
This would make IINA a possible replacement for Kodi and useful as an HTPC option.

**Examples of other projects that have something similar:**
Kodi, possibly Plex?

anohren:
I imagine you mean when IINA is in legacy fullscreen mode -- or is it possible for IINA to change macOS' refresh rate in other modes? 

As an aside I'm having trouble seeing how IINA and Kodi overlap so much as to popularize one as a replacement for the other for HTPCs just by adding this. It has no 3 m interface and no library.

DanielSmedegaardBuus:
Well, I see what you mean, it's just that I use(d) Kodi because of these two features, not because of its library features :) The display refresh change feature is the most important feature to me, and I assume others looking to use a box as a media playback device (which I dubbed "HTPC" here, a little incorrectly).

As it turns out, having spoken to a member of team Kodi, this feature no longer works on Big Sur in Kodi, and neither do they support HDR, so you're already one-upping them there :D If IINA supported auto-refresh rate changing to match the video, that'd be a massive thing for people like me who don't want to watch content with stuttering issues, even if they're minute :)

Thanks for replying! :)

NickStrupat:
I too look for refresh rate matching as my number one feature. All other features are second to displaying the content in its native refresh rate.

23.976Hz for example... almost no set top boxes can do this today (NVidia Shield I believe can do it, and it's coming soon to Apple TV 4K), and the only player I've seen that will change the refresh rate for my is MPC (Windows-only).

In Windows I can control the refresh rate via NVidia Control Panel, but it's pretty tedious. In macOS, I don't think there's any working method of adjusting it at the moment.

DanielSmedegaardBuus:
@NickStrupat — in Windows, Kodi should do this perfectly. There are two settings for this; changing refresh to match video, and a secondary one that is match video playback to screen refresh. The latter is used if it cannot precisely match output refresh with the media content, e.g. 23.976 fps on 24 Hz output, but it does require resampling of the audio which means it cannot do DTS/DD passthrough at the same time.

low-batt:
### Concerns

[...]

PR #3527, opened 2021-10-27
Add an option to match refresh rate in fullscreen

jesec:
Allow to switch to a matching refresh rate (if there is any) when
the player goes fullscreen. This can eliminate stuttering, and, on
some external displays, enable frame interpolation.

Bug: #3414

Match patterns:

```
[23.976, 47.952, 24, 48]
[29.97, 59.94, 30, 60]
[24, 48]
[25, 50]
[30, 60]
```

Screenshot:

image

For reference, refresh rates available in newer Mac and Pro Display XDR:

image

image

and certain external displays:

image

### DIFF
--- a/iina/Base.lproj/PrefCodecViewController.xib
+++ b/iina/Base.lproj/PrefCodecViewController.xib
@@ -27,7 +27,7 @@
         </customView>
         <userDefaultsController representsSharedInstance="YES" id="pxc-7C-SGP"/>
         <customView id="gZf-gF-XoY">
-            <rect key="frame" x="0.0" y="0.0" width="444" height="202"/>
+            <rect key="frame" x="0.0" y="0.0" width="444" height="278"/>
             <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMinY="YES"/>
             <subviews>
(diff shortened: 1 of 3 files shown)


varenc:
I'm just another IINA user looking at PRs, but this is really cool! Sounds like it automatically adjusts the actual macOS system wide display settings to accommodate the video's frame rate and then restores it?  Wish I had a monitor that supported something other than 30 and 60fps so I could actually try this out with 23.976 fps video.  Though adjusting and restoring the display refresh rate would cause my screen to black out for ~3-5 seconds, but I imagine on the new M1 machines it happens instantly.

Have you compared this to just setting mpv's `video-sync` property to `display-resample`?  I experienced stuttering due to the video/display mismatch, but setting `video-sync=display-resample` in my mpv.conf, which I have IINA load, has totally fixed it for me without noticeable side effects. I believe it works be slightly adjusting video/audio playback speed so that the video's frames align with the display refresh rate, though they're subtle enough I've never noticed the speed adjustments.

This PR is the proper solution, but I wonder if just having IINA default to use `video-sync=display-resample` would get 90% of the way there.

PR #3529, opened 2021-10-30
Close add A-B Loop indicator on progress bar, #548

### LINKED ISSUE #548: Adding A-B Loop indicator on progress bar
**System and IINA version:**

- macOS 10.12.4
- IINA 0.0.7

**Expected behavior:**

**Actual behavior:**

It will be intuitive if we can highlight the loop section on progress bar and provide handle for adjusting loop duration
1. Easy to know the looping status.
1. No need to clear loop and start all over again.

- Crash report:
```

```

- MPV log:
```

```

**Steps to reproduce:**

low-batt:
This commit adds two additional thumbs to the progress bar representing
the A and B loop points when the A-B loop feature is active. This allows
the user to adjust the loop points.

This commit will:
- Add a new class PlaySliderLoopKnob that adds an additional thumb
- Add a new class PlaySlider that customizes NSSlider to add the thumbs
- Change MainWindowController.xib to use PlaySlider for the OSC
- Change PlayerCore to allow mpv loop points to be updated
- Change MainWindowController to observe the thumbs and update mpv
- Change PlayerWindowController to synchronize the thumbs with mpv
- Change MainMenuActions to call the window controller for A-B Loop
- Change PlaybackInfo to use an enum class for abLoopStatus
- Add a new message to OSDMessage

- [ x] It implements / fixes issue #548.

---

**Description:**
See the associated issue for screenshots of what the OSC looks like with these changes.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -634,16 +634,16 @@ class MainWindowController: PlayerWindowController {
 
     // Observe the loop knobs on the progress bar and update mpv when the knobs move.
     addObserver(to: .default, forName: .iinaPlaySliderLoopKnobChanged, object: playSlider.abLoopA) { [weak self] _ in
-      guard let strongSelf = self else { return }
-      let seconds = strongSelf.percentToSeconds(strongSelf.playSlider.abLoopA.doubleValue)
-      strongSelf.player.abLoopA = seconds
-      strongSelf.player.sendOSD(.abLoopUpdate(.aSet, VideoTime(seconds).stringRepresentation))
+      guard let self = self else { return }
+      let seconds = self.percentToSeconds(self.playSlider.abLoopA.doubleValue)
+      self.player.abLoopA = seconds
+      self.player.sendOSD(.abLoopUpdate(.aSet, VideoTime(seconds).stringRepresentation))
     }
     addObserver(to: .default, forName: .iinaPlaySliderLoopKnobChanged, object: playSlider.abLoopB) { [weak self] _ in
-      guard let strongSelf = self else { return }
-      let seconds = strongSelf.percentToSeconds(strongSelf.playSlider.abLoopB.doubleValue)
-      strongSelf.player.abLoopB = seconds
-      strongSelf.player.sendOSD(.abLoopUpdate(.bSet, VideoTime(seconds).stringRepresentation))
+      guard let self = self else { return }
+      let seconds = self.percentToSeconds(self.playSlider.abLoopB.doubleValue)
+      self.player.abLoopB = seconds
+      self.player.sendOSD(.abLoopUpdate(.bSet, VideoTime(seconds).stringRepresentation))
     }
 
     player.events.emit(.windowLoaded)
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -632,7 +632,39 @@ class MainWindowController: PlayerWindowController {
       setWindowFrameForLegacyFullScreen()
     }
 
+    // Observe the loop knobs on the progress bar and update mpv when the knobs move.
+    addObserver(to: .default, forName: .iinaPlaySliderLoopKnobChanged, object: playSlider.abLoopA) { [weak self] _ in
+      guard let strongSelf = self else { return }
+      let seconds = strongSelf.percentToSeconds(strongSelf.playSlider.abLoopA.doubleValue)
+      strongSelf.player.abLoopA = seconds
+      strongSelf.player.sendOSD(.abLoopUpdate(.aSet, VideoTime(seconds).stringRepresentation))
+    }
+    addObserver(to: .default, forName: .iinaPlaySliderLoopKnobChanged, object: playSlider.abLoopB) { [weak self] _ in
+      guard let strongSelf = self else { return }
+      let seconds = strongSelf.percentToSeconds(strongSelf.playSlider.abLoopB.doubleValue)
+      strongSelf.player.abLoopB = seconds
+      strongSelf.player.sendOSD(.abLoopUpdate(.bSet, VideoTime(seconds).stringRepresentation))
+    }
+
     player.events.emit(.windowLoaded)
+  }
+
+  /// Returns the position in seconds for the given percent of the total duration of the video the percentage represents.
+  ///
+  /// The number of seconds returned must be considered an estimate that could change. The duration of the video is obtained from
+  /// the [mpv](https://mpv.io/manual/stable/) `duration` property. The documentation for this property cautions that
+  /// mpv is not always able to determine the duration and when it does return a duration it may be an estimate. If the duration is
+  /// unknown this method will fallback to using the current playback position, if that is known. Otherwise this method will return zero.
+  /// - Parameter percent: Position in the video as a percentage of the duration.
+  /// - Returns: The position in the video the given percentage represents.
+  private func percentToSeconds(_ percent: Double) -> Double {
+    if let duration = player.info.videoDuration?.second {
+      return duration * percent / 100
+    } else if let position = player.info.videoPosition?.second {
+      return position * percent / 100
+    } else {
+      return 0
+    }
   }
 
   /** Set material for OSC and title bar */
@@ -1122,13 +1154,21 @@ class MainWindowController: PlayerWindowController {
     guard let w = self.window, let cv = w.contentView else { return }
     if cv.trackingAreas.isEmpty {
       cv.addTrackingArea(NSTrackingArea(rect: cv.bounds,
-                                        options: [.activeAlways, .inVisibleRect, .mouseEnteredAndExited, .mouseMoved],
+                                        options: [.activeAlways, .enabledDuringMouseDrag, .inVisibleRect, .mouseEnteredAndExited, .mouseMoved],
                                         owner: self, userInfo: ["obj": 0]))
     }
     if playSlider.trackingAreas.isEmpty {
       playSlider.addTrackingArea(NSTrackingArea(rect: playSlider.bounds,
-                                                options: [.activeAlways, .inVisibleRect, .mouseEnteredAndExited, .mouseMoved],
+                                                options: [.activeAlways, .enabledDuringMouseDrag, .inVisibleRect, .mouseEnteredAndExited, .mouseMoved],
                                                 owner: self, userInfo: ["obj": 1]))
+    }
+    // Track the thumbs on the progress bar representing the A-B loop points and treat them as part
+    // of the slider.
+    if playSlider.abLoopA.trackingAreas.count <= 1 {
+      playSlider.abLoopA.addTrackingArea(NSTrackingArea(rect: playSlider.abLoopA.bounds, options:  [.activeAlways, .enabledDuringMouseDrag, .inVisibleRect, .mouseEnteredAndExited, .mouseMoved], owner: self, userInfo: ["obj": 1]))
+    }
+    if playSlider.abLoopB.trackingAreas.count <= 1 {
+      playSlider.abLoopB.addTrackingArea(NSTrackingArea(rect: playSlider.abLoopB.bounds, options: [.activeAlways, .enabledDuringMouseDrag, .inVisibleRect, .mouseEnteredAndExited, .mouseMoved], owner: self, userInfo: ["obj": 1]))
     }
 
     // update timer
--- a/iina/PlaySlider.swift
+++ b/iina/PlaySlider.swift
@@ -1,0 +1,88 @@
+//
+//  PlaySlider.swift
+//  iina
+//
+//  Created by low-batt on 10/11/21.
+//  Copyright © 2021 lhc. All rights reserved.
+//
+
+import Cocoa
+
+/// A custom [slider](https://developer.apple.com/design/human-interface-guidelines/macos/selectors/sliders/)
+/// for the onscreen controller.
+///
+/// This slider adds two thumbs (referred to as knobs in code) to the progress bar slider to show the A and B loop points of the
+/// [mpv](https://mpv.io/manual/stable/) A-B loop feature and allow the loop points to be adjusted. When the feature is
+/// disabled the additional thumbs are hidden.
+/// - Requires: The custom slider cell provided by `PlaySliderCell` **must** be used with this class.
+/// - Note: Unlike `NSSlider` the `draw` method of this class will do nothing if the view is hidden.
+final class PlaySlider: NSSlider {
+
+  /// Knob representing the A loop point for the mpv A-B loop feature.
+  var abLoopA: PlaySliderLoopKnob { abLoopAKnob }
+
+  /// Knob representing the B loop point for the mpv A-B loop feature.
+  var abLoopB: PlaySliderLoopKnob { abLoopBKnob }
+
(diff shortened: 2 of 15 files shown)


CarterLi:
I got these kind of errors:

```
/Users/carter/iina/iina/PlaySliderLoopKnob.swift:162:59: error: 'darkKnobColor' is unavailable in macOS
      return slider.window!.effectiveAppearance.isDark ? .darkKnobColor : .lightKnobColor
                                                          ^~~~~~~~~~~~~
/Users/carter/iina/iina/PlaySliderLoopKnob.swift:15:14: note: 'darkKnobColor' was obsoleted in macOS 10.14
  static let darkKnobColor = NSColor(calibratedRed: 0.59, green: 0.59, blue: 0.59, alpha: 1)
             ^
/Users/carter/iina/iina/PlaySliderLoopKnob.swift:162:76: error: 'lightKnobColor' is unavailable in macOS
      return slider.window!.effectiveAppearance.isDark ? .darkKnobColor : .lightKnobColor
                                                                           ^~~~~~~~~~~~~~
/Users/carter/iina/iina/PlaySliderLoopKnob.swift:16:14: note: 'lightKnobColor' was obsoleted in macOS 10.14
  static let lightKnobColor = NSColor(calibratedRed: 0.35, green: 0.35, blue: 0.35, alpha: 1)
```

Any ideas?

low-batt:
Hi!

The offending code from PlaySliderLoopKnob.swift is:
```swift
// These colors are for 10.13- only
@available(macOS, obsoleted: 10.14)
fileprivate extension NSColor {
  // Use different colors to distinguish loop knobs from the primary knob.
  static let darkKnobColor = NSColor(calibratedRed: 0.59, green: 0.59, blue: 0.59, alpha: 1)
  static let lightKnobColor = NSColor(calibratedRed: 0.35, green: 0.35, blue: 0.35, alpha: 1)
}

    // Starting with macOS Mojave 10.14 colors can be configured to automatically adjust for the
    // current appearance.
    if #available(macOS 10.14, *) {
      return NSColor(named: .mainSliderLoopKnob)!
    } else {
      return slider.window!.effectiveAppearance.isDark ? .darkKnobColor : .lightKnobColor
    }
```
I patterned that code after this existing code from PlaySliderCell.swift:
```swift
// These colors are for 10.13- only
@available(macOS, obsoleted: 10.14)
fileprivate extension NSColor {


    if #available(macOS 10.14, *) {
      return NSColor(named: .mainSliderKnobActive)!
    } else {
      return .darkKnobColor
    }
```
I'm not spotting a typo in the new code that would trigger the errors shown.
Do you spot something I've missed?

This relates to the "macOS Deployment Target" build setting in the top level IINA project.
It should be set to "10.11" as IINA still supports that release.

Any local changes to your Xcode project file?
What Xcode version?
I was testing with Xcode 12.5.

Late, so my brain may not be working. I'm coming up empty. I'll think on this some more.

CarterLi:
I was using the latest Xcode 13.1

Strange enough. I didi find a similar usage:

https://github.com/iina/iina/blob/develop/iina/PlaySliderCell.swift#L12

But I didn't get errors there. And If I changed `obsoleted ` to `deprecated`, the error disappeared.

Don't know if it's a xcode bug or not.

low-batt:
All the work on these changes was done on an Intel Mac in a VM running macOS 11.4 using Xcode 12.5. I've since moved my development system to be native on an Apple silicon Mac running macOS 12.0.1 and Xcode Version 13.1 (13A1030d). I have these changes in my current local branch. Xcode is successfully compiling PlaySliderLoopKnob for me. The mystery continues!

This from the [Xcode 13 Release Notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-13-release-notes) is interesting:
> Availability checks in iPhone and iPad apps on a Mac with Apple silicon always return true. This causes iOS apps running in macOS 11 Big Sur to see iOS 15 APIs as available, resulting in crashes. This only affects apps available in the Mac App Store built with the “My Mac (Designed for iPhone)” or “My Mac (Designed for iPad)” run destination. It doesn’t affect Mac Catalyst apps. (83378814)

The description of the problem indicates the failure only affects iPhone and iPad apps, but maybe the problem is worse than they knew when the release notes were published? 

Are you building on an Apple silicon Mac?
If so, are you building for Apple silicon or for Intel (with Rosetta)?
I'm building for Apple silicon.

The release notes under Swift /  Resolved Issues lists this fix:
> The compiler used to erroneously accept @available annotations on enum cases with associated values that were newer than the deployment target. (80238318)

Maybe that "fix" introduced a regression in the availability support?

The [Xcode 13.2 Beta Release Notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-13_2-release-notes) contains this:

> Fixed an issue where availability checks in iPhone and iPad apps running on a Mac with Apple silicon always returned true, which caused iOS apps running in macOS Big Sur to see iOS 15 APIs as available and crash when trying to use those APIs. These availability checks now return the correct result for apps compiled with Xcode 13.2. (83378814)

So the issue with availability checks is definitely still a problem in Xcode 13.1.

None of this provides a definitive answer for the problem at hand. It just makes more suspicious that the issue is due to Xcode 13.

---
By the way, IINA built under Xcode 13 crashed out for me when I bought up Preferences / General. Telegram user xjc pointed me to a fix in this commit ed796c66ce0197293dc19448cc19667d79dda249 that upgrades the [Sparkle](https://sparkle-project.org/) framework to 2.0.0-beta.3. That fix worked for me.

iina/PlaySlider.swift
```diff
@@ -0,0 +1,88 @@
...
+      return
+    }
+    super.mouseDown(with: event)
+  }
+
+  override func viewDidUnhide() {
+    super.viewDidUnhide()
+    // When IINA is not the application being used and the onscreen controller is hidden if the
+    // mouse is moved over an IINA window the IINA will unhide the controller. If the slider is
+    // not marked as needing display the controller will show without the slider. I would of thought
```
svobs: "would have" thought

svobs:
Played around with this and it looks good to me. Although I didn't notice until now that there is some screwiness with the position of the knob on the play slider. If you try clicking on the slider near the beginning of the video, the slider bar is drawn about too far to the right by ~ 1x its full width.  When clicking near the end, it's drawn to the left by ~ half its width. I should probably file a bug report for it sometime. Doesn't seem to be the fault of your PR (I checked and now see it on develop branch and in older versions of IINA), but it's noticeable because the play slider seems to go past the "B" indicator before looping. And for some reason it seems more noticeable with a `Floating` OSC.

iina/PlaySlider.swift
low-batt: Good catch! New commit fixes typo.

iina/MainWindowController.swift
```diff
@@ -630,9 +630,41 @@ class MainWindowController: PlayerWindowController {
       setWindowFrameForLegacyFullScreen()
     }
 
+    // Observe the loop knobs on the progress bar and update mpv when the knobs move.
+    addObserver(to: .default, forName: .iinaPlaySliderLoopKnobChanged, object: playSlider.abLoopA) { [weak self] _ in
+      guard let strongSelf = self else { return }
```
uiryuu:
``` swift
guard let self = self else { return }
```
This is supported starting from Swift 4.2

iina/MainWindowController.swift
```diff
@@ -630,9 +630,41 @@ class MainWindowController: PlayerWindowController {
...
+      let seconds = strongSelf.percentToSeconds(strongSelf.playSlider.abLoopA.doubleValue)
+      strongSelf.player.abLoopA = seconds
+      strongSelf.player.sendOSD(.abLoopUpdate(.aSet, VideoTime(seconds).stringRepresentation))
+    }
+    addObserver(to: .default, forName: .iinaPlaySliderLoopKnobChanged, object: playSlider.abLoopB) { [weak self] _ in
+      guard let strongSelf = self else { return }
```
uiryuu: Ditto

iina/PlaySlider.swift
```diff
@@ -0,0 +1,88 @@
...
+
+  /// Knob representing the B loop point for the mpv A-B loop feature.
+  var abLoopB: PlaySliderLoopKnob { abLoopBKnob }
+
+  /// The slider's cell correctly typed for convenience.
+  var customCell: PlaySliderCell { cell as! PlaySliderCell }
+
+  // MARK:- Private Properties
+
+  private var abLoopAKnob: PlaySliderLoopKnob!
```
uiryuu, line 38: I think we can use `private(set) var abLoopAKnob` instead of creating `var abLoopA`.

iina/PlaySliderLoopKnob.swift
```diff
@@ -0,0 +1,218 @@
...
+  ///
+  /// The height of loop knobs is reduced in order to give prominence to the primary knob.
+  private static let knobHeightAdjustment: CGFloat = 0.75
+
+  // The X coordinate of the last mouse location when dragging.
+  private var lastDragLocation: CGFloat = 0
+
+  private var slider: PlaySlider!
+
+  private var value: Double = 0
```
uiryuu: Is it necessary to use a separate `value` here? I would recommend to only use `doubleValue` with `didSet` if there's no other special consideration.

iina/PlaySliderLoopKnob.swift
```diff
@@ -0,0 +1,218 @@
...
+/// This class adds an additional thumb (knob) to a slider.
+///
+/// This class is used to add thumbs representing the A and B loop points of the [mpv](https://mpv.io/manual/stable/) A-B
+/// loop feature when that feature is in use. When the feature is not being used the thumbs are hidden.
+/// - Requires: The custom slider provided by `PlaySlider` must be used with this class.
+/// - Note: This class is derived from `NSView` in part to gain support for help tags (tool tips).
+final class PlaySliderLoopKnob: NSView {
+
+  // Must accept first responder for help tags to work.
+  override var acceptsFirstResponder: Bool { true }
```
uiryuu: Found a bug with this line:
- Play a video
- Create an AB-loop
- Drag the B knob to adjust the loop
- Make the cursor be in the OSC so that the controller doesn't disappear
- Try to pause the video by pressing space 
- The video is temporary stopped and then resume

Removing this line seems to solve this problem, so I'm guessing this maybe the culprit.

uiryuu (changes requested):
When the playback position is on A, the main knob will surpass knob A by 1(?)px. Don't know whether it's a intended design?

iina/MainWindowController.swift
low-batt: Fixing

iina/MainWindowController.swift
low-batt: Fixing

iina/PlaySliderLoopKnob.swift
low-batt: I remember I had to add this to get tool tips to work. And yet I commented it out and tool tips still seemed to work. I will test some more, but yes it is looking like the fix is to remove this.

low-batt:
I'm also looking into your comment about handling`mouseDown` in `PlaySliderLoopKnob`. I think that will work and simplify the code. Still working on an update.

low-batt:
I decided to go ahead and explore the behavior of the main slider. Changing `knobWidth` in `PlaySliderCell` from 3 to 30 makes the primary knob issues obvious. This is the knob set as far to the left as it can go:

issue-548-min

If I move the knob all the way to the right it disappears completely. In this screen shot it is not moved as far right as it can go:

issue-548-max

My expectation is that knobs should never move outside of the width of the bar. This is a loop knob set all the way to the left:

issue-548-a-min

And all the way to the right:

issue-548-a-max

This behavior is because the code in `PlaySliderLoopKnob` takes into account the width of the knob and does not use the full width of the bar:
```swift
    // The knob is not allowed to slide past the end of the bar, so the usable width is reduced.
    let effectiveWidth = bar.width - cell.knobWidth
```

I will see if I can make the main knob behave.

low-batt:
Here is another part of the problem. From [NSView.isFlipped](https://developer.apple.com/documentation/appkit/nsview/1483532-isflipped):
> The default value of this property is false, which results in a non-flipped coordinate system. In a non-flipped coordinate system, the origin is in the lower-left corner of the view and positive y-values extend upward. In a flipped coordinate system, the origin is in the upper-left corner of the view and y-values extend downward. X-values always extend to the right.

So `isFlipped` controls the y-axis and has no effect on the x-axis. However `knobRect` in `PlaySliderCell` considers it when calculating the x position of the knob:
```swift
    return NSMakeRect(pos - flippedMultiplier * 0.5 * knobWidth, rect.origin.y, knobWidth, height)
```

This is what is preventing the knob from going all the way to the right.

At least I think I am properly understanding this…

svobs:
> So isFlipped controls the y-axis and has no effect on the x-axis. However knobRect in PlaySliderCell considers it when calculating the x position of the knob[...]

Yeah I saw this when I dug into the code a while back. I too thought that was the smoking gun at first, but it turned out to be a red herring, or at best only part of the problem. Unfortunately something like that usually hints that the code around it should not be trusted, but the code around it is doing all kinds of unusual things too. Hopefully you were able to map out part of this 🐇🕳️ already and might have better odds than me!

low-batt:
And that is why I decided to not touch the main slider code back when I originally added this PR. It is not just one problem in one part of the code. I have found a few more things that seem wrong. I have the knobs matching up now. But I still have more work to do. Might not get done tomorrow, but will have an update soon.

low-batt:
I'm still having "fun" with this one. Here is the latest problem I tracked down in the current code. This is where the slider knob ends up when you start playing a video with IINA configured to pause at the start:

issue-548-gap

You would expect the slider knob to be at the very start of the slider. But notice the gap. That is as far left as the knob will go. Why is part of the slider visible? In `PlaySliderCell` we override `barRect`:

```swift
  override func barRect(flipped: Bool) -> NSRect {
    let rect = super.barRect(flipped: flipped)
    return NSMakeRect(0, rect.origin.y, rect.width + rect.origin.x * 2, rect.height)
  }
```

I wish there was a comment about the thinking behind this. I still need to check Catalina, but on Ventura the slider's bar is located at 2. So we are overriding this method to stretch the size of the bar, moving it to 0 and increasing the width by 4. I don't see a problem with that.

But the method `knobRect` contains this line:
```swift
let barRect = super.barRect(flipped: flipped)
```

Because it specified `super` it by passed the local overridden method is therefore using the wrong bar size.

iina/PlaySliderLoopKnob.swift
low-batt: Fixing this. Removing `value` and changing `doubleValue` to use `didSet`.

low-batt:
Based on the excellent review comments I reworked the new code to simplify it. I also updated the existing `PlaySliderCell.knobRect` method to correct problems with determining the x coordinate of the thumb that controls the playback position. These changes correct the problems with the primary knob being unable to move all the way to the start of the bar and not lining up with the loop knobs.

I also removed the method `PlaySliderCell.barRect` that increased the width of the slider's bar. We can put that back in if reviewer's object. It makes it very hard to grab a knob that is at the end of the bar as the bar is running right up against the time label. A lot of the time you will end up changing what the time label is reporting while trying to click on the thumb.

uiryuu (approved):
LGTM

PR #3552, opened 2021-11-29
replace considates with consolidates

nicu-chiciuc:

---

Either I'm wrong or the word "considates" doesn't exist.


nicu-chiciuc:
@saagarjha this seems like a simple change, can it be merged so I don't have this PR hanging?

saagarjha:
Indeed it can. Thanks for the fix :)

Issue #3558, opened 2021-12-05
Camera housing blocks onscreen controller


low-batt:
**System and IINA version:**

- macOS 12.0.1
- IINA 1.2.0

**Expected behavior:**
The IINA window is not blocked by the camera housing.

**Actual behavior:**
When the IINA preference "use legacy full screen" is enabled while running on one of the new MacBook Pros that have a camera housing that intrudes into the screen, a portion of the video window is obscured when IINA is in full screen mode:
![camera-housing](https://user-images.githubusercontent.com/86170219/144761104-7d64a9d9-5827-4e91-a403-6d748b3f1455.jpg)

**Steps to reproduce:**
Obtain  a Mac with a camera housing ("notch") that obscures part of the screen to test on.
Obtain a build of IINA with issue #3543 fixed, otherwise IINA will crash.
Then:
- Enable the "use legacy full screen" IINA preference
- Configure the onscreen controller layout to be "Top"
- Play a video
- Under the "Video" menu select "Enter Full Screen"
- The entire screen will be used including the top portion where the middle is obscured by the camera

- [ ] MPV does not have this problem.

Haven't checked mpv, but this is an issue with IINA.

**How often does this happen?**
Everytime.

I will shortly post more information and submit a proposed fix.

low-batt:
### Workaround

It is possible to workaround this problem by enabling "Scale to fit below built-in camera" as explained in this Apple document: [How to adjust an app's settings to appear below the camera area](https://support.apple.com/en-us/HT212842)

The disadvantage of this workaround is that it shrinks the entire window.

### Analysis

The Apple document [Human Interface Guidelines / Full-Screen Mode](https://developer.apple.com/design/human-interface-guidelines/macos/app-architecture/fullscreen-mode/)  says:

> On some Macs, the camera housing occupies an area at the top-center of the screen. Using the system’s full-screen support automatically accounts for this area; if you use a custom full-screen experience, you need to account for this area as you position your content. 

This explains why IINA's normal full screen mode is working on the new Macs on which the camera housing intrudes into the screen, but the legacy full screen feature which requires custom full-screen behavior is not.

For Monterrey a new [safeAreaInserts](https://developer.apple.com/documentation/appkit/nsscreen/3882821-safeareainsets) property has been added to [NSScreen](https://developer.apple.com/documentation/appkit/nsscreen). The documentation for this new property says:
[...]

PR #3559, opened 2021-12-05
Fix  camera housing blocks controller, #3558

low-batt:
- Extends NSScreen with a new optional computed property,
    cameraHousingHeight
- Adds a new class CameraAreaMaskWindow that is used to black out the
    area of the screen to the left and right of the camera housing
- Changes MainWindowController.legacyAnimateToFullscreen to check if
    cameraHousingHeight indicates a camera housing intrudes into the
    screen and if so reduce the size of the full screen video window
    appropriately and put up a black window to hide the wallpaper on
    either side of the camera housing
- Changes MainWindowController.legacyAnimateToWindowed to close the
    window masking the area around the camera housing if displayed

- [x] It implements / fixes issue #3558.

---

**Description:**
To see this fix work, the fix for #3543 must first be applied as otherwise the once legacy full screen is enabled IINA will crash.

### DIFF
--- a/iina.xcodeproj/project.pbxproj
+++ b/iina.xcodeproj/project.pbxproj
@@ -15,6 +15,7 @@
 		49542986216C34950058F680 /* OpenInIINA.appex in Embed App Extensions */ = {isa = PBXBuildFile; fileRef = 49542973216C34950058F680 /* OpenInIINA.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
 		496B19921E2968530035AF10 /* PIP.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 496B19911E2968530035AF10 /* PIP.framework */; settings = {ATTRIBUTES = (Weak, ); }; };
 		519872FF26879B9B00F84BCC /* AccessibilityPreferences.swift in Sources */ = {isa = PBXBuildFile; fileRef = 519872FE26879B9B00F84BCC /* AccessibilityPreferences.swift */; };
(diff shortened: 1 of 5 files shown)


saagarjha:
Sorry, I've only skimmed this PR, but is there a reason you need to make an entire new window to hide the background? Does changing the window background not work?

low-batt:
My background does not include GUI work. Definitely question anything UI related you see me submit. When it comes to UI I'm a nearly total newbie.

Setting the background of the main window does not work since the main window does not extend into the menu area. To do that we'd have to separate the size of the window and the size of the video area. I vaguely remember some open issue requesting that for full screen, but I thought currently the video takes up all of the main window. We can't have the video extend into the menu area as the camera house will block some of the video, so the main window size is reduced to avoid that.

I tried just creating a black window that was just the size of the menu area, but macOS adjusted its position so it was right below the menu area. The only way I could get macOS to put something in that area of the screen was to create a window that took up the entire screen.

If there is some way to have the main window take up all of the screen, but restrict the video so that it does not use the top area of the window, then that would be a better solution. Did I miss a way to do that?

saagarjha:
The video view is constrained to match the size of the window in `MainWindowController.addVideoViewToWindow()`. Until now we haven't had to add anything fancy there, but what I was expecting is that because of the new safe area inset we would adjust the top constraint by looking at [`NSScreen`'s `safeAreaInsets`](https://developer.apple.com/documentation/appkit/nsscreen/3882821-safeareainsets).

low-batt:
That certainly would be a lot better way to handle this. I will try and get that working.

low-batt:
Switching to changing the view is a big enough change that I am abandoning this PR and replacing it with PR #3618 which is based on the latest develop branch. The new code is a huge improvement over the code in this PR.

Issue #3574, opened 2021-12-13
Media keys doesn't work


TheBlinded:
**System and IINA version:**

- macOS 11.6.1
- IINA 1.1.2 and 1.2

**Expected behavior:**
Clicking the play / pause button on the keyboard should cause the video to start or stop.

**Actual behavior:**
With iina open and clicking the play button, the "Music" system app opens. With all the other programs, however, the button works (VLC, videos on Firefox etc.).
In fact, iina is the only one that never appears in the playback menu (attached image). This happens from Catalina I think. I attach screenshots to make it clear.

**Steps to reproduce:**
Open any audio or video with iina.

**How often does this happen?**
Always.

MPV 0.34.0 have the same problem.

![Schermata 2021-12-13 alle 18 08 34](https://user-images.githubusercontent.com/5796668/145860317-bcc31cc2-6231-46e6-a5cb-120a0cd7019e.jpg)

low-batt:
Similar report in #3340.

I quickly ran a test. Under Catalina and the play / pause was recognized by IINA. Under Monterey, 12.0.1, I experienced the reported behavior. With IINA playing and in focus the play / pause key started the Music app instead of pausing IINA. Seems like something changed in Big Sur.

low-batt:
I found a problem. The property `playbackState` in the `MPNowPlayingInfoCenter` is left set to `unknown` when playing starts instead of being updated to `playing`. That needs to be fixed. Continuing to investigate.

low-batt:
### Analysis

In `AppDelegate.applicationDidFinishLaunching` support for the Now Playing Control Center is initialized, setting the property `playbackState` in the `MPNowPlayingInfoCenter` to `unknown`.  Then in `PlayerCore.fileStarted` the `MPNowPlayingInfoCenter` is updated with information about the media being played such as the title, but the `playbackState` is not updated and is left set to `unknown`. Apparently starting with Big Sur the `playbackState` must be set to something other than `unknown` for the Now Playing Control Center to assign the media keys to the application. Likely this is a result of Apple porting Control Center code from iOS to macOS.

### Fixing

The commit in the pull request:
- Adds the ability to set the state to `NowPlayingInfoManager.updateInfo` 
- Eliminates the method  `NowPlayingInfoManager.updateState`
- Changes  `PlayerCore.fileStarted` to set the state to `playing`
- Changes references to `updateState` to use `updateInfo`

low-batt:
[...]

PR #3579, opened 2021-12-16
Fix media keys doesn't work, #3574

low-batt:
- Adds the ability to set the state to `NowPlayingInfoManager.updateInfo`
- Eliminates the method  `NowPlayingInfoManager.updateState`
- Changes  `PlayerCore.fileStarted` to set the state to `playing`
- Changes references to `updateState` to use `updateInfo`

- [x] It implements / fixes issue #3574.

---

**Description:**

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -167,7 +167,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
       if RemoteCommandController.useSystemMediaControl {
         Logger.log("Setting up MediaPlayer integration")
         RemoteCommandController.setup()
-        NowPlayingInfoManager.updateState(.unknown)
+        NowPlayingInfoManager.updateInfo(state: .unknown)
       }
     }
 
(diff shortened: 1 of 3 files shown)


tueksta:
is this working for everybody? I'm on Monterey 12.6.3 and 1.3.1 Build 133 does not react to media keys at all.

low-batt:
Hi @tueksta,

I just tested IINA 1.3.1 on my MacBookPro18,2 under macOS Ventura 13.2 and media keys worked for me.

Try this:
- Start IINA
- Click on  `Settings…` (preferences) under the `IINA` menu
- The settings panel appears
- On the left side of the panel click on `Key Bindings`
- On the right side of the panel at the top is the setting `Use system media control`
- Is this setting checked (enabled)?

This setting controls IINA's use of the media keys as well as use of the now playing widget in [Control Center](https://support.apple.com/guide/mac-help/quickly-change-settings-mchl50f94f8f/mac). If this setting is not checked, check it, restart IINA and test the media keys again.

If it is checked then we will need to dig into why this is not working.

tueksta:
hmm, seems the service wasn't running after all. got fixed by `launchctl load -w /System/Library/LaunchAgents/com.apple.rcd.plist` thanks for the quick feedback though.

low-batt:
Yikes! Not what I expected. Definitely bring up [Console](https://support.apple.com/guide/console/welcome/mac) and check crash reports to see if macOS background daemons are crashing and restarting.

Thanks for letting us know what was causing the failure.

PR #3581, opened 2021-12-18
Fix pause the video and the program crashes, #3475

low-batt:
- Adds a context property to ViewLayer for the OpenGL context
- Changes MPVController.mpvInitRendering to save the OpenGL context
- Changes ViewLayer.draw to use the saved context

- [x] It implements / fixes issue #3475 .

---

**Description:**

### DIFF
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -366,7 +366,9 @@ class MPVController: NSObject {
         mpv_render_param()
       ]
       mpv_render_context_create(&mpvRenderContext, mpv, &params)
-      mpv_render_context_set_update_callback(mpvRenderContext!, mpvUpdateCallback, mutableRawPointerOf(obj: player.mainWindow.videoView.videoLayer))
+      let layer = player.mainWindow.videoView.videoLayer
+      layer.context = CGLGetCurrentContext()
+      mpv_render_context_set_update_callback(mpvRenderContext!, mpvUpdateCallback, mutableRawPointerOf(obj: layer))
     }
   }
 
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -41,6 +41,8 @@ class MPVController: NSObject {
   var mpv: OpaquePointer!
   var mpvRenderContext: OpaquePointer?
 
+  private var openGLContext: CGLContextObj! = nil
+
   var mpvClientName: UnsafePointer<CChar>!
   var mpvVersion: String!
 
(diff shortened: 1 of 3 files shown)


CarterLi:
Does `CGLLockContext` have negative impact on performance?

low-batt:
Definitely something a programmer should be thinking about whenever they see the word "lock".

Yes, CGLockContext can have a negative impact on performance. This thread will block on that call, waiting until it can obtain the lock and while it has the lock it will block other threads from getting the lock.

From [OpenGL Programming Guide for Mac](https://developer.apple.com/library/archive/documentation/GraphicsImaging/Conceptual/OpenGL-MacProgGuide/opengl_threading/opengl_threading.html): 

> **OpenGL Restricts Each Context to a Single Thread**
> ...
> OpenGL is not reentrant. If you modify the same context from multiple threads simultaneously, the results are unpredictable. Your application might crash or it might render improperly.

So no choice, OpenGL requires us to lock the context.

Another worry with locks is deadlocking, especially when aquiring more than one lock at a time. Threads must acquire multiple locks in the same order to avoid deadlocks. I was careful to follow the pattern in the other `draw` method and lock the context after acquiring the uninitLock.

Yet another worry with locks is failing to unlock. Some path where the code returns early, skipping an unlock. A return statement or an exception. The correct coding pattern for swift is probably to use the `defer` keyword. That is not being done in the other `draw` method, so I did not make use of it in the new code.

low-batt:
Merge PR #3593 first. Then I will update this PR to use the new MPVController methods introduced in that PR.

low-batt:
Wait to review these changes until I finish investigating a deadlock in IINA-Plus that may be related to these changes.

low-batt:
The first fix resulted in a deadlock that could be reproduced by creating a very short A-B loop and letting it run until playback stopped. When making the fix I was careful to lock the `uninitLock` first and then lock the OpenGL context, to match the locking order in the other `draw` method. Or so I thought. The Apple documentation is silent on the state of the OpenGL context being passed to `CAOpenGLLayer` methods, but I set a breakpoint on `CGLLockContext` and confirmed that  `CAOpenGLLayer._display` calls it before calling `draw`. That explains the deadlock. Reversing the order of lock acquisition corrected the deadlock.

However that is not the fix I'm proposing. Locking and setting the OpenGL context is also required before calling `mpv_render_context_free`. Failure to do that is the cause of issue #3593. With the addition of locking the OpenGL context in that case there is no longer a need for the  `uninitLock` lock. So the proposed fix eliminates that lock.

This Apple example from [CALayerEssentials](https://developer.apple.com/library/archive/samplecode/CALayerEssentials/Listings/ExampleCAOpenGLLayer_m.html#//apple_ref/doc/uid/DTS40008029-ExampleCAOpenGLLayer_m-DontLinkElementID_8) is inconsistent:
```swift
// 4)   [Required] Implement this message in order to actually draw anything.
//      Typically you will do the following when you recieve this message:
//      1. Draw your OpenGL content! (the current context has already been set)
//      2. call [super drawInContext:pixelFormat:forLayerTime:displayTime:] to finalize the layer content, or call glFlush().
//      NOTE: The viewport has already been set correctly by the time this message is sent, so you do not need to set it yourself.
//      The viewport is automatically updated whenever the layer is displayed (that is, when the -display message is sent).
//      This is arranged for when you send the -setNeedsDisplay message, or when the needsDisplayOnBoundsChange property is set to YES
//      and the layer's size changes.
-(void)drawInCGLContext:(CGLContextObj)glContext pixelFormat:(CGLPixelFormatObj)pixelFormat forLayerTime:(CFTimeInterval)timeInterval displayTime:(const CVTimeStamp *)timeStamp
{
    // Set the current context to the one given to us.
    CGLSetCurrentContext(glContext);
```

The header comment says:
> 1. Draw your OpenGL content! (the current context has already been set)

And then the code proceeds to set the context. Thanks Apple. I found sample code on the net that follow the documentation and did not lock and set the context. From debugging we know for sure  `CAOpenGLLayer._display`  is locking the context. If locking and setting is required then shouldn't `ViewLayer.canDraw` also be doing it?

So the second simplification in the proposed fix is to eliminate the locking and setting of the OpenGL context in the overridden `draw` method. Let me know if you want a more conservative fix that retains `uninitLock` and the locking and setting of the OpenGL context in that method.

lhc70000:
Are there any prerequisites for this PR, probably #3591? If not, I think we can merge this one first and let the users test; this is usually how we fix locking issues before.

low-batt:
No prerequisites for this PR. Good to merge. 

Merging this one may introduce merge conflicts with other PRs. I will be looking for conflicts in my outstanding PRs and fixing them to be compatible with the latest develop.

This is not the only issue with properly managing the OpenGL context. But this is the one users keep encountering.

lhc70000:
Thanks a lot!

PR #3582, opened 2021-12-18
RFC: Plugin: allow webcontent load http api

qinyuhang:

---

**Description:**
In plugin system, we often use apis in Overlay or StandaloneWindow, in this window we might load http contents or apis
For example retrieve information about LAN emby info.

Also I was wondering if set `NSAllowsArbitraryLoads` rather than `NSAllowsArbitraryLoadsInWebContent`.


lhc70000:
Sorry, I overlooked this PR.

I guess `NSAllowsArbitraryLoadsInWebContent` will suffice. I'm not fond of enabling an app-wise policy like `NSAllowsArbitraryLoads`. It may also create issues in the future, say if we want to submit IINA to AppStore.

I cherry-picked your commit to the plugin-system branch.

PR #3600, opened 2022-01-02
fix: playerCore emit fileLoaded param

qinyuhang:
The fileLoad event callback param here is NSURL, in plugin js, this object cannot use to get any info.
And the NSURL not corresponding d.ts ( says url is a string) https://github.com/iina/iina-plugin-definition/blob/master/iina/index.d.ts#L168

---

**Description:**


lhc70000:
Seems that it's already fixed by f478037. Thanks for all your PRs against the plugin system branch! You will be a contributor once the branch is merged.

Issue #3603, opened 2022-01-03
Plugin: main entry setTimeout/setInterval not working


qinyuhang:
**System and IINA version:**

- macOS 12.1 (21C52)
- IINA 1.2.0 (build from plugin-system branch)

**Expected behavior:**
In plugin entry js file, setTimeout or setInterval works.

etc:
```
// code block 1
setTimeout(function() {
  console.log(1)
}, 1000)
```

code block 1 should print 1 after 1 second

**Actual behavior:**

it dose nothing.

<details>
<summary><b>Crash report:</b></summary>

```
```
</details>

<br>

<details>
<summary><b>mpv log:</b></summary>

```
```
</details>

<br>

**Steps to reproduce:**
1 checkout to system-plugin branch
2 build and run
3 load a dev-plugin
4 use Safari dev tool to open an inspector of Main entry
5 past code block 1 into the inspector's `console` tab

- [ ] MPV does not have this problem.

**How often does this happen?**
Every time.

lhc70000:
See #3604.

PR #3604, opened 2022-01-03
Fix: setTimeout/setInterval not work in plugin entry

qinyuhang:
- [x ] It implements / fixes issue #3603.

---

**Description:**

This patch helps the main plugin entry get setTimeout and setInterval. Because the pluginInstance is in a sub thread,
we must run the RunLoop manually or register to RunLoop.main. This patch choose to run the RunLoop manually.
This patch also include cleanTimer code to prevent too many timer cause `self.timers` full of invalid timers takes too much memory.


lhc70000:
I tested on the current plugin branch, and `setTimeout` works well. It also released the timers correctly. Closing for now; feel free to reopen if you have other concerns.

PR #3618, opened 2022-01-20
Fix camera housing blocks controller, #3558

low-batt:

- Extends NSScreen with a new optional computed property,
  cameraHousingHeight

- Changes MainWindowController.legacyAnimateToFullscreen to check if
  cameraHousingHeight indicates a camera housing intrudes into the
  screen and if so reduces the size of the full screen video window's
  view's frame appropriately.

- [x] It implements / fixes issue #3558.

---

**Description:**
This PR replaces PR #3559 and takes a different approach addressing code review comments in the original PR.

### DIFF
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -600,3 +600,14 @@ extension NSAppearance {
     }
   }
 }
+
+extension NSScreen {
+  /// Height of the camera housing on this screen if this screen has an embedded camera.
+  var cameraHousingHeight: CGFloat? {
+    if #available(macOS 12.0, *) {
+      return safeAreaInsets.top == 0.0 ? nil : safeAreaInsets.top
+    } else {
+      return nil
+    }
+  }
+}
--- a/iina/Info.plist
+++ b/iina/Info.plist
@@ -1153,5 +1153,7 @@ Released under GPLv3.</string>
 			</dict>
 		</dict>
 	</array>
+	<key>NSPrefersDisplaySafeAreaCompatibilityMode</key>
+	<false/>
 </dict>
 </plist>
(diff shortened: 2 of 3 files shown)


lhc70000:
I can't test this fix since I don't have the corresponding device. It makes a lot of sense, but I want to ask how other video players (QuickTime, etc.) handle this issue? Do they also skip the notch area?

(Another solution would be adding some padding to the control bar at the top, but I'm concerned about whether it will look good aesthetically)

low-batt:
For many issues I post an analysis and discussion about the fix in the issue. There are lot of details about this fix there.

A bunch of my friends bought the new MacBook Pros. The peer pressure became overwhelming and I am now typing this on a MacBookPro18,2 with the camera housing. That is my new Mac in the picture attached to the issue. I can test issues related to the newest MacBook Pros.

The very top portion of the screen on these Macs where the camera housing intrudes is "extra" screen space. When you go into full screen in any application that top area of the screen is not used as long as the application calls the full screen method provided by AppKit. IINA does not intrude into this area of the screen if you are not using legacy full screen. The fix needs to make IINA work the same way as the official full screen when legacy full screen is enabled.

The entire screen is available to applications. So if you are doing your own full screen it is up to your application to avoid this area of the screen like AppKit does. Apple documents how to do this. It involves using a new safeAreaInserts property. Full details in the issue along with references to the Apple documentation.

As a workaround Apple added a special mode that can be set using "get info". This scales down the entire screen for the application. Undesirable but better than having part of the screen blocked. Apple recommends applications be updated to properly handle the camera housing and then disable this feature. So this fix also disables that workaround as recommended by Apple. More details in the issue.

I installed the latest VLC and confirmed that in full screen mode you can not move its controller into the top area of the screen.

PR #3645, opened 2022-02-17
Fix formatting issue in README.md

SamZhangQingChuan:

---

**Description:**

The code blocks were not being rendered in the expected way

### Before

image

### After

image


saagarjha:
Yeah, this seems to have broken recently. Thanks for spotting it and sending in a fix!

Issue #3620, opened 2022-01-21
Audio filters with same name cannot be removed.


timlenertz:
**System and IINA version:**

- macOS 12.1
- IINA 1.2.0 Build 129 (mpv 0.32.0)

**Expected behavior:**
Added two audio filters (both name 'scaletempo', different values). Then they can no longer be removed. When clicking the minus button, the overlay "Filter removed" appears, but the filter is not removed. It stays when reopening IINA.
The only way to remove it is to remove/modify the files in ~/Library/Application Support/com.colliderli.iina/watch_later.

**Actual behavior:**
Clicking the minus button in the audio filters list window should delete the selected audio filter.

<br>

**Steps to reproduce:**
- Open a video file in IINA.
- Add two audio filters name 'scaletempo', one with 'stride=8:overlap=1:search=10:scale=2:speed=pitch', the other with 'stride=8:overlap=1:search=10:scale=0.4:speed=pitch' (for example).
- Try to remove them again.

**How often does this happen?**
First try: both could not be removed. (maybe added more filters before)
Second try: one of the filters could not be removed.

low-batt:
### Analysis

Reproduced for me. This is a defect. This does not require adding multiple filters. IINA can have trouble removing even one filter.

Audio and video filters are retrieved from mpv in `MPVController.getFilters`. That method uses `MPVNode.parse` to convert the `mpv_node` returned by `mpv_get_property` into Swift objects. When converting the objects within the `mpv_node` the method parses a `MPV_FORMAT_NODE_MAP` object into a [Swift Dictionary](https://developer.apple.com/documentation/swift/dictionary). The order of key-value pairs in a Swift Dictionary is unpredictable, so this does not preserve the order of keys in the `MPV_FORMAT_NODE_MAP`. This should not be a problem since the documentation for `MPV_FORMAT_NODE_MAP` explicitly documents that _keys are in random order_.

Trouble is the mpv `af` and `vf` `remove` command when matching filters compares _argument order_. The order of arguments returned by `mpv_get_property` in the `MPV_FORMAT_NODE_MAP` does not appear to be random as indicated in the mpv documentation. In a test the keys matched the order of the arguments when the filter was added. But parsing the map into a Swift Dictionary is changing the order of the filter arguments. This causes the mpv `af` `remove` command to fail to find the filter.

### Fixing

[...]

PR #3665, opened 2022-03-15
Fix audio filters cannot be removed, #3620

low-batt:
This commit will:
- Add a new method removeFilter to MPVController that removes a filter
  based on its position within the list of filters
- Add methods for removing audio and video filters based on their position
  within the list of filters to PlayerCore
- Change FilterWindowController.removeFilterAction to use the new methods
- Add a parseRawParamString method to MPVFilter
- Add an isEqual method to MPVFilter
- Change  MPVFilter.init(rawString:) to properly handle filters with labels
- Add a debugDescription  property to  MPVFilter
- Change FilterWindowController.toggleSavedFilterAction to locate the saved
  filter using MPVFilter objects instead of strings
- Change FilterWindowController.reloadTable to locate saved filters using
  MPVFilter objects instead of strings
- Add a filtersAsString method to FilterWindowController
- Add a debugDescription property to SavedFilter

With these changes IINA removes filters based on their position within the
filter list. For saved filters IINA parses the parameters in the saved string
representation into a swift dictionary so the parameters of filters can be
compared by comparing dictionaries instead of using strings which can fail if
the order of parameters does not match.

- [x] It implements / fixes issue #3620.

---

**Description:**
This is a significant change. An intense mean detailed review of the PR is called for.

### DIFF
--- a/iina/FilterWindowController.swift
+++ b/iina/FilterWindowController.swift
@@ -87,13 +87,12 @@ class FilterWindowController: NSWindowController, NSWindowDelegate {
     filters = PlayerCore.active.mpv.getFilters(filterType)
     filterIsSaved = [Bool](repeatElement(false, count: filters.count))
     savedFilters.forEach { savedFilter in
-      savedFilter.isEnabled = false
-      for (index, filter) in filters.enumerated() {
-        if filter.stringFormat == savedFilter.filterString {
-          filterIsSaved[index] = true
-          savedFilter.isEnabled = true
-          break
-        }
+      if let asObject = MPVFilter(rawString: savedFilter.filterString),
+         let index = filters.firstIndex(of: asObject) {
+        savedFilter.isEnabled = true
+        filterIsSaved[index] = true
+      } else {
+        savedFilter.isEnabled = false
       }
     }
     currentFiltersTableView.reloadData()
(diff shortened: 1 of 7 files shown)


lhc70000:
Thanks for all the work and the detailed description. I almost completely forgot how the filters work and that helped a lot. If the only problem here is to find which filter to remove, I have another idea. Since it's possible to add a label to the filter, maybe we can:

- assign a unique label to each `SavedFilter` (also saved in the user defaults to make sure it never changes)
- add/remove filters based on their labels

What do you think?

low-batt:
I considered doing that, but I didn't go that way as it feels like "trespassing on data belonging to the user". Since mpv supports unlabeled filters it is desirable for IINA to support them as well. The user can use the mpv IPC interface to add and remove filters while IINA is running. Would be odd to add a filter that way and have IINA change it to have a label.

I'm not that happy with my proposed solution. I still find the mpv behavior of caring about order and yet using an unordered hash map inconsistent. But the mpv project has ruled on that, so I followed their recommendation as to how to delete filters.

What is most worrisome to me in the proposed solution is the parsing of the filters IINA saves as strings. Would be better to store the object form to avoid string parsing. That would require careful data migration. I did not look into that.

If you want to postpone fixing filters for this next release to dig into this some more, I'm good with that.

Issue #1168, opened 2017-10-25
Aspect ratio change with rotation doesn't works with custom shortcut


pighair:
**System and IINA version:**

- macOS 10.13
- IINA 0.0.14.1

**Expected behavior:**

Portrait and landscape toggles along "video-rotation +90" or -90 action in keyboard shortcut.

**Actual behavior:**

Window does not change on video rotation defined in keyboard shortcut.

**Steps to reproduce:**

1. Add a keyboard shortcut with action "add video-rotate 90".
2. Play any video which does not have square aspect ratio.
3. Use keyboard shortcut to rotate video.

**How often does this happen?**
Always.

pighair:
I found I can set keyboard shortcuts with multiple commands using ";".
So I tried to assign a keyboard with command below.

set video-rotation 90;set video-aspect "-1"

But iina set video aspect not considering rotation. Please refer following.

Steps:
- Open a 4:3 video
- Change aspect to 16:9
- Use the keyboard shortcut

Result:
- Video rotates 90 degrees and video is resized to 4:3

Expected behavior:
- Video rotates 90 degrees and video is resized to 3:4

lhc70000:
@pighair The multiple command syntax is not supported currently, but now that someone is trying to use it, I'll implement it in the next version.

pighair:
I meant multiple command syntax is working now even thouth UI does not handling it properly and it is not intended. Enhancement I want is to consider rotation status when modifying video’s aspect ratio.
Also I want iina maintains window scale when aspect ratio changes, which I might create another post.

gontthias:
mark

low-batt:
With the fix supplied by @mkyt in PR #3671 (THANKS!) rotation using the methods mentioned by @pighair initially work and then fail:
```text
17:50:53.983 [iina][w] mpv log: [autorotate] error: Video rotation with this format not supported
17:50:54.084 [iina][w] mpv log: [vo/libmpv] warn: Video is flagged as rotated by 359 degrees, but the video output does not support this.
```

The way to get this to completely work is to use this this Lua script: [cycle-video-rotate.lua](https://github.com/VideoPlayerCode/mpv-tools/blob/master/scripts/cycle-video-rotate.lua)

And bind keys to:
```test
script-message Cycle_Video_Rotate 90
script-message Cycle_Video_Rotate -90
```

This is one of the **unofficial 3rd party scripts** listed on the mpv Wiki [User Scripts](
[...]

PR #3671, opened 2022-03-23
Fix wrong aspect ratio when video rotation changed outside GUI, #1168

mkyt:
Observe `MPVOption.Video.videoRotate` to make sure `PlayerInfo.rotation` and `MPVOption.Video.videoRotate` are in sync, even if video rotation is changed outside IINA UI (e.g. commands emitted through hotkeys).

- [X] It implements / fixes issue #1168.

### DIFF
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -72,6 +72,7 @@ class MPVController: NSObject {
     MPVProperty.chapter: MPV_FORMAT_INT64,
     MPVOption.Video.deinterlace: MPV_FORMAT_FLAG,
     MPVOption.Video.hwdec: MPV_FORMAT_STRING,
+    MPVOption.Video.videoRotate: MPV_FORMAT_INT64,
     MPVOption.Audio.mute: MPV_FORMAT_FLAG,
     MPVOption.Audio.volume: MPV_FORMAT_DOUBLE,
     MPVOption.Audio.audioDelay: MPV_FORMAT_DOUBLE,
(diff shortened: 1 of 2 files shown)


iina/MPVController.swift
```diff
@@ -827,6 +828,9 @@ class MPVController: NSObject {
         player.sendOSD(.hwdec(player.info.hwdecEnabled))
       }
 
+    case MPVOption.Video.videoRotate:
+      player.info.rotation = Int(getInt(MPVOption.Video.videoRotate))
```
low-batt, line 832: This is getInt method that is being called:
```swift  
func getInt(_ name: String) -> Int {
    var data = Int64()
    mpv_get_property(mpv, name, MPV_FORMAT_INT64, &data)
    return Int(data)
  }
```
So I'm not understanding the need to nest the call to getInt inside "Int(...)"? This is the same pattern used in the other calls to `getInt`, so good to be consistent, no need to update this commit. Possibly something to eventually clean up. Or am I confused about Swift?

low-batt (approved):
Thumbs up from another IINA user. I tested these changed and confirmed they work.

lhc70000:
Thanks for the contribution!

I will merge it locally and remove the `Int` issue as pointed out by @low-batt.

lhc70000:
Merged via 09467a4054e238906ee2dcfabdc39c69f686b3a4.

lhc70000:
Sorry @mkyt, I realized that a squash merge will not list you as a contributor in the commit history. I replaced the previous commit by 4606359e2fad0907c7dd67da9fcd0c0984c93702 to keep your original commit.

Issue #3692, opened 2022-04-13
Custom key binding ignored


low-batt:
**System and IINA version:**

- macOS 12.3.1
- IINA 1.2.0

**Expected behavior:**
When a key is pressed IINA executes the commands assigned in a custom key binding.

**Actual behavior:**
IINA ignored the key binding.

<br>

**Steps to reproduce:**
- Open IINA / Preferences
- Click on `Key Bindings`
- Select `IINA Default`
- Click `Duplicate`
- Click `+` to add a new binding
- Add an additional binding for the `RIGHT` (arrow) key with the action `seek 60 exact`
- Add an additional binding for the `LEFT` (arrow) key with the action `seek -60 exact`
- This adds a **second** definition for these keys
- Restart IINA to insure new preferences are being used
- Play video
- Open `Playback` menu
- Notice that the menu item `Step Forward 5s` is still bound to the right arrow key
- Notice that the menu item `Step Backward 5s` is still bound to the left arrow key
- Test the right and left arrow keys
- New binding are ignored and IINA continues to seek by 5 seconds

- [ ] MPV does not have this problem.

This is a problem in IINA.

**How often does this happen?**
Every time.

I will post an analysis of the root cause and a proposed fix.

low-batt:
### Analysis

Custom IINA key bindings are stored in a mpv [input.conf](https://mpv.io/manual/master/#input-conf) configuration file. These files can be found in the directory `~/Library/Application\ Support/com.colliderli.iina/input_conf`:
```bash
low-batt@gag ~$ cd ~/Library/Application\ Support/com.colliderli.iina/input_conf/
low-batt@gag input_conf$ ls
issue-3692.conf
low-batt@gag input_conf$
```
Each line of an `input.conf` file maps a key to an input command. Neither IINA nor mpv enforce a requirement that the configuration file only contains one entry for a particular key even though the result of processing the configuration file results in a one to one binding between a key and an input command (ignoring mpv's ability to bind a sequence of keys). So what happens if the configuration file contains duplicate entries for a key? IINA should match mpv's behavior. I did not find this discussed in the mpv manual. This  `input.conf` file was tested with mpv:
```text
RIGHT seek 5
LEFT seek -5
RIGHT quit
LEFT quit
```
When the right arrow key was clicked mpv shutdown. The convention seems to be that if there are duplicates, last definition wins. What you would expect if each line of the configuration file is being processed and added to a unique key to input command mapping without checking for an existing entry.

[...]

PR #3693, opened 2022-04-13
Fix custom key binding ignored, #3692

low-batt:
- Add swift-algorithms to the list of packages
- Change MenuController.updateKeyEquivalentsFrom to ignore key bindings
  that have been overridden when searching for keys to be used as
  shortcuts for menu items

- [x] It implements / fixes issue #3692.

---

**Description:**
Note the addition of [Swift Algorithms](https://github.com/apple/swift-algorithms).

### DIFF
--- a/iina.xcodeproj/project.pbxproj
+++ b/iina.xcodeproj/project.pbxproj
@@ -15,6 +15,7 @@
 		49542986216C34950058F680 /* OpenInIINA.appex in Embed App Extensions */ = {isa = PBXBuildFile; fileRef = 49542973216C34950058F680 /* OpenInIINA.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
 		496B19921E2968530035AF10 /* PIP.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 496B19911E2968530035AF10 /* PIP.framework */; settings = {ATTRIBUTES = (Weak, ); }; };
 		519872FF26879B9B00F84BCC /* AccessibilityPreferences.swift in Sources */ = {isa = PBXBuildFile; fileRef = 519872FE26879B9B00F84BCC /* AccessibilityPreferences.swift */; };
+		51AE489128065AA90077378A /* Algorithms in Frameworks */ = {isa = PBXBuildFile; productRef = 51AE489028065AA90077378A /* Algorithms */; };
 		6100FF2B1EDF9806002CF0FB /* dsa_pub.pem in Resources */ = {isa = PBXBuildFile; fileRef = 6100FF2A1EDF9806002CF0FB /* dsa_pub.pem */; };
 		8400D5C41E17C6D2006785F5 /* AboutWindowController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8400D5C21E17C6D2006785F5 /* AboutWindowController.swift */; };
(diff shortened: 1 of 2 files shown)


lhc70000:
Please see my comments in #3692, I think that's a better fix. Would you like to update your PR or I push my proposed fix instead?

low-batt:
Caught me on a rare super busy day. Not sure how well the brain is working tonight. I say we toss my commit in the trash and go with your fix. **BUT** we should be coordinating with @svobs who has just proposed several fixes for key binding issues #3710,  #3716 and #3718.  See the PRs he posted. So lets close this PR.

svobs:
> Caught me on a rare super busy day. Not sure how well the brain is working tonight. I say we toss my commit in the trash and go with your fix. **BUT** we should be coordinating with @svobs who has just proposed several fixes for key binding issues #3710, #3716 and #3718. See the PRs he posted. So lets close this PR.

Hi @lhc70000! Let me just preface this by saying I'm a longtime Mac user, longtime developer, but very recent Mac developer. I'd tried several video players over the years and been kind of shocked at how mediocre they all were, and in fact had settled on bare-bones no-UI MPV for a long time before I discovered IINA :) Anyway, I'd like to help keep it going, and it's also a good chance to learn more about MPV integration.

So I understand the need to have a separate layer on top of MPV - must have been tons of work - but the hot reload of key bindings and smoother MacOS integration is excellent. Still, I get the sense that there are a small but very influential group of power users which it might be worth catering to in order to ensure IINA's ultimate success. 

So my presumption is: IINA should behave as much like MPV as possible, unless there's a good reason not to. It seems like it should be able to support Lua scripts and other advanced things but still have a nice user experience for the more casual user. I'm willing to work on a lot of this stuff.

So, sorry for being so long-winded - was kinda making the case for my other PRs here as well since I just went and did them without much discussion, so let me know what you think.

As for this one - @lhc70000 would it be possible to implement both fixes for this issue? I agree that bumping the range to 60sec would be an improvement. But larger values will still lead to a discrepancy between IINA and MPV, like this:

```
f seek 5 exact     # iina will use this
f seek 80 exact    # mpv will use this
```
Specifically, the discrepancy arises when there are two lines with the same keysym, and the first is inside the matching range but the second is outside of it. The first one becomes attached to a menu item while the second does not, and at keyDown a menu item is always higher priority than a non menu item.

In the PR for #3710 I thought I independently fixed the problem by picking the last binding which matched, instead of the first. But that only fixed it in the case where they both matched. I did some testing and there is a small merge conflict with @low-batt's fix but it should be easy to remedy.

lhc70000:
> would it be possible to implement both fixes for this issue?

Yes, of course, we don't want such a discrepancy. It's a bug. I will push a fix later to make sure the behavior is consistent with mpv as well as bumping the range (to make the shortcut displayed in the menu).

BTW,
> It seems like it should be able to support Lua scripts and other advanced things but still have a nice user experience for the more casual user.

IINA supports mpv lua scripts and mpv config files in Preferences - Advanced. It supports all mpv arguments in the bundled `iina-cli`. I want to minimize the discrepancy between IINA and mpv, so if you found one feature that could be implemented in IINA, feel free to report it (or even submit a PR)!

There's a Javascript-based plugin system on the `[plugin-system](https://github.com/iina/iina/tree/plugin-system)` branch. It includes all mpv features, and you can also add video overlays, displaying custom window and sidebar, add subtitle downloaders, and even manage multiple player instances. Hopefully we can release it in 1.4.0…

low-batt:
The fix in this PR is not correct. This PR has been replaced by PR #3728. Closing this one.

Issue #3695, opened 2022-04-20
Crash on open when previous position was on a now-disconnected display


qtist:
IINA generally lives on my secondary display. After using it in this state, when only one display is connected and I try to open a video, IINA will consistently crash every time (as if it tries to resume into a position that no longer exists). I must have done this 100 times now and keep forgetting to expect it. Recovery is simply opening the file again and it works as normal. 

- I'm using the "initial window size" setting, but not "initial window position"
- Also using "use physical resolution on retina displays"
- Machine is 14" M1 PRO
- Main and secondary displays are both 4K using 2x scaling. Internal display is always off

low-batt:
Hello from another IINA user,

When an application crashes macOS usually creates a crash report. If the crash happened recently the report may still be available. Open `/Applications/Utilities/Console`. Look for an IINA crash report. Post the entire report here.

With computers there is a vast amount of complexity going on "under the hood". The computer industry works hard to hide this from users. But when a failure occurs. it is frequently necessary to dig into all the behind the scenes details to fully understand and correct the root cause of the problem. With respect to crashes, the crash report is a critical piece of evidence and required for diagnosing the problem.

qtist:
Apologies, I hoped that someone who had worked in the area may have quickly recognised this as an unhandled case or something. Here is the latest crash report:

```{"app_name":"IINA","timestamp":"2022-04-20 13:23:34.00 +0100","app_version":"1.2.0","slice_uuid":"cb0ad903-780e-3761-97a3-a2cdd353ea16","build_version":"129","platform":1,"bundleID":"com.colliderli.iina","share_with_app_devs":0,"is_first_party":0,"bug_type":"309","os_version":"macOS 12.1 (21C52)","incident_id":"F1CAB947-5141-41DA-A27F-5D756A87ACC8","name":"IINA"}
{
  "uptime" : 60000,
  "procLaunch" : "2022-04-18 11:15:33.9174 +0100",
  "procRole" : "Foreground",
  "version" : 2,
  "userID" : 501,
  "deployVersion" : 210,
  "modelCode" : "MacBookPro18,3",
  "procStartAbsTime" : 89782995429,
  "coalitionID" : 1549,
  "osVersion" : {
    "train" : "macOS 12.1",
    "build" : "21C52",
    "releaseType" : "User"
  },
  "captureTime" : "2022-04-20 13:23:30.8612 +0100",
  "incident" : "F1CAB947-5141-41DA-A27F-5D756A87ACC8",
[...]

PR #3698, opened 2022-04-23
Fix crash on open when external display disconnected, #3695

low-batt:
- Add a new method selectDefaultScreen to MainWindowController that
  selects the screen to use as the default for a window
- Add a new method logScreen to MainWindowController for screen
  related log messages
- Change windowWillOpen to log information about screen selection
- Change windowWillOpen to use selectDefaultScreen

The new selectDefaultScreen method searches for a screen in this order:
 - window!.screen: The screen where most of the window is on; it is
   nil when the window is offscreen.
 - NSScreen.main: The screen containing the window that is currently
   receiving keyboard events.
 - NSScreeen.screens[0]: The primary screen of the user’s system.

If the screen associated with the window is nil this method will fall
back to using the main screen. Apple documentation is silent concerning
when NSScreen.main is nil.  If that is encountered the method will fall
back to using the primary screen.

- [x] It implements / fixes issue #3695.

---

**Description:**
See issue for full description of the problem and fix.

### DIFF
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -602,6 +602,7 @@ extension NSAppearance {
 }
 
 extension NSScreen {
+
   /// Height of the camera housing on this screen if this screen has an embedded camera.
   var cameraHousingHeight: CGFloat? {
     if #available(macOS 12.0, *) {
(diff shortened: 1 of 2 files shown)


lhc70000 (commented):
I'd prefer making these methods extensions. There're already too many methods in MainWindowController, so I think it's better to put these general utility functions somewhere else, like `Extensions.swift`.

```swift
extension NSScreen {
  func log() {
  }
}

extension NSWindow {
  func selectDefaultScreen() {
  }
}
```

low-batt:
Yes, MainWindowController too big.

I rebased, resolved merge conflicts and moved the methods to extensions.

Issue #3626, opened 2022-01-27
File Loop mode cannot be toggled off via menu command after it is turned on


vassudanagunta:
### System and IINA version:

- macOS 12.1
- IINA 1.2.0 build 129

### Expected behavior:

Selecting the File Loop menu item will toggle File Loop mode between `on` and `off` (with a ✔︎ appearing next to the menu item when `on`, see #3625)

### Actual behavior:

File Loop mode does not toggle off. One is forced either to restart IINA or figure out how to create a key binding to do so (very hard!).

### Workaround:

Create a key binding with the action: `cycle-values loop "no"`.  This reveals another bug: one cannot create a simple *toggle* keybinding. I have to create a separate `cycle-values loop "yes"`  A `cycle-values loop` binding does not work as expected.

low-batt:
Reproduced for me. This is a defect. 

IINA code is treating the mpv `loop-file` property as a boolean flag. It is not a boolean. That is the source of these problems. This is the definition of the property from the mpv manual:
> **--loop-file=<N|inf|no>, --loop=<N|inf|no>**
>  &nbsp;&nbsp;&nbsp;Loop a single file N times. inf means forever, no means normal playback.

@CarterLi has a fix for this. He is currently away. I will coordinate with him when he gets back.

vassudanagunta:
Thanks! A side note: you guys might want to consider using Issue Labels to mark issues that have been verified, was actually an mpv issue, etc. That way you could get a filtered list of the open 1.2k (‼️) issues.

low-batt:
I'm an IINA user that noticed the core developers were overloaded and have been helping by investigating problems and posting proposed solutions. I do not have the GitHub privileges required to add labels.

The core developers are currently reviewing and merging proposed fixes as a part of preparing the next release. That may still take a while as there are a lot of fixes outstanding and we do want to be careful and not send out a release with regressions in it.

I will discuss this issue with CarterLi. He just announced he is going to be unavailable for a couple of weeks, so don't expect a quick update. This issue is simple enough that it certainly should be fixed in this coming release in my opinion.

vassudanagunta:
Thanks for contributing to FOSS!

low-batt:
Fix has been merged into the develop branch.

low-batt:
Closing. Fixed in IINA 1.3.0.

PR #3701, opened 2022-04-24
Add HDR playback support

CarterLi:
- [x] This change has been discussed with the author.
- [x] It implements / fixes issue #3526.

---

**Description:**

* This PR contains all HDR related commits.

Several small changes were included too in order not to introduce merge conflicts when cherry-picking:

1. fix(pref): Don't load advanced settings when `enableAdvancedSettings` is disabled.
2. feat(pref): Support disabling ICC profile ( for SDR mode )
3. feat(perf): Support configuring thumb width

* UI related changes:

1. image
2. image
3. image
4. image

* TODO: Detect [maximum luminance supported by current monitor](https://mpv.io/manual/master/#options-target-peak) and [enable manual tone mapping](https://mpv.io/manual/master/#options-tone-mapping)

PR #3704, opened 2022-04-28
PlaylistView: Fix loop* issues

### LINKED ISSUE #3625: File Loop menu item does not show ✔︎ when toggled on
**System and IINA version:**

- macOS 12.1
- IINA 1.2.0 build 129

**Expected behavior:**

When File Loop playback mode is enabled, the `File Loop` menu item should have a ✔︎ to show that it is `on` -- just like `Playlist Loop` menu item.

**Actual behavior:**
No ✔︎ ever appears.

CarterLi:
- [x] It implements / fixes issue #3625 and #3626.

---

**Description:**

1. Store preference for playlist loop status
2. Make `loop file` actually work and add a `loop file` button

TODO: The style of the new `loop file` button in active mode doesn't
match other buttons in playlist view, which needs rework

Related PR: #3700

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,3 +109,4 @@ browser/Firefox_Open_In_IINA/*
 *.so
 
 *.xcarchive
+/.vscode
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -70,7 +70,7 @@ class MPVController: NSObject {
     MPVOption.TrackSelection.sid: MPV_FORMAT_INT64,
     MPVOption.Subtitles.secondarySid: MPV_FORMAT_INT64,
     MPVOption.PlaybackControl.pause: MPV_FORMAT_FLAG,
-    MPVOption.PlaybackControl.loopPlaylist: MPV_FORMAT_FLAG,
+    MPVOption.PlaybackControl.loopPlaylist: MPV_FORMAT_STRING,
     MPVProperty.chapter: MPV_FORMAT_INT64,
     MPVOption.Video.deinterlace: MPV_FORMAT_FLAG,
     MPVOption.Video.hwdec: MPV_FORMAT_STRING,
@@ -420,7 +420,7 @@ class MPVController: NSObject {
   }
 
   // MARK: - Command & property
-  
+
   private func makeCArgs(_ command: MPVCommand, _ args: [String?]) -> [String?] {
     if args.count > 0 && args.last == nil {
       Logger.fatal("Command do not need a nil suffix")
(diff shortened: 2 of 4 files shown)


CarterLi:
Ref: https://github.com/iina/iina/pull/3701#issuecomment-1111797805

low-batt:
Well I'm confused. It is definitely not working for me. I checked IINA+ and not working for me there either. As can be seen above both kinds of looping are enabled in the persisted user defaults, but when I start IINA and play a file, neither items are checked on the `Playback` menu. I hacked in the following code to `MPVController.mpvInit`:
```swift
setUserOption(PK.enableFileLoop, type: .other, forName: MPVOption.PlaybackControl.loopFile) { key in
  return Preference.bool(for: key) ? "inf" : "no"
}
setUserOption(PK.enablePlaylistLoop, type: .other, forName: MPVOption.PlaybackControl.loopPlaylist) { key in
  return Preference.bool(for: key) ? "inf" : "no"
}
```

And restarted IINA and then the menu items were both initially enabled. Maybe some changes were lost at some point during all the merging? Can you confirm it is still working for you?

The example of when preserving this setting is useful for video playback would be continuous playback for advertising, kiosk, demos, that kind of thing.

I am embarrassed to say that I am a multi-window user. Usually it is because I was in the middle of watching some instructional video and got interrupted. I know I can depend upon watch later to remember where I was, but I usually leave the window open so I don't forget to come back to it.

CarterLi:
> It is definitely not working for me

I find it strangely only works for music mode. Let me find what's going wrong...

CarterLi:
I removed preference related code. I'll just keep it in iina-plus

lhc70000:
I have two quick comments:

- I do hope we have vector images for all UI elements, and the icon should derive from the current file loop icon. I will try to draw a new one.
- The standard logic in a music player should be: we have only one button for looping, and clicking the button switches among "off", "playlist loop" and "file loop" mode. It makes no sense to have two buttons, with both playlist loop and file loop enabled.

lhc70000:
Hi @CarterLi, could you remove the changes to the button for now? I hope the playlist related fixes can be merged into 1.3.0, but we need to discuss more on the UI changes.

CarterLi:
> Hi @CarterLi, could you remove the changes to the button for now? I hope the playlist related fixes can be merged into 1.3.0, but we need to discuss more on the UI changes.

Done.

However I want to say that loop file is very useful for a music player. Please add it sometime

lhc70000:
Thanks. Sure!

Issue #579, opened 2017-04-17
Update "Date last opened" Finder metadata


patrickrushton:
When a file is opened, the "Date last opened" Finder data should be updated. At the moment, this only happens when the file is opened from Finder (e.g. double-click). It could also be updated when

- file is reached in a playlist
- file is dragged onto the player window
- file opened via File > Open dialog

fmaluse:
I think it makes sense that this data should be updated regardless of from where the file is opened. It would make it easier to keep track of which media has been viewed. I'd very much like to see this implemented!

IsaacInsoll:
I'd love this too

ccj242:
Me three. Would love this.

smoyn:
Implementing this would make the Finder an excellent video manager rather just a good one. I have not come across any player that implements this feature. You could be the first.

low-batt:
PR #3707 with support for this feature has been merged into the develop branch.

low-batt:
Closing. Fixed in IINA 1.3.0.

smoyn:
Like the new v3 build 132. There is bad news and good news:

The Bad:
Finder still NOT updating videos opened by Iina playlist mode on my machine running Mojave. Only videos manually opened from the Finder  update "date last opened" in Finder. Same as before the update. Playlist mode in Iina does work in Iina history but never updates the Finder.

The good:
"Date last opened" in Finder WILL work now for files opened in Iina playlist mode, IF the User has a Saved (spotlight) Search  placed in the sidebar. That is new, it didn't work in previous versions of Iina. 
The parameters of my search are simple: Kind is: any or movie, (any includes folders containing the movies). 
(Last Opened Date) is (within last X days weeks etc)

So the work around is good enough for me, still would be nice to see it in the Finder too. One other note is I started to use Iina History window as the controller for serialized files, and found it works as well as the Saved search results without having to switch back and forth between Iina and the Finder. The issue with Iina History window is the tiny ass font and no way to change it. All in all I'm happy with the way Iina works following the updates. Thanks

PR #3707, opened 2022-05-02
Update "data last opened" attribute on file start

lhc70000:
- [x] It implements / fixes issue #579.

---

**Description:**

Seems that we can only use the raw key `"com.apple.lastuseddate#PS"`. P and S are `XATTR_FLAG_NO_EXPORT` and `XATTR_FLAG_SYNCABLE` flags ([ref](https://opensource.apple.com/source/copyfile/copyfile-138/xattr_flags.c.auto.html)).

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -1052,6 +1052,23 @@ class PlayerCore: NSObject {
       URL(fileURLWithPath: path)
     info.isNetworkResource = !info.currentURL!.isFileURL
 
+    // set "date last opened" attribute
+    if let url = info.currentURL, url.isFileURL {
+      // the required data is a timespec struct
+      var ts = timespec()
+      let time = Date().timeIntervalSince1970
+      ts.tv_sec = Int(time)
+      ts.tv_nsec = Int(time.truncatingRemainder(dividingBy: 1) * 1_000_000_000)
+      let data = Data(bytesOf: ts)
+      // set the attribute; the key is undocumented
+      let name = "com.apple.lastuseddate#PS"
+      url.withUnsafeFileSystemRepresentation { fileSystemPath in
+        let _ = data.withUnsafeBytes {
+          setxattr(fileSystemPath, name, $0.baseAddress, data.count, 0, 0)
+        }
+      }
+    }
+
     if #available(OSX 10.13, *), RemoteCommandController.useSystemMediaControl {
       NowPlayingInfoManager.updateInfo(state: .playing, withTitle: true)
     }


low-batt (commented):
This will be a nice add!

I think the code needs to be updated to use the official key. The key is documented. The documentation follows industry best practices and is poor and hard to find especially since the column name in the Finder does not reflect the name of the attribute. This is the key in question: [kMDItemLastUsedDate](https://developer.apple.com/documentation/coreservices/kmditemlastuseddate)

I tested using that key:
```swift
      // set the attribute
      url.withUnsafeFileSystemRepresentation { fileSystemPath in
        let _ = data.withUnsafeBytes {
          let name = String(kMDItemLastUsedDate)
          let result = setxattr(fileSystemPath, name, $0.baseAddress, data.count, 0, 0)
          if result != 0 {
            let error = String(validatingUTF8: strerror(errno)) ?? "Unknown errno code \(errno)"
            Logger.log("Unable to update last opened date: \(error)")
          }
        }
      }

```

Before:
```bash
low-batt@gag pr3707$ mdls earth.mp4 | grep -m 1 kMDItemLastUsedDate
kMDItemLastUsedDate                    = 2022-05-01 19:18:16 +0000
```

After:
```bash
low-batt@gag pr3707$ mdls earth.mp4 | grep -m 1 kMDItemLastUsedDate
kMDItemLastUsedDate                    = 2022-05-02 19:06:45 +0000
```

And confirmed the Finder `Date Last Opened` column value matches the date seen in the terminal.

I'm unsure on whether the error logging is needed. This is what a failure looks like:
```text
15:05:45.765 [iina][d] Unable to update last opened date: Permission denied
```
That is what happens if the video file is readonly. I would expect that most users do not bother to protect their video files by marking them readonly, but a few might. And of course this will happen with readonly media, like optical disks.

lhc70000:
Seems that `mdls` reads/writes Spotlight Metadata Attributes. I don't know how they are stored in the system; some old
sources mentioned that they are stored as extended attributes, which is a filesystem-level feature. They claimed you can find these attributes using keys like `com.apple.metadata:kMDItemLastUsedDate`. But that's not (no longer?) true in my test.

Here are the behaviors I observed in my system:

- If I try to get the extended attributes through `xattr file` or `getxattr()`, it only returns these two entries:
  ```
  com.apple.lastuseddate#PS
  com.apple.macl
  ```
- (At least for a file with the `com.apple.lastuseddate#PS` attribute), setting `kMDItemLastUsedDate` through `setxattr()` won't change the date displayed in Finder.
- Setting `com.apple.lastuseddate#PS` will also change the `kMDItemLastUsedDate` displayed in `mdls`.

There's also no public API to modify Spotlight Metadata Attributes, so `com.apple.lastuseddate#PS` was the only thing I could do.

Could you try inspecting the xattr of the file after setting `kMDItemLastUsedDate` using `setxattr()`? I don't know why that worked for you, because even they are stored in the xattr, the key name should have some namespace prefix. The value of `kMDItemLastUsedDate` is just `"kMDItemLastUsedDate"`, and this theoretically shouldn't be the correct key.

low-batt:
I think this is definitive proof you can not trust this reviewer. And this reviewer needs to less trusting of internet posts. I remember checking the date last opened in the Finder and thinking it was updated. Easy to see today that it is still set to yesterday. I guess we are stuck with the hardcoded attribute string.

Issue #3715, opened 2022-05-04
Window not resized when screen changes in legacy full screen mode


low-batt:
**System and IINA version:**

- macOS 12.3.1
- IINA 1.2.0

**Expected behavior:**
When IINA is playing in legacy full screen mode on an external monitor and that monitor is disconnected when macOS moves the window to the internal monitor the window should be resized to fit the internal monitor.

**Actual behavior:**
The window is not resized.

**Steps to reproduce:**
- Plug in an external monitor
- Start IINA
- Enable `Use legacy full screen` is in preferences 
- Restart IINA
- Start playing a video
- Move the window to the external monitor
- Under the `Video` menu select `Enter Full Screen`
- Unplug the external monitor
- The IINA window moves to the internal monitor

At this point if the size of the external monitor differs from the internal monitor the IINA window is not resized to properly fit the internal monitor.

- [x] MPV does not have this problem.

Legacy full screen is an IINA feature.

**How often does this happen?**
Every time.

low-batt:
The commit in the pull request will:
- Add a method `setFrame` to `Extensions.NSScreen` to set a window's frame and content view frame to use the full screen
- Add call to `setFrame` in the observer for `didChangeScreenParametersNotification`

low-batt:
The fix in PR #3719 has been merged into the develop branch.

low-batt:
Closing. Fixed in IINA 1.3.0.

PR #3719, opened 2022-05-04
Fix window not resized when screen changes, #3715

low-batt:
The commit updates MainWindowController to resize the window
appropriately when the screen the window is on changes if in legacy
full screen mode.

- Add a method setFrame to Extensions.NSScreen to set a window's
  frame and content view frame to use the full screen
- Add call to setFrame in observer for
  didChangeScreenParametersNotification

- [x] It implements / fixes issue #3715.

---

**Description:**

### DIFF
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -630,22 +630,6 @@ extension NSScreen {
       Logger.log("\(label): visible frame \(screen.visibleFrame)")
     }
   }
-
-  /// Set the frame of the given window and its content view to use all the usable area of this screen.
-  /// - parameter window: The window to set the frame of.
-  /// - parameter displayFlag: Specifies whether the window redraws the views that need to be displayed. When true the
-  ///   window sends a [displayIfNeeded()](https://developer.apple.com/documentation/appkit/nswindow/1419096-displayifneeded)
-  ///   message down its view hierarchy, thus redrawing all views.
-  /// - parameter animateFlag: Specifies whether the window performs a smooth resize. true to perform the animation, whose
-  ///   duration is specified by [animationResizeTime(_:)](https://developer.apple.com/documentation/appkit/nswindow/1419655-animationresizetime)
-  func setFrame(_ window: NSWindow, display displayFlag: Bool, animate animateFlag: Bool) {
-    window.setFrame(frame, display: displayFlag, animate: animateFlag)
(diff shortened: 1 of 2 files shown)


iina/Extensions.swift
```diff
@@ -630,6 +630,22 @@ extension NSScreen {
...
+  /// - parameter animateFlag: Specifies whether the window performs a smooth resize. true to perform the animation, whose
+  ///   duration is specified by [animationResizeTime(_:)](https://developer.apple.com/documentation/appkit/nswindow/1419655-animationresizetime)
+  func setFrame(_ window: NSWindow, display displayFlag: Bool, animate animateFlag: Bool) {
+    window.setFrame(frame, display: displayFlag, animate: animateFlag)
+    guard let unusable = cameraHousingHeight  else { return }
+    // This screen contains an embedded camera. Shorten the height of the window's content view's
+    // frame to avoid having part of the window obscured by the camera housing.
+    let view = window.contentView!
+    view.setFrameSize(NSMakeSize(view.frame.width, frame.height - unusable))
+  }
```
lhc70000, line 648: Here are my thoughts:

- It's redundant to have both `screen` and `window`. The screen should always be `window.screen ?? NSScree.main`; it makes little sense to call this method on a screen different from where the window is.
- We don't see much potential to use this method somewhere else. This shouldn't be a general method; Apart from setting the window frame, it also changes the content view, which is not expected for a method named `setFrame`.
- So, since we only use it in this specific case (legacy screen), it's better to rename the method to reflect the actual use and remove the argument. (Do we need to override `display` or `animate` in the future? If not, we can remove them.)
- If we expect the target window to be the `MainWindow`, it's better to move it to `MainWindowController`.

Therefore, I suggest putting it in `MainWindowController` with name
```swift
setWindowFrameForLegacyFullScreen()
```

Alternatively, if write as an extension, it should be an `NSWindow` extension, and its name should reflect the purpose:
```swift
window.setFrameAndContentViewToFullUseableSize()
```

low-batt:
I agree with all of that and I'm rather embarrassed as to how bad the proposed commit was. See if I properly addressed the comments in the new commit.

In testing this I sometimes noticed a lot of `didChangeScreenParametersNotification` notifications. That appears to be due to that notification being posted when macOS swaps the GPU handling the screen. I did not find a way to tell macOS to not notify us of a GPU change.

Issue #3716, opened 2022-05-04
Add support for key sequence bindings


svobs:
While MPV's input config syntax supports several complex features that IINA does not, this seems like a low-hanging fruit.

_Originally posted by @varenc in https://github.com/iina/iina/issues/3692#issuecomment-1105888951_

For example, here's a wonderful/horrible hack in my input.conf that illustrates some of these issue:

```
d-d {mpv-only} add chapter 1;
d {mpv-only} ignore
d keypress "Meta+Alt+F12"
Meta+Alt+F12-Meta+Alt+F12 add chapter 1;
```

The goal here to bind the key sequence "d - d" to `add chapter 1` and have that **work in IINA or mpv**.

Notes:
- `d-d` means "press d, then press d again" and is supported by mpv, but not IINA
- IINA's input.conf parsing doesn't understand the `{mpv-only}` section marks so it just ignores those lines.
- The line with `Meta+Alt+F12-Meta+Alt+F12` is also ignored by IINA's parsing, since IINA doesn't understand sequences, but it is not ignored by mpv's input parsing, which also runs inside of IINA.
- I have things configured so the `{mpv-only}` section is only enabled when vanilla mpv is ran, not IINA. Those just exist to support the same behavior when I use the same input.conf with mpv.

This ridiculous hack makes `d-d` work in IINA by relying on the fact that behind the scenes IINA still runs the mpv input.conf parsing and keybinding, which supports sequences, but it just never fires them.   When IINA receives a keypress is just pairs that up with its input.conf parsing of what the command should be instead of relying on mpv's version. So when IINA sees `d` it fires a keypress at the mpv level by just calling `keypress "Meta+Alt+F12"`.  IINA doesn't support key sequences, but mpv does, so when `d` is pressed again, another `keypress "Meta+Alt+F12"` is called, which mpv then correctly interprets as the second press and runs `add chapter 1`.

I will say one nice advantage of the IINA approach is that you can swap out all of your keybindings without restarting. This is something mpv doesn't support.  Though that does put you in a weird state if you change your bindings because now if you run a `keypress ...` command that's a keypress at the mpv level, where the keybindings haven't been updated, so that'll run the old command binding and not the updated bindings at the IINA level.  Puts you in a state where pressing "x" and firing running the command `keypress x` could do totally different things.

Anyway, that was a lot of text. Just wanted to brain dump some thoughts/observations about IINA's keybinding treatment here since it seems relevant.  Cheers!

svobs:
[...]

Issue #2140, opened 2018-12-25
MPV scripts: can't setup key binding


shaddysignal:
I was going to try mpv-webm (https://github.com/ekisu/mpv-webm) lua script since IINA probably should support this, but it only kind of works. If I add key binding in the settings with `W script-binding webm/display-webm-encoder`, I can get script to start at least. But then the script setup the binding for themselves with `self/command` so I can't even make custom binding myself to call it.
I assume it have something to do with changes in #569. But I don't know whether it worked back then.
I understand that crop feature probably will not work right anyway, but I'm interested in plans for scripts support.

**System and IINA version:**

- macOS version 10.13.6
- IINA 1.0.0-rc build 95

**Expected behavior:**
I expected it to work as described in script readme.

**Actual behavior:**
Pressing the right combination doesn't do anything.

Putting it there to show that the script is actually loading.
<details>
<summary><b>mpv log:</b></summary>

```
[   0.075][d][webm] Loading lua script /Users/foo/.config/mpv/scripts/webm.lua...
...
[   0.089][d][cplayer] Run command: define-section, flags=0, args=[input_webm, W script-binding webm/display-webm-encoder
[   0.089][d][cplayer] , default]
...
[   0.089][d][cplayer] Run command: enable-section, flags=0, args=[input_webm, allow-hide-cursor+allow-vo-dragging]
...
[   0.089][d][cplayer] Run command: define-section, flags=0, args=[input_forced_webm, , force]
...
[   0.090][d][cplayer] Run command: enable-section, flags=0, args=[input_forced_webm, allow-hide-cursor+allow-vo-dragging]
```
</details>

<br>

**Steps to reproduce:**
Enable `use config directory`.
Put web.lua in scripts directory there.
Open IINA and try to use it as described.

- [x] MPV does not have this problem.
I tested it with 0.29.1.

**How often does this happen?**
Every time.

shaddysignal:
> But then the script setup the binding for themselves with `self/command` so I can't even make custom binding myself to call it.

Apparently I can, somehow I didn't think about checking logs to see how exactly does it put new bindings. Issue still stands until further comment. It just now has a somewhat workaround.

low-batt:
This issue is about support for this mpv Lua scripting function:

> **mp.add_key_binding(key, name|fn [,fn [,flags]])**
>
>  Register callback to be run on a key binding. 

[...]

PR #3722, opened 2022-05-05
Add support for key sequences via 4-item buffer

svobs:
- [x] It implements / fixes issue #3716.

---

**Description:**
Added a simple circular buffer which tracks the last 4 pressed keys/key-combos and sends the shortest sequence which matches, mimicking MPV's own logic.

### DIFF
--- a/iina/PlayerWindowController.swift
+++ b/iina/PlayerWindowController.swift
@@ -11,7 +11,7 @@ import Cocoa
 class PlayerWindowController: NSWindowController, NSWindowDelegate {
 
   unowned var player: PlayerCore
-  
+
   var videoView: VideoView {
     fatalError("Subclass must implement")
   }
@@ -254,10 +254,15 @@ class PlayerWindowController: NSWindowController, NSWindowDelegate {
   }
 
   override func keyDown(with event: NSEvent) {
-    let keyCode = KeyCodeHelper.mpvKeyCode(from: event)
-    if let kb = PlayerCore.keyBindings[keyCode] {
-      handleKeyBinding(kb)
+    if let keyBinding = player.keyInputController.resolveKeyEvent(event) {
+      if !keyBinding.isIgnored {  // if "ignore", do nothing. No beep, no send
+        if !handleKeyBinding(keyBinding) {
+          // beep if cmd failed
+          super.keyDown(with: event)
+        }
+      }
     } else {
+      // invalid key
       super.keyDown(with: event)
     }
   }
--- a/iina.xcodeproj/project.pbxproj
+++ b/iina.xcodeproj/project.pbxproj
@@ -158,6 +158,9 @@
 		B4E446F725CB54EF0069F06E /* Mustache in Frameworks */ = {isa = PBXBuildFile; productRef = B4E446F625CB54EF0069F06E /* Mustache */; };
 		B4E4470125CE3F930069F06E /* Sparkle in Frameworks */ = {isa = PBXBuildFile; productRef = B4E4470025CE3F930069F06E /* Sparkle */; };
 		C789872F1E34EF170005769F /* InfoPlist.strings in Resources */ = {isa = PBXBuildFile; fileRef = C78987311E34EF170005769F /* InfoPlist.strings */; };
+		D18333BC2834CDF40060A4E9 /* RingBuffer.swift in Sources */ = {isa = PBXBuildFile; fileRef = D18333BB2834CDF40060A4E9 /* RingBuffer.swift */; };
+		D18333BE2834CE620060A4E9 /* KeyInputController.swift in Sources */ = {isa = PBXBuildFile; fileRef = D18333BD2834CE620060A4E9 /* KeyInputController.swift */; };
+		D19E23E9284F0A8400FD6999 /* IINAApplication.swift in Sources */ = {isa = PBXBuildFile; fileRef = D19E23E8284F0A8400FD6999 /* IINAApplication.swift */; };
 		E301EFDA21312AB300BC8588 /* KeychainAccess.swift in Sources */ = {isa = PBXBuildFile; fileRef = E301EFD921312AB300BC8588 /* KeychainAccess.swift */; };
 		E322A4F820A8442E00C67D32 /* PlaylistPlaybackProgressView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E322A4F720A8442E00C67D32 /* PlaylistPlaybackProgressView.swift */; };
 		E32712B024EAA14500359DAB /* ScreenshootOSDView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E32712AE24EAA14500359DAB /* ScreenshootOSDView.swift */; };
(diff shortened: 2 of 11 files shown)


iina/PlayerWindowController.swift
```diff
@@ -255,11 +258,51 @@ class PlayerWindowController: NSWindowController, NSWindowDelegate {
...
+      Logger.log("KeyDown: \(keyCode)", level: .verbose)
+      #endif
+      if matchAndHandle(keyCode) {
+        return
+      }
+
+      // try to match key sequences, up to 4 values. shortest match wins
+      var keyCodeSequence = keyCode
+      for i in 0..<3 {
+        let prevKeyCode = lastKeysPressed[(lastKeyPressedIndex+4-i)%4]
```
low-batt, line 270: Remember I'm the newbie. Might want to wait and see what @lhc70000 thinks...

Instead of implementing our own ring buffer, maybe it would be easier to understand if the code used a Deque from swift-collections. Something like:
```swift
internal var lastKeysPressed = Deque<String>()

private func rememberKeyPress(_ keyCode: String) {
  if lastKeysPressed.count == 4 {
    lastKeysPressed.removeLast()
  }
  lastKeysPressed.prepend(keyCode)
}
```
And:
```swift
 rememberKeyPress("a")
 rememberKeyPress("b")
 var keyCodeSequence = "c"
 var iterator = lastKeysPressed.makeIterator()
 while let prevKeyCode = iterator.next() {
   keyCodeSequence = "\(prevKeyCode)-\(keyCodeSequence)"
   print(keyCodeSequence)
 }
```
Something along those lines...

iina/PlayerWindowController.swift
```diff
@@ -255,11 +258,51 @@ class PlayerWindowController: NSWindowController, NSWindowDelegate {
...
-    if let kb = PlayerCore.keyBindings[keyCode] {
+    if keyCode != "" {
+      #if DEBUG
+      Logger.log("KeyDown: \(keyCode)", level: .verbose)
+      #endif
+      if matchAndHandle(keyCode) {
+        return
+      }
+
+      // try to match key sequences, up to 4 values. shortest match wins
```
low-batt, line 267: A developer who is unaware of mpv's support of key sequences may be very confused by this code as we are not adding full support for this feature. If you go to Preferences / Key Bindings and try and add a key sequence you can't, yes? A developer thinking about the IINA key binding UI may wonder what this is about. Might want to indicate this is for users that edit their input.conf file by hand to take advantage of mpv features not yet supported by the IINA key binding UI.

low-batt (commented):
I pulled the PR locally and tested it and it worked.

It differs from the mpv behavior in a slighly annoying way. When I typed the first two characters in a three character sequence in mpv it was quiet. With IINA each key press of the first two characters emits the "bad operation" noise. Not sure what Apple calls that. As if the key is not accepted. That didn't seem right.

See what @lhc70000 thinks.

iina/PlayerWindowController.swift
saagarjha: I agree with the concern, although perhaps not the implementation. What we really want here is a fixed-size buffer, which your array is, but inlining the details here isn't great. @svobs, would you be up for writing a RingBuffer Collection and sticking it in our utilities folder, so we can reuse it for other things too?

low-batt: A RingBuffer Collection in utilities like @saagarjha suggests sounds good to me.

svobs:
> A developer who is unaware of mpv's support of key sequences may be very confused by this code as we are not adding full support for this feature. If you go to Preferences / Key Bindings and try and add a key sequence you can't, yes? A developer thinking about the IINA key binding UI may wonder what this is about. Might want to indicate this is for users that edit their input.conf file by hand to take advantage of mpv features not yet supported by the IINA key binding UI.

You actually can, sort of, in a not-very-pretty way. In the Key Bindings window you can check "Display raw values" and then double-click the Key column for an existing key binding. There's no way to record a new key sequence directly though - you'd have to click + and record some dummy keybinding using the dialog, then edit it in the table afterwards.

> Instead of implementing our own ring buffer, maybe it would be easier to understand if the code used a Deque from swift-collections

Yeah I would love to just use a Deque, but I was trying to avoid adding a dependency on a new library since that potentially can come with a lot of baggage.

A little story - I'll admit that I'm fairly new to Swift (last 12 months), and had added swift-collections to a project. It was working well for a while, but I ran into problems when trying to create a universal binary - apparently their team hadn't released a build for Apple Silicon. I was left having to choose between making my own custom build of it, or migrating to another solution. I ended up doing the latter - copying & pasting a LinkedList class I found from another freeware project and using that. It's kind of mind-boggling that Swift doesn't include these standard data structures with Foundation. 

> I agree with the concern, although perhaps not the implementation. What we really want here is a fixed-size buffer, which your array is, but inlining the details here isn't great. @svobs, would you be up for writing a RingBuffer Collection and sticking it in our utilities folder, so we can reuse it for other things too?

Sure, I will work on those changes and notify when done

>It differs from the mpv behavior in a slighly annoying way. When I typed the first two characters in a three character sequence in mpv it was quiet. With IINA each key press of the first two characters emits the "bad operation" noise. Not sure what Apple calls that. As if the key is not accepted. That didn't seem right.

That is a good call - I hadn't noticed that until now. That happens when super.keyDown() is called and (best guess) it goes through the responder chain until the end which defaults to a system beep. So we don't call super.keyDown() if we recognize the key.

I'll see if I can fix that. I think the solution will be to parse the sequences in the keybindings beforehand and keep a separate set of their subsequences ready. So e.g., if the sequence is "d-d-d", then "d" and "d-d" have no beep even if doing nothing, but "s" and "d-s" will beep (assuming they have no other bindings). We'd only check against this set on keyDown after no other bindings matched. Functionally, they would be equivalent to an "ignore", but we'd need to keep it as a separate set from the player's primary keybindings dict because we wouldn't want it to create any side effects in other functionality.

low-batt:
I too am new to Swift. I think you are ahead of me by a month. Think newbie when you read my comments. I do not speak with authority. Always feel free to question my comments.

In this case I believe Saagar would prefer not to add a dependency on swift-collections, so no need to use Deque. It doesn't match what is needed here as it doesn't have an option to limit the size and perform like a ring buffer. The desire is to separate out the generic rung buffer operations.

And yes, it is calling super.keyDown() that emits the noise.

svobs:
Just pushed a big update to this PR based on feedback. This turned out to be bigger than I originally envisioned. Some possible conversation points:

1. Added a new class, `KeyInputController`, which has both a singleton/static part (for generating partial sequences with which to identify whether an otherwise non-bound key should beep or not) and a non-static part - each instance of `KeyInputController` is associated with a single `PlayerCore` instance. This instance currently only keeps track of the last 4 keys typed, but I have plans for later keeping track of other things such as dynamic key bindings from Lua scripts.
2. Added a new class, `RingBuffer`, to the `Utils` group as suggested.
3. Added a modulo operator: `%%` to `Extensions.swift`. This was needed to fix bugs in `RingBuffer` - I did not realize that Swift's `%` is the "remainder" operator, not the modulo!
4. Unit tests! I created a new target called `IINA-Tests`. I needed to write some and there was no other place to put them. Was going to attempt to write a UI functional test but ran out of steam / time; maybe at some point. Feedback appreciated.

svobs:
Update: Not quite finished. There is some more work to do here...

I found a case which can result in incorrect behavior, if the user pressed a menu item-bound key in the middle of a sequence.

### Test case:
Say a user has only these key bindings:
```
a-b-c  playlist next
c      quit
p      cycle pause
```

IINA will recognize `c` and `p`  and associate them to its menu items. This means that when these two keys are pressed, their menu item action will be immediately called, with no `keyDown` notification sent to the player window. As we'll see, this messes up the state tracking.

Now the user has a window focused and video playing in it, and types:
```
a
b
p
c
```

**What should happen:**
1. Video is paused
2. IINA quits

**What actually happens:**
1. Video is paused
2. Next video in playlist

The reason is that `KeyInputController` has no idea that `p` had been pressed, because no `keyDown` event was sent. If it had, it would have cleared its typed key buffer.

### Proposed solution:
The only robust solution I can think of is to refactor `MenuItemController` so that instead of setting each action directly to its selector, each action calls into a new central method with takes the selector as an argument. That way, the central method can act as a decorator and add extra functionality (in this case notifying the `KeyInputController` of the typed key) before actually calling the action.

svobs:
After many iterations and much head-banging, I finally got this 👍 

I found some sample code which showed me how to subclass `NSApplication` to intercept the key events at the source, which makes everything so much easier, although I doubt there was no other way to do it. I named that subclass `IINAApplication`. Right now it mostly leaves things as they were before - `MainMenuController` still is first responder, but the new class explicitly sends `keyDown` events to it  so that it can find out if they were handled and then notify `KeyInputController`.

I also created a bunch of automated UI tests to make sure this worked, but I pulled out all the testing code and put it into PR #3787 (`KeySequences_UITests.swift`).  Thought that was kind of a clever solution too - I mapped different key bindings to `set display-scale` commands of different sizes and then tested to see if the size of the player window changed after each keypress.

Didn't intend for this PR to get so much more complicated, but it was a big hurdle to clear on the way to my long-term goal - to tackle issue #2140 :)

svobs:
Well, here goes again...

I discovered that (a) my solution which involved subclassing `NSApplication` ended up breaking text input fields in the Quick Settings View and other window controls which use the keyboard. However - (b) I have found a much better and more direct solution. All that was needed was a change to [`MainWindow.swift`](https://github.com/iina/iina/pull/3847/commits/df2c6f4e14995a60c408917ab446b82ec6348720#diff-e93c2b9f08c8ad3bb4cd2300c9c26a3fa8fdc997b2d3ef3bb287240b03c22bec):

Previous code:
```swift

  // this method is called for either ESC or command-period
  override func cancelOperation(_ sender: Any?) {
    let controller = windowController as! MainWindowController
    if let kb = PlayerCore.keyBindings["ESC"] {
      controller.handleKeyBinding(kb)
    } else {
      super.cancelOperation(sender)
  }
```

Proposed change:
```swift
  override func keyDown(with event: NSEvent) {
    // Forward all key events which the window receives to controller. This fixes:
    // (a) ESC key not otherwise sent to window
    // (b) window was not getting a chance to respond before main menu
    if let controller = windowController as? MainWindowController {
      controller.keyDown(with: event)
    }
  }
```

The problem apparently was that `NSWindow` was the first responder the whole time, but just not passing them on to the window controller (which is where our keyDown listener is)!

So the player window was already earlier in the responder chain than the menubar, which is what we wanted. It's just that the when the player window was sending them back unhandled, the menubar was given a chance to respond.

It makes sense that the menubar should always be the last responder - it's the least focused. But apparently by default, window controllers are even further down the responder chain, and in some cases not even part of the chain for the `ESC` and `DEL` keys. I don't see why they should even be included in the chain by default - seems like doing so just adds to confusion.

I'm going to close this PR as #3847 now supercedes this. I'm going to focus my testing there.

Issue #3718, opened 2022-05-04
The mpv default key bindings set is missing keys mpv added


low-batt:
**System and IINA version:**

- macOS 12.3.1
- IINA 1.2.0

**Expected behavior:**
The `mpv Default` key bindings found in `Preferences` / `Key Bindings` includes all the mpv key assignments appropriate for IINA.

**Actual behavior:**
The key bindings for showing statistics are missing. From the [mpv manual](https://mpv.io/manual/stable/#keyboard-control):
> i and I
    Show/toggle an overlay displaying statistics about the currently playing file such as codec, framerate, number of dropped frames and so on. See [STATS](https://mpv.io/manual/stable/#stats) for more information.
	
The mpv key bindings should be reviewed to see if any other bindings are missing from the set.

**Steps to reproduce:**
Click on the `mpv Default` key bindings in `Preferences` and it is celar i and I are missing.

- [x] MPV does not have this problem.

**How often does this happen?**
Every time.

low-batt:
Fix in PR #3727 has been merged into the develop branch.

low-batt:
Closing. Fixed in IINA 1.3.0.

PR #3723, opened 2022-05-06
Update mpv bindings

svobs:
- [x] It implements / fixes issue #3718.

---

**Description:**
Did a manual merge of mpv 0.34's default input.conf. Lotsa updates. Also added missing commands to MPVCommand.swift, I suppose just for the sake of staying current, even if they are not used for much.


lhc70000:
Thanks for the PR!

We have a script `other/parse_doc.rb` to crawl the latest mpv document and generate related files like `MPVCommand.swift`. We usually do that when upgrading libmpv. If you edited `MPVCommand.swift` manually, could you discard the changes for now? We expect a 1.3.0 release soon, and I will update libmpv after the changes are finalized.

svobs:
Ah, good to know! OK, I will close this PR and resubmit with only the input.conf changes.

PR #3728, opened 2022-05-07
Fix inconsistency with mpv when encountered duplicated keybindings

lhc70000:
- [x] It implements / fixes issue #3692.

---

Replacement of #3693.


low-batt (approved):
I reviewed the changes. I also pulled the PR locally and tested the case reported in #3692 and confirmed it works. I also tested changing the key to seek 90 seconds and confirmed the menu item is no longer bound to a shortcut key as expected. Changes look good to me.

PR #3730, opened 2022-05-09
Expand "Quit after all windows are closed" into popup, add "Show Welcome window" option

svobs:
- [x] It implements / fixes issue #3797.

---

**Description:**
I really wanted IINA to show the Welcome window again after I closed my last video window. While it was possible to leave it open, I'd have to click on the IINA icon again to bring it up. The obvious change to make seemed to be just changing the "Quit after all windows are closed" checkbox into a popup with 3 options, and default to "Do nothing", to match the existing behavior.
Screen Shot 2022-05-09 at 01 02 50
I just went ahead and coded it. It's not super complicated and most of the changes were in the XIBs, getting things to line up properly. I managed to remove a bunch of warnings in the process.

Hope this can make it into the official branch at some point! And of course I'm open to making changes to it as well.

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -306,21 +306,16 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
   @objc
   func checkForShowingInitialWindow() {
     if !openFileCalled {
-      showWelcomeWindow()
+      showWindowAfterLaunch()
     }
   }
 
-  private func showWelcomeWindow(checkingForUpdatedData: Bool = false) {
+  private func showWindowAfterLaunch(checkingForUpdatedData: Bool = false) {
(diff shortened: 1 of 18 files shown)


svobs:
Re-developed from more recent code, making the assumption that e2b1495a3aac18233093436d26aaabbcdc6dc240 from PR #4107 will be merged first and therefore its XIB can be added to (let's live dangerously!)

Added a new func to `AppDelegate` to avoid a crash:
  ```Swift
func terminateSafely() {
    // Prevent infinite loop
    guard !isTerminating else { return }

    NSApp.terminate(nil)
  }
```

Where possible, it might be a good idea to use APIs owned by IINA rather than AppKit's, to have more control.

svobs:
Found some regressions in this new code, but getting hard to merge. Will close this and work off of #4164 from here on.

Issue #3717, opened 2022-05-04
Preference panes keep shifting vertically


svobs:
Go to IINA > Preferences..., then try clicking on different sections on the left. Watch the vertical alignment of the pane on the right relative to the Search field. About half the time, the first row in the pane is level with the Search field, and half the time it's below it. See pictures below

- macOS 12.3.1
- IINA 1.2.0
- 2019 16" Intel Macbook Pro

**Expected behavior:**
First row should be roughly level with Search field.

**Actual behavior:**
Sometimes correct, sometimes too low, seemingly random

![1_General_Bad](https://user-images.githubusercontent.com/2213815/166633867-bb2da1cf-2354-4134-8780-150361395543.png)
![2_UI_Good](https://user-images.githubusercontent.com/2213815/166633869-152ae36d-a626-4040-8d37-24029298a96d.png)
![3_VideoAudio_bad](https://user-images.githubusercontent.com/2213815/166633870-5ef03203-409a-4ec9-b09a-a63b2edf07eb.png)
![4_General_Good](https://user-images.githubusercontent.com/2213815/166633872-2f8917b9-f035-44e3-8b9f-ce21a7312f46.png)

low-batt:
This is fixed in the released version of IINA.

PR #3731, opened 2022-05-09
Fix inconsistent vertical offset in Prefs Window panels: remove unnec…

svobs:
- [x] It implements / fixes issue #3717.

---

**Description:**
Fix inconsistent vertical offset in Prefs Window panels: removed an unnecessary and unreliable scroll(.zero) request. This resulted in a consistent vertical offset which was too large and easily reduced. The pref panes should always be fixed size and not scrollable, because their superview is an NSScrollView which will provide the scroll. Also, they do not need an explicit scroll() because their superview will always have an implicit scroll of zero when they are redrawn in it.

### DIFF
--- a/iina/PrefAdvancedViewController.swift
+++ b/iina/PrefAdvancedViewController.swift
@@ -26,6 +26,10 @@ class PrefAdvancedViewController: PreferenceViewController, PreferenceWindowEmbe
     return NSImage(named: NSImage.Name("pref_advanced"))!
   }
 
+  var preferenceContentIsScrollable: Bool {
+    return false
+  }
+
   var hasResizableWidth: Bool = false
 
   var options: [[String]] = []
--- a/iina/PreferenceWindowController.swift
+++ b/iina/PreferenceWindowController.swift
@@ -160,7 +160,7 @@ class PreferenceWindowController: NSWindowController {
     completionTableView.delegate = self
     completionTableView.dataSource = self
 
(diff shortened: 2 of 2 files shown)


low-batt (approved):
I reproduced the bad behavior with a build of the develop branch. I fetched the PR locally and confirmed the changes correct the issue with the vertical offset. Changes look good to me.

lhc70000:
This is an old problem that we do want to fix.

However, I think the `preferenceContentIsScrollable` property was designed for the key bindings pane - we want the table always fill the whole super view and resize with the window. This PR breaks the key bindings pane's layout. Is it possible to keep the current behavior but only remove the extra space at the top?

svobs:
> However, I think the preferenceContentIsScrollable property was designed for the key bindings pane - we want the table always fill the whole super view and resize with the window. This PR breaks the key bindings pane's layout. Is it possible to keep the current behavior but only remove the extra space at the top?

Gotcha. I see that now...

I'll try to find time this weekend to see what can be done.

svobs:
I think I fixed it. I ended up leaving in the `preferenceContentIsScrollable` logic but took out the problematic calls to change `scrollView`, and just adjusted the vertical offsets for a couple of constraints.

I also noticed that the Advanced page could benefit from having its table expand down with the window as well, so figured better to beg forgiveness than ask for permission? 😇 It's the single reason why `PrefAdvancedViewController.swift` has changes.

Issue #3710, opened 2022-05-02
Keybindings for `seek exact` are silently overridden to `seek keyframes`


svobs:
I was playing with building IINA myself, and was tailing the MPV log, and discovered some issues with seek keybindings.

In IINA's Key Bindings config, I've got the L key bound to "seek 10 exact", and the J key bound to "seek -10 exact". But looking at the commands MPV is logging:

```
tail -f /Users/msvoboda/Library/Logs/com.colliderli.iina/2022-05-02-12-49-20_TLOoov/mpv.log | grep -A 1 'Run command: seek'
[ 100.964][d][cplayer] Run command: seek, flags=64, args=["-5.000000", "relative", "unused"]
[ 100.964][v][mkv] queuing seek to 15.854000
--
[ 115.514][d][cplayer] Run command: seek, flags=64, args=["5.000000", "relative", "unused"]
[ 115.514][v][mkv] queuing seek to 15.427000
--
[ 117.548][d][cplayer] Run command: seek, flags=64, args=["5.000000", "relative", "unused"]
[ 117.548][v][mkv] queuing seek to 25.854000
--
[ 119.141][d][cplayer] Run command: seek, flags=64, args=["5.000000", "relative", "unused"]
[ 119.141][v][mkv] queuing seek to 36.281000
```

I noticed that to some extent this is a feature: I see "Step Forward 10s" and "Step Backward 10s" with my key bindings listed in the Playback menu. But the MPV logs were not lying, and I managed to trace part of the problem to hard-coded values in **MainMenuActions.swift**:

```
  @objc func menuStep(_ sender: NSMenuItem) {
    if sender.tag == 0 { // -> 5s
      player.seek(relativeSecond: 5, option: .relative)
    } else if sender.tag == 1 { // <- 5s
      player.seek(relativeSecond: -5, option: .relative)
    }
  }
```

In addition, there appear to be several menu localizations which have the 5s value hard-coded. For example, MainMenu.strings (it):
```
/* Class = "NSMenuItem"; title = "Step Forward 5s"; ObjectID = "6yE-op-DTS"; */
"6yE-op-DTS.title" = "Avanti 5s";
```

[...]

PR #3732, opened 2022-05-10
Fix: honor "exact" seek in user bindings

svobs:
- [x] It implements / fixes issue #3710.

---

**Description:**
I rewrote the fix to this issue to minimally impact the existing code and make it more robust (e.g. using ? instead of !)

### DIFF
--- a/iina/MainMenuActions.swift
+++ b/iina/MainMenuActions.swift
@@ -90,11 +90,10 @@ extension MainMenuActionHandler {
   }
 
   @objc func menuStep(_ sender: NSMenuItem) {
-    let seconds = Double(abs((sender.representedObject as? Int) ?? 5))
-    if sender.tag == 0 { // -> 5s
-      player.seek(relativeSecond: seconds, option: .relative)
-    } else if sender.tag == 1 { // <- 5s
-      player.seek(relativeSecond: -seconds, option: .relative)
+    if let args = sender.representedObject as? (Double, Preference.SeekOption) {
+      player.seek(relativeSecond: args.0, option: args.1)
+    } else {
+      player.seek(relativeSecond: 5, option: Preference.SeekOption.defaultValue)
     }
   }
 
--- a/iina/MenuController.swift
+++ b/iina/MenuController.swift
@@ -8,36 +8,65 @@
 
 import Cocoa
 
-fileprivate func sameKeyAction(_ lhs: [String], _ rhs: [String], _ normalizeLastNum: Bool, _ numRange: ClosedRange<Double>?) -> (Bool, Double?) {
+fileprivate func sameKeyAction(_ lhs: [String], _ rhs: [String], _ normalizeLastNum: Bool, _ numRange: ClosedRange<Double>?) -> (Bool, Double?, Any?) {
   var lhs = lhs
-  if lhs.first == "seek" && (lhs.last == "exact" || lhs.last == "keyframe") {
-    lhs = [String](lhs.dropLast())
+  var extraData: Any? = nil
+  if lhs.first == "seek" {
+    if let last = lhs.last {
+      if let _ = Double(last) {
+        // "relative+keyframes" is implicit default.
+        extraData = Preference.SeekOption.defaultValue
+      } else {
+        // matching values are one of: "relative", "keyframes", "exact", "relative+keyframes", "relative+exact"
+        let splitArray = last.split(whereSeparator: { $0 == "+" })
+        if splitArray.count == 2 && splitArray[0] == "relative" {
+          if splitArray[1] == "exact" {
+            lhs = [String](lhs.dropLast())
+            extraData = Preference.SeekOption.exact
+          } else if splitArray[1] == "keyframes" {
+            lhs = [String](lhs.dropLast())
+            extraData = Preference.SeekOption.relative
+          }
+        } else if splitArray.count == 1 {
+          if splitArray[0] == "exact" {
+            lhs = [String](lhs.dropLast())
+            extraData = Preference.SeekOption.exact
+          } else if splitArray[0] == "keyframes" {
+            lhs = [String](lhs.dropLast())
+            extraData = Preference.SeekOption.relative
+          } else if splitArray[0] == "relative" {
+            lhs = [String](lhs.dropLast())
+            extraData = Preference.SeekOption.defaultValue
+          }
+        }
+      }
+    }
   }
   guard lhs.count > 0 && lhs.count == rhs.count else {
-    return (false, nil)
+    return (false, nil, nil)
   }
   if normalizeLastNum {
     for i in 0..<lhs.count-1 {
       if lhs[i] != rhs[i] {
-        return (false, nil)
+        return (false, nil, nil)
       }
     }
     guard let ld = Double(lhs.last!), let rd = Double(rhs.last!) else {
-      return (false, nil)
+      return (false, nil, nil)
     }
     if let range = numRange {
-      return (range.contains(ld), ld)
+      return (range.contains(ld), ld, extraData)
     } else {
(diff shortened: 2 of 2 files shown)


low-batt:
I'm going to defer to @lhc70000 on the code changes as it has been proven that I don't know this part of the code well enough to comment. But I do want to comment on this:
> make it more robust (e.g. using ? instead of !)

Indeed I remember reading that one large company specifies that use of the swift forced unwrap operator is a [code smell](https://en.wikipedia.org/wiki/Code_smell).

I have a different view. I think of force unwrapping as a sanity check and like most sanity checks it will result in a fatal error, in this case a crash. Therefore in the case where an optional must be non-nil for the code to function as intended, I think of using ? instead of ! as removing a sanity check. That can turn a crash that might be noticed during development into subtle malfunctioning behavior that isn't noticed until some user encounters the obscure problem.

So now a user has reported the problem. If the application had crashed and the user provided the crash report there is a good chance that points to the source of the problem. But it didn't crash and we were not able to reproduce the malfunction. We suspect something was nil when it shouldn't have been and are desk checking the code. Think of every ? as an if statement. The code is now filled with many different execution paths. Which property was nil when it shouldn't be?

For this reason I don't see avoiding the force unwrapping operator as a black and white issue. I don't think it is appropriate to always consider it a code smell. In primary features that persist data you might prefer the application crashes rather than corrupt a database. In secondary features of the application you might prefer the feature malfunctions rather than take down the application.

My impression is that mine is a minority viewpoint. The "fear of crash" philosophy is winning. I have been reducing my use of the force unwrap operator.

With respects to IINA, we need to hear what @lhc70000 thinks about use of ! and follow his guidance.

saagarjha:
Not @lhc70000 but +1 to the use of force unwrapping when appropriate

svobs:
Re: use of force-unwrapping issue, yeah, I can understand the need to fail fast, but bringing down the whole program seems like harsh punishment! Maybe adding an "else" and writing an error to the log would be good? Although that does tend to clutter up the code... anyway I'll go with the official decree whatever that may be.

low-batt:
My comment is about the push for an absolute ban on use of forced unwrapping which seems to be taking hold in some parts of the industry. Not an opposition to use of optional chaining. Just trying to bring awareness to the downsides of excessive use of optional chaining.

svobs:
Rebranding this one so it doesn't get lost. (Excited by the 1.3.0 release by the way! It will be a huge help). Yes, this isn't a crasher but I'd like to push for it more than anything else, since it's very high visibility, it's not very complicated, and it's only been partially fixed.

The original issue was that forward/backward seeks which qualified for menu items were being hard-coded to `keyframes` and "5 sec". The duration is no longer hard-coded, but it remains hard-coded to `keyframes`, so the actual seek time will vary widely. A lot of modern videos I have are highly compressed and only have keyframes every 8s or so. 

Let's say there is a video with a keyframe only every 10s. This is how `keyframes` seeks appear to work:
```
KF                   KF                  KF
|--------------------|--------------------|
0s                   10s                 20s
 ^A               ^B   ^C
```
Point A is at `00:01`. A `seek 5` from there will jump to `00:10` because that's the next closest keyframe after the seek - almost double what they requested.
Point B is at `00:08`. A `seek 5` from there will jump over the next keyframe, then try to find the next closest, which will end up at `00:20` - 12 seconds!
Point C is at `00:11`. A `seek 5` from there will go back 5 seconds and then scan forward to the next keyframe, which will result in `00:10` - only 1s.

I don't like keyframe seeks. The MPV team seemed to make this the default because they were worried about performance. And they may have been a valid concern 10 or 15 years ago, but IINA has the luxury of assuming relatively modern Mac hardware. I've been using `exact` seeks for the last couple of years and have had no problems at all. If it were my choice, it would be the default mode, but currently IINA is effectively **forcing** users to use `keyframes`. Let me know if there is some underlying issue with `exact` that I'm missing here.

low-batt:
Yes, there are still lots of interesting proposed changes outstanding. Merging will continue.
This release includes a lot more than originally planned. Way past time to get something out there.
With all the changes in 1.3.0 we are expecting a follow up maintenance release will soon be required.
There is already a 1.3.0 issue posted related to the mpv 0.34.1 upgrade.
Was hoping we'd at least make it 24 hours before that happened...

svobs:
Didn't realize this had fallen into conflict. Fixed and rebased.

Really hope this can make it into develop soon.

low-batt (approved):
Pulled PR, corrected as needed to get build to work under Xcode 15 with current deps, tested with key binding that used seek exact, checked mpv log file and confirmed the exact flag is now being passed to mpv.

svobs:
> Pulled PR, corrected as needed to get build to work under Xcode 15 with current deps, tested with key binding that used seek exact, checked mpv log file and confirmed the exact flag is now being passed to mpv.

@low-batt do you have the ability to commit your changes? I don't know if I will have time in the next few days.

low-batt:
No need to commit those changes. This PR does not have conflicts preventing merge. Those changes are only needed if you want to build and test this PR. The changes are quite simple.

Anyway, don't worry, I can take care of it should the other reviewers want changes.

iina/MainMenuActions.swift
```diff
@@ -90,11 +90,10 @@ extension MainMenuActionHandler {
...
   @objc func menuStep(_ sender: NSMenuItem) {
-    let seconds = Double(abs((sender.representedObject as? Int) ?? 5))
-    if sender.tag == 0 { // -> 5s
-      player.seek(relativeSecond: seconds, option: .relative)
-    } else if sender.tag == 1 { // <- 5s
-      player.seek(relativeSecond: -seconds, option: .relative)
+    if let args = sender.representedObject as? (Double, Preference.SeekOption) {
+      player.seek(relativeSecond: args.0, option: args.1)
+    } else {
+      player.seek(relativeSecond: 5, option: Preference.SeekOption.defaultValue)
```
uiryuu, line 96: You forgot to consider `sender.tag`, if `sender.tag == 1` the `relativeSecound` should be -5. When the menu item is not matched by keybindings, we will have forward and **backward** menu items by default.

iina/MenuController.swift
```diff
@@ -8,36 +8,65 @@
...
+  var extraData: Any? = nil
+  if lhs.first == "seek" {
+    if let last = lhs.last {
+      if let _ = Double(last) {
+        // "relative+keyframes" is implicit default.
+        extraData = Preference.SeekOption.defaultValue
+      } else {
+        // matching values are one of: "relative", "keyframes", "exact", "relative+keyframes", "relative+exact"
+        let splitArray = last.split(whereSeparator: { $0 == "+" })
+        if splitArray.count == 2 && splitArray[0] == "relative" {
```
uiryuu, line 22: Here you assumed "relative" always appears first, however it seems things like "keyframes+relative" are also valid mpv flags. I tried to add `seek -5 keyframes+relative` and `seek 5 relative+keyframes`, both shortcuts can be used in IINA, but only the latter one is recognized as the same action as in the default menu.

uiryuu (changes requested):
I have a problem compiling on this PR after rebasing to develop. On line 901 of `MenuController.swift`, Xcode doesn't let me compile

```swift
let (_, value) = sameKeyAction(actions, actions, normalizeLastNum, numRange)
```
The error message is "Cannot convert value of type '(Bool, Double?, Any?)' to specified type '(_, _)'". So I have to change it to:
```swift
let value = sameKeyAction(actions, actions, normalizeLastNum, numRange).1
```

low-batt:
As @svobs is busy I will attempt to address the feedback.

low-batt:
Replacing this PR with PR #4739

The new PR attempts to address the feedback from @uiryuu. Lets continue the discussion in that PR.

Issue #3669, opened 2022-03-18
[Feature Request] Play/resume media from the welcome window via menu and keyboard


veloxo:
**What you want IINA to do:**

In the **welcome window,** enable the "Resume" action under "Playback" in the menu bar, and accordingly any `cycle pause` or other `play`-like key bindings,  to _open-and-resume_ the highlighted media item at the top of the recent-media list.

It may be sensible to enable similar actions like "Next Media"/"Previous Media", too.

**What IINA does currently:**

AFAICT, in the welcome window (_Preferences > General > Behavior > At launch: "Show welcome window"_), there is no menu-bar action or keyboard shortcut to play/resume the highlighted media item. Only cursor-clicking the item will resume it.

Note that the `up` and `down` keys will play next/previous media items from the welcome window. This seems like the equivalent of an *open-and-next/previous-media* action.

**Why you think this should be added:**

This would extend keyboard-driven operation to application startup and help enable automation via AppleScript, MacOS Shortcuts, BetterTouchTool, etc. (Personally, I'm trying to configure a remote-control button that launches IINA and resumes the last-watched video.)

Thanks!

svobs:
It does look like keyboard navigation is missing from the Welcome screen. Actually there seems to be a bug in which pressing either the UP or DOWN arrows resumes playback of the last video (doesn't matter what key bindings are set). You might be able to use either of those in your script for now ;)

It would be good if at a minimum, pressing ENTER on the Welcome screen resumes playback of the last video. Even better would be for the UP / DOWN / TAB keys to change the highlighted item, with ENTER resuming whatever item is highlighted.

veloxo:
>It would be good if at a minimum, pressing ENTER on the Welcome screen resumes playback of the last video

Agreed, this would be the most straightforward solution!

>pressing either the UP or DOWN arrows resumes playback of the last video

To clarify: AFAICT in v1.2.0, when there's several items in the Welcome-Screen playback history, `UP` or `DOWN` start playing the earliest or latest item in the history list _below_ the highlighted/most-recent/partially-watched item. A simple "resume highlighted item" command seems to be unavailable.

>Even better would be for the UP / DOWN / TAB keys to change the highlighted item, with ENTER resuming whatever item is highlighted.

Makes sense to me (a random user), as it would allow navigation to any item in history.

low-batt:
The implementation of this enhancement in PR #3739 has been merged into the develop branch.

svobs:
[...]

PR #3739, opened 2022-05-12
Enable up/down & enter key in Welcome window (issue #3669)

svobs:
- [x] It implements / fixes issue #3669.

---

**Description:**
_[second attempt - prev PR got deleted when I renamed the branch]_
The Welcome window actually is displaying a button with the most recently played file, followed by a table of Recent File 2, Recent File 3, etc. The reason the up/down arrow keys caused a file to play was that the "selectionChanged" callback was being used as a proxy for detecting a user event.

I changed this so that:
- When window opens, we "pretend" the button with the most recent entry is selected, which makes sense because the table has no items selected. Thus, ENTER or RETURN at this point will open the most recent file, or if for some reason that is no longer available, tries to open the next most recent one.
- User can use UP and DOWN arrow keys to select recent files, which will highlight in the table, and then press ENTER or RETURN to open that file.
- Mouse clicks behave the same as before

One snag: once the user scrolls down to Recent File 2 or below, they can't scroll back up to Most Recent File. That's due to the fact that that entry is actually outside the table, and would have required a bunch of custom code to make it look good. Which I was prepared to do, but I also ran into a problem where once I give the table control of the arrow keys, it doesn't seem to want to give them back, and I'd need to intercept them to make that work. So maybe that's something for the future, but this should be much improved over the previous behavior.

### DIFF
--- a/iina/Base.lproj/InitialWindowController.xib
+++ b/iina/Base.lproj/InitialWindowController.xib
@@ -338,7 +338,7 @@
             </view>
             <connections>
                 <outlet property="delegate" destination="-2" id="0bl-1N-AYu"/>
-                <outlet property="initialFirstResponder" destination="VeK-rB-yjt" id="GMr-dI-Npg"/>
+                <outlet property="initialFirstResponder" destination="se5-gp-TjO" id="F1N-a0-gN8"/>
             </connections>
             <point key="canvasLocation" x="135" y="275"/>
         </window>
--- a/iina/Base.lproj/InitialWindowController.xib
+++ b/iina/Base.lproj/InitialWindowController.xib
@@ -338,6 +338,7 @@
             </view>
             <connections>
                 <outlet property="delegate" destination="-2" id="0bl-1N-AYu"/>
+                <outlet property="initialFirstResponder" destination="VeK-rB-yjt" id="GMr-dI-Npg"/>
             </connections>
             <point key="canvasLocation" x="135" y="275"/>
         </window>
--- a/iina/InitialWindowController.swift
+++ b/iina/InitialWindowController.swift
@@ -66,6 +66,35 @@ fileprivate extension NSColor {
   }()
 }
 
+fileprivate class GrayHighlightRowView: NSTableRowView {
+  private func fillWithColor(_ color: NSColor) {
+    let selectionRect = NSInsetRect(self.bounds, 2.5, 2.5)
+    color.setFill()
+    let selectionPath = NSBezierPath.init(roundedRect: selectionRect, xRadius: 4, yRadius: 4)
+    selectionPath.fill()
+  }
+  
+  override func drawSelection(in dirtyRect: NSRect) {
+    if self.selectionHighlightStyle != .none {
+      let selectionRect = NSInsetRect(self.bounds, 0, 0)
+      NSColor.initialWindowLastFileBackground.setFill()
+      let selectionPath = NSBezierPath.init(roundedRect: selectionRect, xRadius: 4, yRadius: 4)
+      selectionPath.fill()
+    }
+  }
+
+  func setHoverHighlight() {
+    self.wantsLayer = true
+    self.layer?.cornerRadius = 6
+    self.layer?.backgroundColor = NSColor.initialWindowActionButtonBackgroundHover.cgColor
+  }
+
+  func unsetHoverHighlight() {
+    self.wantsLayer = true
+    self.layer?.cornerRadius = 6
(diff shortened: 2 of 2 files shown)


low-batt:
On this:
> One snag: once the user scrolls down to Recent File 2 or below, they can't scroll back up to Most Recent File.

That sounds like a problem that needs to be fixed. However, I pulled the PR locally and tested and did not encounter that issue? I was able to scroll to the end and back up to the beginning without any problems. All seemed to be working well. What did I miss?

svobs:
> > One snag: once the user scrolls down to Recent File 2 or below, they can't scroll back up to Most Recent File.
> 
> That sounds like a problem that needs to be fixed. However, I pulled the PR locally and tested and did not encounter that issue? I was able to scroll to the end and back up to the beginning without any problems. All seemed to be working well. What did I miss?

Welcome-Screen_Recently-Played-List
Here's a screenshot of the recently used items, where "1" = last recently used, "2" = 2nd to last recently used, etc. As you can see, 1 is displayed differently than the rest - it's actually a button made to look like a table entry. Only 2 and below are actually in a table.

It actually wouldn't be too hard for me to put everything in a table, and add icons and/or time listings for the rest same as the "first", or simply make it look similar to how it looks now (maybe change the table highlight color to gray, to match LRU item 1). I spent a great deal of time on another personal project recently which does that. But that seemed out of scope for this bug request where the requestor just wanted keyboard navigation for scripting. Changing everything into one table would probably actually be less work than trying to make the button behave like the first entry in the table. But I'm willing to do whatever will get approved 😉

low-batt:
My confusion continues as this is what I see when the window first comes up:
issue-3669
No highlighting, no button?

low-batt:
That is not with your current changes. I'll update.

svobs:
> No highlighting, no button?

Interesting! Does that list match the list in File > Open Recent... ? Are you using Apple Silicon? 
I made some style changes early this morning so that the highlight color of the list matches that first value. But if you weren't see that button, then I guess getting that straightened out is the most important issue.

low-batt:
> Does that list match the list in File > Open Recent... ?

Yes.

> Are you using Apple Silicon?

Yes and No.

I recently upgraded my development Mac. I have a model MacBookPro18,2, one of the new 16" Macs. I'm running macOS 12.3.1. I'm building with Xcode 13.3. So usually building and testing a native Apple Silicon version of the app.

I ran some tests...

Running IINA 1.1.2 on my development machine (under Rosetta) I do not see the button. It looks just like I posted above.

Under my Apple Silicon Mac I configured Xcode to build for Intel and built the current develop branch. I took the resulting app over to my older Intel MacBook Air running macOS 10.15.7. The welcome screen showed the button.

Running that same app under Rosetta on my development machine does not show the button.

IINA preferences could differ between those two machines. I took a very quick look at the code and didn't see it keying off a preference?

What version of macOS are you testing with?

low-batt:
Maybe this code is the issue:
```swift
    if Preference.bool(for: .recordRecentFiles),
      Preference.bool(for: .resumeLastPosition),
```
I will check my preference settings...

low-batt:
Ok. That explains the behavior. I did not have resume last position set on my development machine. Now I can reproduce the problem with not being able to get back to the first entry using up and down keys.

low-batt:
Not being able to up arrow back to the first entry when it is a button definitely feels wrong. Would be good to address that if possible.

svobs:
> Ok. That explains the behavior. I did not have resume last position set on my development machine. Now I can reproduce the problem with not being able to get back to the first entry using up and down keys.

Ahh good to know. I wasn't aware of that preference! I'm developing on the last Intel Macbook Pro and my knee-jerk reaction is to blame architecture differences, but surprisingly it seems there are hardly any issues with that.

I wasn't aware of that preference! But I was able to make the button disappear now. I did some testing and fortunately, none of the stuff I've added in this PR seems to misbehave when the button is gone, or if both button and table are gone.

I did push another update just an hour ago which adds hover highlight to match the other elements in the window. It was quite a bit of code but it looks a lot better. I suppose now I just need to get the UP arrow to highlight the button.

svobs:
@low-batt I think I got it. Take a look at my latest commit and let me know what you think

low-batt (approved):
I tested these changes both with resume enabled and without. All working well now.

lhc70000:
Great job! I like it very much and it works amazing.

lhc70000:
I added a small commit 2dc760d5 to use the key names in `KeyCodeHelper.keyMap` when handling the key code.

PR #3745, opened 2022-05-16
Close support automatic screen refresh, #3414

low-batt:
This commit builds on top of the core code developed by @jesec to
add support for detecting additional cases where the display needs to
be adjusted, notifying the user using the OSD and logging the decisions
made by the feature.

The original code has been refactored out of MainWindowController and
into a new DisplaySynchronizer class. That class contains the core code
that manages a display a selects an appropriate refresh rate. That class
is used by a new RefreshRateMatcher class that handles the higher level
aspects of the feature.

Control of the feature has been moved from IINA Preferences to a menu
item in the Video menu. This provides a way to quickly turn the feature
off if a display proves problematic.

IINA must be running under macOS 12 or above for this feature to be
available. Older versions of macOS return zero for a display's refresh
rate preventing this feature from working. When running on an older
version of macOS the menu item will be hidden.

The commits in the pull request will:
- Add a new item "Match Refresh Rate" to video menu
- Bind the menu item to the ctrl+meta+r key
- Add a new class DisplaySynchronizer to control displays
- Add a new class RefreshRateMatcher to control the feature
- Change MainWindowController to use RefreshRateMatcher
- Add a new OSD message to show the display refresh rate
- Change PlayerCore to support short internal pause/resume cycles
- Add the ability to suppress pause/resume OSD messages

- [x] It implements / fixes issue #3414.

---

**Description:**


low-batt:
**DO NOT MERGE**
Discussions about this feature are ongoing in the issue.

low-batt:
Closing this PR. See issue #3414 for details. Just a couple of the issues raised with the changes in this PR:
- Should not need to pause/resume
- Too complex
- Does not address the issue with the UI being tiny

Issue #3729, opened 2022-05-07
On-screen displays do not update sliders for keys bound to menu items until keyUp


svobs:
I mentioned this problem in #3686, but I was finally able to reproduce it more reliably.

**System and IINA version:**
- macOS 12.3.1
- IINA 1.2.0

**Steps to reproduce**
1. Open System Preferences, and go to Keyboard. Set Key Repeat all the way to the right ("Fast").
2. Open IINA
3. Go to Preferences > Key Bindings.
4. Select "IINA Default", then click "Duplicate...", and pick a name like "IINA 2".
5. Confirm that your new config is selected. Confirm that key "UP" = "add volume 5" and "DOWN" = "add volume -5".
6. Add two new mappings: "1" = "add volume 5" and "2" = "add volume -5"
7. Close prefs window. Check "Audio" menu in menu bar and confirm that "Volume +5%" shows UP and "Volume -5%" shows DOWN.
8. Open any video. Video should start to play.
9. Press and hold DOWN for several seconds. Then press and hold UP for several seconds.
10. Press and hold 2 for several seconds. Then press and hold 1 for several seconds.

**Expected behavior:**
UP and DOWN keys should update volume sliders as they are repeated, as "1" and "2" keys do.

**Actual behavior:**
"1" and "2" keys behave correctly, but UP and DOWN do not update during key repeat. They only update once they are released.

This seems to only happen for keys which are mapped to menu items.

- [x] MPV does not have this problem. But MPV does not show mapped keys in menu items.

**How often does this happen?**
Always

svobs:
Update: I tested this again in MacOS Monterey, and it looks like my original fix was not the magic bullet.

I'm now seeing in 1.2.0, 1.3.0, and with the previous fix:

1. The seek slider now updates consistently, even with keyboard repeat.
2. The volume slider now consistently does not update when keyboard repeat is used.

This would seem to indicate that these slider animations are updated at the mercy of OS scheduling decisions.

Back to the previous suspect...

By adding custom logging to the code, I was previously able to see that:

1. On main `DispatchQueue`, keyboard repeats are causing 20-30 events per second to be sent to mpv (i'm guessing it's based on how fast the CPU can handle them)
2. On `MPVController`'s DQ, IINA correctly receives events back from MPV notifying that it has processed the commands.
3. The problem occurs here:
[...]

PR #3747, opened 2022-05-18
Clean up MainMenuActions responder chain

svobs:
Note: this was originally intended as the fix for #3729, but testing on MacOS Monterey shows that is no longer the case. I'm keeping this open mainly because it simplifies the code, which is a good thing, but now much lower priority. 

_From the commit message:_
Simplify MainMenuActions: change it to an extension of PlayerWindowController rather than a separate class, since it was already performing that role. Removes redundant lifecycle management and custom code for tinkering with the player window responder chain

### DIFF
--- a/iina/MainMenuActions.swift
+++ b/iina/MainMenuActions.swift
@@ -8,19 +8,7 @@
 
 import Cocoa
 
-
-class MainMenuActionHandler: NSResponder {
-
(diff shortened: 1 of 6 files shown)


svobs:
Closing this PR: I've rolled this into PR #3847 which reworks the whole key bindings system.

Issue #1415, opened 2018-01-24
cursor does not auto-hide after switching app windows


delusionxb:
**System and IINA version:**

- macOS 10.11.6
- IINA 0.0.15

**Expected behavior:**
cursor does not auto-hide

**Actual behavior:**
cursor should auto-hide

- Crash report:
```

```

- MPV log:
```

```

**Steps to reproduce:**
1. open a video file by iina
2. see cursor and status bar auto-hide
3. switch between current running apps by cmd+tab key
4. hold cmd key and keep switching back to iina, see cursor does not auto-hide
5. swing cursor a little to make status bar show up, then cursor will auto-fade together with status bar again

- [x] MPV does not have this problem.

**How often does this happen?**
very easy to reproduce every time

uiryuu:
I vaguely remember this has been fixed and I couldn't reproduce this either. Feel free to reopen or file a new issue if the issue still occurs.

PR #3758, opened 2022-05-27
Focus the window to ensure cursor can be hidden in fullscreen

jesec:
"NSCursor.setHiddenUntilMouseMoves" is only effective when the
calling window is the focused window.

When user moves back from another screen or space, the window may
not be focused. As a result, the cursor can't be hidden before the
user clicks the window.

It can be reasonably expected that the user would want to focus on
the window when the mouse moves inside the window and the window is
in fullscreen.

As such, focus the window on mouse movement inside the window when
the window is in fullscreen.

Bug: #1415

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -947,6 +947,9 @@ class MainWindowController: PlayerWindowController {
 
   override func mouseMoved(with event: NSEvent) {
     guard !isInInteractiveMode else { return }
+    if fsState.isFullscreen {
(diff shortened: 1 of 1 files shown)


uiryuu:
Firstly, I believe that the issue you mentioned is not related with this PR, and should have been already fixed. Besides, I think it is not reasonable to change the focus window just because the mouse moved into the window in fullscreen.

uiryuu:
I think this has been fixed; now even the window is not focused, the UI still fade and appear based on cursor movements.

PR #3773, opened 2022-05-31
Update German Localization …

Aeneon:
Updated German Translation, so it will match IINA's latest 1.3.0 Build …


uiryuu:
Please update localization on [IINA's instance of Crowdin](https://crowdin.com/project/iina), thanks!

PR #3779, opened 2022-06-03
How to Refresh, Reload, or load cache when opening a stream online

Jerry080801:
Sometime when I opening a online mp4 stream, it cache from about 20-30 minutes, when I normally play to that point, I cannot play it and it is not going to continue streaming. And it would be great if I can reload the online streaming videos like in a browser, thank you


uiryuu:
Please report as an issue, closing.

PR #3780, opened 2022-06-03
Update the network prefetch preference's default label

Mnpn:

---

**Description:**
The default value for this preference appears to have changed in 14f785b (iina/Preference.swift line 755), but the label wasn't updated, leading to a mismatch:

image

This PR updates the label's text, though translations may also need to be updated on Crowdin.

(This is a copy of #3658. The author appears to have deleted their fork, permanently closing the PR.)


low-batt:
Thanks for posting this PR. I created an issue for it to give it more visibility.

I'm unsure about the IINA conventions on using Crowdin, but I think the convention on translations is to include the english translation in the commit and not include other translations to avoid merge conflicts with Crowdin. If I am correct about that then this line in the English version of PrefNetworkViewController.strings needs to be updated:
```text
/* Class = "NSTextFieldCell"; title = "Default: 100"; ObjectID = "fab-7d-mMQ"; */
"fab-7d-mMQ.title" = "Default: 100";
```

See if others confirm that is the convention.

uiryuu:
Thanks for your PR; I'll change the en.lproj accordingly.

Issue #3788, opened 2022-06-04
Symlinks in recently opened files can cause duplicates in Welcome Window


svobs:
**System and IINA version:**

- macOS 12.4
- IINA 1.3.0

![IINA-InitialWindow-Screenshot](https://user-images.githubusercontent.com/2213815/172028727-190d020e-36db-4710-aad0-7dfff9212489.jpeg)

**Steps to reproduce:**

1. In IINA Preferences > General, make sure "Enable playback history" and "Resume last playback position" are selected.
2. In the terminal, copy a video into /tmp and change its permissions like this:

```
sudo cp myVideo.mp4 /tmp/myVideo.mp4
sudo chown `whoami` /tmp/myVideo.mp4
```

Then open it with IINA-cli:
```
/Applications/IINA.app/Contents/MacOS/iina-cli /tmp/myVideo.mp4
```

When it starts to play, quit IINA. Then open it again without a video file, so it displays the Welcome window.

**Expected behavior:**
Recent video appears first in the list with the option to "Resume"

**Actual behavior:**
Recent video appears first in the list with the option to "Resume", and again right below it.

<br>

**Steps to reproduce:**

**How often does this happen?**
Always for these paths.

svobs:
Forgot to point out where the symlink comes in. In modern MacOS, `/var` and `/tmp` have been relocated under `/private`:
```
$ ls -l /tmp
lrwxr-xr-x@ 1 root  wheel  11 May  9 14:30 /tmp -> private/tmp
```

low-batt:
This is fixed in the released version of IINA.

PR #3789, opened 2022-06-04
Resolve symlinks when comparing recent files against last played file

svobs:
- [X] It implements / fixes issue #3788.

---

**Description:**
The last played file is checked against the recent files list so that it doesn't appear twice, but if one is a symlink, it can give a false negative and result in a duplicate listing. This seems to be a bigger problem for files in /tmp and /var, which Apple now links to /private/tmp and /private/var, so they can have different paths depending on which MacOS API they came from.

This fix adds `resolvingSymlinksInPath()` to both sides of the comparison, which is needed to normalize both paths. Apple treats /var and /tmp a bit strangely. See [this discussion](https://stackoverflow.com/questions/43031045/for-what-paths-is-resolvingsymlinksinpath-wrong).

### DIFF
--- a/iina/InitialWindowController.swift
+++ b/iina/InitialWindowController.swift
@@ -138,7 +138,7 @@ class InitialWindowController: NSWindowController {
   }
 
   lazy var recentDocuments: [URL] = {
-    NSDocumentController.shared.recentDocumentURLs.filter { $0 != lastPlaybackURL }
+    makeRecentDocumentsList()
   }()
   private var lastPlaybackURL: URL?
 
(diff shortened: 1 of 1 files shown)


uiryuu (approved):

uiryuu:
Pending one more approve then we can merge it!

iina/InitialWindowController.swift
```diff
@@ -240,8 +247,21 @@ class InitialWindowController: NSWindowController {
 
   func reloadData() {
     loadLastPlaybackInfo()
-    recentDocuments = NSDocumentController.shared.recentDocumentURLs.filter { $0 != lastPlaybackURL }
+    recentDocuments = makeRecentDocumentsList()
     recentFilesTableView.reloadData()
+
+    if Logger.enabled && Logger.Level.preferred >= .verbose {
+      let last = lastPlaybackURL == nil ? "<none>" : lastPlaybackURL!.resolvingSymlinksInPath().path
```
saagarjha, line 254: Also an option

```suggestion
      let last = lastPlaybackURL.flatMap { $0.resolvingSymlinksInPath().path } ?? "<none>"
```

saagarjha (approved):

uiryuu:
Thanks!

PR #3793, opened 2022-06-05
feat: allow user to override file extension filter

### LINKED ISSUE #3792: User should be allowed to override file extension filters
**What you want IINA to do:**

Before setting the file extension filter in https://github.com/iina/iina/commit/9937f452bba5a25c41679741a1d5987b269102f0, IINA was able to load an mkv and extract the subtitles dynamically.
Now it forces a user to use ffmpeg/etc to extract the subtitle first.

**What IINA does currently:**

The open file dialog should suggest file types but ultimately allow the user to override it.

The best UX would be to make sure a beginning user doesn't fail, but allow an intermediate+ user to do what they want.

s22chan:
Before, load subtitles allowed you to load an mkv and use it as a subtitle. I find it inconvenient now to use ffmpeg to extract the subtitle first.

Fixes #3792

- [x] It implements / fixes issue #.

---

**Description:**
See issue

### DIFF
--- a/iina/Utility.swift
+++ b/iina/Utility.swift
@@ -125,6 +125,7 @@ class Utility {
     panel.canChooseDirectories = chooseDir
     panel.resolvesAliases = true
     panel.allowedFileTypes = allowedFileTypes
+    panel.allowsOtherFileTypes = true
     panel.allowsMultipleSelection = false
     panel.level = .modalPanel
     if let dir = dir {


uiryuu:
I wanted to merge this fix, but it seems has no effect and I have no idea why...

s22chan:
Doesn't fix it. I must've applied it to the wrong commit.

Issue #3798, opened 2022-06-06
Allow changing "logLevel" in Advanced Preferences


svobs:
**What you want IINA to do:**
Add a drop-down for log level next to the "Enabled logging" checkbox in Preferences > Advanced Settings.

![Screen Shot 2022-06-05 at 21 08 41](https://user-images.githubusercontent.com/2213815/172094005-b054504d-0f03-498c-8a4f-72ae0b2d89c2.png)

It is grayed out if "Enable logging" is disabled:
![Screen Shot 2022-06-05 at 21 08 58](https://user-images.githubusercontent.com/2213815/172094036-c5c04e8e-a6c8-4a88-b3ec-84761dd590a7.png)

It's also grayed out with everything else if "Enable advanced settings" is disabled:
![Screen Shot 2022-06-05 at 21 09 05](https://user-images.githubusercontent.com/2213815/172094098-7c4c164b-31a1-4834-9126-9272359e2a3a.png)

**What IINA does currently:**
Although there is an option to enable/disable logging, and there is a preference value `logLevel` which controls the verbosity of logging which is already fully functional, there is currently no way to configure it in the UI or set it permanently, except for editing IINA's .plist file.

**Why you think this should be added:**
I wrote this code a while back in order to learn how to use Interface Builder. It isn't super important but does work and it is somewhat useful, and I'd rather not have the code be wasted.

**Examples of other projects that have something similar:**
Various other developer-friendly projects, probably

svobs:
Didn't realize this was still open. Closing. See comments in PR #3799.

PR #3799, opened 2022-06-06
Add "log level" drop-down to Advanced prefs page

svobs:
- [x] It implements / fixes issue #3798.

---

**Description:**

- Adds "Log level" dropdown which allows configuration of existing "logLevel" preference key.
- Is grayed out if "Enable logging" is unchecked or if "Enable advanced settings" is unchecked

### DIFF
--- a/iina/Base.lproj/PrefAdvancedViewController.xib
+++ b/iina/Base.lproj/PrefAdvancedViewController.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="16097" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="20037" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="16097"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="20037"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <objects>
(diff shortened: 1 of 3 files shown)


low-batt:
I see a couple of minor concerns with the issue.

#### mpv Log Level

No mention of the mpv log level. Should that be included? I've been controlling it by adding to the "Additional mpv options" section:
```text
msg-level all=trace
```

By the way, there seems to be a regression in mpv 0.34.1 with respects to logging. Apparently there is a window of vulnerability in the mpv logger. It is especially noticeble if you turn up logging to full as shown above. IINA will hang. Profiling shows all threads blocked on the mpv logger lock. Issue #3560 is an example of this. See the spindump included in that issue.

I tried reproducing this with mpv and failed. I don't know if IINA is doing something to trigger this, or if mpv introduced a regression that can only be triggered using libmpv. I did confirm that mpv made changes to logger locking. I reviewed their code, but did not spot an error.

#### Users Posting Logs

There already is an issue with users failing to post logs. If we introduce this feature will users post logs with the level turned down such that we don't get the debugging information we need?

svobs:
> mpv Log Level
Ah yes, that does make sense. I hadn't considered the MPV logs at all at the time. The dropdown should be labelled "IINA log level" if nothing else.

Does it does make sense to add an option for "MPV log level"? It might have to be simply "IINA level" and "MPV level" since it will start to run out of space. Or: move the "Open log directory" button to the next line, and then add a separator between the logging options and "Use mpv's OSD" checkbox so that all the logging stuff is in its own section. Something **roughly** like this?
Screen Shot 2022-06-06 edit

MPV's levels are interesting:
```
no: | complete silence
-- | --
fatal messages only
error messages
warning messages
informational messages
status messages (default)
verbose messages
debug messages
very noisy debug messages
```

MPV's "verbose" level is actually less verbose than its "debug", which is the opposite of IINA's. It would probably help to assign a number for both to make it obvious.
I think (for developers at least) it's expected that "trace"-level logging will be almost unusably slow. And this section is labelled "Advanced", after all.

> Users Posting Logs
I believe logging is disabled by default, right? So the users are already being told to go to Preferences > Advanced and check "Enable logging". It doesn't seem like a big hurdle to add "and make sure 'IINA log level' is set to at least 'Debug' and ideally 'Verbose'". In fact, it might help to increase logging from the default of 'Debug' for some issues.

We can also add a message in the preference pane itself, in the space below right below the options, along the lines of, "Setting MPV level to Trace may be unusably slow and is not recommended for everyday use." As well as a message saying something like, "If submitting a bug report, please try to attach the logs for the time when the program occurred, while 'Enable Logging' was checked and both log levels are at least 3", or something along those lines.

So difficult to cater to both advanced and gen-pop users. It's ultimately a philosophical opinion. Though I tend to err on the side of allowing more options and the dangers that brings.

low-batt:
I like proposed ways to address the issues I raised.

I believe officially it is "mpv", not "MPV". From the [mpv](https://mpv.io/) web site:
> mpv is a free (as in freedom) media player for the command line.

svobs:
> I believe officially it is "mpv", not "MPV". From the [mpv](https://mpv.io/) web site:
> 
> > mpv is a free (as in freedom) media player for the command line.

Heh, oh yeah. Not sure I'm so comfortable leaving it lowercase in the beginning of a sentence. But if it comes up in UI elements, I will definitely lowercase it. Probably I will work extra hard to place it in each sentence so that I can lowercase it without it clashing with grammar rules...

Maybe will revisit this PR at some point in the future. It's too bad there isn't a "shelved for now" status for PRs.

svobs:
Closing. Realistically not gonna happen

Issue #3800, opened 2022-06-06
Playback History searches don't work for capital letters


svobs:
**System and IINA version:**

- macOS 12.4
- IINA 1.3.0 build 131

**Expected behavior:** 
If a file is listed in Playback History window, searching for its exact name should find it.

**Actual behavior:**
Searches that include a capital letter will never find anything, even if the filename has that exact capital letter.

**Steps to reproduce:**
1. Play a file
<img width="673" alt="Screen Shot 2022-06-06 at 00 14 23" src="https://user-images.githubusercontent.com/2213815/172114788-b8ba81b9-9e19-4bf8-bb90-ed5735706013.png">

2. Search using a capital letter
<img width="673" alt="Screen Shot 2022-06-06 at 00 14 39" src="https://user-images.githubusercontent.com/2213815/172114800-03be7617-31a6-4336-a48a-9df7704098be.png">

**How often does this happen?**
Always

svobs:
I was able to identify the problem right away and thus the specific name. Will submit PR

low-batt:
The fix in PR [#3801](https://github.com/iina/iina/pull/3801) has been merged into the develop branch.

Keeping this issue open until the fix is in a released version of IINA.

low-batt:
Fixed in IINA 1.3.1

PR #3801, opened 2022-06-06
Fix bug: Playback History search returns no results for uppercase letters

svobs:
- [x] It implements / fixes issue #3800.

---

**Description:**
Looks like playback history searches are meant to be case-insensitive. But `String.lowercased()` was only called for one side of the comparison.


svobs:
UPDATE: I discovered there is a better way: `String.localizedStandardContains(String)`. I have updated the fix to use this method, which in addition to being case insensitive, is also diacritic insensitive (e.g., `é` vs `ë`) and should work better for different locales.

uiryuu:
Thanks for your fix!

PR #3832, opened 2022-06-14
Fix in Install Plugin from GitHub window: Install button not enabled …

svobs:
**Description:**
`controlTextDidChange` does not fire when text changed via code - need to execute manually when setting from table entry


svobs:
Already fixed

Issue #3831, opened 2022-06-14
Multiple keys bound to the same menu item results in a random key being chosen


svobs:
**System and IINA version:**
- macOS 12.4
- IINA 1.3.0 Build 131

Take a look at the bindings for `seek 5` (mapped to both `l` and `RIGHT`), and `seek -5` (mapped to both `j` and `LEFT`):
<img width="906" alt="Screen Shot 2022-06-06 Forward Backward One" src="https://user-images.githubusercontent.com/2213815/173501044-b24db6f8-dd69-43e0-9c82-fdd3d3f0be4b.png">

<img width="906" alt="Screen Shot 2022-06-06 Forward Backward Two" src="https://user-images.githubusercontent.com/2213815/173501056-84adb2a9-dfff-4489-910c-cdb54a311d01.png">

Choose a binding which is known to show up as a menu item (like `seek 5` in this example), then add two or more lines to the input bindings mapping different keys to the same binding.

**Expected behavior:**
When a binding is mapped to multiple keys, the bottom-most-listed key should always be the one displayed in the menu

**Actual behavior:**
Which key is chosen seems random, and usually changes every time the list of bindings is changed in the Preferences (even for changes unrelated to these bindings)

**Steps to reproduce:**

1. Open IINA and go to Preferences > Key Bindings.
2. As shown above, add two or more actions for something in the menu, like `seek 5`, mapping them to different keys.
3. Observe which key is listed in the menu.
4. Remove some unrelated binding
5. Observe that a different key is listed in the menu. If not, remove another binding and observe again, and it will usually change.

- [X] MPV does not have this problem.

**How often does this happen?**
Most of the time

svobs:
This appears to be a regression via the code which was committed to fix #3692 (4abeb5b95113c9efe96f0503ff1540c633f44862). Storing the key in an unsorted dictionary did remove the duplicates, but caused the ordering to be lost.

low-batt:
Thanks for entering this issue! Not sure this is a regression?

Likely somewhere a Swift `Dictionary` is being used resulting in random ordering. If a `Dictionary` style mapping is required you might have to make use of  [OrderedDictionary](https://github.com/apple/swift-collections/blob/main/Documentation/OrderedDictionary.md).

The fix should follow mpv conventions for handling keys that have multiple different assignments in the `input.conf` file. I believe the file is treated as ordered and the first definition wins. Check that.

So in this case IINA should always pick for the menu item the key that first appears in the conf file for the associated command.

[...]

PR #3833, opened 2022-06-14
Fix regression: key for menu items randomly chosen from bindings mapped to multiple keys

svobs:
- [X] It implements / fixes issue #3831

---

**Description:**
As @low-batt has pointed out, using an `OrderedDictionary` from swift-collections would be the most elegant solution, but if adding a dependency to that is not favorable, here is a solution which uses only the built-in data structures.


svobs:
Already fixed

svobs:
Closing this in favor of #3887. The fix given here has a logic hole which allows for duplicate entries to be stored which (though not harming the functionality) is sub-optimal.

PR #3847, opened 2022-06-20
Overhaul key input handling: support runtime script bindings, key sequences, streamline event flow, fix parsing issues

### LINKED ISSUE #2140: MPV scripts: can't setup key binding

### LINKED ISSUE #3716: Add support for key sequence bindings

### LINKED ISSUE #3831: Multiple keys bound to the same menu item results in a random key being chosen

### LINKED ISSUE #3846: Cmd-period (⌘.) triggers ESC key binding in player window
**System and IINA version:**

- macOS 12.4
- IINA 1.3.0 build 132

**Steps to reproduce:**

1. Go to Preferences > Key Bindings and add/set a binding for the ESC (escape) key. Make sure there is no binding set for `Meta+.`
2. Open a video. Hold the `⌘` (command) key and press the`.` (period) key.

**Expected behavior:**
Nothing

**Actual behavior:**
The binding for the ESC key gets activated.

Always

svobs:
I've successfully reverse-engineered mpv's key input handling and reproduced the important parts so that Lua scripts can set/unset bindings in IINA as they would in mpv, following the priorities which they expect. Substantial changes had to be made to IINA's existing key handling mechanism to accomplish this. At the same time, this directly builds on the work I did to add support for key sequences (#3716) and also adds some critical fixes, rendering it obsolete, so in the interest of time I'd like to focus testing and address concerns with this issue going forward. I realize this probably won't make it into IINA all that soon, but I'd like to get the ball rolling in that direction (hopefully) and will try to keep it up to date in the meantime.

- [X] It implements / fixes issues
  - #2140 
  - #3716 
  - #3846
  - #3831
  - #3881
  - #3851

---

**Description:**

To determine what and when bindings are being set by scripts, I initially tried subscribing to `MPV_EVENT_PROPERTY_CHANGE`, and also tried polling the `input-bindings` property via `mpv_get_property`, but these proved to be inadequate. I filed a feature request with the mpv project which goes into complete detail (see: https://github.com/mpv-player/mpv/issues/10275), and based on the discussion there it seems unlikely that such event support will be added in the near future, but I did get confirmation that, in fact, all the needed information can be gathered by parsing the `MPV_EVENT_LOG_MESSAGE` event stream.

Quick summary of design:

- Adds a new class, `KeyInputController`, which encapsulates all the key binding state for a single `PlayerCore`.
- Organizes key bindings into "input sections" as mpv does (see comments at the top of `KeyInputController.swift`), with all the bindings set by IINA into the "default" section.
- `MPVController` was already subscribed to `MPV_EVENT_LOG_MESSAGE` events at the `warning` level. This increases the threshold to the `debug` level so that input section data can be extracted. The amount of log data received from mpv is still quite small and performance impact is expected to be negligible.
  - Note: by default only `warning` level events are output to the IINA log, which preserves the previous behavior.
  - Yes, there is a small risk that future updates to mpv may change the format of the logs which may break this feature, but this not super likely given its current stable state, and it shouldn't be hard to cope.
- Adds a new class, `MPVLogHandler`, to parse the log messages for the 3 relevant input section event types and send them to `KeyInputController`.
- `KeyInputController` keeps track of which input sections are defined and enabled, in what order they were enabled, and whether they are "weak" or strong. Each time there is a change to an input section or to the global key bindings set via IINA, the list of key bindings is rebuilt for the given player. It retains information about which section they originated from, which also includes the script name. It may be useful to display this info to the user at some point.
- Also made many parsing & normalization fixes as detailed in #3887: this PR is a superset of that one. These fixes were needed to get the bindings from the built-in console & stats scripts to work properly since they are using unusual (mostly lowercase) names of some keys.

### DIFF
--- a/iina.xcodeproj/project.pbxproj
+++ b/iina.xcodeproj/project.pbxproj
@@ -158,6 +158,12 @@
 		B4E446F725CB54EF0069F06E /* Mustache in Frameworks */ = {isa = PBXBuildFile; productRef = B4E446F625CB54EF0069F06E /* Mustache */; };
 		B4E4470125CE3F930069F06E /* Sparkle in Frameworks */ = {isa = PBXBuildFile; productRef = B4E4470025CE3F930069F06E /* Sparkle */; };
 		C789872F1E34EF170005769F /* InfoPlist.strings in Resources */ = {isa = PBXBuildFile; fileRef = C78987311E34EF170005769F /* InfoPlist.strings */; };
+		D140E81C285A0BEF00067094 /* LinkedList.swift in Sources */ = {isa = PBXBuildFile; fileRef = D140E81B285A0BEF00067094 /* LinkedList.swift */; };
+		D18333BC2834CDF40060A4E9 /* RingBuffer.swift in Sources */ = {isa = PBXBuildFile; fileRef = D18333BB2834CDF40060A4E9 /* RingBuffer.swift */; };
+		D18333BE2834CE620060A4E9 /* KeyInputController.swift in Sources */ = {isa = PBXBuildFile; fileRef = D18333BD2834CE620060A4E9 /* KeyInputController.swift */; };
+		D1FD2DD828521AD500CF7B35 /* MPVLogHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = D1FD2DD728521AD500CF7B35 /* MPVLogHandler.swift */; };
+		D1FD2DDA28521B5900CF7B35 /* MPVLogLevel.swift in Sources */ = {isa = PBXBuildFile; fileRef = D1FD2DD928521B5900CF7B35 /* MPVLogLevel.swift */; };
+		D1FD2DDC28533EE700CF7B35 /* MPVInputSection.swift in Sources */ = {isa = PBXBuildFile; fileRef = D1FD2DDB28533EE700CF7B35 /* MPVInputSection.swift */; };
 		E301EFDA21312AB300BC8588 /* KeychainAccess.swift in Sources */ = {isa = PBXBuildFile; fileRef = E301EFD921312AB300BC8588 /* KeychainAccess.swift */; };
 		E322A4F820A8442E00C67D32 /* PlaylistPlaybackProgressView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E322A4F720A8442E00C67D32 /* PlaylistPlaybackProgressView.swift */; };
 		E32712B024EAA14500359DAB /* ScreenshootOSDView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E32712AE24EAA14500359DAB /* ScreenshootOSDView.swift */; };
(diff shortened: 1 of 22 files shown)


svobs:
Closing because this is obsolete. Will soon replace with a much larger PR.

Issue #3861, opened 2022-07-01
Plugin: event iina.file-loaded callback type is not string


whlsxl:
**System and IINA version:**

- macOS 12.2.1
- IINA [63caeb4a81d39ef44f78747302f027503480a072](https://github.com/iina/iina/tree/63caeb4a81d39ef44f78747302f027503480a072) plugin system

**Expected behavior:**
**Actual behavior:**

`on(event: "iina.file-loaded", callback: (url: string) => void): string;`

event `iina.file-loaded` callback url type is not string, it's `URL` in swift.

PR #3862, opened 2022-07-01
Plugin: fix event.on "iina.file-loaded" url type URL => String

whlsxl:
- [x] It implements / fixes issue #3861.


lhc70000:
Thanks for your contribution!

PR #3887, opened 2022-07-22
Multiple fixes to key binding parsing

### LINKED ISSUE #3851: Using different forms of uppercase letters can cause wrong key binding to be set
**System and IINA version:**

- macOS 12.4
- IINA 1.3.0 build 132

**Steps to reproduce:**

1. Open IINA, go to Preferences > Key Bindings.
2. In the current config, make sure there are no existing bindings containing `set window-scale`.
3. Add these lines to the current config:

```
   Shift+F set window-scale 1.0
   F set window-scale 0.5
   Shift+f set window-scale 2.0
```
4. Play a video, then hold Shift and type "f"
5. Also check Video menu

**Expected behavior:**

- Video goes to double size
- Video menu has only one binding for Shift+F, and that's the last one

**Actual behavior:**

- Video goes to half size or regular size (usually)
- Same binding is set in the menu 3 times

Screen Shot 2022-06-21 at 21 40 54

### LINKED ISSUE #3881: Cannot Quit with Q-Key
Hello,

The keybinding for q to Quit doesn't work on version 1.3.0. 
It was working fine on previous versions before.

I have had trying to changed the Key Bindings configuration profiles but still can't.
Also duplicated profile and manual config to q still the same result.

- macOS 
macOS 12.4 & macOS 13 beta 3

- IINA 
 1.3.0 Build 132

**Expected behavior:**
Press Q then application Quit 

**Actual behavior:**
It doesn't do anything.

**Steps to reproduce:**
Open any video
Press q  to quit -> no action
Press shift + q to quit -> application quit

Everytime

svobs:
- [x] It implements / fixes issues:
   #3831 
   #3881 
   #3851

---

**Description:**
This patch is a subset of (and is actually back-ported from) #3847, and contains only the changes to parsing and mapping key bindings. It is the successor to #3833: it fixes a loophole in that logic which result in unused duplicate entries being stored, while also adding other fixes which couldn't be merged separately.

It does the following
- Splits `KeyMapping`'s `key` field into rawKey and `normalizedMpvKey`, similar to how its `action` field has raw & parsed version.
- Moves the existing mpv normalization logic out of the `KeyMapping` constructor and into `KeyCodeHelper`.
- Expands the mpv normalization logic to ensure that in normal form (fixes #3851):
  - Special key symbols are capitalized
  - A key's uppercase symbol is used if it is available, instead of containing "Shift+"
  - Key sequences containing the `-` (minus) key are properly parsed
- Use normal form always when identifying key bindings; use raw form when editing in the UI, to ensure consistent functionality while keeping a smooth user experience
- Add check for ignoring the line `default-bindings begin`, rather than binding it to a ke
- Add check for `{section_name}` (section name in curly braces): if it is "default" then strip the section and include the binding; for other section names do not include the binding (fixes #3881)

### DIFF
--- a/iina/AppData.swift
+++ b/iina/AppData.swift
@@ -105,9 +105,12 @@ extension Notification.Name {
   static let iinaMediaTitleChanged = Notification.Name("IINAMediaTitleChanged")
   static let iinaVFChanged = Notification.Name("IINAVfChanged")
   static let iinaAFChanged = Notification.Name("IINAAfChanged")
-  static let iinaKeyBindingInputChanged = Notification.Name("IINAkeyBindingInputChanged")
+  static let iinaInputConfListChanged = Notification.Name("IINAInputConfListChanged")
+  static let iinaCurrentInputConfChanged = Notification.Name("IINACurrentInputConfChanged")
+  static let iinaKeyBindingInputChanged = Notification.Name("IINAKeyBindingInputChanged")
   static let iinaFileLoaded = Notification.Name("IINAFileLoaded")
   static let iinaHistoryUpdated = Notification.Name("IINAHistoryUpdated")
   static let iinaLegacyFullScreen = Notification.Name("IINALegacyFullScreen")
+  static let iinaGlobalKeyBindingsChanged = Notification.Name("iinaGlobalKeyBindingsChanged")
   static let iinaKeyBindingChanged = Notification.Name("iinaKeyBindingChanged")
 }
--- a/iina/AppData.swift
+++ b/iina/AppData.swift
@@ -105,8 +105,6 @@ extension Notification.Name {
   static let iinaMediaTitleChanged = Notification.Name("IINAMediaTitleChanged")
   static let iinaVFChanged = Notification.Name("IINAVfChanged")
   static let iinaAFChanged = Notification.Name("IINAAfChanged")
-  static let iinaInputConfListChanged = Notification.Name("IINAInputConfListChanged")
-  static let iinaCurrentInputConfChanged = Notification.Name("IINACurrentInputConfChanged")
   static let iinaKeyBindingInputChanged = Notification.Name("IINAKeyBindingInputChanged")
   static let iinaFileLoaded = Notification.Name("IINAFileLoaded")
   static let iinaHistoryUpdated = Notification.Name("IINAHistoryUpdated")
--- a/iina/KeyCodeHelper.swift
+++ b/iina/KeyCodeHelper.swift
@@ -9,9 +9,22 @@
 import Foundation
 import Carbon
 
-fileprivate let modifierSymbols: [(NSEvent.ModifierFlags, String)] = [(.control, "⌃"), (.option, "⌥"), (.shift, "⇧"), (.command, "⌘")]
-
 class KeyCodeHelper {
+
+  // mpv modifiers in normal form:
+  static let CTRL_KEY = "Ctrl"
+  static let ALT_KEY = "Alt"
+  static let SHIFT_KEY = "Shift"
+  static let META_KEY = "Meta"
+
+  fileprivate static let modifierOrder: [String: Int] = [
+    CTRL_KEY: 0,
+    ALT_KEY: 1,
+    SHIFT_KEY: 2,
+    META_KEY: 3
+  ]
+
+  fileprivate static let modifierSymbols: [(NSEvent.ModifierFlags, String)] = [(.control, "⌃"), (.option, "⌥"), (.shift, "⇧"), (.command, "⌘")]
 
   static let keyMap: [UInt16 : (String, String?)] = [
     0x00: ("a", "A"),
@@ -185,10 +198,10 @@ class KeyCodeHelper {
   }()
 
   static let mpvSymbolToKeyName: [String: String] = [
-    "META": "⌘",
-    "SHIFT": "⇧",
-    "ALT": "⌥",
-    "CTRL":"⌃",
+    META_KEY: "⌘",
+    SHIFT_KEY: "⇧",
+    ALT_KEY: "⌥",
+    CTRL_KEY:"⌃",
     "SHARP": "#",
     "ENTER": "↩︎",
     "KP_ENTER": "↩︎",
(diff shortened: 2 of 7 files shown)


iina/KeyCodeHelper.swift
```diff
@@ -265,37 +278,157 @@ class KeyCodeHelper {
...
     splitted.dropLast().forEach { k in
       switch k {
-      case "Meta": modifiers.insert(.command)
-      case "Ctrl": modifiers.insert(.control)
-      case "Alt": modifiers.insert(.option)
-      case "Shift": modifiers.insert(.shift)
+      case META_KEY: modifiers.insert(.command)
+      case CTRL_KEY: modifiers.insert(.control)
+      case ALT_KEY: modifiers.insert(.option)
+        case SHIFT_KEY: modifiers.insert(.shift)
```
uiryuu, line 431: Remove the space

svobs:
I made another commit based on more recent code - this was a little out of date - and fixed typos. The only significant change was that it needed a check for the special line `default-bindings start` so it didn't look ugly in the UI (even though IINA ignores it otherwise).

svobs:
Ack - found a bug - and just committed the fix. It incorrectly parsed if the rawKey included ended in the minus key (`-`) but also had modifiers (example: `Meta+-`". Updated my tests to include that...

svobs:
Found one more corner case with key sequences that wasn't handled (`--d--`). Swift's substring handling is really clunky...

I cleaned up the dash checks in `getNextEndIndex()` and added some comments. Sorry. Final answer this time.

iina/KeyCodeHelper.swift
```diff
@@ -263,39 +276,169 @@ class KeyCodeHelper {
...
+    // this is a hard-coded special case in mpv
+    if mpvKeystrokes == "default-bindings" {
+      return [mpvKeystrokes]
+    }
+    let keystrokesList = splitKeystrokes(mpvKeystrokes)
+
+    var normalizedList: [String] = []
+    for keystroke in keystrokesList {
+      normalizedList.append(normalizeSingleMpvKeystroke(keystroke))
+    }
```
uiryuu, line 390:
```suggestion
    let normalizedList = keystrokesList.map({ normalizeSingleMpvKeystroke($0) })
```

iina/KeyCodeHelper.swift
```diff
@@ -263,39 +276,169 @@ class KeyCodeHelper {
...
+    return normalizedList
+  }
+
+  /*
+   MPV accepts several forms for the same keystroke. This ensures that it is reduced to a single standardized form
```
uiryuu, line 395:
```suggestion
   mpv accepts several forms for the same keystroke. This ensures that it is reduced to a single standardized form
```

iina/AppData.swift
```diff
@@ -105,9 +105,12 @@ extension Notification.Name {
   static let iinaMediaTitleChanged = Notification.Name("IINAMediaTitleChanged")
   static let iinaVFChanged = Notification.Name("IINAVfChanged")
   static let iinaAFChanged = Notification.Name("IINAAfChanged")
-  static let iinaKeyBindingInputChanged = Notification.Name("IINAkeyBindingInputChanged")
+  static let iinaInputConfListChanged = Notification.Name("IINAInputConfListChanged")
+  static let iinaCurrentInputConfChanged = Notification.Name("IINACurrentInputConfChanged")
```
lhc70000, line 109: These names are unused?

iina/PlayerCore.swift
```diff
@@ -282,11 +282,49 @@ class PlayerCore: NSObject {
...
-    keyMappings.forEach { keyBindings[$0.key] = $0 }
-    PlayerCore.keyBindings = keyBindings
-    (NSApp.delegate as? AppDelegate)?.menuController.updateKeyEquivalentsFrom(Array(keyBindings.values))
+    Logger.log("Set key bindings (\(keyMappings.count) mappings)")
+    // If multiple bindings map to the same key, choose the last one
+    var keyBindingsDict: [String: KeyMapping] = [:]
+    var orderedKeyList: [String] = []
+    keyMappings.forEach {
+      if $0.rawKey == "default-bindings" && $0.action.count == 1 && $0.action[0] == "start" {
+        Logger.log("Skipping line: \"default-bindings start\"", level: .verbose)
```
lhc70000, line 291: Why don't we drop the `default-bindings` line when parsing the conf file?

iina/PlayerCore.swift
```diff
@@ -282,11 +282,49 @@ class PlayerCore: NSObject {
...
+      } else {
+        if let kb = filterSectionBindings($0) {
```
lhc70000, line 293: Can be `} else if let kb = ... {`

lhc70000 (changes requested):
Regarding the mpv key sequence part: It's great to handle key sequences (and we probably should add support for them later), but since IINA doesn't support key sequences, perhaps currently we should filter them out before displaying/setting the key bindings?

uiryuu (changes requested):
Didn't test for the issues linked, hopefully that has been done by someone else. Just some suggestions based on the code.

svobs:
I pushed a new commit, non-forced so you can see the changes.

> Why don't we drop the default-bindings line when parsing the conf file?

I agree it's not pretty to put it here, but it's hard to find another place to put it. If we drop the line while parsing the file, we'll end up removing it from the file the first time the user makes any other changes to it, because of the way config files are handled. It's one thing to strip the comments out of a file, but it's bigger thing to strip valid lines out of it, and I'd rather avoid doing that (and this is basically a section start token, so the consequences would be big). There are apparently some users who use the same config files for both IINA and mpv, and they would be adversely affected by it. 

> Regarding the mpv key sequence part: It's great to handle key sequences (and we probably should add support for them later), but since IINA doesn't support key sequences, perhaps currently we should filter them out before displaying/setting the key bindings?

Fair point. I've updated the normalization code to just pass back anything with more than one`-` in it. In practice this will cause key sequences to always fail to match against typed characters (as it was doing before).

svobs:
Now I see that changes are still retained either way, so I force pushed against the original commit, for tidiness.

uiryuu (approved):

PR #3891, opened 2022-07-28
Fix #3857: Enter Full Screen exists under 2 different menu items

### LINKED ISSUE #3857: The Full-screen Option Exists As 2 Instances Under Different Menu Items.
**System and IINA version:**

- macOS 10.14
- IINA 1.3.0

**Expected behaviour:**

Should be only 1 full-screen mode option. 2 - is redundant.

**Actual behaviour:**

There're 2 duplicates each toggling the full screen.

**Steps to reproduce:**
 See the screenshot.

Always
iina_2-full-screen
.

svobs:
- [x] It implements / fixes issue #3857.

---

**Description:**
Since the "Enter Fullscreen" option already exists under the Video menu, this patch hides the instance under the Window menu, to conform with Apple HIG. Note:
- This change has no effect on the built-in `fn-F` shortcut, which still works.
- This is set as a preference value which MacOS recognizes, which means:
  - Although it's set when IINA starts, it will not take effect until the app is closed & reopened.
  - After being set, even old versions of IINA will hide the full-screen option in the Window menu (unless IINA's prefs are manually deleted or edited to remove this key)

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -128,6 +128,9 @@ class AppDelegate: NSObject, NSApplicationDelegate {
       guideWindow.show(pages: [.highlights])
     }
 
+    // Hide Window > "Enter Full Screen" menu item, because this is already present in the Video menu
+    UserDefaults.standard.set(false, forKey: "NSFullScreenMenuItemEverywhere")
+
     SUUpdater.shared().feedURL = URL(string: Preference.bool(for: .receiveBetaUpdate) ? AppData.appcastBetaLink : AppData.appcastLink)!
 
     // handle arguments


iina/AppDelegate.swift
```diff
@@ -201,6 +201,9 @@ class AppDelegate: NSObject, NSApplicationDelegate {
       }
     }
 
+    // Hide Window > "Enter Full Screen" menu item, becuase this is already present in the Video menu
```
CarterLi, line 204: because

iina/AppDelegate.swift
```diff
@@ -201,6 +201,9 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
+    UserDefaults.standard.set(false, forKey: "NSFullScreenMenuItemEverywhere")
```
low-batt, line 205: This is the wrong place to set this. That is what is causing this:
> Although it's set when IINA starts, it will not take effect until the app is closed & reopened.

The only Apple documentation on this seems to be [AppKit Release Notes (macOS 10.12 and Earlier)](https://developer.apple.com/library/archive/releasenotes/AppKit/RN-AppKitOlderNotes/index.html#10_11FullScreen) which says:
>AppKit automatically creates an "Enter Full Screen" menu item after the application finishes launching if an equivalent menu item isn't found. If this menu item should not be created for your app, before NSApplicationDidFinishLaunchingNotification is sent you may set the NSFullScreenMenuItemEverywhere default to NO.

This statement has been added to the method `applicationDidFinishLaunching` which is after the `NSApplicationDidFinishLaunchingNotification` notification. I tested moving this statement into the `applicationWillFinishLaunching` method right after the "App will launch" log is emitted. I used `defaults` to make sure that key was not set. The menu item was hidden on the first run. I did not have to restart IINA for it to take effect.

saagarjha:
I think the correct solution here should be to get AppKit to recognize our fullscreen menu item as the legitimate one, rather than using the user default to suppress the addition of the second one in the Window menu.

svobs:
I've force-re-submitted an improved commit based on the above feedback, though honestly @low-batt should get full credit, since he did do more than 50% of the work for this one-liner.

> I think the correct solution here should be to get AppKit to recognize our fullscreen menu item as the legitimate one, rather than using the user default to suppress the addition of the second one in the Window menu.

@saagarjha I did a deep-dive into this...

Of the affected open source projects I found, the majority seem to have settled on the above preference setting ([Electron](https://github.com/electron/electron/issues/3038) being the biggest name I found) but I did discover that the Chromium was able to trigger Apple's matching logic by changing the name of the menu item action to `toggleFullScreen` and is[ using that](https://bugs.chromium.org/p/chromium/issues/detail?id=538872).

I was able to do the same in [this branch](https://github.com/iina/iina/compare/develop...svobs:iina:pr-hide-redundant-fullscreen-menu-item-ALT), but there are still some problems. I was only able to trigger the override by hard-coding the action selector in the XIB file. But I still can't figure out how to create a selector for MainWindowController inside the XIB -- IINA is currently doing that in code after the app has started. Best I could do was to make the XIB call an action in MenuController, which in turn calls the action in MainWindowController. And everything works this way - *except* the "Enter Full Screen" menu item is never grayed out because it's using a MenuController selector instead of a MainWindowController selector, and I'm not ready to commit a lot more time to figure out how to make that work.

IMHO it seems like using a special selector name is the most robust solution, especially because I haven't found any documentation from Apple to support it. At least the preference key is documented.

low-batt:
@svobs Full credit goes to you for providing the implementation. With Apple specifying a clear requirement on how the menus must be structured, but apparently no documentation on how to implement it, I thought this needed a detailed review.

Due to the feedback from @saagarjha I looked again for guidence from Apple on how to implement the HIG requirement of suppressing the menu item in the `Window` menu. Nothing useful.

One reason why developers fail to document things is because "it is obvious". So I tried a quick experiment. I first made sure to delete the preference:
```text
defaults delete com.colliderli.iina NSFullScreenMenuItemEverywhere
```

The documentation for [NSWindow.toggleFullScreen](https://developer.apple.com/documentation/appkit/nswindow/1419527-togglefullscreen/) says:
> If an application supports fullscreen, it should add a menu item to the View menu with toggleFullScreen: as the action, and nil as the target.

IINA's main menu does not contain a "Enter Full Screen" menu item. In a quick test I added that menu item to the `Window` menu and tied the action to `toggleFullScreen` in `First Responder` and checked `Hidden`. That quick test seemed to work. Busy today, so I only tested Monterey. 

Given the only documentation for `NSFullScreenMenuItemEverywhere` is from that archived old AppKit release notes from 10.12, maybe adding a menu item as a solution should be further explored?

I followed this post in order to properly attach the action:
[How to connect IBAction and menu item to First Responder?](https://stackoverflow.com/questions/37168040/how-to-connect-ibaction-and-menu-item-to-first-responder)

If the menu item action is not attached to `toggleFullScreen` then AppKit does not recognize it as the standard `Enter Full Screen` menu item.

saagarjha:
My understanding is that using the toggleFullScreen: action is the correct way to do this, yes. Is it not being propagated correctly through the responder chain?

low-batt:
The responder chain is not being used. The full screen menu item under the `Video` menu is tied to `MenuController` which sets the action to `MainWindowController.menuToggleFullScreen`. That calls the method `toggleWindowFullScreen` which will either call `NSWindow.toggleFullScreen` or legacy full screen methods in `MainWindowController`.

The task at hand is to eliminate the full screen menu item under the `Window` menu. Right now IINA does not have such a menu item. AppKit is adding a disabled one. By adding a full screen menu item under the `Window` menu, using `NSWindow.toggleFullScreen` and marking that menu item as hidden we can eliminate the disabled `Enter Full Screen` as required by the HIG.

Issue #3849, opened 2022-06-20
Total time / remaining time display update


decodism:
**System and IINA version:**

- macOS 13.0 beta
- IINA 1.3.0

**Expected behavior:**

When the total time / remaining time is clicked, they should switch.

**Actual behavior:**

If the video is paused, the time display is only updated when the video is resumed.

**Steps to reproduce:**

- Pause a video
- Click on the total time / remaining time

**How often does this happen?**

Always

low-batt:
I reproduced the behavior. Seems wrong to me. I will propose a fix when I get a chance.

decodism:
It seems the issue has been mostly solved by #3602. However, there is still an edge case: position the cursor on the time display, pause with the keyboard, click on the time display without moving the cursor. I have updated my PR accordingly.

low-batt:
I reproduced the problem running a build from develop and following the instructions to pause using the keyboard.

PR #3913, opened 2022-08-17
Update rightLabel when clicked while player paused

decodism:
- [x] It implements / fixes issue #3849.

### DIFF
--- a/iina/DurationDisplayTextField.swift
+++ b/iina/DurationDisplayTextField.swift
@@ -16,7 +16,9 @@ class DurationDisplayTextField: NSTextField {
   }
 
   static var precision : UInt = UInt(Preference.integer(for: .timeDisplayPrecision))
-  var mode: DisplayMode = .duration
+  var mode: DisplayMode = .duration { didSet { updateText() } }
+  var duration: VideoTime = .zero
+  var current: VideoTime = .zero
 
   /** Switches the display mode between duration and remaining time */
   func switchMode() {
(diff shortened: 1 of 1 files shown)


iina/MPVController.swift
```diff
@@ -1025,7 +1025,7 @@ class MPVController: NSObject {
             // video is paused to avoid wasting energy with needless processing. If paused shutdown
             // the timer that synchronizes the UI and the high priority display link thread.
             if paused {
-              player.invalidateTimer()
+              if ![.willShow, .shown].contains(player.mainWindow.animationState) { player.invalidateTimer() }
```
low-batt, line 1028: This fixes the problem, but is undesirable from an energy use perspective. We absolutely do not want the timer to run when playback is paused and therefore the video position is not changing.

low-batt (changes requested):
The timer that updates the UI is still running when it should not, wasting energy. That is why this problem seemed to be fixed.  The fix for this issue should not rely upon that timer running. We want to get to the point where that timer is only run when playback is active and the OSC is visible. We have been moving in that direction, but more changes are needed. So this fix can not depend up the timer updating the label.

I suggest the fix follow the normal pattern for handling changes in a control. 

One way to do this would be to add a `didSet` property observer to the `DurationDisplayTextField` `mode` property. This observer would call `updateText` to update the `stringValue` `NSTextField` property. Then the observer would set [needsDisplay](https://developer.apple.com/documentation/appkit/nsview/1483360-needsdisplay) to true.

To be able to call `updateText` `DurationDisplayTextField` needs to know the duration and position. That means adding properties and changing `updateText` to remember the values used to generate the string.

Something like that.

decodism:
I reverted to my previous fix and implemented your suggestion.

low-batt (approved):
Changes look good. I pulled the PR locally and confirmed the fix works.

uiryuu (approved):

uiryuu:
Thanks for your PR!

Issue #3911, opened 2022-08-14
Top of progress bar not behaving as such


decodism:
**System and IINA version:**

- macOS 13.0 beta
- IINA 1.3.0

**Expected behavior:**

Clicking on the progress bar (on the height of its handle) should change the playback time.

**Actual behavior:**

Clicking on the top of the progress bar (on the height of its handle) does not change the playback time and dragging while holding down moves the control panel.

**Steps to reproduce:**

- Click on the top of the progress bar

**How often does this happen?**

Always

decodism:
@low-batt maybe you could also look at this?

low-batt:
I believe this screenshot shows the issue being reported:
![issue-3911-seek](https://user-images.githubusercontent.com/86170219/184557404-383d11fc-265c-454e-851e-2e43ec840e3c.jpg)

Notice the pointer is below the height of the seek bar's knob. Given that position the user would expect that clicking would seek playback to that position. Instead it initiates repositioning of the on screen controller (OSC).

The above screenshot was taken under Monterey 12.4. The same problem can be seen under Catalina 10.15.7, but not quite as bad. [NSSlider](https://developer.apple.com/documentation/appkit/nsslider) was changed by Apple for Big Sur. That may have made this problem slightly worse. I'm assuming Ventura behavior matches Monterey.

Another concern seen in the screenshot is that the time label overlaps the top of the seek bar's knob.

I will look into this.

decodism:
Thank you for investigating this issue.

decodism:
> Another concern seen in the screenshot is that the time label overlaps the top of the seek bar's knob.

That was the issue. The label was intercepting the clicks. Sending it to the back fixes the issue.

low-batt:
That fully exposes the slider, but then the slider overlaps part of the time.

I'm close to posting a full analysis explaining what went wrong along with a more involved fix.
I might be able to finish it and post it tomorrow.

This is an important issue. Thank you for taking the time to report it!

low-batt:
I have labeled this issue as a bug because the code is not working as intended.
I have labeled this issue as a high priority because it is a regression in macOS 11+.

### Analysis

Changing the `NSSlider` to have a green background makes it easier to see that the `NSTextField` showing the time intrudes into the slider when running under macOS 12.4:
![issue-3911-green](https://user-images.githubusercontent.com/86170219/185295031-7e7201d7-179a-41e1-82c8-b9113b6c86e5.jpg)


[...]

Issue #3863, opened 2022-07-05
Enable scroll wheel scrubbing with Controller UI in Top position


PilotiE36:
**System and IINA version:**

- macOS  10.14.6
- IINA  1.30

**Expected behavior:**

The scroll wheel should scrub the video when the pointer is hovering over the controller, for all controller layout options.

**Actual behavior:**

When the Top layout is chosen as the controller UI option, scrubbing does not occur as it does when the Bottom or Floating layout is chosen.

**Steps to reproduce:**

Set UI position to Top, open video file, hover pointer over controller, and attempt to scrub using the scroll wheel.

**How often does this happen?**

Always.  This issue was present in previous versions as well, and 1.30 exhibits the same behavior.

low-batt:
Reproduced for me with IINA 1.3.0 running on macOS 12.4 on a M1 Mac.

This is a defect. Such scroll wheel behavior should not be dependent upon the on screen controller layout specified in preferences.

From a very brief look at the code, the scroll wheel event is being dropped by this guard statement in `MainWindowController`:
```swift
override func scrollWheel(with event: NSEvent) {
  guard !isInInteractiveMode else { return }
  guard !isMouseEvent(event, inAnyOf: [sideBarView, titleBarView, subPopoverView]) else { return }
```

The mouse event is found to be in `titleBarView` because in `MainWindowController.setupOnScreenController` the height of that view is extended to include the OSC when top layout is configured:
```swift
titleBarHeightConstraint.constant = TitleBarHeightWithOSC
```

I'm not sure what the best fix is. Hopefully someone more familiar with the UI will chime in.

low-batt:
This problem is fixed in the development version of IINA. The fix is expected to be included in 1.4.0-beta2.

PR #3915, opened 2022-08-18
Fix top of progress bar not behaving as such, #3911

low-batt:
- Change MainWindowController methods playSliderChanges and
  updateTimeLabel to position the time preview relative to the progress
  bar
- Add method canShowThumbnailAbove to MainWindowController
- Change updateTimeLabel to use canShowThumbnailAbove
- Change PlaySliderCell.awakeFromNib to set slider control size to small
  for macOS 11+
- Increase bottom and top constraints around progress bar from 4.5 to 6

These changes stop the time preview from intruding into the progress bar
in the on screen controller.

- [x] It implements / fixes issue #3911.

---

**Description:**
Correct regression introduced by Big Sur.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -2059,11 +2059,35 @@ class MainWindowController: PlayerWindowController {
     }
   }
 
+  /// Determine if the thumbnail preview can be shown above the progress bar in the on screen controller..
+  ///
+  /// Normally the OSC's thumbnail preview is shown above the time preview. This is the preferred location. However the
+  /// thumbnail preview extends beyond the frame of the OSC. If the OSC is near the top of the window this could result
+  /// in the thumbnail extending outside of the window resulting in clipping. This method checks if there is room for the
+  /// thumbnail to fully fit in the window. Otherwise the thumbnail must be displayed below the OSC's progress bar.
+  /// - Parameters:
+  ///   - timnePreviewYPos: The y-coordinate of the time preview `TextField`.
+  ///   - thumbnailHeight: The height of the thumbnail.
+  /// - Returns: `true` if the thumbnail can be shown above the slider, `false` otherwise.
+  private func canShowThumbnailAbove(timnePreviewYPos: Double, thumbnailHeight: Double) -> Bool {
+    guard oscPosition != .bottom else { return true }
+    guard oscPosition != .top else { return false }
+    // The layout preference for the on screen controller is set to the default floating layout.
+    // Must insure the top of the thumbnail would be below the top of the window.
+    let topOfThumbnail = timnePreviewYPos + timePreviewWhenSeek.frame.height + thumbnailHeight
+    // Normally the height of the usable area of the window can be obtained from the content
+    // layout. But when the legacy full screen preference is enabled the layout height may be
+    // larger than the content view if the display contains a camera housing. Use the lower of
+    // the two heights.
+    let windowContentHeight = min(window!.contentLayoutRect.height, window!.contentView!.frame.height)
+    return topOfThumbnail <= windowContentHeight
+  }
+
   /** Display time label when mouse over slider */
   private func updateTimeLabel(_ mouseXPos: CGFloat, originalPos: NSPoint) {
-    let timeLabelXPos = playSlider.frame.origin.y + 15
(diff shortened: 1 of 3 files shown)


low-batt:
See the issue for a detailed discussion of the issue and the proposed fix.

iina/MainWindowController.swift
```diff
@@ -2059,11 +2059,35 @@ class MainWindowController: PlayerWindowController {
     }
   }
 
+  /// Determine if the thumbnail preview can be shown above the progress bar in the on screen controller..
+  ///
+  /// Normally the OSC's thumbnail preview is shown above the time preview. This is the preferred location. However the
+  /// thumbnail preview extends beyond the frame of the OSC. If the OSC is near the top of the window this could result
+  /// in the thumbnail extending outside of the window resulting in clipping. This method checks if there is room for the
```
saagarjha, line 2066: I feel like the long-term solution here would be to pull the thumbnail into its own window.

saagarjha (commented):
Another comment I have is that it seems like the text field should just have user interaction disabled

iina/MainWindowController.swift
low-batt: On the text field and user interaction, in Xcode the behavior attribute, which can be selectable, editable or none, is set to none. I think user interaction is already disabled? So the mouse click goes up the responder chain and is handled by `ControlBarView`.

If the thumbnail was it its own window how would what work in full screen mode?

And of course I ignored the question of whether IINA should increase the size of its sliders in the spirt of Apple's Big Sur changes. For this fix I tried to make the minimum changes required to restore what I think was the intended behavior. Part of the reason for limiting the fix is due to my lack of UI experience.

saagarjha: Hmm, my understanding is that if the view does not have user interaction enabled it should just remove itself from hit testing, causing whatever's under it to get the event. Is that not happening?

Also, there are various window flags that allow windows to show up in fullscreen mode.

low-batt: I found a `UIView` [hitTest](https://developer.apple.com/documentation/uikit/uiview/1622469-hittest) method. Not sure what the equivalent AppKit method is. That method does take into account the state of user interactions:
> This method ignores view objects that are hidden, that have disabled user interactions, or have an alpha level less than 0.01.

However I have not found anything about the mouse event being delivered to the view underneath. The UIKit method refers to traversing the view higharchy. In the case at hand the superview is the OSC. This results in the `mouseDown` method of `ControlBarView` being called which initiates dragging of the OSC. That was as far as I traced the behavior as I realized the intention was that these OSC elements not overlap.

On converting the screenshot preview to use a `NSWindow`, although I am usually in favor of adding some cleanup/refactoring in this case I'd rather postpone that past the planned bug fix release. There are a lot of issues that it would be nice to fix in the bug fix release that still need PRs. For example issue #3863. I've tracked down what is going wrong. I was hoping to hear why the code is treating the OSC as a part of the title bar before I remove that. And of course I really need to get back to reviewing the plugin system as that is blocking the creation of PRs to avoid generating merge conflicts.

If you really want me to I'll see if I can switch screenshot previews to use a `NSWindow`.

low-batt: Found the AppKit [hitTest](https://developer.apple.com/documentation/appkit/nsview/1483364-hittest) method. Unlike the UIKit method the documentation for the AppKit method only mentions ignoring hidden views, nothing about user interaction.

saagarjha: I'll ask around but I would be very surprised if views that have user interaction disabled do anything meaningful on clicks

lhc70000: FYI, I used `hitTest` in the plugin overlay view: https://github.com/iina/iina/blob/c9dcd4061ca14e0f0eb0b41d9a1bba88daf85f92/iina/PluginOverlayView.swift#L17

Overriding this method and returning `nil` should disable all interactions.

low-batt: Yes. But no need to do that for this issue?

The root cause of the problem is the miss-alignment of UI components caused by Apple increasing the size of sliders in Big Sur. Everything works as intended once the alignment problems are corrected.

low-batt: Be sure to look at the screenshots in the issue. They show the OSC before the changes in this commit and after.

uiryuu:
Tested on macOS 10.15, seems fine.

uiryuu (approved):

Issue #3842, opened 2022-06-17
🔴 IINA 1.3.0 BUG: Error Cannot prevent display sleep!


Thomas-Harrell:
**System and IINA version:**

- macOS Big Sur 11.6.2
- IINA 1.3.0

**Expected behavior:**

**Actual behavior:**

- After running IINA for a period of time, about 2-3 days, there is a high probability of appearing

- After this error, opening the media file, continuing playback after a pause, and dragging the progress bar will all result in this error

- Restore settings and reinstall software did not fix it

- After restarting the computer, the problem can be solved, but after a period of operation, it will reopen again

![SCR-20220617-fif](https://user-images.githubusercontent.com/106761803/174218360-a31c70df-635c-491f-9257-11c7a95127ac.png)

<details>
<summary><b>Crash report:</b></summary>

```
Error Cannot prevent display sleep!
```
</details>

<br>

<details>
<summary><b>mpv log:</b></summary>

```
```
</details>

<br>

**Steps to reproduce:**

- After running IINA for a period of time, about 2-3 days, there is a high probability of appearing

- [x] MPV does not have this problem.

svobs:
It looks like IINA should be logging the details of the error. Can you try to enable logging (Preferences > Advanced > Enable logging, the quit & reopen IINA) and then if it happens again, grab the iina.log file and attach it here?

low-batt:
## Summary

This is not a bug in IINA. This is how IINA informs you that **macOS power management is broken** on the Mac IINA is being run on.

Please take the follow actions:

Check if you have any products from [Rogue Amoeba](https://rogueamoeba.com/) installed on your Mac. If any are found check to see if there are pending updates to those products that have not been installed. It is critical that Rogue Amoeba's Audio Capture Engine (ACE) framework is up to date. 

The ACE framework may be included in products from other companies as discussed in the post [Licensing Rogue Amoeba's ACE SDK](https://rogueamoeba.com/licensing/ace/). If you do not have products from Rogue Amoeba installed check for any audio related applications and see if they use the ACE framework. Make sure any audio related applications are up to date.

Check if you have Zoom installed on your Mac. If you do and are familiar with the history of security problems (see articles like tom's guide [Zoom security issues: What's gone wrong and what's been fixed](https://www.tomsguide.com/news/zoom-security-privacy-woes)) and still want to use it, check to see if there are pending updates that have not been installed. Make certain the latest version is installed.

Please let us know if these actions resolve the issue.

[...]

PR #3924, opened 2022-08-23
Fix Error Cannot prevent display sleep!, #3842

low-batt:
- Change SleepPreventer.preventSleep to only display an alert once per
  IINA invocation
- Change Utility.showAlert to support a suppression button
- Add a button to the alert to allow the user to permanently suppress
  the alert
- Change the alert to include the error code returned by macOS

Additional text was added to the alert's message requiring localization.

This fixes how IINA reports a macOS power management failure.
The root cause of the failure is in macOS and must be fixed by Apple.

- [x] It implements / fixes issue #3842.

---

**Description:**
This commit includes new text that needs localization.
See the issue for screenshots of the new alert.


iina/SleepPreventer.swift
```diff
@@ -15,8 +15,44 @@ class SleepPreventer: NSObject {
 
   static private var assertionID = IOPMAssertionID()
 
+  static private var haveShownAlert = false
+
   static private var preventedSleep = false
 
+  /// Ask macOS to not dim or sleep the display.
```
saagarjha, line 22: A brief description is fine but I'd strongly recommend just linking to the issues themselves rather than copying their contents here, because they'll remain up-to-date and are a better place to put extended discussion

iina/SleepPreventer.swift
```diff
@@ -28,11 +64,21 @@ class SleepPreventer: NSObject {
                                               &assertionID)
     if success == kIOReturnSuccess {
```
saagarjha, line 65: It might be worth converting this to a guard

iina/Base.lproj/Localizable.strings
```diff
@@ -143,7 +143,7 @@
 
 "alert.fatal_error" = "Fatal error: %@ \nThe application will exit now.";
 "alert.mpv_error" = "Internal error %@ (%@) when setting option %@.";
-"alert.sleep" = "Cannot prevent display sleep!";
+"alert.sleep" = "Cannot prevent display sleep!\nmacOS error 0x%0X8";
```
saagarjha, line 146: I think I'd prefer a "Cannot prevent display sleep! (error code)" format, as this better matches how macOS displays errors

iina/SleepPreventer.swift
low-batt: That came from my bias for keeping information with the code after encountering broken links on work projects. I've changed that portion of the comment to just refer to the two issues.

iina/SleepPreventer.swift
low-batt: Done.

iina/Base.lproj/Localizable.strings
low-batt: I first considered doing it like that and changed to this to put emphasis on this being a macOS problem. With it now not adding any words to the string should I be updating more of the lanuages, or just base and EN? The alert now looks like:
issue-3842-review-change

iina/Utility.swift
```diff
@@ -70,12 +71,24 @@ class Utility {
       alert.informativeText = String(format: format)
     }
 
+    if let suppressionKey = suppressionKey {
```
uiryuu, line 74: Can we move the check right after `let alert` line?

uiryuu (changes requested):
Since it's first time we have alert suppression, I think we should also add a "Reset all dialog warnings" function, just like other apps.

Music.app:
image

Xcode:
image

iina/Utility.swift
low-batt: Good catch. Will move.

low-batt: On the resetting of suppressed alerts, I will work on adding a new section to Preferences/Utilities that looks like:

---

#### Restore Alerts

The button below will restore all alerts that have been suppressed using the "Do not show this message again" checkbox.
`Restore Suppressed Alerts`

iina/Base.lproj/Localizable.strings
saagarjha: lgtm. I prefer to not pass judgement on where the error is coming from, in general we should just present it as it is.

iina/SleepPreventer.swift
saagarjha: We are not likely to migrate to JIRA anytime soon :P

iina/Utility.swift
low-batt: The latest commit adds a `Restore Suppressed Alerts` button in the `Utilities` section of `Preferences` giving users the ability to easily change IINA back to showing alerts. Remember I am a total newbie when it comes to GUI programming. Be on the lookout for newbie mistakes.

The `Utilities` section of `Preferences` now looks like:
issue-3842-util

Clicking the button asks for confirmation:
issue-3842-util-alert

If `OK` is clicked `Restored.` is shown similar to other buttons in `Utilities`:
issue-3842-restored

uiryuu (approved):

uiryuu:
LGTM

PR #3938, opened 2022-09-09
Fix NSFileHandleOperationException crash in logger, #3590

low-batt:
- Add a new Lock class
- Change Logger to use a lock to coordinate closing of the log file
- Change Logger to not use `Utilities` methods that call back into the logger
- Change Logger to catch and handle exceptions thrown by file operations
- Remove logDirURL from Utilities
- Change PrefAdvancedViewController to use Logger.logDirectory instead of Utilities.logDirURL

These changes address multiple ways in which the logger can cause IINA to crash.

- [x] It implements / fixes issue #3590.

---

**Description:**
A new proposed fix for this issue that directly addresses the logger crash.

### DIFF
--- a/iina/Lock.swift
+++ b/iina/Lock.swift
@@ -1,0 +1,83 @@
+//
+//  Lock.swift
+//  iina
+//
+//  Created by low-batt on 9/5/22.
+//  Copyright © 2022 lhc. All rights reserved.
+//
+
+import Foundation
+
+/// An object that coordinates the operation of multiple threads of execution.
+///
+/// This class hides the need for different implementations depending upon the macOS version.
+/// An [os_unfair_lock](https://developer.apple.com/documentation/os/os_unfair_lock) is preferred for its efficiency,
+/// however it was introduced in macOS 10.12, so an
+/// [NSLock](https://developer.apple.com/documentation/foundation/nslock) is used when running under earlier
+/// macOS versions.
+///
+/// The interface exposed is patterned after
+/// [OSAllocatedUnfairLock](https://developer.apple.com/documentation/os/osallocatedunfairlock)
+/// which is expected to be included in macOS 13.
+///
+///- Warning: This isn’t a recursive lock. Attempting to lock an object more than once from the same thread without unlocking in
+///    between will either trigger a runtime exception (macOS 10.12+) or will block your thread permanently (macOS 10.11).
+class Lock {
+
+  private let lock: LockImpl
+
+  init() {
+    if #available(macOS 10.12, *) {
+      lock = OSUnfairLockImpl()
+    } else {
+      lock = NSLockImpl()
+    }
+  }
+
+  /// Executes a closure while holding a lock.
+  /// - Parameter body: A closure that contains the code to execute using the lock.
+  /// - Returns: The value that body returns.
+  /// - Throws: The exception that body throws.
+  func withLock<R>(_ body: () throws -> R) rethrows -> R {
+    return try lock.withLock(body)
+  }
+}
+
+// MARK: - Implementation
+
+private protocol LockImpl {
+  func withLock<R>(_ body: () throws -> R) rethrows -> R
+}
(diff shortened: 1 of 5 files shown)


low-batt:
The new `Lock` class in the commit is the one I posted a preview of in the plugin-system PR. There were two comments on that:

>Looks reasonable, except I'd make two changes: make the underlying lock type a typealias rather than a generic type, and using deinitialize with no parameters.

I corrected the lock property declaration to use `os_unfair_lock_t`. I was unable to remove the parameter in the call to `deinitialize` as that resulted in the compilation error shown below. That version of `deinitialize` was removed in Swift 5.
```text
/Users/low-batt/Documents/builds/iina/iina/iina/Lock.swift:63:10: error: 'deinitialize()' is unavailable: the default argument to deinitialize(count:) has been removed, please specify the count explicitly
    lock.deinitialize()
         ^~~~~~~~~~~~
Swift.UnsafeMutablePointer:6:17: note: 'deinitialize()' was obsoleted in Swift 5.0
```

low-batt:
The new plugin system is triggering issue #3590, so it is desirable to get this PR reviewed and merged. Another reason to get this merged early is that the new `Lock` class is needed for other fixes.

The fix attempts to insure that the logger will never crash IINA.

The problems with IINA's termination sequence the original PR attempted to address will be fixed for issue #3939. I'm nearly completion on a fix for that. For the moment I'm turning my attention to other issues now that the plugin system has been merged, but will come back to fixing the termination sequence soon.

iina/Lock.swift
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  func withLock<R>(_ body: () throws -> R) rethrows -> R {
+    os_unfair_lock_lock(lock)
+    defer { os_unfair_lock_unlock(lock) }
+    return try body()
+  }
+}
+
+private class NSLockImpl: LockImpl {
```
saagarjha, line 74:
```suggestion
private struct NSLockImpl: LockImpl {
```

iina/Logger.swift
```diff
@@ -45,31 +56,97 @@ struct Logger {
   static let enabled = Preference.bool(for: .enableAdvancedSettings) && Preference.bool(for: .enableLogging)
 
   static let logDirectory: URL = {
+    // get path
+    let libraryPath = FileManager.default.urls(for: .libraryDirectory, in: .userDomainMask)
+    guard libraryPath.count >= 1 else {
+      fatalDuringInit("Cannot get path to Logs directory: \(libraryPath)")
+    }
+    let logsUrl = libraryPath.first!.appendingPathComponent("Logs", isDirectory: true)
```
saagarjha, line 64:
```suggestion
    let libraryPaths = FileManager.default.urls(for: .libraryDirectory, in: .userDomainMask)
    guard libraryPath = libraryPaths.first else {
      fatalDuringInit("Cannot get path to Logs directory: \(libraryPath)")
    }
    let logsUrl = libraryPath.appendingPathComponent("Logs", isDirectory: true)
```

saagarjha (approved):

iina/Lock.swift
```diff
@@ -0,0 +1,83 @@
...
+// MARK: - Implementation
+
+private protocol LockImpl {
+  func withLock<R>(_ body: () throws -> R) rethrows -> R
+}
+
+@available(macOS 10.12, *)
+private class OSUnfairLockImpl: LockImpl {
+
+  // Use a pointer to insure the lock, which is a struct, is not copied.
```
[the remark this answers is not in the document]
low-batt: Fixed typo.

iina/Lock.swift
low-batt: Good catch. Fixed.

iina/Logger.swift
low-batt: This was code I moved out of Utility and didn't cleanup. Fixed.

PR #3957, opened 2022-09-29
Improve the file api performance for plugin system

zurl:

---

**Description:**

The current plugin file system read api is very slow, it takes ~8s to read 16MB file. The bottleneck is createUInt8Array which are copying in JavaScript.

In this PR we can directly use JavaScriptCore API JSObjectMakeTypedArrayWithBytesNoCopy to create type array directly in Swift, which is much faster

Test on Mac Studio (M1 Max), read 16 MB File to javascript
Before Change: 7338ms
After Change: 3.94 ms


lhc70000:
Thank you very much!

PR #3961, opened 2022-10-02
Fix random failure to resume playback position, #3939

low-batt:
This commit will:
- Add isStopped and isShutdown properties to PlayerCore
- Rename PlayerCore property isMpvTerminated to isShuttingDown
- Add playbackStopped and mpvHasShutdown methods to PlayerCore
- Add savePlayerState method to PlayerCore
- Add iinaPlayerStopped and iinaPlayerShutdown notifications
- Change PlayerCore to post the new notifications
- Change savePlaybackPosition to not send a command to write the watch later file if playback has been stopped
- Move code in MainWindowController method windowWillClose related to stopping playback to the PlayerCore stop method
- Change MPVController to inform PlayerCore when the stop command finishes and when mpv shuts down
- Change MPVController to remove IINA preference observers before sending a quit command to mpv
- Change MPVController to not use the main dispatch queue when calling NSApp.terminate to avoid blocking the queue
- Add an isDisabled property to MenuController
- Add disableAllMenus and disableAllMenuItems methods to MenuController
- Change menuWillOpen to not update menus if menus are disabled
- Add a disableAllCommands method to RemoteCommandController
- Add isTerminated property to AppDelegate
- Add a removeAllMenuItems method to AppDelegate
- Change AppDelegate method applicationShouldHandleReopen to not permit reopening during termination
- Change AppDelegate method applicationShouldTerminate to return terminateLater to Cocoa to delay termination until mpv has finished shutting down
- Change applicationShouldTerminate to shutdown further user input and coordinate application shutdown using observers for the new iinaPlayerStopped and iinaPlayerShutdown notifications

From a high level perspective these changes:
- Stop further input from the user once application termination starts
- Disallow reopening during termination
- Wait for the asynchronous mpv stop command to complete before sending the mpv quit command
- Wait for the asynchronous quit command to complete before allowing Cocoa to proceed with termination

These changes are needed because mpv may decide to write to the watch later file and then fail to store the playback position if the quit command is sent to mpv while a stop command is still executing. This breaks the "Resume last playback position" IINA feature. To correct this IINA must move to a controlled clean shutdown.

- [x] It implements / fixes issue #3939.

---

**Description:**
This proposed fix replaces the original thread based fix with one that is event driven.

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -41,6 +41,8 @@ class AppDelegate: NSObject, NSApplicationDelegate {
 
   private var commandLineStatus = CommandLineStatus()
 
+  private var isTerminating = false
+
   // Windows
 
   lazy var openURLWindow: OpenURLWindowController = OpenURLWindowController()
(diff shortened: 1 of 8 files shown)


low-batt:
This proposed fix is the same fix in closed PR #3952 from a high level perspective. But the new code is event driven eliminating the background thread that was being used by `AppDelegate` and replacing it with listening for new notifications being posted by `PlayerCore`. The updated code also queues tasks to the main dispatch queue so that access to the new flags always occurs from the main thread. This eliminates the need for special handling to coordinate access to the flags from multiple threads.

iina/AppDelegate.swift
```diff
@@ -296,13 +298,136 @@ class AppDelegate: NSObject, NSApplicationDelegate {
...
+      // command to mpv while it is still asynchronously executing the stop command can result in a
+      // watch later file that is missing information such as the playback position. See issue #3939
+      // for details.
+      player.shutdown()
+    }
+    observers.append(observer)
+
+    // Establish an observer for a player core shutting down.
+    observer = center.addObserver(forName: .iinaPlayerShutdown, object: nil, queue: .main) { _ in
+      guard !timedOut else {
```
uiryuu, line 390: I don't see the meaning of this guard and the `timeOut` variable. Soon after `timeOut` is set to true, the application is terminated forcibly, no?

iina/MPVController.swift
```diff
@@ -760,20 +774,53 @@ class MPVController: NSObject {
...
+        // Submit the following task synchronously to ensure it is done before application
+        // termination is started.
         DispatchQueue.main.sync {
-          NSApp.terminate(nil)
+          self.player.mpvHasShutdown(isMPVInitiated: true)
+        }
+        // Initiate application termination. AppKit requires this be done from the main thread,
+        // however the main dispatch queue must not be used to avoid blocking the queue as per
+        // instructions from Apple.
+        if #available(macOS 10.12, *) {
```
uiryuu, line 810: Terminating one mpv instance should not cause the whole app to be terminated. I know this was the behavior before, but I think it's a good time to correct that behavior.

uiryuu (changes requested):

iina/AppDelegate.swift
low-batt: Maybe I should rephrase that "Forcing application termination" log message. The timeout tells Cocoa to proceed with termination. This does not happen immediately. Tasks on the main queue are still run after this. This guard statement insures IINA does not reply again to Cocoa telling it a second time to proceed with termination. Was worried about what Cocoa would do if IINA replied a second time. This also avoids outputing the log message that indicates a normal termination.

Eventually after IINA tells Cocoa to proceed with termination AppKit will call back to IINA, executing applicationWillTerminate. At that point IINA closes the log file. Even after that the main queue may run other tasks before the application is terminated. We know that happens because it was causing crashes when a task tried to log a message after the log file was closed.

Maybe the log message should be: "Timeout. Proceeding with application termination"?

uiryuu: Okay, got that. I also think is reasonable to change the log message.

PR #3971, opened 2022-10-07
Close log build date, git branch and commit hash, #3970

### LINKED ISSUE #3970: Log build date, git branch and commit hash
**What you want IINA to do:**
During startup emit a log message giving information about the build including the date, git branch and commit hash.

**What IINA does currently:**
From an IINA log file it is not obvious the executable is from a development build.

For developers that take development builds to other machines for testing it is desirable to have a way to distinguish between development builds.

I will be posting a PR containing the changes needed to emit such a log message.

I looked, but did not find a standard convention for addressing this.

If you build FFmpeg this is what you get for a version number:
```text
ffmpeg version N-107541-g81ebf40efa Copyright (c) 2000-2022 the FFmpeg developers
  built with Apple clang version 13.1.6 (clang-1316.0.21.2.5)
```

The `100221` is the branch's commit count, the `81ebf40efa` is a shortened commit hash.

Some Apple developers change the version number to be an ISO formatted build date with all characters stripped out. I thought it was better to avoid using the application version given Apple's restrictions on what the version can contain.

low-batt:
This commit will:
- Add custom keys to Info.plist to contain build information
- Enable preprocessing of the Info.plist file in the Xcode build
- Enable use of an Info.plist header file in the Xcode build
- Add a shell script write_info_plist_header.sh that generates the Info.plist header file
- Add a new Xcode build phase to execute the script
- Add a new logBuildDetails method to AppDelegate

With these changes IINA will log a message during startup containing the build date, git branch and commit hash.

- [x] It implements / fixes issue #3970.

---

**Description:**


iina.xcodeproj/project.pbxproj
```diff
@@ -2744,6 +2745,27 @@
...
+			name = "Write Info.h";
+			outputFileListPaths = (
+			);
+			outputPaths = (
+				"$(INFOPLIST_PREFIX_HEADER)",
+				"$(INFOPLIST_FILE)",
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "iina/write_info_plist_header.sh\n";
```
uiryuu, line 2767: Is this \n intended?

iina/Info.plist
```diff
@@ -611,7 +617,7 @@ Released under GPLv3.</string>
 				<string>public.audio</string>
 			</array>
 			<key>UTTypeDescription</key>
-			<string>Monkey's Audio Lossless Audio</string>
+			<string>Monkey&apos;s Audio Lossless Audio</string>
```
uiryuu: Also please revert this. I've seen Xcode automatically jumped between ' and &apos; multiple times...

uiryuu (approved):

iina.xcodeproj/project.pbxproj
low-batt: Xcode demands it. I was unable to remove it within Xcode editing the build phase. So with Xcode shutdown I removed it using Emacs. Confirmed with git diff. Brought up Xcode. Confirmed build still worked. Checked git status. No change. Xcode put the newline back in.

iina/Info.plist
low-batt: Reverted.

iina.xcodeproj/project.pbxproj
uiryuu: Then it should be ok as long as it works

PR #3973, opened 2022-10-07
Fix white window after entering PIP when pausing

uiryuu:
We need to redraw a frame after entering PIP when pausing, and previously we perform such redraw in the next `layout`. However, it now takes two `layout`s before finishing entering PIP. This was tested on macOS 13 beta and macOS 12.6.

Changing the pending flag into a counter and redrawing twice solve the problem.

### DIFF
--- a/iina/VideoView.swift
+++ b/iina/VideoView.swift
@@ -37,7 +37,7 @@ class VideoView: NSView {
   // cached indicator to prevent unnecessary updates of DisplayLink
   var currentDisplay: UInt32?
 
-  var pendingRedrawAfterEnteringPIP = false;
+  var pendingRedrawAfterEnteringPIPCount = 0;
 
   lazy var hdrSubsystem = Logger.Subsystem(rawValue: "hdr")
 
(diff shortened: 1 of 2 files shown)


low-batt:
I pulled the PR and tested. I don't usually use PIP, so I'm unfamiliar with this feature.

The bad behavior I encountered was that with playback paused exiting PIP would sometimes leave me with a fully black window.

This is running on my MacBookPro18,2 under macOS 12.6 and built with Xcode 14.0.

uiryuu:
> The bad behavior I encountered was that with playback paused exiting PIP would sometimes leave me with a fully black window.

Did you test on dev branch? This pr doesn't change behavior when exit, so the behavior should be the same.

> This is running on my MacBookPro18,2 under macOS 12.6 and built with Xcode 14.0.

I tested on macOS 13 beta and 12.6 virtual machine, didn't reproduce.

low-batt:
I did this to test the PR:
```bash
git fetch origin pull/3973/head:pr3973
git checkout pr3973
```

In `MainWindowController` I see this:
```Swift
      videoView.pendingRedrawAfterEnteringPIPCount = 2
```

I think I have the PR sources.

I tried reproducing the black screen on develop and it didn't reproduce for me.

It doesn't do it every time. I pause the video and then go in and out of PIP until I'm left with a black screen. For the following screen recording it happened on the first try:

https://0x0.st/oy5J.mov

iina/VideoView.swift
```diff
@@ -37,7 +37,7 @@ class VideoView: NSView {
   // cached indicator to prevent unnecessary updates of DisplayLink
   var currentDisplay: UInt32?
 
-  var pendingRedrawAfterEnteringPIP = false;
```
lhc70000, line 40: Could be just `pendingRedrawsAfterEnteringPIP`

PR #3980, opened 2022-10-12
Fix toggle menus to adhere to HIG, #3116

low-batt:
This commit will change these toggle menus to use "Hide" in their title instead of "Show" when selecting the menu item hides the panel:
- Show Chapters Panel
- Show Playlist Panel
- Show Quick Settings Panel

This commit adds text that will require localization.

This fix only corrects one part of the reported issue. The behavior of buttons still needs to be updated to adhere to the HIG.

- [x] It implements / fixes issue #3116.

---

**Description:**
This PR replaces the original PR, removing localization files.

### DIFF
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -122,8 +122,12 @@
 "menu.hide_chapters" = "Hide Chapters Panel";
 "menu.playlist" = "Show Playlist Panel";
 "menu.hide_playlist" = "Hide Playlist Panel";
-"menu.quicksettings" = "Show Quick Settings Panel";
-"menu.hide_quicksettings" = "Hide Quick Settings Panel";
+"menu.video" = "Show Video Panel";
+"menu.hide_video" = "Hide Video Panel";
+"menu.audio" = "Show Audio Panel";
+"menu.hide_audio" = "Hide Audio Panel";
+"menu.subtitles" = "Show Subtitles Panel";
+"menu.hide_subtitles" = "Hide Subtitles Panel";
 
 "menu.seek_forward" = "Step Forward %.0fs";
 "menu.seek_backward" = "Step Backward %.0fs";
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -118,6 +118,12 @@
 "menu.pause" = "Pause";
 "menu.resume" = "Resume";
 "menu.speed" = "Speed: %.2fx";
+"menu.chapters" = "Show Chapters Panel";
+"menu.hide_chapters" = "Hide Chapters Panel";
+"menu.playlist" = "Show Playlist Panel";
+"menu.hide_playlist" = "Hide Playlist Panel";
+"menu.quicksettings" = "Show Quick Settings Panel";
+"menu.hide_quicksettings" = "Hide Quick Settings Panel";
 
 "menu.seek_forward" = "Step Forward %.0fs";
 "menu.seek_backward" = "Step Backward %.0fs";
--- a/iina/AppData.swift
+++ b/iina/AppData.swift
@@ -89,8 +89,12 @@ struct Constants {
     static let hideChaptersPanel = NSLocalizedString("menu.hide_chapters", comment: "Hide Chapters Panel")
     static let playlistPanel = NSLocalizedString("menu.playlist", comment: "Show Playlist Panel")
     static let hidePlaylistPanel = NSLocalizedString("menu.hide_playlist", comment: "Hide Playlist Panel")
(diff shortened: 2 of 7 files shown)


iina/Base.lproj/Localizable.strings
```diff
@@ -118,6 +118,12 @@
 "menu.pause" = "Pause";
 "menu.resume" = "Resume";
 "menu.speed" = "Speed: %.2fx";
+"menu.chapters" = "Show Chapters Panel";
```
uiryuu, line 121: Please also add new strings into `en.lproj/Localizable.strings`, otherwise Crowdin won't pick them up.

uiryuu (changes requested):
The inconsistency annoys me. Currently there are `Show Playlist Panel` and `Show Chapter Panel` in the playback menu, but in it is `Show Quick Settings Panel` in Video, Audio, and Subtitle menu.

We should rename them into `Show Video/Audio/Subtitle Panel` and `Hide Quick Settings Panel` correspondingly.

low-batt:
The inconsistency in menu item names is due to the inconsistency in the menu structure. Using Video, Audio and Subtitles in the item names seems like duplicate text. Menu items according to the HIG are supposed to use "a condensed style". Of course you can argue that we are already violating that with items such as "Audio filters...". I can go ahead and make this change and we can see how it looks.

At the moment I am distracted by the fact that this fix is not working. MainWindowController has changed. The sideBarStatus used to indicate which panel was set, .settings or .playlist. That is no longer working. It is set to .settings when the playlist/chapter panel is being displayed. So the menus are malfunctioning.

Was hoping to post an additional commit that as requested adds the english translations. But I need to investigate why the sideBarStatus seems to be incorrect. A quick fix did not work. Need to dig into what happened in the MainWindowController code.

Tomorrow busy. Might be able to post an update Saturday.

uiryuu:
> Using Video, Audio and Subtitles in the item names seems like duplicate text. 

Sorry I didn't make it clear. I meant we should rename them into
- `Show Audio Panel`
- `Show Video Panel`
- `Show Subtitle Panel`
in the corresponding menu

uiryuu:
> The sideBarStatus used to indicate which panel was set, .settings or .playlist. That is no longer working.

@low-batt This is caused by 60b944f when copy-pasting the reused code. I've pushed the fix, please pull and try.

low-batt:
I picked up the commit and confirmed that fixes the problem.
I need to rename the menu items. Tied up today. I should be able to get that done tomorrow.

low-batt:
I've renamed the menu items. This does create a slight inconsistency with the buttons used in the OSC:

issue-3116

Should `Quick Settings` be renamed to  `Video, audio and subtitles`?

saagarjha:
What’s the inconsistency?

low-batt:
I remembered this morning that I forgot to update MainMenu.xib. Even though the text in the xib is not used it should be updated to not confuse developers. So that needs to be changed before this PR is ready to be merged.

On the question about the button inconsistency...

Currently the menu items are:
- Show Quick Settings Panel
- Show Chapters Panel
- Show Playlist Panel

And the OSC buttons are:
- Quick Settings
- Playlist and chapters

So it is clear that the `Quick Settings` button corresponds to `Show Quick Settings Panel`.

By renaming the menu items to make them more consistent:
- Show Video Panel
- Show Audio Panel
- Show Subtitles Panel
- Show Chapters Panel
- Show Playlist Panel

We have eliminated the term "Quick Settings" and broken the correspondence with the OSC button.

If we want to proceed with the change to the menu items then I think the buttons should be renamed as well:
- Video, audio and subtitles
- Playlist and chapters

Should I go ahead and change the button name as well?

uiryuu:
> Should Quick Settings be renamed to Video, audio and subtitles?

No, "Quick Settings" means the video panel + audio panel + subtitle panel.

uiryuu (approved):

low-batt:
How does the user know what "Quick Settings" means now that we removed that text from the menus? The panel itself has no such label. That was my worry.

Should I also update the menu item text in MainMenu.xib? It gets overwritten, but if I leave it set to "Show Quick Settings Panel" it could confuse a developer.

uiryuu:
> How does the user know what "Quick Settings" means now that we removed that text from the menus? The panel itself has no such label. That was my worry.

I think this is a valid concern, but since we have used "Quick Settings" for years in IINA, removing that will also surprise some of the users. I think we can discuss this in another issue, let's get this merged first!

> Should I also update the menu item text in MainMenu.xib? It gets overwritten, but if I leave it set to "Show Quick Settings Panel" it could confuse a developer.

Yes please do that. That should change the corresponding text in `Base.lproj`.

uiryuu (approved):

PR #3981, opened 2022-10-13
Fix app quits when quitting an instance of mpv

uiryuu:
**Description:**
Quitting one mpv instance should not cause the whole app to quit, close the corresponding window instead. Not sure what I've done is sufficient or did I miss something, but the behavior seems correct.

### DIFF
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -799,21 +799,9 @@ class MPVController: NSObject {
         // now that mpv has shut down. This is not needed when IINA sends the quit command to mpv
         // as in that case the observers are removed before the quit command is sent.
         removeOptionObservers()
-        // Submit the following task synchronously to ensure it is done before application
-        // termination is started.
-        DispatchQueue.main.sync {
+        player.closeMainWindow()
+        DispatchQueue.main.async {
           self.player.mpvHasShutdown(isMPVInitiated: true)
-        }
-        // Initiate application termination. AppKit requires this be done from the main thread,
-        // however the main dispatch queue must not be used to avoid blocking the queue as per
-        // instructions from Apple.
-        if #available(macOS 10.12, *) {
-          RunLoop.main.perform(inModes: [.common]) {
-            self.terminateApplication()
-          }
-        } else {
-          RunLoop.main.perform(#selector(self.terminateApplication), target: self,
-                               argument: nil, order: Int.min, modes: [.common])
         }
       } else {
         mpv_destroy(mpv)


low-batt:
IINA should function like mpv unless there is a definite need to deviate. The mpv commands in question are:

[quit](https://mpv.io/manual/stable/#command-interface-quit-[%3Ccode%3E]):
> **quit [\<code>]**
>    Exit the player. If an argument is given, it's used as process exit code.

[stop](https://mpv.io/manual/stable/#command-interface-stop-[%3Cflags%3E])
> **stop [\<flags>]**
>    Stop playback and clear playlist. With default settings, this is essentially like quit. > Useful for the client API: playback can be stopped without terminating the player.
>
>    The first argument is optional, and supports the following flags:
>
>    **keep-playlist**
>        Do not clear the playlist.

The IINA code in question is processing the `MPV_EVENT_SHUTDOWN` event:
```c
    /**
     * Happens when the player quits. The player enters a state where it tries
     * to disconnect all clients. Most requests to the player will fail, and
     * the client should react to this and quit with mpv_destroy() as soon as
     * possible.
     */
    MPV_EVENT_SHUTDOWN          = 1,
```

And IINA did not initiate the shutdown. There are two ways that can happen that I know of. The default IINA key bindings maps "q" to sending the `quit` command to mpv. The user can also use the mpv IPC interface to send a `quit` command to mpv totally behind the back of IINA.

From the mpv manual above the `quit` command definitely means exit the application. That command even takes a process exit code to use. So I disagree with the premise that IINA should not initiate application termination in reaction to a  `MPV_EVENT_SHUTDOWN` event that was not initiated by IINA.

The behavior of just closing the current window sounds like what IINA should be doing when the user sends a `stop` command to mpv. That command means stop playing and unload the file. That sounds like what should be tied to the proposed behavior. My guess is that IINA is not properly handling that case.

BUT...  There are other problems in this area....

We do not want mpv to quit behind IINA's back. From the mpv comments above IINA must not send commands to mpv while it is quitting. We have experienced crashes in mpv code when IINA has done that. That is why the recent changes try and shutdown activity before sending a `quit` command to mpv. If mpv can start quitting behind the back of IINA then it would seem there is a window of vulnerability where IINA can send a command to a mpv core that is quitting, possibly triggering a crash.

We could change the code that handles key mapping to not send the `quit` command, but the user can always use the mpv IPC interface and send the command totally behind the back of IINA. I've not investigated is mpv provides a way for an application using the library to intercept IPC commands. Something has to be done to allow IINA to control shutdown. I've not had time to look into this.

Anyway, if we wanted to treat the `MPV_EVENT_SHUTDOWN` event as just closing the current window then we need to consider how unhappy some users will be who are used to typing "q" in the window to quit IINA. I don't think we can make such a change without changing the key mapping code to preserve the current "q" behavior.

We'd also have to make quite a few more code changes to handle player core caching. Once mpv has quit we now have a bad player core in the cache. Normally players in the cache have a mpv core that has been stopped. But now we have a player with an unusable mpv core. Easiest solution is probably to discard that player. We'd need to insure there isn't any references to the player that keep it from being fully released and the memory returned to the heap.

uiryuu:
@low-batt Thanks for the detailed explanation and discussion.

> Happens when the player quits. The player enters a state where it tries to disconnect all clients. Most requests to the player will fail, and the client should react to this and quit with mpv_destroy() as soon as possible.

It's my misunderstanding of the `quit` command. Since the API documentation explicitly says to disconnect *all* clients, let's obey it.

> My guess is that IINA is not properly handling that case.

I tried to send a `stop` command, IINA just close one window, so it seems work fine.

PR #3993, opened 2022-10-16
Fix postMessage type not valid JSON object will crash

whlsxl:

---

**Description:**

`postMessage` only support valid JSON object, not test before convert.


lhc70000:
Merged in 8415cd5.

Issue #3995, opened 2022-10-16
Audio continues after window closed (in develop)


low-batt:
**System and IINA version:**

- macOS 12.6
- IINA a build of the current develop branch

**Expected behavior:**
When a window is closed IINA does not continue playing audio.

**Actual behavior:**
With the preference `Quit after all windows are closed` disabled audio continues playing after the window playing the video is closed.

This defect is not in a released version of IINA. This is a regression currently in the develop branch, introduced by the commit in PR #3961.

I will be posting a PR with a fix shortly.

**Steps to reproduce:**
- Start IINA
- Choose `Preferences...` from the `IINA` menu
- In the `Behavior` section disable  `Quit after all windows are closed`
- In the `Playlist` section disable `Add files in the same folder automatically`
- **Restart IINA** to active the new settings  
- Open video1
- Choose `Show Playlist Panel` in the `Playback` menu
- Click `+` in the `PLAYLIST` side panel
- Choose `Add File...` 
- Select a second video, video2
- Click open
- The video is added to the playlist
- Select video1 in the `PLAYLIST` side panel
- Click `-` in the `PLAYLIST` side panel
- IINA starts playing video2
- Press command ⌘-W to close the window
- IINA closes the window
- Audio from video2 continues to play

- [x] MPV does not have this problem.

**How often does this happen?**
Everytime.

low-batt:
Added high priority label as this is a regression.

low-batt:
### Analysis

When `-` is clicked IINA sends the  [playlist-remove](https://mpv.io/manual/stable/#command-interface-playlist-remove) command to mpv:
> **playlist-remove \<index>**
>    Remove the playlist entry at the given index. Index values start counting with 0. The special value current removes the current entry. Note that removing the current entry also stops playback and starts playing the next entry.

In the sequence that reproduces the defect the playlist entry being removed is the current entry. So mpv stops playback and starts playing the next entry. As the follow log messages show mpv optimizes this by concurrently starting to load the next entry while stopping playback of the current entry:
```text
15:08:31.118 [player0][d] File started
15:08:31.120 [player0][d] Playback has stopped
15:08:31.148 [player0][d] File loaded
```

The changes in PR #3961 reset the new `isStopped` property to `false` in the `PlayerCore` method `fileStarted`. In the failing senario `isStopped` is then reset to `true` when mpv finishes stopping playback of the previous file. The property then incorrectly indicates playback has already stopped when the window is closed.

[...]

PR #3996, opened 2022-10-16
Fix audio continues after window closed, #3995

low-batt:
The fix in the commit moves the resetting of the isStopped flag from the fileStarted method in PlayerCore to the fileLoaded method.

The method fileStarted was the wrong place to reset this flag because mpv will start loading the next file in the playlist while still stopping playback of the previous file.

- [x] This implements/fixes issue #3995.

---

**Description:**
This fixes a regression recently added to the develop branch.


uiryuu:
Thanks

PR #3999, opened 2022-10-16
Fix bug: Shift modifier incorrectly added to filter shortcuts

### LINKED ISSUE #3998: Saved filter key shortcuts: lowercase letters have shift modifier added
Screen Shot 2022-10-16 at 15 07 08

**System and IINA version:**

- macOS 12.6
- IINA 1.3.0 develop branch, commit 6da6a3f410d8733119aa29303fad306552575f31

**Expected behavior:**
If a saved filter has a key equivalent, it should only include `Shift` if the user specified it.

**Actual behavior:**
When a filter is saved, the`Shift` modifier is added for both lowercase and uppercase letters.

**Steps to reproduce:**

1. Pick a lowercase letter which is easy to distinguish from its uppercase variant, like 't', and make sure you haven't already assigned it to a keybinding in the Preferences > Key Bindings table.
2. From the `Video` menu, select `Video Filters...`
3. Click the `+` button to create a new filter. A new dialog should open with options.
4. Leave `Crop` selected on the left. On the right, enter 200 for Width, and 200 for Height, and leave Origin X & Origin Y empty. Click `Add`.
5. The dialog should close, and the new filter should appear in the list, with a `Save` button. Click `Save'.
6. Enter a name for the new filter, like "Custom Crop", then click the blank area under `Shortcut Key:`. Type 't' (or your chosen lowercase key). It should appear in the box in uppercase. Click Add.
7. The filter should now appear in the list under `Saved filters`.
8. Now go back to the `Video` menu, and hover over `Saved Video Filters`. See that the key will appear with a shift symbol in front of it.
9. Open a video and type the lowercase letter. See that IINA beeps at you. Type it again holding down `Shift`. See that the filter is activated and the video is cropped.

Always, given that the key is a letter which can have an uppercase form.

svobs:
- [x] This implements/fixes issue #3998.

---

**Description:**
See above issue for problem description.

**Summary / conclusions**
This terrifying block of code appears to be the culprit:
```Swift
extension NSEvent {
  var readableKeyDescription: (String, String) {
    get {

      let rawKeyCharacter: String
      if let char = NSEventKeyCodeMapping[Int(self.keyCode)] {
        rawKeyCharacter = char
      } else {
        let inputSource = TISCopyCurrentASCIICapableKeyboardLayoutInputSource().takeUnretainedValue()
        if let layoutData = TISGetInputSourceProperty(inputSource, kTISPropertyUnicodeKeyLayoutData) {
          let dataRef = unsafeBitCast(layoutData, to: CFData.self)
          let keyLayout = unsafeBitCast(CFDataGetBytePtr(dataRef), to: UnsafePointer<UCKeyboardLayout>.self)
          var deadKeyState = UInt32(0)
          let maxLength = 4
          var actualLength = 0
          var actualString = [UniChar](repeating: 0, count: maxLength)
          let error = UCKeyTranslate(keyLayout,
                                     UInt16(self.keyCode),
                                     UInt16(kUCKeyActionDisplay),
                                     UInt32((0 >> 8) & 0xFF),
                                     UInt32(LMGetKbdType()),
                                     OptionBits(kUCKeyTranslateNoDeadKeysBit),
                                     &deadKeyState,
                                     maxLength,
                                     &actualLength,
                                     &actualString)
          if error == 0 {
            rawKeyCharacter = String(utf16CodeUnits: &actualString, count: maxLength).uppercased()

          } else {
            rawKeyCharacter = KeyCodeHelper.keyMap[self.keyCode]?.0 ?? ""
          }
        } else {
          rawKeyCharacter = KeyCodeHelper.keyMap[self.keyCode]?.0 ?? ""
        }
      }

      return (([(.control, "⌃"), (.option, "⌥"), (.shift, "⇧"), (.command, "⌘")] as [(NSEvent.ModifierFlags, String)])
        .map { self.modifierFlags.contains($0.0) ? $0.1 : "" }
        .joined()
        .appending(rawKeyCharacter), rawKeyCharacter)
    }
  }
}
```

The value it returns will always be in uppercase, which is fine for display purposes, but it shouldn't have been using it to store the "raw" key value (The code was already grabbing the raw value from the user and putting it in `currentKey`, and it was far simpler to just use that). It has another problem: it seems to add 3 null characters to the string it returns. The debugger shows that `t` is stored as `T\0\0\0`.

This is not good, but I'm going to punt and declare it to be outside the scope of this defect. 

After this fix, `readableKeyDescription` is only used in two places:
1. In `FilterWindowController` for display. At least this doesn't seem to be breaking anything.
2. In `KeyCodeHelper.mpvKeyCode()`, where it is checked for inequality against another string, causing a block of code to always evaluate to `true` - which is ok because that block should always be executed.

This should be cleaned up at some point. Rather than try to debug that block, a more prudent solution would be to remove all use of it and find a simpler substitute if needed. At least this PR makes strides in that direction.

### DIFF
--- a/iina/FilterWindowController.swift
+++ b/iina/FilterWindowController.swift
@@ -249,7 +249,7 @@ class FilterWindowController: NSWindowController, NSWindowDelegate {
     currentSavedFilter = savedFilters[row]
     editFilterNameTextField.stringValue = currentSavedFilter!.name
     editFilterStringTextField.stringValue = currentSavedFilter!.filterString
-    editFilterKeyRecordView.currentRawKey = currentSavedFilter!.shortcutKey
+    editFilterKeyRecordView.currentKey = currentSavedFilter!.shortcutKey
     editFilterKeyRecordView.currentKeyModifiers = currentSavedFilter!.shortcutKeyModifiers
     editFilterKeyRecordViewLabel.stringValue = currentSavedFilter!.readableShortCutKey
     window!.beginSheet(editFilterSheet)
@@ -322,7 +322,7 @@ extension FilterWindowController {
     if let currentFilter = currentFilter {
       let filter = SavedFilter(name: saveFilterNameTextField.stringValue,
                                filterString: currentFilter.stringFormat,
-                               shortcutKey: keyRecordView.currentRawKey,
+                               shortcutKey: keyRecordView.currentKey,
                                modifiers: keyRecordView.currentKeyModifiers)
       savedFilters.append(filter)
       reloadTable()
@@ -339,8 +339,7 @@ extension FilterWindowController {
     if let currentFilter = currentSavedFilter {
       currentFilter.name = editFilterNameTextField.stringValue
       currentFilter.filterString = editFilterStringTextField.stringValue
-      // FIXME: shouldn't be shift-modified; should examine this carefully
-      currentFilter.shortcutKey = editFilterKeyRecordView.currentRawKey.lowercased()
+      currentFilter.shortcutKey = editFilterKeyRecordView.currentKey
       currentFilter.shortcutKeyModifiers = editFilterKeyRecordView.currentKeyModifiers
       reloadTable()
       syncSavedFilter()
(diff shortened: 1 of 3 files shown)


low-batt (approved):
Reproduced problem. Pulled PR locally, built, confirmed it fixed the problem.

uiryuu (approved):

uiryuu:
Thanks!

PR #4002, opened 2022-10-18
Fix CPU is consumed by OSC while shown and player paused

decodism:
- [ ] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)

---

**Description:**

Tries to fix an issue discovered in #3849.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -254,7 +254,7 @@ class MainWindowController: PlayerWindowController {
     case shown, hidden, willShow, willHide
   }
 
-  var animationState: UIAnimationState = .shown
+  var animationState: UIAnimationState = .hidden
   var osdAnimationState: UIAnimationState = .hidden
   var sidebarAnimationState: UIAnimationState = .hidden
 
@@ -1657,7 +1657,7 @@ class MainWindowController: PlayerWindowController {
   }
 
   private func showUI() {
-    if player.disableUI { return }
+    guard !player.disableUI, ![.willShow, .shown].contains(animationState) else { return }
     animationState = .willShow
     fadeableViews.forEach { (v) in
       v.isHidden = false
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -1521,6 +1521,7 @@ class PlayerCore: NSObject {
 
   func createSyncUITimer() {
     invalidateTimer()
+    guard !info.isPaused else { return }
     syncPlayTimeTimer = Timer.scheduledTimer(
       timeInterval: TimeInterval(DurationDisplayTextField.precision >= 2 ? AppData.syncTimePreciseInterval : AppData.syncTimeInterval),
       target: self,


iina/PlayerCore.swift
```diff
@@ -1521,6 +1521,7 @@ class PlayerCore: NSObject {
 
   func createSyncUITimer() {
     invalidateTimer()
+    guard !info.isPaused else { return }
```
low-batt, line 1524: Should the guard statement be before or after the call to `invalidateTimer`?

My thoughts...

If playback is paused then the timer should not be running and therefore there should be no need to call `invalidateTimer`.

Now if the timer is running with playback paused then something has gone wrong and there is a defect somewhere that needs to be fixed. By calling `invalidateTimer` the impact of that defect is reduced. That is good, but it also makes it harder to notice and track down the problem.

See what the other reviewers think. They may like the code the way it is.

iina/MainWindowController.swift
```diff
@@ -1657,7 +1657,7 @@ class MainWindowController: PlayerWindowController {
   }
 
   private func showUI() {
-    if player.disableUI { return }
+    guard !player.disableUI, ![.willShow, .shown].contains(animationState) else { return }
```
low-batt, line 1660: I'm confused about why this change is needed and how it relates to the CPU issue.

iina/PlayerCore.swift
svobs: Generally agree. A function with a name `create*` shouldn't end up doing the opposite. The ideal way for code to behave when it breaks is for it to write something to the log for the developers's benefit, but avoid disturbing the user. And I think the main effect of leaving the timer running is a message getting written to the logs, right (along with some extra CPU usage which the user probably won't notice)?

low-batt: In the case at hand broken code only means wasting energy. If playback is paused then the times representing the playback position are not changing. If the timer is running it is obtaining the playback position from mpv and updating the OSC time labels with the same time they are already set to display. So everything works fine, but the CPU time to update the UI components with exactly the same information is all wasted. If the user changes the precision of the times displayed in the OSC then the timer runs faster and wastes even more information.

This part the of code is free of logging due to the frequency at which it gets executed.

Where to put the guard statement is definitely a grey area.

decodism: Happy to change if needed.

iina/MainWindowController.swift
decodism: This function is executed a lot of times when the cursor is moved even though it is not necessary. This is not related to the timer issue.

iina/PlayerCore.swift
svobs: > ...This part the of code is free of logging due to the frequency at which it gets executed.

I see. I'm fine with leaving it the way it is - we've probably said more than enough about it.

PR #4003, opened 2022-10-18
Fix quit triggers "invalid context state" error, #3593

low-batt:
This commit will change MPVController.mpvUninitRendering to lock and set the OpenGL context before calling the mpv render methods.

- [x] This implements/fixes issue #3593.

---

**Description:**

### DIFF
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -435,6 +435,8 @@ class MPVController: NSObject {
 
   func mpvUninitRendering() {
     guard let mpvRenderContext = mpvRenderContext else { return }
+    lockAndSetOpenGLContext()
+    defer { unlockOpenGLContext() }
     mpv_render_context_set_update_callback(mpvRenderContext, nil, nil)
     mpv_render_context_free(mpvRenderContext)
   }


low-batt:
Very simple change. Let's get this one merged ASAP as the already merged fix to make IINA properly shutdown mpv has the possibility of making this issue more likely.

The OpenGL context is passed in thread local storage. Before calling mpv render methods we must insure the context has been set. This is a mpv/OpenGL requirement.

Of course if you fail to lock and fail to set, things might work because the thread just happens to have the context already in its local storage and no other thread happens to try and use the context while the thread is using the context without having locked it.

So this is one of those threading issues that randomly causes crashes. Previously we were terminating a lot of the time before properly executing shutdown. Now that we are doing that properly this code will be getting more use. Time to fix it.

I suspect this is not the only place where IINA fails to do this. This is the case I was able to reproduce and confirm this change fixes it. I will take a closer look into this later, but for now let's get this one fixed.

uiryuu (approved):

uiryuu:
Thanks!

PR #4008, opened 2022-10-21
Fix column selection warnings

svobs:

---

**Description:**
This removes `Warning: Column selection is not supported in view-based table and outline views` which are being output to the logs.

Reference:  https://developer.apple.com/forums/thread/693375


low-batt (approved):
Pulled PR locally, tested, confirmed fixed.

I don't think we were getting the warning for the table in advanced preferences? It was actually selectable. But that should not have been selectable as there isn't any action you can do with that table having selected a column. So good that we are disabling it for that table as well.

uiryuu:
Thanks

Issue #4014, opened 2022-10-26
Dark image when player HDR in Fullscreen after upgrade to Ventura


buntupana:
**System and IINA version:**

- macOS 13.0
- IINA 1.3.0
- Macbook pro 14 2021

**Expected behavior:**
Watch HDR content in Fullscreen

**Actual behavior:**
When I try to play a content in HDR in fullscreen the image will go dark when the controller display fade out. If I play no HDR content or I disable HDR in options this won't happen. I tried legacy and native fullscreen with the same result.

<br>

**Steps to reproduce:**

Play any HDR content

**How often does this happen?**

Always

singh-jagjot:
The issue can be replicated on Macbook Pro 16 2021

buntupana:
Just discoverd a workaround:

Keeping the pointer in the bottom right/left corner of the screen while in fullscreen will fix the problem

low-batt:
Duplicate of #3844, #3880, #3906.

Liferme:
> 

I don't think it is only a matter of HDR videos, I watched a normal 2160P Bluray content today, and the issue remains.

singh-jagjot:
> > 
> 
> I don't think it is only a matter of HDR videos, I watched normal 2160P Bluray content today, and the issue remains.

For me, it's only limited to HDR videos, tested non-HDR(AV1, VP9) videos till 2160p and they are working as expected.

royj:
I made a workaround here: https://github.com/iina/iina/pull/4017

I only tested it on my MBP16, so additional testing would be useful.

Liferme:
@low-batt @singh-jagjot  **Sorry guys, I just updated to Ventura today, and I didn't do much testing when I commented.**

After a thorough test, it appears that we can confirm the issue is tied to HDR enabling in the quick settings panel. And it has nothing to do with hardware decoding.

**Behavior:**
The screen will become so dim that the video becomes unwatchable when the controller bar fade out in HDR fullscreen mode.

**My Device Specification:**
- macOS Ventura 13.0
- IINA 1.3.0
- Macbook pro 14 2021

low-batt:
@Liferme Thanks for the testing and clarifying your report. This does line up with what we know about the problem.

low-batt:
This same problem is reported in multiple issues. Closing duplicates in favor of the first report of this problem in issue #3844.

Issue #3906, opened 2022-08-11
HDR video become very dark in hdr mode on macbook air m2


YuntianCheng:
**System and IINA version:**

- macOS 12.5
- IINA 1.3.0

**Expected behavior:**
Play HDR video normally

**Actual behavior:**

Video become very dark

**Steps to reproduce:**
 play a hdr video on macbook air m2, turn on the hdr switch

**How often does this happen?**
everytime

hannesaronsson:
The Macbook Air does not have a screen that supports HDR. But i guess you could handle the behaviour differently anyway.

KrisCris:
Sometimes if I got lucky enough, I am able to play the same HDR content correctly, but most of the time it is too dark to see

KrisCris:
I want to take a screenshot, but the screenshot is not darken at all

YuntianCheng:
> The Macbook Air does not have a screen that supports HDR. But i guess you could handle the behaviour differently anyway.
The build-in screen on MacBook Air definitely support HDR
![3527D3B2-E978-4FC7-A39F-4360BEE98100](https://user-images.githubusercontent.com/29092622/185258513-c0b1dac3-edf7-4ba6-87f7-a11b4505eb35.png)

hannesaronsson:
> The Macbook Air does not have a screen that supports HDR. But i guess you could handle the behaviour differently anyway.
> The build-in screen on MacBook Air definitely support HDR
> ![3527D3B2-E978-4FC7-A39F-4360BEE98100](https://user-images.githubusercontent.com/29092622/185258513-c0b1dac3-edf7-4ba6-87f7-a11b4505eb35.png)

This sounds kinda confusing to me as the screen is an edge lit LCD (400 nits) which does not have any local dimming zones. What exactly does HDR even do when there is no brightness levels that can change within the picture?

KrisCris:
Actually, things like hdr400 don't require local dimming. Without local dimming just means it looks not that good

Zer0xPoint:
accessbilty -> display -> Pointer Size 
Switch the toggle
u will be fine

YuntianCheng:
> accessbilty -> display -> Pointer Size Switch the toggle u will be fine
> 
> it seem like an macOS's Bug rither then IIna, non beta version working fine

I tried, it didn't work.

JHNunderwood:
[...]

Issue #3880, opened 2022-07-19
The screen automatically dims when playing


HurstSi:
**System and IINA version:** macOS 13.0 iina 1.3.0

**Expected behavior:**
The screen should keep its brightness level.

**Actual behavior:**
When playing HDR video, the screen automatically dims when the mouse stops moving.

**Steps to reproduce:**
Paly an HDR video on MacBook Pro 14 inch's screen.

**How often does this happen?**
everytime

low-batt:
You are experiencing this running the macOS Ventura (13.0) public beta?

If so, I suggest reporting this behavior to Apple. See this iMore article: [How to send feedback to Apple about macOS betas](https://www.imore.com/how-send-mac-os-public-beta-feedback)

HurstSi:
> You are experiencing this running the macOS Ventura (13.0) public beta?
> 
> If so, I suggest reporting this behavior to Apple. See this iMore article: [How to send feedback to Apple about macOS betas](https://www.imore.com/how-send-mac-os-public-beta-feedback)

Yes, that's happening on macOS Ventura. OK, I'll reporting to Apple. Thank you.

low-batt:
Thanks! Please let us know what Apple as to say.

One possibility is that this macOS feature is malfunctioning in Ventura:
[If the brightness is limited on your MacBook Pro with Liquid Retina XDR display or Apple Pro Display XDR](https://support.apple.com/en-us/HT212877)

Hopefully Apple will explain the behavior.

poorest1212:
beta2 still sucks... cannot watch HDR video in fullscreen, only works in window mode

low-batt:
Thank you for reporting this. It is helpful for the IINA project to know that there is trouble when running under the Ventura Public Beta. _However_ the point of macOS Beta releases is for _Apple_ to gather feedback. Have you reported this behavior to Apple as well? From [Apple Beta Software Program FAQ](https://beta.apple.com/sp/betaprogram/faq):
> **How do I provide my feedback to Apple?**
>
[...]

Issue #4015, opened 2022-10-27
[Ventura] Non-IINA monitor momentarily blackened while using Firefox or Chromium in full screen while IINA is also full screen


dbenjaminmiller:
**System and IINA version:**

- macOS 13.0
- IINA - Both 1.3.0 (132) and nightly@f8811b6

**Expected behavior:** When Firefox/Chromium is open on monitor in full screen, and IINA is in full screen on the other, there should be no momentarily blackening of any screens at any time.

**Actual behavior:** When I open Firefox full-screen on one monitor, and IINA full-screen on the other, the screen with Firefox will sometimes be blackened. Sometimes this occurs somewhat randomly when a key is pressed on the keyboard while I have the Firefox window open. But the most consistent way to get this to occur is by clicking in the Firefox screen and then moving the cursor to the IINA monitor, which usually (though not 100% of the time) causes the Firefox monitor to go black, until I either click on the IINA monitor or move my mouse back (usually), though the screen will un-black after a moment in many cases. The screen also sometimes will turn black when I make a keystroke in the Firefox screen without moving my cursor to the IINA screen. This often happens when I am typing in a text box on a webpage, especially when I press the space bar.

I have tried this using both Firefox 106.0.2 (current stable) and Firefox's latest nightly build as of the time of my writing this, with the same results. I also tried using Chromium 106, which seems susceptible to the same problem. With most other programs, this does not occur (e.g., with Safari and Apple Mail). I don't know why it occurs with Firefox and Chromium and not those other programs, but it clearly has something to do with IINA, because it occurs with multiple other programs on the other screen, and it does not happen with MPV, nor with any other video player I try (e.g., QuickTime and VLC).

This phenomenon only occurs when a video is playing, not when one is open but paused. This does occur even if the video and/or audio tracks are disabled. More than that, whenever there is a UI panel of some kind open in IINA (playlist, settings sidebar, about window, etc.) above the IINA full-screen window, in any way, then the phenomenon does not occur even while playback is actually active.

The phenomenon does not occur when there are two windows split-screen on the same screen. This includes when IINA and Firefox are split on halves of the same screen, or even in fact when IINA is on one window and two firefox windows are split-screen in full-screen mode on the other.

The issue occurs no matter what hardware acceleration settings are chosen in Firefox or IINA. It also occurs no matter what codec is used in the video played in IINA and no matter its framerate (including whether it has an integer or non-integer framerate).

[...]

PR #4017, opened 2022-10-27
Workaround for dimmed HDR content in full screen on macOS 13

royj:
- [x] This implements/fixes issue #3844, #3880, #3906 and #4014

---

**Description:**

Starting with macOS 13, some strange behavior can be seen when viewing videos in full screen in HDR mode. The screen will be dimmed when the player controls disapear. It looks like it might be a bug on Apple's side, although I'm not sure. From what I've found was if the views do not set the property `isHidden`, but rather an almost invisible alpha level, the video displays normally in its full HDR glory. The end result for the user is the same, setting the player controls to such a low alpha is invisible to the human eye.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -1621,8 +1621,16 @@ class MainWindowController: PlayerWindowController {
     NSCursor.setHiddenUntilMouseMoves(true)
   }
   
-  // Workaround for macOS 13 where HDR content becomes dimmed when overlaying views are fully hidden.
-  private let almostInvisibleAlpha: CGFloat = 0.01
+  // Workaround for macOS 13 where HDR content becomes dimmed when overlaying views are fully hidden. (#3844)
+  private var almostInvisibleAlpha: CGFloat {
+    if #available(macOS 13, *) {
+      if fsState.isFullscreen && player.info.hdrEnabled {
+        return 0.01
+      }
+    }
+    // This issue does not occur when running in non HDR / non full screen or a macOS version lower than 13.
+    return 0
+  }
 
   private func hideUI() {
     // Don't hide UI when in PIP
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -1621,7 +1621,7 @@ class MainWindowController: PlayerWindowController {
     NSCursor.setHiddenUntilMouseMoves(true)
   }
   
-  // Fixes a bug in macOS 13 where HDR content becomes dimmed overlaying views are fully hidden.
+  // Workaround for macOS 13 where HDR content becomes dimmed when overlaying views are fully hidden.
   private let almostInvisibleAlpha: CGFloat = 0.01
 
   private func hideUI() {
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -1620,6 +1620,9 @@ class MainWindowController: PlayerWindowController {
     hideUI()
     NSCursor.setHiddenUntilMouseMoves(true)
   }
+  
+  // Fixes a bug in macOS 13 where HDR content becomes dimmed overlaying views are fully hidden.
+  private let almostInvisibleAlpha: CGFloat = 0.01
 
   private func hideUI() {
     // Don't hide UI when in PIP
@@ -1631,15 +1634,15 @@ class MainWindowController: PlayerWindowController {
     player.invalidateTimer()
     animationState = .willHide
     fadeableViews.forEach { (v) in
-      v.isHidden = false
+      v.alphaValue = almostInvisibleAlpha
     }
     NSAnimationContext.runAnimationGroup({ (context) in
       context.duration = UIAnimationDuration
       fadeableViews.forEach { (v) in
-        v.animator().alphaValue = 0
+        v.animator().alphaValue = almostInvisibleAlpha
       }
       if !self.fsState.isFullscreen {
-        titleTextField?.animator().alphaValue = 0
+        titleTextField?.animator().alphaValue = almostInvisibleAlpha
       }
     }) {
       // if no interrupt then hide animation
@@ -1648,7 +1651,7 @@ class MainWindowController: PlayerWindowController {
           if let btn = v as? NSButton, self.standardWindowButtons.contains(btn) {
             v.alphaValue = 1e-100
           } else {
-            v.isHidden = true
+            v.alphaValue = self.almostInvisibleAlpha
           }
         }
         self.animationState = .hidden


inzar98:
I cant wait to building and approval !!! Thxxx <3

low-batt:
@royj I was about to point out the comment should say "workaround" not "fix"...

Currently writting some review comments...

low-batt (commented):
Thanks for coding this up! Nice commit comment!

Normally I prefer to really track down and understand an issue before reaching for a workaround. However in this case I keep getting stuck trying to figure out what Apple has done to cause this problem. We are probably going to have to go with a workaround.

My own rules for adding a workaround:
- Clearly document in the code that this is a workaround
- Limit the scope as much as possible

A developer in the future might look at this code and expect it to use `isHidden`. There is a good comment on `almostInvisibleAlpha` that should give them an idea that this is some sort of workaround. I would suggest referencing one of the issues reporting the problem.

In this case limiting the scope would imply:
- Only apply to macOS 13.0+
- Only apply when in full screen mode
- Only apply when HDR is active

Assuming our current understanding of when the problem arrises is correct.

Setting `isHidden` has other impacts such as changing how events are delivered.
I can't think of a problem not fully hiding these views would cause, but this means by applying the workaround to the existing views the scope of the workaround might be larger than needed.

Another way to do this would be to create a view specifically for the workaround, something like:
```swift
  private lazy var dimmingWorkroundView = {
    let frame = NSRect(origin: .zero, size: NSSize(width: 0.1, height: 0.1))
    let view = NSView(frame: frame)
    view.wantsLayer = true
    view.layer?.backgroundColor = NSColor.black.cgColor
    view.layer?.opacity = 0.01
    return view
  }()
```

Add that subview in `windowDidEnterFullScreen`:
```swift
    if #available(macOS 13.0, *) {
      window?.contentView?.addSubview(dimmingWorkroundView)
    }
```

And remove it in `windowWillExitFullScreen`:
```swift
    if #available(macOS 13.0, *) {
      dimmingWorkroundView.removeFromSuperviewWithoutNeedingDisplay()
    }
```

Maybe something like that. Would also need a check that HDR is active.

I'm the junior newbie. See what the other reviewers think. They may prefer applying the workaround to the existing views.

royj:
Hello, thanks for your feedback. A dimming overlay view might also fix the problem but in my opinion this increases complexity (and thus, chances for bugs). I followed your advice and applied the fix only for the scenarios you mentioned.

I read some information on setting the alpha lower than 0.02. Apperantly, this also prevents system events from occuring on the controls, the same as using `isHidden`. I'm not sure on the accuracy of this but even if it does still receive events, I believe this is fine. The user moves the cursor anyway to tap an element on the player controller. It is very hard to click on something without moving the mouse a single pixel. 

I created this pull request because this was an issue I was running into when watching a movie yesterday. As you can imagine, I don't have a lot of time to perfect it. 😄 If needed, someone else can pick up the pace from here and clean it up further.

svobs:
@royj this is awesome, thank you! Hopefully this can get merged soon!

Thought of a suggestion (not necessarily for this case - you've done plenty) but in general might be good for things like this - is to add a new preference key, something like `enableHDRDimmingWorkaround`, which defaults to `true`, which controls whether the workaround is enabled. It wouldn't need to be represented in the Preferences UI or even documented; it would just be a safety switch to future-proof the workaround, a potential workaround for the workaround, if it comes up. That way if it turns out to cause problems in the future, users could either set the pref key to false using launch arguments, or change it permanently using a prefs editor, rather than have to wait for a new release.

low-batt:
Currently we are waiting on the other developers for a review. They may approve and merge the current proposed fix as it is a priority to get this regression fixed.

I do like being able to control some of the implementation through preferences. IINA already has a few such preferences.

inzar98:
Guys is there anyway to try that without building myself ?

low-batt:
At the moment you would have to build IINA yourself.

I have just confirmed that issue #4015 seems to not be reproducible with this workaround. That problem is seen when playing a normal, NOT HDR video. Since I had @royj restrict this workaround to HDR videos I think it is on me to do the work to update the workaround to be applied to all videos. I'm currently waiting for comments from other reviewers incase additional changes are needed.

Sure seems like something is very broken in macOS.

aidan-gibson:
This worked for me, but had to enable "Use legacy full screen" else it didn't work. Also have Advanced > "Use mpv osd" enabled; did not test disabling that option.

low-batt:
@aidan-gibson Thanks for reporting trouble with the workaround.

I tried to reproduce the failure. I fetched the PR:
```text
git fetch origin pull/4017/head:pr4017
git checkout pr4017
```

I confirmed I was working with the sources that contain the workaround.
In Xcode I cleaned and rebuilt. 
I changed my IINA preferences, enabling `Use mpv's OSD` under `Advanced` and disabling `Use legacy full screen` under `General`.
I restarted IINA.
I tested using the YouTube HDR video: [The World in HDR in 4K (ULTRA HD)](https://www.youtube.com/watch?v=tO01J-M3g0U) I had previously downloaded and reproduced the problem with.
I did not experience the diming of the screen in full screen mode when the OSC disappears.

I'm very concerned that for some reason the workaround did not work for you. I'm testing on a MacBookPro18,2 with the M1 chip.

Can you think of anything else about your environment that might have played a role?

aidan-gibson:
Tested disabling mpv osd; dimming still occurs with legacy fullscreen disabled but fixed with legacy fullscreen enabled.

Just tested on an external HDR-capable display (Samsung Neo G9), confirmed HDR was enabled system-wide and in IINA; experienced no dimming regardless of settings.

I'm running Ventura 13.0 on a 2021 Macbook Pro 16-inch, M1 Max 32GB.
Honestly there's a boatload of weird customizations to my mac setup, is there anything I should specifically look for?

aidan-gibson:
I followed the build instructions, ie `./other/download_libs.sh` and building in Xcode (Xcode is up to date, latest public version 14.0.1)

low-batt:
Your setup is really close to what I'm testing with. I'm testing on my 16" MacBook Pro with the M1 Max chip. I'm building with Xcode 14.0.1.

When the problem reproduces do you see the same behavior where the screen dims when the OSC disappears and brightens up if you move the mouse and the OSC appears again?

The idea behind the workaround is to make the OSC extremely transparent instead of completely hiding it so playback will always act like it does when the OSD is displayed.

I was also unable to reproduce the odd problem reported in issue #4015 when running with this workaround. That does not involve HDR. It involves multiple monitors. So Ventura graphics problems extend beyond this HDR problem.

There are significant changes related to graphics in Ventura. From [macOS Ventura is now available](https://www.apple.com/newsroom/2022/10/macos-ventura-is-now-available/):
> macOS Ventura brings new technologies like Metal 3 

Likely the problems we are experiencing are related to those changes.

The workaround is setting the [alphaValue](https://developer.apple.com/documentation/appkit/nsview/1483560-alphavalue) to `0.01`:

```swift
  // Workaround for macOS 13 where HDR content becomes dimmed when overlaying views are fully hidden. (#3844)
  private var almostInvisibleAlpha: CGFloat {
    if #available(macOS 13, *) {
      if fsState.isFullscreen && player.info.hdrEnabled {
        return 0.01
 ```

You could try increasing that value and see if that makes a difference. Of course if that works we'd be confused as to why the current setting works on my Mac and not yours.

I'm still thinking about what other Mac settings might be coming into play to cause the difference in behavior between our Macs.

**Comment by svobs** [comment]
I've started looking into this and have managed to acquire an HDR display (a Sony XBR-43X800H) and testing on a 2021 M1 Macbook Air, using the develop branch. 

I'm seeing a few issues. I'm launching IINA and playing an HDR video on the HDR display by clicking on the video in Welcome window. I find that if I move the Welcome window into the HDR display before clicking the link, it activates HDR. If I click the Welcome window while it's on my Macbook display, even though it opens the video on the HDR display, HDR is disabled and can't be selected from the side panel.

Also, I'm seeing the dimming issue in fullscreen - only when HDR has been enabled. I've been adding logging, and I also see the following when coming out of fullscreen (this might end up being a different issue...not sure). I do not see these message s except when HDR has been enabled.

```
19:25:29.330 [iina][d] Deactivating fullscreen <-- my logging
19:25:29.922 [iina][w] mpv log: [cplayer] warn: 
19:25:29.922 [iina][w] mpv log: [cplayer] warn: Audio/Video desynchronisation detected! Possible reasons include too slow
19:25:29.922 [iina][w] mpv log: [cplayer] warn: hardware, temporary CPU spikes, broken drivers, and broken files. Audio
19:25:29.922 [iina][w] mpv log: [cplayer] warn: position will not match to the video (see A-V status field).
19:25:29.922 [iina][w] mpv log: [cplayer] warn: 
19:25:36.574 [iina][d] Activating fullscreen  <-- my logging
```

Some questions:

1. It looks like there are bug fixes for HDR in the iina-plus project.  Is there a timeline for merging those into the upstream project? I suppose I'll try to merge the changes there and try to use those. 
2. Curious about the workaround...can the translucent layer made smaller and still work? Would it be possible to reduce it to one pixel, or does it need to be the whole screen?

First guess is that the fullscreen bug has something to do with Apple's new whiz-bang EDR which is supposed to produce HDR displays out of not-HDR and seems sort of ill-defined.

I also see some potential issues with the HDR code, with nil boolean values getting interpreted as false, and some possible race conditions. Those probably are outside the scope of this bug, but shouldn't be ruled out. More research is needed. Will continue to investigate...

**Comment by svobs** [comment]
OK just saw https://github.com/iina/iina/issues/4015#issuecomment-1296054995 so feeling less overwhelmed. But looks like there are some other potential issues which I'll work on getting sorted out and filing separate tickets for.

**Comment by low-batt** [comment]
Issue #4015 is interesting. I reproduced that one running IINA on the builtin display and running Firefox in fullscreen on an external monitor. The problem causes the external monitor to sometimes go black. That problem exhibits similar behavior. IINA needs to be in full screen mode, playing a video and not displaying any UI components. BUT it does not require HDR. Running IINA with the workaround I was unable to reproduce the problem.

Issue #4015 was just updated with a report that the problem has been reproduced without using IINA.

The behavior with the welcome window, HDR and an external monitor definitely sounds like a defect. See if that is reproducible with IINA+.

Momentary A/V desynchronization can occur when going in and out of full screen depending upon the amount of animations. If you enable "Reduce motion" in macOS settings and enable "Use legacy full screen" in IINA then desynchronization is less likely.

Two HDR related changes from IINA+ have just been merged. At the moment most merging is paused as it is looking like we will need to address this Ventura issue and cut a release, assuming this workaround is viable. We need to explain why the workaround is failing for @aidan-gibson. 

On the question about a smaller translucent layer see [this comment](https://github.com/iina/iina/pull/4017#pullrequestreview-1158954877) above.

My guess is these macOS problems are a result of faulty changes related to the new Metal 3, rather than EDR which has been in macOS for a while.

One of the HDR two fixes that were merged into develop addresses a race condition. If there are others then we definitely want to fix them as well.

I see you replied while I was writing this. That this kind of behavior can be reproduced without using IINA is definitely a relief. I was concerned that changes in Ventura exposed some sort of lurking problem in IINA.

The important issue at the moment is why did this workaround only work for @aidan-gibson if "Use legacy full screen" is enabled.

**Comment by svobs** [comment]
I've been shotgunning a bunch of tests and I found some random results which may or may not be helpful.

## 1
I ran a custom build where I hard-coded the HDR colorspace to `nil`:

```Swift
// MARK: - HDR

@available(macOS 10.15, *)
extension VideoView {
  func refreshEdrMode() {

    //  ... 

    Logger.log("Will activate HDR color space instead of using ICC profile", level: .debug, subsystem: hdrSubsystem);

//    videoLayer.colorspace = CGColorSpace(name: name!)
    videoLayer.colorspace = nil  // <-- this
    mpv.setString(MPVOption.GPURendererOptions.targetTrc, gamma)
    mpv.setString(MPVOption.GPURendererOptions.targetPrim, primaries)
    return true;
  }
}
```

The result looks the same as when the fullscreen bug presents itself. I can't know for sure, but when I go into fullscreen and let the on-screen elements settle, there's no change. So it's possible this might be what's occurring. Although it's not a very strong clue, it's something...

## 2
So, seeing as the workaround involves putting an alpha-blending element on the screen, I decided to see if maybe the bug has to do with any of the on-screen displays. So I removed all the OSC, OSD, loading indicators, and other stuff from the window so I could see whether that would help or hurt it. So I set everything except the video to hidden, commented out all the bodies of all the methods which use them and all the mouse listeners. I also reverted the changes to the previous test so that the colorspace was again being set for HDR.

The result was:

- When I enter fullscreen, my mouse cursor isn't hidden (oops?), but the bug doesn't present itself and the video plays as it should - until I move the cursor even slightly, and then the video goes dim and drab. And this seems to be its new stable state. Leaving the cursor alone won't bring back the vibrancy.
- Although - moving the cursor up and displaying the menubar will bring the video back, but as soon as the menu bar is hidden again, it goes drab.

---

Maybe I just re-treaded the path @aidan-gibson took. It's easy to speculate that this bug is happening when the EDR daemons are checking whether there's any EDR content on the screen...and assuming it ignored the video, if it doesn't see anything else it may conclude that EDR processing isn't needed and shut it down...or something along those lines.

I still haven't run any tests with the workaround, or peered into legacy fullscreen - I used up my time quota for this for now.

aidan-gibson:
Rebuilt it and it's working now?? Legacy fullscreen is off, all is well! No idea what I fucked up before. Sorry about that! Full steam ahead.

low-batt:
@aidan-gibson No problem. Happens. Really appreciate the update as the developers were chatting today and were very worried about the report of the workaround not always working. We will be moving ahead with the workaround.

svobs:
Today I saw the defect affect my whole screen. I had it plugged in and was going to do some more testing and realized the screen seemed dim. I double-checked the settings on my display, then unplugged its Thunderbolt cable from my MacBook Air M1, and then plugged it in again. I saw the following:

1. Connected the display to my Mac.
2. ~5 sec later: it displays my desktop and a couple of windows including an HDR video in IINA.
3. ~1 sec later: it gets brighter, and the IINA video has more contrast.
4. ~0.5 sec later: it dims again.

I wasn't able to restore it to full brightness unless I disconnected & reconnected it again, but still got the same flicker and death.

Restarted, and things were back to normal...

I don't think I'm going to do any more testing for this. Apple should have fixed it, or else delayed the release. Never trust MacOS updates which occur simultaneously with iOS updates, I suppose.

Still, I'm not opposed to implementing the workaround if it helps 95% of the time. Can always remove it in a later build. Just need to remember it's important to manage expectations.

low-batt:
Another person had the screen going into an infinite dim/bright cycle while the video was paused in fullscreen. The graphics portion of macOS Ventura is definitely broken in a very odd way.

low-batt:
Reviewers asked for changes to the workaround. I've adjusted the workaround accordingly and created a new PR, #4031. Closing this PR in favor of that one.

yanhuichao:
这个修复版本预计什么时候会放出来吖

low-batt:
以英语为母语的人使用在线翻译器进行了回应。

替换的PR #4031已经被审核并合并到开发分支。目前，该修正只在未签名的[夜间构建](https://nightly.iina.io/)中可用，或者如果你从开发分支中自己构建IINA的话。IINA维护版本的工作正在进行中。

这个 "修复 "是一个变通方案，我们希望它在大多数情况下都能发挥作用。问题的实际来源似乎是macOS Ventura的一个退步。希望苹果公司正在进行修复工作。

The replacement PR #4031 has been reviewed and merged into the develop branch. At the moment the fix is only available in an unsigned [nightly build](https://nightly.iina.io/) or if you build IINA yourself from the develop branch. Work is in progress on an IINA maintenance release.

The "fix" is a workaround that we hope will be effective in most cases. The actual source of the problem appears to be a regression in macOS Ventura. Hopefully Apple is working on a fix.

Issue #4025, opened 2022-10-31
HDR status not immediately updated on screen change


svobs:
**System and IINA version:**

- macOS 13.0
- IINA develop branch 15d0451e9c8ba9113f85081d6586adee80def4df

**Expected behavior:**
Dragging an HDR video from an HDR screen to SDR (and vice versa) should immediately reconfigure its HDR status

**Actual behavior:**
It doesn't. The video needs to be paused & started again, or some other trigger.

**Steps to reproduce:**

As the difference in visual quality between HDR and SDR can vary widely and is hard to quantify, the best way to verify is to watch the log. It's easier if first applying #4024.

Use a display which supports HDR, and one which does not. Note: "EDR" displays such as recent Macbooks are considered HDR by IINA.

1. Open a video in IINA. It may open in either screen. Either don't play it, or start playing it but make sure it's long enough to run the whole test.
2. Drag and drop it from its original screen into the other one.
3. Drag and drop it back into its original screen.
4. Check the IINA log to see if it updated. If you included PR #4024, the correct result would be to see a line that starts with `Refreshing HDR` each time it changed screens. If not, you should still expect to see `Will activate HDR color space instead of using ICC profile` when the window moved _into_ the HDR screen, but nothing the other way. With this defect there won't be anything.

svobs:
I did more research, and it looks like this was a regression which is present in neither the last release nor in IINA-Plus; rather, it was created when the plugin-system code was merged into `develop`.  I'll put more comments in the PR.

PR #4026, opened 2022-10-31
Fix: PlayerWindowController.windowDidChangeScreen not getting called (issue 4025)

svobs:
- [x] This implements/fixes issue #4025.

---

**Description:**
Looks like this slipped through the cracks somehow.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -1536,6 +1536,8 @@ class MainWindowController: PlayerWindowController {
   }
   
   override func windowDidChangeScreen(_ notification: Notification) {
+    super.windowDidChangeScreen(notification)
+
     player.events.emit(.windowScreenChanged)
   }
 


svobs:
Some more explanation:

- Full-sized player windows use `MainWindowController` which is a child class of `PlayerWindowController`
- In the last IINA release,`PlayerWindowController` implemented `windowDidChangeScreen()`, which notifies it that the window changed screens, and calls `videoView.updateDisplayLink()`, which calls `refreshEdrMode().
- When the `plugin-system` code was merged, it implemented `windowDidChangeScreen()` in `MainWindowController`, which was not previously implemented. It didn't call `super` because at its time of writing, the other file didn't include it before.

This seems to make a good case against avoiding inheritance which I didn't think about before.

low-batt (approved):
I reproduced the problem using a build of develop and my MacBook Pro with XDR screen and an external monitor. I confirmed IINA 1.3.0 correctly enables or disables HDR as IINA's window moves between displays. I pulled the PR locally and confirmed the fix corrects the regression.

Issue #4013, opened 2022-10-26
全屏看视频会变暗


smartes0:
#一些视频全屏以后不动鼠标就会突然屏幕变暗怎么办

singh-jagjot:
Similar to issue 4014 [https://github.com/iina/iina/issues/4014](url)

low-batt:
以英语为母语的人使用在线翻译器回答......

正如@singh-jagjot所指出的，描述的行为与macOS Ventura下全屏播放HDR视频时出现的问题相吻合。这个问题已经在多个问题中被报告，包括#4014。

我们目前正在努力为IINA增加一个解决这个问题的方法。

As @singh-jagjot indicated the described behavior matches a problem seen under macOS Ventura when playing HDR videos in full screen. This has been reported in multiple issues including #4014.

We are currently working on adding a workaround to IINA for this problem.

lty2008one:
暂时可以通过

uiryuu:
See #4014, closing

PR #4031, opened 2022-11-03
Workaround for dimmed HDR content in full screen on macOS 13, #3844

low-batt:
This workaround also addresses another macOS Ventura regression where an external monitor goes black, #4015.

This same dimming problem was reported in issues #3880, #4013 and #4014.

The workaround changes the windowDidLoad method of MainWindowController such that when running under macOS Ventura it will add a tiny subview with such a low alpha level it is invisible to the human eye.

Issue #4015 has been reproduced without IINA, so we are certain this is a regression in macOS 13. At this time we do not know the root cause and can not explain why this workaround works.

- [x] This implements/fixes issue #3844.

---

**Description:**
Starting with macOS 13, some strange behavior can be seen when viewing videos in full screen in HDR mode. The screen will be dimmed when the player controls disappear. This workaround adds a tiny invisible subview when running under macOS 13 so that there is always more than one "visible" view in the window.

Many thanks to @royj  who came up with the original workaround. This PR addresses feedback from reviewers.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -550,6 +550,19 @@ class MainWindowController: PlayerWindowController {
     // gesture recognizer
     cv.addGestureRecognizer(magnificationGestureRecognizer)
 
+    // Workaround a bug in macOS Ventura where HDR content becomes dimmed when playing in full
+    // screen mode once overlaying views are fully hidden (issue #3844). After applying this
+    // workaround another bug in Ventura where an external monitor goes black could not be
+    // reproduced (issue #4015). The workaround adds a tiny subview with such a low alpha level it
+    // is invisible to the human eye.
+    if #available(macOS 13, *) {
+      let view = NSView(frame: NSRect(origin: .zero, size: NSSize(width: 0.1, height: 0.1)))
+      view.wantsLayer = true
+      view.layer?.backgroundColor = NSColor.black.cgColor
+      view.layer?.opacity = 0.01
+      cv.addSubview(view)
+    }
+
     player.initVideo()
 
     // init quick setting view now
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -550,11 +550,11 @@ class MainWindowController: PlayerWindowController {
     // gesture recognizer
     cv.addGestureRecognizer(magnificationGestureRecognizer)
 
-    // Workaround a bug in macOS Ventura where HDR content becomes dimmed when playing in full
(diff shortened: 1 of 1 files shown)


iina/MainWindowController.swift
```diff
@@ -550,6 +550,19 @@ class MainWindowController: PlayerWindowController {
     // gesture recognizer
     cv.addGestureRecognizer(magnificationGestureRecognizer)
 
+    // Workaround a bug in macOS Ventura where HDR content becomes dimmed when playing in full
```
saagarjha, line 553:
```suggestion
    // Work around a bug in macOS Ventura where HDR content becomes dimmed when playing in full
```

iina/MainWindowController.swift
```diff
@@ -550,6 +550,19 @@ class MainWindowController: PlayerWindowController {
...
+    // screen mode once overlaying views are fully hidden (issue #3844). After applying this
+    // workaround another bug in Ventura where an external monitor goes black could not be
+    // reproduced (issue #4015). The workaround adds a tiny subview with such a low alpha level it
+    // is invisible to the human eye.
```
saagarjha, line 557: I'd like a short comment mentioning that this workaround isn't perfect

royj:
Hopefully this can be merged soon. Glad I could help 🙂

iina/MainWindowController.swift
low-batt: Added note that it might not be effective in all cases.

lhc70000:
@uiryuu Please confirm that it works on your machine, then you can merge it.

uiryuu (approved):

ownia:
Is there a more elegant solution? IMO this workaround sucks.

low-batt:
@ownia I'm pretty certain all of us agree with your assessment of this fix.

With this workaround in place I was unable to reproduce issue #4015. That problem has been reproduced just using Firefox. This indicates the regression in macOS Ventura is not tied to something special about IINA.

Both the dimming problem and the black screen problem occur when IINA is in full screen mode and not showing the on screen controller. At least we think that behavior is consistent.

We do not know the root cause of the problem in Ventura. We can not explain why the problem does not occur when the change in this PR in place other than we are making it seem like the OSC is always being displayed to avoid one of the conditions we think are required to trigger the macOS 13 regression.

The term for this kind of fix is [hackaround](https://en.wiktionary.org/wiki/hackaround).

Any additional information concerning this problem or suggestions for alternative workarounds welcome.

Issue #4022, opened 2022-10-30
Crash when passed - in parameter list


xrisk:
**System and IINA version:**

- macOS 13.0
- IINA 1.3.0

**Expected behavior:**

`iina -` does not cause crash. If possible, it should try to read input from stdin (equivalent to —stdin flag) for parity with mpv and vlc. At the very least, it should not cause a crash.

**Actual behavior:**

Causes crash.

<details>
<summary><b>Crash report:</b></summary>

```
{"app_name":"IINA","timestamp":"2022-10-30 20:12:00.00 +0530","app_version":"1.3.0","slice_uuid":"f7a5a99b-6f8a-3f45-b5a3-47523f809c97","build_version":"132","platform":1,"bundleID":"com.colliderli.iina","share_with_app_devs":1,"is_first_party":0,"bug_type":"309","os_version":"macOS 13.0 (22A380)","roots_installed":0,"name":"IINA","incident_id":"1C2219BA-428C-4644-8BC2-2680DB98D75F"}
{
  "uptime" : 64000,
  "procRole" : "Background",
  "version" : 2,
  "userID" : 501,
  "deployVersion" : 210,
  "modelCode" : "MacBookPro18,3",
  "coalitionID" : 1230,
  "osVersion" : {
    "train" : "macOS 13.0",
    "build" : "22A380",
    "releaseType" : "User"
  },
  "captureTime" : "2022-10-30 20:12:00.6265 +0530",
  "incident" : "1C2219BA-428C-4644-8BC2-2680DB98D75F",
  "pid" : 31725,
  "translated" : false,
  "cpuType" : "ARM-64",
  "roots_installed" : 0,
  "bug_type" : "309",
  "procLaunch" : "2022-10-30 20:12:00.5286 +0530",
  "procStartAbsTime" : 1539299812096,
  "procExitAbsTime" : 1539302116005,
  "procName" : "IINA",
  "procPath" : "\/Applications\/IINA.app\/Contents\/MacOS\/IINA",
  "bundleInfo" : {"CFBundleShortVersionString":"1.3.0","CFBundleVersion":"132","CFBundleIdentifier":"com.colliderli.iina"},
  "storeInfo" : {"deviceIdentifierForVendor":"6F45AADE-299B-5292-AF09-76C91C812612","thirdParty":true},
  "parentProc" : "launchd",
  "parentPid" : 1,
  "coalitionName" : "com.googlecode.iterm2",
[...]

PR #4032, opened 2022-11-03
Fix crash when single dash ("-") arg is supplied to IINA via cmd line

svobs:
- [x] This implements/fixes issue #4022.

---

**Description:**
Adds a simple check to avoid a fatal error at startup, if a single dash (`-`) is included in the args.

It was easy to confirm the crash, and also confirm the fix, by just supplying an arg in XCode's Scheme Editor:
Dash

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -158,7 +158,11 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         continue
       }
       if arg.first == "-" {
-        if arg[arg.index(after: arg.startIndex)] == "-" {
+        let indexAfterDash = arg.index(after: arg.startIndex)
+        if indexAfterDash == arg.endIndex {
+          // single '-'
+          commandLineStatus.isStdin = true
+        } else if arg[indexAfterDash] == "-" {
           // args starting with --
           iinaArgs.append(arg)
         } else {
@@ -774,10 +778,13 @@ struct CommandLineStatus {
       let name = String(splitted[0])
       if (name.hasPrefix("mpv-")) {
         // mpv args
-        if splitted.count <= 1 {
-          mpvArguments.append((String(name.dropFirst(4)), "yes"))
+        let strippedName = String(name.dropFirst(4))
+        if strippedName == "-" {
+          isStdin = true
+        } else if splitted.count <= 1 {
+          mpvArguments.append((strippedName, "yes"))
         } else {
-          mpvArguments.append((String(name.dropFirst(4)), String(splitted[1])))
+          mpvArguments.append((strippedName, String(splitted[1])))
         }
       } else {
         // other args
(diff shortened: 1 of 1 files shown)


svobs:
> Then IINA instead of translating "-" into "--stdin" or "/dev/stdin" just ignores the "-"? Because of what iina-cli did, that is helpful in this case. But also seems wrong.

I think there are two separate points here.

1. It sounds like you are expecting a lone `-` to be interpreted as equivalent to `--stdin`, but I don't see anything documenting that behavior. Sometimes apps will use `--` in this way, but IINA seems to be honoring this correctly:
```
msvoboda|17:25:59|MacOS$ cat ~/LocalHome/Downloads/bbb_sunflower_2160p_60fps_normal.mp4 | ./iina-cli --
...
17:44:20.057 [iina][d] Got arguments ["--stdin"]
17:44:20.057 [iina][d] IINA arguments: ["--stdin"]
17:44:20.057 [iina][d] Filenames from arguments: []
```

2. It looks like `iina-cli` ignores all tokens it doesn't recognize. While not strictly wrong, it certainly goes against standard practice  for command line tools, which preaches the need for "fail fast" behavior. But it's not doing anything unusual by ignoring `-`. Though I too have issue filing fatigue, this should probably be logged as a separate issue as it's out of the scope of this defect.

low-batt:
I believe in unix it is up to commands as to how they want to interpret "-", so yes, it does not always mean read from stdin.

For IINA we always want to match up with mpv unless we have a good reason to deviate.  From the mpv manual, section [usage](https://mpv.io/manual/stable/#usage):
> One exception is the lone - (without anything else), which means media data will be read from stdin. Also, -- (without anything else) will make the player interpret all following arguments as filenames, even if they start with -. (To play a file named -, you need to use ./-.)

I don't see a reason for IINA to not match up with mpv in this case.

I didn't request changes as the PR fixes a crash and supports reading from stdin like the user requested. Just wanted to hear your thoughts on this aspect of the behavior. We can definitely treat this as a separate issue. There are definitely other issues in regard to CLI behavior.

svobs:
OK, I dug deeper into it, and you are right.

I pushed an updated version which should handle the single dash correctly, and also should handle `--mpv--`.

Seems like it would be easier & less error prone to just use a stricter form of parsing which requires that all non-file arguments are in the form "--key=value", and anything that doesn't start with dashes is an input file. And especially since mpv has indicates that's where they are trying to go in the future. But this isn't really a feature I use anyway.

low-batt (approved):
Pulled PR, built and confirmed running IINA like this works now and does not crash:
```bash
low-batt@gag MacOS$ cat ~/Movies/resume-failure.mp4 | ./iina-cli -
```

uiryuu:
Thanks!

Issue #4020, opened 2022-10-29
Crash during quit in windowDidExitFullScreen


low-batt:
**System and IINA version:**

- macOS 13.0
- IINA develop 15d0451

**Expected behavior:**
IINA does not crash when quitting.

**Actual behavior:**
During termination MainWindowController.windowDidExitFullScreen was called. MainWindowController called PlayerCore which called MPVController.getDouble. That method called mpv_get_property triggering a crash as IINA had already sent a quit command to mpv.

This is with the develop branch that contains changes to try and correct shutdown problems.

<details>
<summary><b>Xcode Console:</b></summary>

```
14:10:23.356 [iina][d] IINA 1.3.0 Build 131
14:10:23.358 [iina][d] Copyright © 2017-2022
14:10:23.358 [iina][d] Collider LI, et al.
14:10:23.358 [iina][d] Released under GPLv3.
14:10:23.358 [iina][d] FFmpeg 4.4.2
14:10:23.358 [iina][d]   libavcodec 58.134.100
14:10:23.358 [iina][d]   libavformat 58.76.100
14:10:23.358 [iina][d]   libavutil 56.70.100
14:10:23.358 [iina][d]   libswscale 5.9.100
14:10:23.358 [iina][d] App will launch
14:10:23.358 [iina][d] Got arguments ["-NSDocumentRevisionsDebugMode", "YES"]
14:10:23.358 [iina][d] IINA arguments: []
14:10:23.358 [iina][d] Filenames from arguments: []
IINA 1.3.0 Build 131
This binary is not intended for being used as a command line tool. Please use the bundled iina-cli.
Please ignore this message if you are running in a debug environment.
14:10:23.379 [iina][d] App launched
14:10:23.380 [iina][d] Loading key bindings
14:10:23.382 [iina][d] Set key bindings
2022-10-29 14:10:23.385423-0400 IINA[9185:878426] [logging] open flag(s) 0x01000000 are reserved for VFS use and do not affect behaviour when passed to sqlite3_open_v2
2022-10-29 14:10:23.385491-0400 IINA[9185:878426] [logging-persist] cannot open file at line 46922 of [6bf7a27121]
2022-10-29 14:10:23.385507-0400 IINA[9185:878426] [logging-persist] os_unix.c:46922: (0) open(/private/var/db/DetachedSignatures) - Undefined error: 0
2022-10-29 14:10:23.486631-0400 IINA[9185:877912] Metal API Validation Enabled
[...]

PR #4035, opened 2022-11-03
Add more checks to prevent requests to mpv during shutdown (fix for issue #4020)

svobs:
- [x] This implements/fixes issue #4020.

---

**Description:**

The crash in the issue appears to be caused by a null pointer somewhere inside libmpv. According to the stack trace, `windowDidExitFullScreen()` called `showUI()`, which called `syncUI()`, which called `mpv_get_property`, which was bad because the mpv core was in the process of shutting down and setting all its pointers to null.

This update adds `isShuttingDown` checks to 4 places where `mpv_get_property` normally gets called and thus could be problematic. It expects that `isShuttingDown==true` only happens at application shutdown, and that it will still be true even when `isShutdown==true`.

As a side effect, IINA's shutdowns might be a tiny bit faster.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -1268,6 +1268,8 @@ class MainWindowController: PlayerWindowController {
   }
 
   func windowDidExitFullScreen(_ notification: Notification) {
+    guard !player.isShuttingDown else { return }
+    
     if AccessibilityPreferences.motionReductionEnabled {
       // When animation is not used exiting full screen does not restore the previous size of the
       // window. Restore it now.
@@ -1659,7 +1661,8 @@ class MainWindowController: PlayerWindowController {
   }
 
   private func showUI() {
-    if player.disableUI { return }
+    guard !player.disableUI && !player.isShuttingDown else { return }
+
     animationState = .willShow
     fadeableViews.forEach { (v) in
       v.isHidden = false
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -1604,6 +1604,7 @@ class PlayerCore: NSObject {
   func syncUI(_ option: SyncUIOption) {
     // if window not loaded, ignore
     guard mainWindow.loaded else { return }
+    guard !disableUI && !isShuttingDown else { return }
     Logger.log("Syncing UI \(option)", level: .verbose, subsystem: subsystem)
 
     switch option {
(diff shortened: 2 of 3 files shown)


low-batt:
Caught me working on issue #4020. That is why I assigned it to myself. I have several objections to the proposed fix. The first is a code design one. The current attempt to get to a clean shutdown is following the philosophy that the high level code has the primary responsibility for controlling the shutdown. Now this approach might prove to be problematic and have to be abandoned. But that is the current design.

An alternative low level approach would tackle this at the low level. For example we could refactor the code so that the mpv context is private and all calls must go though `MVPController` (that should be done at some point anyway). Then change `MPVController` to not permit calls to mpv while it is shutting down, somewhat like what was done for the `Logger`. Easy to do for calls that do something like set a property, you just make the method do nothing. But what do you do for calls that return a value? The technique I've used in the past was exactly what you suggested, throwing some sort of shutdown exception. As you touched on, the trouble with that technique is that you must insure no code is confused by that exception. The other issue with the low level approach is that you end up with checks all over the place. The low level approach also brings up issues with coordinating threads. This is why the current design tries to push shutdown coordination to the high level code. So I would prefer to try and avoid making middle and low level code deal with shutdown.

The crash points out another thing the code is doing wrong. The code is not just accessing mpv while it is quitting, it is also accessing mpv while mpv is asynchronously processing a stop command. That is happening when you close a window. As this is a thread race condition we don't know if there is some sort of bad behavior that is sufficiently rare we have not yet identified it. So I was hoping for a fix that would avoid accessing mpv once `windowWillClose` has send a `stop` command to mpv. Something like adding a `isClosing` property to `MainWindowController` and having the higher level code in that class recognize the window is closing and avoid needless work, including calls to mpv. That is not as easy as it sounds as window objects are reused through player core caching, so they must be left in a valid state.

I was hoping to get a fix for this one ready ASAP as there is a chance we might cut a release to get the partial workaround for the Ventura dimming out there. Do you want to continue working on this one?

svobs:
Ah, I'm sorry, in my zealotry I didn't notice you were assigned to the task! I'll close this PR and let you tackle it. As I indicated in my issue comment I didn't think it was the best solution anyway; just an expedient one.

I agree that there are a bunch of timing issues related to the stop command too, but it looks to me like they will be both harder to fix than shutdown and with provide less payoff. For example, it looks like there's a good chance that the syncUI timer will slip through and keep repeating after a stop command is issued, but when that happens the damage will just slightly higher CPU usage, not a crash.

Definitely agree that an architectural change would be best, but looking forward to seeing what you can come up with!

PR #4048, opened 2022-11-08
Ignore warnings from FFmpeg' and mpv's header files

uiryuu:

---

**Description:**
Ref: https://stackoverflow.com/questions/59720526/how-can-i-ignore-all-warnings-from-included-headers-in-clang-gcc

### DIFF
--- a/iina/FFmpegController.m
+++ b/iina/FFmpegController.m
@@ -10,7 +10,7 @@
 #import <Cocoa/Cocoa.h>
 
 #pragma clang diagnostic push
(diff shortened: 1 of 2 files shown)


saagarjha:
Do we really need something this blunt?

uiryuu:
Do you have better idea?

low-batt:
The warnings are all about documentation comments, so this works:
```c
#pragma clang diagnostic ignored "-Wdocumentation"
```

PR #4066, opened 2022-11-11
Close add build info to About window, #3977

### LINKED ISSUE #3977: Add hidden info about the build to the About window
**What you want IINA to do:**
Provide a way to view information that identifies a development build in the UI.

**What IINA does currently:**
Depending upon the nature of changes in a development build it may be very difficult to tell from the UI that you are not running the official release.

Some developers take development builds to other Macs in order to:
- Test under a different version of macOS
- Test for a long time under normal use

When doing such testing it is desirable to have a way to easily identify development builds.

One way to address this would be to add the build date and the git branch from which the build was done to the `About` window. As this information is only needed by developers it would be hidden by default. To reveal it a developer would option-click between the mpv version and the `License` button.

This proposal builds upon the logging of such information proposed by issue #3970.

I looked, but did not find a standard convention for identifying developer builds of Mac applications.

[Semantic Versioning](https://semver.org/#semantic-versioning-200) specifies a way to include [build metadata](https://semver.org/#spec-item-10) in a version number. But such techniques can not be used in an Apple version number.

Revealing hidden information in an about window can be seen in the `About This Mac` window. Clicking on the macOS version reveals the build number as discussed in [Find out which macOS your Mac is using](https://support.apple.com/en-us/HT201260)

Using option-click to reveal additional information is also used in macOS. If you option-click the Wi-Fi icon in the menu bar additional details about the connection are shown.

As this build information is not useful to the user using the released version of IINA revealing should require more than a simple click.

I have changes that add this to IINA's `About` window. As this builds upon the changes in PR #3971 I will prepare a PR once that PR has been merged (assuming it is approved).

low-batt:
This commit will:
- Add a new class InfoDictionary that exposes some Info.plist entries as properties
- Change AppDelegate, GuideWindowController, InitialWindowController, and JavascriptAPICore to use InfoDictionary instead of methods in the Utility class
- Remove iinaCopyright and iinaVersion methods from Utility class
- Add FFmpeg version to About window
- Add build date and build branch as hidden entries in About window that are exposed via an option-click

- [x] This implements/fixes issue #3977.

---

**Description:**

### DIFF
--- a/iina/AboutWindowController.swift
+++ b/iina/AboutWindowController.swift
@@ -59,27 +59,6 @@ class AboutWindowController: NSWindowController {
   private lazy var contributors = getContributors()
   private lazy var translators = loadTraslators()
 
-  /// The user has released the left mouse button.
-  ///
-  /// The `About Window` supports displaying the date and time when the IINA executable was built along with the Git branch from
-  /// which the build was made. As this information is only interesting to developers that take builds to other machines for long term
-  /// testing this information is hidden and only reveled if the user option-clicks between the mpv and FFmpeg versions and the
-  /// `License` button.
-  override func mouseUp(with event: NSEvent) {
-     guard event.modifierFlags.contains(.option), buildDateLabel.isHidden,
-          buildView.isMousePoint(event.locationInWindow, in: buildView.frame),
-          let buildDate = InfoDictionary.shared.buildDate,
-          let buildBranch = InfoDictionary.shared.buildBranch else {
-      super.mouseUp(with: event)
-      return
-    }
-    buildDateLabel.stringValue = buildDate
-    buildBranchLabel.stringValue = buildBranch
-    buildDateLabel.isHidden = false
-    buildBranchLabel.isHidden = false
-    window?.contentView?.needsDisplay = true
-  }
-
   override func windowDidLoad() {
     super.windowDidLoad()
 
@@ -94,6 +73,14 @@ class AboutWindowController: NSWindowController {
 
     mpvVersionLabel.stringValue = PlayerCore.active.mpv.mpvVersion
     ffmpegVersionLabel.stringValue = "FFmpeg \(String(cString: av_version_info()))"
+
+    if let buildDate = InfoDictionary.shared.buildDate,
+       let buildBranch = InfoDictionary.shared.buildBranch {
+      buildDateLabel.stringValue = buildDate
+      buildDateLabel.isHidden = false
+      buildBranchLabel.stringValue = buildBranch
+      buildBranchLabel.isHidden = false
+    }
 
     if let contrubutionFile = Bundle.main.path(forResource: "Contribution", ofType: "rtf") {
       detailTextView.readRTFD(fromFile: contrubutionFile)
--- a/iina/AboutWindowController.swift
+++ b/iina/AboutWindowController.swift
@@ -38,6 +38,11 @@ class AboutWindowController: NSWindowController {
   }
   @IBOutlet weak var versionLabel: NSTextField!
   @IBOutlet weak var mpvVersionLabel: NSTextField!
+  @IBOutlet weak var ffmpegVersionLabel: NSTextField!
+  @IBOutlet weak var buildView: NSView!
+  @IBOutlet weak var buildBranchLabel: NSTextField!
+  @IBOutlet weak var buildDateLabel: NSTextField!
+
   @IBOutlet var detailTextView: NSTextView!
   @IBOutlet var creditsTextView: NSTextView!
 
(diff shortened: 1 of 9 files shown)


low-batt:
The About window now shows the FFmpeg version:
issue-3977-before-click

An option-click between the versions and the `License` button reveals information about the build:
issue-3977-after-click

uiryuu (approved):

iina/InfoDictionary.swift
```diff
@@ -0,0 +1,56 @@
...
+//  Copyright © 2022 lhc. All rights reserved.
+//
+
+import Foundation
+
+/// Wrapper for the dictionary constructed from the bundle’s `Info.plist` file.
+///
+/// This class exposes some of the entries contained in `Bundle.main.infoDictionary` as properties to provide for easier access
+/// to information contained in the dictionary from other classes.
+class InfoDictionary {
```
saagarjha:
```suggestion
struct InfoDictionary {
```

iina/InfoDictionary.swift
```diff
@@ -0,0 +1,56 @@
...
+  let buildDate: String?
+  
+  let bundleIdentifier: String
+  
+  var copyright: String { Bundle.main.infoDictionary!["NSHumanReadableCopyright"] as! String }
+
+  var version: (String, String) {
+    let infoDic = Bundle.main.infoDictionary!
+    return (infoDic["CFBundleShortVersionString"] as! String, infoDic["CFBundleVersion"] as! String)
+  }
```
saagarjha: Why are these two computed and the others aren't?

iina/InfoDictionary.swift
low-batt: Darn. Forgot about structs yet again. Fixed.

iina/InfoDictionary.swift
low-batt: Originally were stored properties. Then I began thinking reviewers might question why we are storing these values that are rarely referenced. So I converted these to match up with the `Utility` methods they replaced. I've pushed a commit that changes them all to be computed. If you would prefer they are all initialized once in an init method just let me know and I'll change it to be all stored properties.

iina/AboutWindowController.swift
```diff
@@ -54,6 +59,27 @@ class AboutWindowController: NSWindowController {
   private lazy var contributors = getContributors()
   private lazy var translators = loadTraslators()
 
+  /// The user has released the left mouse button.
+  ///
+  /// The `About Window` supports displaying the date and time when the IINA executable was built along with the Git branch from
+  /// which the build was made. As this information is only interesting to developers that take builds to other machines for long term
+  /// testing this information is hidden and only reveled if the user option-clicks between the mpv and FFmpeg versions and the
+  /// `License` button.
+  override func mouseUp(with event: NSEvent) {
```
saagarjha: I don't think it is appropriate to hide this information

saagarjha (changes requested):

iina/AboutWindowController.swift
low-batt: Fixed. The build date and branch are no longer hidden. The code is "friendly" and will just not log or display the build information if it is missing from `Info.plist`.

I copied macOS's convention of hiding information in `About This Mac` as I thought I would hear lots of objections to including this information during review such as:
- Normal users do not know what a source control branch is
- The additional information makes the window look cluttered
- Both the build date and branch are well known for the release version
- The build date and branch are not labeled
- Reminding users of the release date may increase the calls for a new release
- The branch for the release duplicates the version number that is already listed
- Other applications do not display such information in their `About` windows

Of those the one that bothers me is the inclusion of the branch name. That is an implementation detail that most users will not understand.

If you have any second thoughts about showing the branch name do not hesitate to ask for more changes. I'm happy to tweak this some more if needed. If there is additional information you think should be included I can look into that as well, or we could plan to add that in the future as I do need to get back to fixing bugs.

Most applications do not display this kind of information, but I dug around some more and found a few examples of applications that include information for developers.

LibreOffice has a link to the git commit along with other information of interest to developers on its `About` window:

issue-3977-LibreOffice

Wireshark lists a lot of information for developers on its `About` window:
issue-3977-Wireshark

KeePassXC has a `Debug Info` tab on its `About` window. I was unable to take a screen shot of that window. I used the `Copy to clipboard` button to capture what it shows:
```text
KeePassXC - Version 2.7.1
Revision: 5916a8f

Qt 5.15.2
Debugging mode is disabled.

Operating system: macOS 13.0
CPU architecture: arm64
Kernel: darwin 22.1.0

Enabled extensions:
- Auto-Type
- Browser Integration
- SSH Agent
- KeeShare
- YubiKey
- Quick Unlock

Cryptographic libraries:
- Botan 2.19.1
```

I definitely prefer the cleaner more stylish look of IINA's `About` window over these windows.

I do plan on eventually enhancing startup to log more information about the Mac IINA is running on.

PR #4084, opened 2022-11-17
UX improvements to the “Jump to…” command

### LINKED ISSUE #4547: "Playback" > "Jump To..." should allow seconds-only input
**What you want IINA to do:**

The `Jump To...` item the Playback menu doesn't support entering only seconds.

**What IINA does currently:**

If I open the `Jump To...` dialog and type `40` and hit `RETURN`, it does nothing. Could even be considered a bug.
Same result when typing `:40` (nothing). To go to t=40s, I need to type `0:40`.

Would be easier & more user friendly. I don't like unnecessary typing. Especially typing colons. Using the Shift key is laborious. Sometimes would be nice to type `300` instead of `5:00` to go to t=5m.

YouTube's `t` URL param allow different time formats, for example either `t=121` or `t=2m1s`

sabberworm:
- [✔] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)

---

This PR adds small UX improvements to the *Playback* → *Jump to…* command:

1. Allows inputting sub-second precision timecodes to jump to
2. Displays the current video time code as the initial value
3. Has the added benefit of making the current time code easy to copy to the clipboard

Initially I had wanted a quick way to copy the current time code. I was already starting to implement a dedicated command for that when I realized the same could be accomplished more synergistically by beefing up the existing “Jump to…” command.

Will also fix #4547

### DIFF
--- a/iina/MainMenuActions.swift
+++ b/iina/MainMenuActions.swift
@@ -130,7 +130,7 @@ extension MainMenuActionHandler {
   @objc func menuJumpTo(_ sender: NSMenuItem) {
     Utility.quickPromptPanel("jump_to", inputValue: self.player.info.videoPosition?.stringRepresentationWithPrecision(3)) { input in
       if let vt = VideoTime(input) {
-        self.player.seek(absoluteSecond: Double(vt.second))
+        self.player.seek(absoluteSecond: vt.second)
       }
     }
   }
--- a/iina/MainMenuActions.swift
+++ b/iina/MainMenuActions.swift
@@ -128,7 +128,7 @@ extension MainMenuActionHandler {
   }
 
   @objc func menuJumpTo(_ sender: NSMenuItem) {
-    Utility.quickPromptPanel("jump_to") { input in
+    Utility.quickPromptPanel("jump_to", inputValue: self.player.info.videoPosition?.stringRepresentationWithPrecision(3)) { input in
       if let vt = VideoTime(input) {
         self.player.seek(absoluteSecond: Double(vt.second))
       }
--- a/iina/VideoTime.swift
+++ b/iina/VideoTime.swift
@@ -49,21 +49,13 @@ class VideoTime {
   }
 
   convenience init?(_ format: String) {
-    let split = format.split(separator: ":").map { (seq) -> Int? in
-      return Int(String(seq))
-    }
-    if !(split.contains {$0 == nil}) {
-      // if no nil in array
-      if split.count == 2 {
-        self.init(0, split[0]!, split[1]!)
-      } else if split.count == 3 {
-        self.init(split[0]!, split[1]!, split[2]!)
-      } else {
-        return nil
-      }
-    } else {
-      return nil
-    }
+    let split = Array(format.split(separator: ":").map { String($0) }.reversed())
+
+    let hour : Int? = split.count > 2 ? Int(split[2]) : nil
+    let minute : Int? = split.count > 1 ? Int(split[1]) : nil
+    let second : Double? = !split.isEmpty ? Double(split[0]) : nil
+
+    self.init(hour ?? 0, minute ?? 0, second ?? 0.0)
   }
 
   init(_ second: Double) {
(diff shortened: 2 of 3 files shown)


iina/MainMenuActions.swift
```diff
@@ -124,7 +124,7 @@ extension MainMenuActionHandler {
   }
 
   @objc func menuJumpTo(_ sender: NSMenuItem) {
-    Utility.quickPromptPanel("jump_to") { input in
+    Utility.quickPromptPanel("jump_to", inputValue: self.player.info.videoPosition?.stringRepresentationWithPrecision(3)) { input in
```
sabberworm, line 131: 3 sub-second digits is the highest precision selectable in the UI (and commonly used in video editing). I think it’s a good default. Alternatively, we could use the precision the user has previously chosen to be displayed in the player controls.

iina/VideoTime.swift
```diff
@@ -49,30 +49,22 @@ class VideoTime {
...
-        self.init(0, split[0]!, split[1]!)
-      } else if split.count == 3 {
-        self.init(split[0]!, split[1]!, split[2]!)
-      } else {
-        return nil
-      }
-    } else {
-      return nil
-    }
+    let split = Array(format.split(separator: ":").map { String($0) }.reversed())
```
sabberworm: I don’t do a lot of swift coding so I’m not sure there isn’t a better way to do this.

sabberworm: Probably the new regex builder would be cool to use for parsing but I gather it’s only available with newer versions of swift.

sabberworm:
Is there anything I can do to improve the chances of this getting merged? This feature would have been helpful to me numerous times in the past year.

svobs:
I pulled and tested this fix, and it looks good to me. Unfortunately I can't do anything to get it merged, so 🤞

iina/VideoTime.swift
saagarjha: I think you should be ok with just `let split = format.split(separator: ":").reversed()`?

iina/VideoTime.swift
```diff
@@ -49,30 +49,22 @@ class VideoTime {
...
+
+    let hour : Int? = split.count > 2 ? Int(split[2]) : nil
+    let minute : Int? = split.count > 1 ? Int(split[1]) : nil
+    let second : Double? = !split.isEmpty ? Double(split[0]) : nil
```
saagarjha: I would think that if we couldn't get a proper timestamp with even a seconds value we should probably return nil

iina/VideoTime.swift
sabberworm: I was able to simplify to 

```swift
let split = Array(format.split(separator: ":").reversed())
```

thanks.

iina/VideoTime.swift
sabberworm: I couldn’t return `null` because an initializer cannot return anything but I changed it to throw in this case.

uiryuu:
> Is there anything I can do to improve the chances of this getting merged? This feature would have been helpful to me numerous times in the past year.

@sabberworm Thanks for your contribution. Noted this PR in the backlog. Will merge this before the next beta version.

iina/VideoTime.swift
sabberworm: Ah, of course I can return `nil`, since it’s a fallible initializer. Sorry, I still am new learning idiomatic swift. Will change it.

iina/VideoTime.swift
```diff
@@ -49,30 +48,26 @@ class VideoTime {
...
-        self.init(split[0]!, split[1]!, split[2]!)
-      } else {
-        return nil
-      }
-    } else {
+    let split = Array(format.split(separator: ":").reversed())
+
+    let hour : Int? = split.count > 2 ? Int(split[2]) : nil
+    let minute : Int? = split.count > 1 ? Int(split[1]) : nil
+    let second : Double? = !split.isEmpty ? Double(split[0]) : nil
```
uiryuu:
```suggestion
    let hour: Int? = split.count > 2 ? Int(split[2]) : nil
    let minute: Int? = split.count > 1 ? Int(split[1]) : nil
    let second: Double? = !split.isEmpty ? Double(split[0]) : nil
```

iina/MainMenuActions.swift
uiryuu: 3 digits is good IMO

uiryuu (changes requested):

uiryuu (approved):
LGTM

PR #4074, opened 2022-11-12
Fix crash & other LanguageTokenField issues

svobs:
- [x] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)
- [x] This implements/fixes issue #3505.

---

**Description:**

This was a really tricky one.

The crash log wasn't helpful at all, and the problem wasn't at all related to constraints, but it was accurate in hinting about an infinite loop which AppKit was detecting and killing. By placing log statements all over the code and running it, I was able to determine that the infinite loop was coming from the `LanguageTokenField` class.

For some reason, the `tokenField(displayStringForRepresentedObject)` and `tokenField(representedObjectForEditing)` methods were getting called endlessly and at a ratio of about 5:1. I know that AppKit is kind of messy and tends to call methods a lot more than it needs to, so I wasn't really concerned by all the `displayStringForRepresentedObject` calls, but the smoking gun was the repeated `representedObjectForEditing` calls, which is where AppKit asks the code what the represented object ("token") should be for the text in the text field. The fact that it was calling it over and over suggested that it either wasn't happy with the result it was getting, or thought it hadn't called it. From my experience doing object-mapping in the past, and seeing that a new Token() was being returned each time, I had a hunch that AppKit wasn't able to establish an identity for the Token objects it was getting back - and that turned out to be correct. I was able to get it to stop crashing by adding an override of `isEqual()` to the `Token` class which used its `code` field to check whether one token was identical to another. (Since `isEqual` was not overridden before, the system would default to comparing object references, but since a new `Token` was being returned each time, that would be seen as a new object.)

I don't know why it's doing what it's doing; it's AppKit and the source code isn't available, but during my research I found a commenter who seemed to sum it up well: `NSTokenField` doesn't really like to be customized. As far as I know, there's nothing documented about the need for a returned token to guarantee identity, and that is a glaring omission on Apple's part.

During the course of this investigation I tore the `LanguageTokenField` apart and pieced it back together again. There were a number of things which didn't seem like good ideas, like overriding `stringValue` and the strange use of `LayoutManager` looking for a Unicode attachment char as a way of figuring out whether the user finished typing. I refactored to get rid of the former...but the latter turned out to be essential, because it was catching cases where the standard events were not getting fired.

But there were a few other issues which I discovered and did fix in this update:
1. There were some cases where adding or removing tokens were not getting written to the prefs, because `controlTextDidChange` wasn't getting called. Fixed by adding more listeners.
[...]

PR #4107, opened 2022-11-30
Streamline layout of Prefs window margins to improve consistency & reliability

### LINKED ISSUE #4100: Preferences' first item is too close to the top
**System and IINA version:**

- macOS 10.15.7
- IINA 1.3.1

**Expected behavior:**
In Ventura , it looks right:
截屏2022-11-27 17 22 27

**Actual behavior:**
In Catalina , it looks like :

截屏2022-11-27 下午5 24 59

<details>

```
```
</details>

<br>

<details>

```
```
</details>

<br>

**Steps to reproduce:**

1.3.1 is not correct,1.3.0 is fine

svobs:
- [x] This implements/fixes issue #4100.

---

**Description:**
My fix to #3717 was sort of a Band-Aid, and there were still aspects of the layout which didn't make sense, including relying on a mysterious 36px offset from the top of the Prefs' window's Detail panel which wasn't specified anywhere.

Among other things, this patch should be the final fix for the vertical layout problem. Instead of adding sections to the FlippedView of the NSScrollView, it creates a vertical Stack View inside the FlippedView and appends to that, which is far more predictable.

But this also includes a bunch of work to simplify the Prefs panel margins and make them more consistent. The new Stack View does the work of adding 28px to all sides, so that the pref views which are added to it should just set their leading & trailing offsets to 0, and set their top & bottom offsets to between 0 and 8px (unfortunately still kind of fuzzy for these because there's more mixed content and unreliable ways of getting their content height).

It also cleans up the Prefs window's minimum size, making it smaller but also fixing various visual bugs & undesirable behavior that showed up when changing between Prefs tabs.

### DIFF
--- a/iina/PreferenceWindowController.swift
+++ b/iina/PreferenceWindowController.swift
@@ -129,13 +129,17 @@ class PreferenceWindowController: NSWindowController {
   @IBOutlet weak var searchField: NSSearchField!
   @IBOutlet weak var tableView: NSTableView!
   @IBOutlet weak var maskView: PrefSearchResultMaskView!
-  @IBOutlet weak var scrollView: NSScrollView!
-  @IBOutlet weak var contentView: NSView!
+  @IBOutlet weak var prefDetailScrollView: NSScrollView!  // contains the prefs detail panel (on right)
+  // Check `prefDetailContentView` constraints in the XIB for window content insets
+  @IBOutlet weak var prefDetailContentView: NSView!       // contains the sections stack
+  @IBOutlet weak var prefSectionsStackView: NSStackView!  // add prefs sections to this
   @IBOutlet var completionPopover: NSPopover!
   @IBOutlet weak var completionTableView: NSTableView!
   @IBOutlet weak var noResultLabel: NSTextField!
 
-  private var contentViewBottomConstraint: NSLayoutConstraint?
+  @IBOutlet weak var navTableSearchFieldSpacingConstraint: NSLayoutConstraint!
+
+  private var detailViewBottomConstraint: NSLayoutConstraint?
 
   private var viewControllers: [NSViewController & PreferenceWindowEmbeddable]
 
@@ -160,7 +164,13 @@ class PreferenceWindowController: NSWindowController {
     completionTableView.delegate = self
     completionTableView.dataSource = self
 
-    contentViewBottomConstraint = contentView.bottomAnchor.constraint(equalTo: contentView.superview!.bottomAnchor, constant: -28)
+    detailViewBottomConstraint = prefDetailContentView.bottomAnchor.constraint(equalTo: prefDetailContentView.superview!.bottomAnchor)
+
+    // NSTableView's "Source List" style is only available with MacOS 11.0+ and includes a built-in 10pt offset for its highlights.
+    // But for older MacOS versions, the style will default to "full width" with no highlight offset, which will touch the Search field.
+    if #unavailable(macOS 11.0) {
+      navTableSearchFieldSpacingConstraint.constant = 10.0
+    }
 
     var viewMap = [
       ["general", "PrefGeneralViewController"],
@@ -186,7 +196,6 @@ class PreferenceWindowController: NSWindowController {
       self.isIndexing = false
     }
 
-    loadTab(at: 0)
   }
 
   override func mouseDown(with event: NSEvent) {
@@ -249,14 +258,14 @@ class PreferenceWindowController: NSWindowController {
     if index != tableView.selectedRow {
       tableView.selectRowIndexes(IndexSet(integer: index), byExtendingSelection: false)
     }
-    contentView.subviews.forEach { $0.removeFromSuperview() }
+    prefSectionsStackView.subviews.forEach { $0.removeFromSuperview() }
     guard let vc = viewControllers[at: index] else { return }
-    contentView.addSubview(vc.view)
-    Utility.quickConstraints(["H:|-20-[v]-20-|", "V:|-0-[v]-20-|"], ["v": vc.view])
+    prefSectionsStackView.addSubview(vc.view)
+    Utility.quickConstraints(["H:|-0-[v]-0-|", "V:|-0-[v]-0-|"], ["v": vc.view])
 
     let isScrollable = vc.preferenceContentIsScrollable
-    contentViewBottomConstraint?.isActive = !isScrollable
-    scrollView.verticalScrollElasticity = .none
+    detailViewBottomConstraint?.isActive = !isScrollable
+    prefDetailScrollView.verticalScrollElasticity = .none
 
     // find label
     if let title = title, let label = findLabel(titled: title, in: vc.view) {
--- a/iina/Base.lproj/PrefAdvancedViewController.xib
+++ b/iina/Base.lproj/PrefAdvancedViewController.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="21225" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="21507" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="21225"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="21507"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <objects>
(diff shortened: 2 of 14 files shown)


low-batt:
Tested under macOS 10.15.7 and 13.0.1.

The problem with not leaving a margin at the top under 10.15.7 is fixed. Yeah! However I have two concerns:

In 1.2.0 for preference tabs that have a scroll bar, the bar was right at the edge of the panel. With this change the scroll bar is inset. Was that intended?

I found another regression in 1.3.1 that is still present with these changes. This one is slightly hard to reproduce. Follow these instructions:
- Open `Preferences…`
- Click on `Network`
- Reduce the width of the panel as much as IINA allows you to
- Click on `Utilities`
- Using two fingers on the trackpad you can scroll horizontally

I am able to scroll horizontally under both macOS 10.15.7 and 13.0.1. Under Ventura it is more "springy" and likes to slide back into place. The following screenshot was taken under 10.15.7:
issue-4100

I was not able to scroll horizontally under IINA 1.2.0.

svobs:
> In 1.2.0 for preference tabs that have a scroll bar, the bar was right at the edge of the panel. With this change the scroll bar is inset. Was that intended?

Thank you for pointing that out - I hadn't noticed it. Should be a trivial fix though.

Part of the improvement here was moving the offsets for the outermost margins into one of the outer elements, rather than requiring all the contained elements to set it individually. But looks I should have added them to the `NSClipView` instead of the `NSScrollView`. This wasn't the hard part, so I had set these set to 0 until almost the end to simplify the remaining problems, so I guess it didn't pop out at me...

> I am able to scroll [`Utilities`] horizontally under both macOS 10.15.7 and 13.0.1. Under Ventura it is more "springy" and likes to slide back into place. 

This is interesting, especially since I can't reproduce it, but I do get a different problem. I can't seem to produce horizontal scrolling for any of the Preference tabs: they just get squeezed (which I think is the desired behavior - more on this in a moment). But while all the other tabs seem to have the same minimum width, `Utilities` lets me go smaller. Then when I switch to another tab, the window's width gets expanded back to the standard minimum...which is strange.
Screenshot 2022-12-01 at 17 57 49
Screenshot 2022-12-01 at 17 57 59

I'll try to track this down. Hopefully we are experiencing different manifestations of the same problem.

But that also leads me to examine the minimum width of the window, which is set at 480px. The navigation(?) panel on the left has a fixed width of 220px, leaving a min 260px for the detail(?) panel on the right. Can you think of a reason not to set the window width to at least 620px? I'm not sure it's even possible to go lower than 640px on MacOS nowadays.

low-batt:
I see the same thing with the `Utilities`  tab allowing you to compress the panel width smaller than when on other tabs. Seems a bit odd. I also see the same behavior of the panel expanding the width if you switch to another tab. I would think it would be better to enforce a consistent minimum width.

On this question:
> But that also leads me to examine the minimum width of the window, which is set at 480px. The navigation(?) panel on the left has a fixed width of 220px, leaving a min 260px for the detail(?) panel on the right. Can you think of a reason not to set the window width to at least 620px? I'm not sure it's even possible to go lower than 640px on MacOS nowadays.

I'm not a GUI developer so I can't comment on what should be used for the minimum size. Tried to find some guidance on this, but struck out.

The issue with being able to scroll horizontally in the `Utilities` tab is slightly hard to reproduce. I have to first go to the `Network` tab and compress the width to the minimum that tab allows and then click on the `Utilities` tab. Then I can move the contents around using two fingers on the touchpad. So it definitely has something to do with the handling of the width of the panel.

low-batt:
I did some more experiments with the horizontal scrolling issue. Try this:
- Open `System Settings…`
- Click on `Appearance`
- Change `Show scroll bars` to `Always`
- Open `IINA`
- Open `Preferences…`
- Reduce the height of the panel so that the scroll bar is present under the `Utilities` tab, but not shown in the `Network` tab
- Click on `Network`
- Reduce the width of the panel as much as IINA allows you to
- Click on `Utilities`
- Using two fingers on the trackpad you can scroll horizontally

See if it reproduces for you with these instructions.

svobs:
The last couple of days have been very productive. I pushed an update which should address all of these concerns.

> The issue with being able to scroll horizontally in the Utilities tab is slightly hard to reproduce. I have to first go to the Network tab and compress the width to the minimum that tab allows and then click on the Utilities tab. Then I can move the contents around using two fingers on the touchpad. So it definitely has something to do with the handling of the width of the panel.

I was able to reproduce it (thanks), and i think I found the cause.

For the window to work as we intend:

1. Each row of controls should be constrained so that its left ("leading") edge is aligned to the leading edge of the view which contains it (`MyComponent.leading = Superview.leading`)
2. In general, all the components in the row should be aligned with each other using `Y.leading = X.trailing + offset`.
3. If we don't have any controls or views in the row which we want to change their width with the window width, then its trailing (right) edge should be aligned to the trailing edge of the view which contains it using Equal or Less Than, so that when it grows, the extra space goes to the window  (`MyComponent.trailing <= Superview.trailing`)
4. If we have views which do change width with the window, they should attach an `=` constraint to the window as in (2), but you should try to have only one constraint per row which uses a `<=` (or `>=`, and the rest use `=`.  (If you want more than one component space to grow with the window, it gets a lot more complicated, because you need to deal with priorities or add stacks or custom views)).

Anyway, the `Utility` tab had a typo where one of the rows ended with `X.trailing >= Superview.trailing` when it should have been `<=`. And when the rest are correctly using `>=`, I suppose the result would be undefined, which explains that behavior.

I also thought a lot more about the min dimensions issue, and realized that it didn't make sense to fret about whether a min window width of 640 was too big, because it's already has a de facto min width a lot bigger than that. Each tab's min width was different, but between ~700 and ~1200 depending on its contents, and because they were all different, that's why changing tabs would resize the window, which isn't desirable.

So I put a lot of work into making each component compress as much as reasonably possible, reduced min widths, etc. Then I set the window's min size to slightly larger than the larger one, so that they are all uniform. The min size measures 932 x 498 on my Mac which I think is reasonable:
Screenshot - latest prefs win

This required me to make a ton of changes though, as every row needed changing. There were a ton of components set to `=` the superview width when they should have been `<=`.

Along the way I fixed a bunch more constraint ambiguities as I found them, and fixed various minor spacing inconsistencies. For example the Search field in the Key Bindings page should align to the right side of the table, instead of having a small gap. And the vertical spacing for checkboxes in `General` was off by a few pixels each in places.

Also deleted these things, which were not visible and not referenced anywhere:
Screenshot 2022-11-30 Phantoms 2Screenshot 2022-11-30 Phantoms 1

So basically every prefs-related XIB file was affected. I went over it and think it looks good, but it probably could use another set of eyes.

While I was at it, I also set internal XCode labels for some of the components to help for future development with these files, so they could be identified and weren't all labelled "Custom View". Working with constraints is hard enough without having unique names for the things you're referring to!

svobs:
Also, maybe this is out of scope for this issue, but was wondering....
Screenshot 2022-12-03 at 00 45 03

1. Is there a reason for the rather large gap between the Search field and the table below it? I have been wondering about this for some time now. To me it seems like when `General` is highlighted, the margin above its highlight and below Search should roughly match the margin to the left of the highlight. Or even match the gap above the Search field. Overall the placement of the Search field looks off to me.
2. Do you know if it's intentional that the text boxes in the Prefs are rectangular and not rounded like in the rest of IINA? It's trivial to change this and I think it would look a lot better. (Even the Search box on the left is rounded...)

low-batt:
This new build warning needs to be addressed:
```text
/Users/low-batt/Documents/builds/pr-tests/iina/iina/Base.lproj/PreferenceWindowController.xib:F0z-JX-Cv5: warning: Minimum size {820, 320} is larger than content size {800, 480} [9]
```

I'm also still seeing the issue with being able to scroll horizontally in the `Utilities` tab. The behavior changed in that I no longer need to shrink the size of the panel while in the `Network` tab. In fact the panel no longer allows me to do that. What I now see is that if I bring up the preferences panel and go directly to the `Utilities` tab I can not scroll horizontally. But if I first click on the `Network` tab and then the `Utilities` tab the contents becomes horizontally scrollable. This is with the macOS system preferences set to always display scroll bars. The content in the `Network` tab fits, so no scroll bar is displayed. The `Utilities` tab content is longer, so a vertical scroll bar is shown.

This testing was all under macOS 13.0.1.

On the questions, remember I am a junior newbie and also not a GUI programmer. So I have no clue as to why certain things are the way they are in the UI. We can definitely say that some of this represents problems that have not been addressed due to lack of developer resources. Here is my personal opinion on the questions:

1. Is there a reason for the rather large gap between the Search field and the table below it?

    - I don't know. The gap seems very excessive to me. I see no reason for that.
 
2. Do you know if it's intentional that the text boxes in the Prefs are rectangular and not rounded like in the rest of IINA?

   - I don't know. I looked over the UI and it seems inconsistent to me as well.

svobs:
Well, I went ahead and changed the text fields to rounded corners, and restyled the Search box to be bigger, and aligned with the table below it, so it somewhat resembles the Search field in the System Settings window. See how you like it. Let's see if anyone disagrees with it.

I fixed the above-mentioned build warning, but I honestly can't figure out how to fix the extra space being created when Scrolling="Always". I see it happening sometimes for `Subtitle` and `Video/Audio` as well, when navigating from  `Network`. I tried several strategies, but most ended up with worse results than before. The way that part of the page is being assembled, from individual sections via a mix code & XIB, and auto-resizing masks & constraints, makes it really hard. I just don't have a clue how to fix it. And I'm not getting paid enough to try rebuilding it from scratch.

I did manage to get rid of the springiness for the horizontal scroll, so hopefully it's a little more bearable at least.

uiryuu:
> Also deleted these things, which were not visible and not referenced anywhere:

@svobs These are used for searching. For example, if you type and search "trackpad", the trackpad box will be highlighted. These invisible labels are used as anchors to locate the boxes when searching. @lhc70000 can explain more about this.

uiryuu:
For your questions @svobs, I'm going to add my 2 cents here:

> Is there a reason for the rather large gap between the Search field and the table below it? I have been wondering about this for some time now. To me it seems like when General is highlighted, the margin above its highlight and below Search should roughly match the margin to the left of the highlight. Or even match the gap above the Search field. Overall the placement of the Search field looks off to me.

I think we can definitely adjust the gap size. I also feel that the search field is a little bit too narrow in terms of its height. See the screenshot below:

image

I can see the similarity of our search field of that of Safari's, but I think our scenario is more like App Store. I would like to make it wider and higher like the search field in App Store.

Another thing is that, there is no space on top of the blinking cursor, but there's 1 px (or 2?) under it. If we still want to maintain the current design, at least we should remove the 1 px under the cursor.

App Store search field:
image
image


Safari search field:
image



> Do you know if it's intentional that the text boxes in the Prefs are rectangular and not rounded like in the rest of IINA? It's trivial to change this and I think it would look a lot better. (Even the Search box on the left is rounded...)

For this one, I'm not sure. I tried to make them rounded and see the result, but it seems off. Then I checked other settings in the OS, they are all rectangular:

image
image

I'd like to keep them rectangular.

svobs:
@uiryuu here's what I came up with for the Search field. It increases the text field from "Regular" to "Large", which I think fits in better with the large-sized table rows below it.
Screenshot 2022-12-06 at 16 52 23


Here's a comparison of the current Codec view:
Screenshot 2022-12-06 at 15 44 49
With the latest commit in this PR, including the rounded text fields:
Screenshot 2022-12-06 at 16 53 29

> I'd like to keep them rectangular.

Sure. Personally I think the rounded version looks better, but I don't feel strongly about it. I'll change them back.

> @svobs These are used for searching. For example, if you type and search "trackpad", the trackpad box will be highlighted. These invisible labels are used as anchors to locate the boxes when searching. @lhc70000 can explain more about this.

Hmm...this seems like a less-than-ideal solution. It would be better to have a list of items in code. (And the Key Bindings tab is supposed to show up when the user searches for "settings"?) But I don't want to dig into this right now - this PR already has had massive scope creep. I will add them back in.

With regard to 1-pixel differences in Search field height - I don't consider myself a front-end developer, and don't want to stress out over things like this. I'd be happy to hand it over for additional minor tweaks 😉

uiryuu:
> Hmm...this seems like a less-than-ideal solution.

Yes, it is. I've discussed it with @lhc70000 before. If you have better implementations, we are happy to change it.

svobs:
OK, I made the necessary updates, so unless I missed anything it should be good to go

uiryuu:
@svobs I pulled and noticed you removed bold font for the titles in the table? Is there any reason for that?

I also noticed some clipping in the key binding page:
image
previously:
image

In Advanced, the position of the question mark button is wrong:
image
previously:
image

svobs:
@uiryuu Didn't notice the clipping - thanks. Looks like it's a lot more obvious in Light Mode. I found and fixed another instance of it.

Also saw there were a few headings in the Subtitles tab which were not capitalized consistently - added capitals to those.

Fixed the alignment of the question mark button too.

> @svobs I pulled and noticed you removed bold font for the titles in the table? Is there any reason for that?

Did I? Which table? I can't find what you're referring to.

uiryuu:
I meant this one:
image

> Also saw there were a few headings in the Subtitles tab which were not capitalized consistently - added capitals to those.

Please don't forget to update the corresponding strings in en.lproj.

svobs:
Didn't see that 😵‍💫 I guess it must have gotten reset when I was wrestling with the Search button alignment.

OK - made another push. Hopefully everything looks good now.

uiryuu (approved):
LGTM

lhc70000:
Tested on 10.13.6. The top margin problem still exists.

@svobs If we can't find a universal solution, I can install another copy of macOS 10.14 to see where's the boundary. Then we can use the code in 1.3.0 for these old systems.

lhc70000:
For `PrefSubViewController.strings`, you still need to change the words in the base xib. The `en.lproj` was created solely for Crowdin translation and won't be used at runtime. (@uiryuu can you confirm this?)

svobs:
> Tested on 10.13.6. The top margin problem still exists.

Ack 😖 I wish there was a way for me to test this myself. @low-batt does it still show up for you?

I forgot to mention that when testing the updated version, I found it was necessary to make sure no other instances of IINA were running, or it may still show the layout from that instance. I'm not sure whether parts of the layout are getting improperly cached, or some other quirk.

If it's still a problem, I guess my last attempt would be to add an empty `NSView` with 28px height as a spacer, instead of relying on offset constraints around the `NSStackView`. I mean, when 10.13.6 was released, stack views had existed for less than 4 years, so maybe Apple hadn't gotten around to fixing its bugs yet...

svobs:
@uiryuu Sorry, don't mean to fall into an infinite loop here! - but I was working with `NSTokenField` for #4074, and was wondering - would you be opposed to using the rounded style for the two `Preferred language` fields (Sub & Codec tabs)? Look at how the tokens just seem to pop into place.

Before:
Screenshot 2022-12-08 at 14 57 51

After:
Screenshot 2022-12-08 at 14 57 13

I wanted to update the XIB for that issue already, to include a trailing "..." for when text overflows the field, and it would be a lot easier to tack on any XIB changes to the edits here.

uiryuu:
@svobs For the sake of consistency, please change them back to rectangular.

low-batt:
I tested the latest commit on macOS 13.0.1 and 10.15.7. I found one minor issue on 10.15.7. The spacing changes under Catalina such that the `General` tab is too close to the search field resulting in the top of the tab being truncated. This is especially evident when the tab is selected as the blue box that normally surrounds it is clearly missing the top of the box:
issue-4100-general

svobs:
@low-batt OK I think I see what happened. That table is using the "Source List" style, which was only available starting with MacOS 11.0. But for older versions it looks like it just falls back to the default "full" style. The "Source List" style includes a 10-point vertical gap above the highlight, whereas the other style did not.

I came up with a solution which adds some glue code, which just checks the MacOS version and adds a 10-point gap for prior to 11.0. See if that is satisfactory. The only minor imperfection is that the scroller now goes up a little higher for 11.0+. But that's only when it's visible, which should be rarely. And in fact, it seems feasible to set the window's minimum height to a large enough value to guarantee that the list will never need a scroller.

low-batt:
On this:
> The only minor imperfection is that the scroller now goes up a little higher for 11.0+. But that's only when it's visible, which should be rarely.

Remember that is not true for users that in `System Settings` under `Appearance` set `Show scroll bars` to `Always`. The following screenshots show the difference.

With the changes in the PR:
issue-4100-new-scroll

With IINA 1.3.1:
issue-4100-oild-scrool

I also see this under 10.15.7. I don't see this as a problem. See what the other reviewers think.

iina/PreferenceWindowController.swift
```diff
@@ -160,7 +164,13 @@ class PreferenceWindowController: NSWindowController {
...
     completionTableView.dataSource = self
 
-    contentViewBottomConstraint = contentView.bottomAnchor.constraint(equalTo: contentView.superview!.bottomAnchor, constant: -28)
+    detailViewBottomConstraint = prefDetailContentView.bottomAnchor.constraint(equalTo: prefDetailContentView.superview!.bottomAnchor)
+
+    // NSTableView's "Source List" style is only available with MacOS 11.0+ and includes a built-in 10pt offset for its highlights.
+    // But for older MacOS versions, the style will default to "full width" with no highlight offset, which will touch the Search field.
+    if #unavailable(macOS 11.0) {
+      navTableSearchFieldSpacingConstraint.constant = 10.0
+    }
```
low-batt, line 173: I confirmed this is working under 10.15.7. Also looks good under 13.0.1.

low-batt (approved):
Tested under 13.0.1 and 10.15.7. Changes look good to me.

uiryuu:
Please check the constrains of these two checkboxes:
image
image

The behavior of the page of Utilities is a little bit odd: when you are on Utilities page, the width cannot be adjusted. And if you change to other pages, the width remains narrow.

svobs:
> Please check the constrains of these two checkboxes

@uiryuu You mean the clipping on the right? I don't think it is possible to make checkbox text wrap around. I went ahead and increased the min window width a bit more... now 932 points (although Interface Builder seems to think it's 820. Seems like it has been a little weird since the last update).

> The behavior of the page of Utilities is a little bit odd

I don't know how that slipped by. I thought I specifically tested for that... OK, should be fixed now. Please verify.

svobs:
Housekeeping: rebased and squashed the changes. (Still haven't figured out a smoother way to replace the previous squash than by using `push --force`)

PR #4118, opened 2022-12-05
Fix incorrect playlist item moved to trash (issue #4061)

### LINKED ISSUE #4061: File matching in playlist is wrong when video/audio mixed.
**System and IINA version:**

- macOS 13.0.1 (x64)
- IINA 1.3.1 Build 133

**Expected behavior:**
When right click the item in playlist, select "Move to Trash", the file should be removed.

**Actual behavior:**
The selected menu item and another file is removed.

**Steps to reproduce:**
1 Place one video and some audio file in the same folder. Make sure the audio files are sorted before the video -- names are "smaller" than the video file.
2 Double click the video file to play.
3 Wait for the video playing to finished.
4 The audio file is started to play automatically. Note: The UI is changed, playlist is not visible like the previous version.
5 Change UI back to full mode. Open the play list.
6 Right click the video item (the 1st one), select "Move to Trash".
7 The selected menu item and the video file specified should be removed. BUT the 1st audio file is removed instead of the video file.

Always.

svobs:
- [x] This implements/fixes issue #4061.

---

**Description:**
In Playlist context menu, `Move to Trash` was removing the item from the playlist before getting the filename from it, which resulted in either the incorrect file being sent to trash, or IINA crashing. Changed so that it first attempts to move each file to the trash, and only removes from the playlist at the end, and only removes the ones were successfully trashed.

### DIFF
--- a/iina/PlaylistViewController.swift
+++ b/iina/PlaylistViewController.swift
@@ -637,20 +637,23 @@ class PlaylistViewController: NSViewController, NSTableViewDataSource, NSTableVi
 
   @IBAction func contextMenuDeleteFile(_ sender: NSMenuItem) {
     guard let selectedRows = selectedRows else { return }
-    var count = 0
+    Logger.log("User chose to delete files from playlist at indexes: \(selectedRows.map{$0})")
+
+    var successes = IndexSet()
     for index in selectedRows {
-      player.playlistRemove(index)
       guard !player.info.playlist[index].isNetworkResource else { continue }
       let url = URL(fileURLWithPath: player.info.playlist[index].filename)
       do {
+        Logger.log("Trashing row \(index): \(url.standardizedFileURL)")
         try FileManager.default.trashItem(at: url, resultingItemURL: nil)
-        count += 1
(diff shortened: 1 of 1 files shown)


uiryuu:
Please split your commits into multiple PRs next time, thanks!

svobs:
Updated the code to remove everything except the bug fix

svobs:
Hopefully this can make it into the next release? This is a serious bug because it can result in unexpected loss of user data, and this fix is not too complicated. Attn @uiryuu

uiryuu:
Thanks for reminding. Rest assured, this will definitely go into the next release.

uiryuu:
Thanks!

jiripospisil:
> Thanks for reminding. Rest assured, this will definitely go into the next release.

@uiryuu Is there an ETA for the new release? This bug is bad enough to warrant a quick patch release on its own. Thanks.

uiryuu:
@jiripospisil We are building the dependencies for the new release. Should be soon, but no eta still.

PR #4068, opened 2022-11-11
Fix  crash during quit in windowDidExitFullScreen, #4020

low-batt:
This commit will:
- Add a new isClosing property to MainWindowController
- Change windowWillClose to set isClosing to true
- Change windowWillExitFullScreen and windowDidExitFullScreen to not access mpv if window is closing
- Change fsState didSet to not access mpv if window is closing
- Change windowWillOpen to set isClosing to false

These changes prevent IINA from calling mpv while mpv is asynchronously unloading the file or shutting down.

- [x] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)
- [x] This implements/fixes issue #4020.

---

**Description:**

PR #4162, opened 2023-01-05
Move guard before switch statement

danshao:
- [x] This implements/fixes issue #4068 

---

**Description:**

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -241,12 +241,11 @@ class MainWindowController: PlayerWindowController {
 
   var fsState: FullScreenState = .windowed {
     didSet {
+      guard !isClosing else { return }
       switch fsState {
       case .fullscreen: player.mpv.setFlag(MPVOption.Window.fullscreen, true)
       case .animating:  break
-      case .windowed:
-        guard !isClosing else { return }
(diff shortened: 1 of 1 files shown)


danshao (commented):
Moving guard statement to before switch as requested.

low-batt:
I've been away from IINA due to the holidays. Trying to catch up now. Thanks so much for coding this up!

Back when I coded the original PR I had trouble testing the fix as I could not reliably reproduce the crash. I could start and quit IINA over and over again without any failure. Now that I have more time to investigate I found that was because I had the `Use legacy full screen` setting enabled. Once I disabled that I was able to reliably crash IINA when quitting while in full screen. I'm now using a bash script to repeat the start - quit sequence over and over again to test the fix.

That testing showed this fix is not good enough. An additional check is needed.

I will be updating the original PR with an additional fix as well as the change requested in the review. I will probably get that pushed tonight or tomorrow. Need to do more testing. Once I get that pushed I'll try and get people to finish the review and merge.

Thanks again for preparing this PR!

Coeur:
Thank you @danshao, but at this point, the original PR #4068 is now better than yours since it addresses `windowDidEndLiveResize`.
So best to close #4162 for giving exclusive visibility to #4068.

PR #4165, opened 2023-01-08
Fixes & improvements to Settings > UI > OSC toolbar preview & drag/drop

svobs:

---

Before:

https://user-images.githubusercontent.com/2213815/211179483-27ad7c9b-2542-4ed0-bee0-1776deb0b3c0.mp4

---

After:

https://user-images.githubusercontent.com/2213815/211179426-28d75d9d-27d0-4feb-b163-b496ecfc6062.mp4

---

**Description:**

_Note: requires #4107 to be merged first._

The `settings` toolbar icon (looks more like a chat bubble to me) comes from the MacOS system icons, whereas all the other OSC toolbar icons are built into IINA. It looks like probably when Apple changed its icon, its displayed size got messed up in IINA. This fix started from an attempt to fix that, which was surprisingly challenging because I wanted to avoid hard-coding an icon size. Along the way I added some more improvements, detailed below.

The `settings` toolbar icon is drawn in 6 different places:
1. The Player window
2. Settings > UI > On Screen Controller > Toolbar preview
3. Settings > UI > On Screen Controller > Toolbar > Customize window > Current Items view
4. Settings > UI > On Screen Controller > Toolbar > Customize window > Available Items view
5. Settings > UI > On Screen Controller > Toolbar > Customize window > Dragging from Current Items view
6. Settings > UI > On Screen Controller > Toolbar > Customize window > Dragging from Available Items view

Only the 1st location was the correct size; the others were too big. It turns out that the Player window was drawing it inside of a button, which fixed its size; the others were just using naked `NSImage`s.

My fix was to replace the naked images with "dead" buttons (disabled but not grayed out visually), which guarantees that the drawing will be the same.

1. Fixed oversized `settings` toolbar icon so it matches the other icons.
2. Color of dragged icon (items 5 and 6 above) was always black, which made it hard to see in Dark Mode. Changed it so that the color is light in Dark mode (& matches its non-dragged color).
3. Changed drag start coordinates so that the icon does a move animation from its visual location to under the user's cursor.
4. Added standard MacOS "poof" animation when dragging out of the toolbar.
5. Generally improved the layout strategy of the toolbar icons so that they will be super robust to changes in icon size (and should make it much easier to swap the icons for different sizes in the future, if the desire ever arises.)

### DIFF
--- a/iina/OSCToolbarButton.swift
+++ b/iina/OSCToolbarButton.swift
@@ -1,0 +1,48 @@
+//
+//  OSCToolbarButton.swift
+//  iina
+//
+//  Created by Matt Svoboda on 11/6/22.
+//  Copyright © 2022 lhc. All rights reserved.
+//
+
+import Foundation
+
+// Not elegant. Just a place to stick common code so that it won't be duplicated
+class OSCToolbarButton {
+  static func setStyle(of toolbarButton: NSButton, buttonType: Preference.ToolBarButton) {
+    toolbarButton.translatesAutoresizingMaskIntoConstraints = false
+    toolbarButton.bezelStyle = .regularSquare
+    toolbarButton.image = buttonType.image()
+    toolbarButton.isBordered = false
+    toolbarButton.tag = buttonType.rawValue
+    toolbarButton.refusesFirstResponder = true
+    toolbarButton.toolTip = buttonType.description()
+    let sideSize = Preference.ToolBarButton.frameHeight
+    Utility.quickConstraints(["H:[btn(\(sideSize))]", "V:[btn(\(sideSize))]"], ["btn": toolbarButton])
+  }
+
+  static func buildDragItem(from toolbarButton: NSButton, pasteboardWriter: NSPasteboardWriting,
+                            buttonType: Preference.ToolBarButton) -> NSDraggingItem? {
+    // seems to be the only reliable way to get image size
+    guard let imgReps = toolbarButton.image?.representations else { return nil }
+    guard !imgReps.isEmpty else { return nil }
+    let imageSize = imgReps[0].size
+
+    let dragItem = NSDraggingItem(pasteboardWriter: pasteboardWriter)
+    let iconSize = Preference.ToolBarButton.frameHeight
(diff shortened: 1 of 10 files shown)


CobraFighter (approved):

svobs:
Updated to resolve conflicts

svobs:
Rebased and fixed conflicts, improved layout further, improved commit message.

iina/OSCToolbarButton.swift
```diff
@@ -0,0 +1,48 @@
...
+    let sideSize = Preference.ToolBarButton.frameHeight
+    Utility.quickConstraints(["H:[btn(\(sideSize))]", "V:[btn(\(sideSize))]"], ["btn": toolbarButton])
+  }
+
+  static func buildDragItem(from toolbarButton: NSButton, pasteboardWriter: NSPasteboardWriting,
+                            buttonType: Preference.ToolBarButton) -> NSDraggingItem? {
+    // seems to be the only reliable way to get image size
+    guard let imgReps = toolbarButton.image?.representations else { return nil }
+    guard !imgReps.isEmpty else { return nil }
+    let imageSize = imgReps[0].size
```
lhc70000: With our extension we can write
```swift
guard let imageSize = toolbarButton.image?.representations[at: 0]?.size else { return nil }
```

lhc70000 (commented):
Looks good to me, only have one small suggestion

iina/OSCToolbarButton.swift
svobs: Nice. Added a commit with the change.

uiryuu:
@svobs This PR causes a bug where the volume control and the toolbar buttons are not shown in the floating OSC.

image

It seems like a constraint issue, but I failed to find out the root cause.

svobs:
Will investigate tonight

lhc70000:
I've already found the reason, can fix it if you don't mind.

svobs:
Don't mind at all! Please go ahead 👍

low-batt:

PR #4174, opened 2023-01-11
Use float OpenGL buffer and extended linear colorspaces for HDR playback

### LINKED ISSUE #3772: Luminance is incorrect for HDR videos taken on iPhone
**System and IINA version:**

- macOS 12?
- IINA 1.3.0

**Expected behavior:**
When playing HDR video taken on iPhone in IINA on a MacBook 16 2021 (M1 Max) the picture matches what is seen with Quicktime.

**Actual behavior:**
The luminance is inappropriate, resulting in a catastrophic loss of details.

Quicktime:
Alt text

IINA:
Alt text

**Steps to reproduce:**
Take a video with an iPhone 13 Pro and play the video in IINA.

The issue is more noticeable in indoor clips and/or at higher brightness levels.

This issue was reported by @jesec in discussions about PR #3701, the commit that added HDR playback support in 1.3.0. This was also reported in the IINA [Telegram Group](https://t.me/IINAUsers).

Everytime.

ashen0xa:
- [x] This implements/fixes issue #3772.

---

**Description:**
[This WWDC 2021 presentation](https://developer.apple.com/videos/play/wwdc2021/10161/) suggested to use floating point OpenGL buffers (see 20:56) and extended linear colorspaces (see 18:08) for EDR content. 

One may also note that [original mpv mac output code](https://github.com/mpv-player/mpv/blob/25b66256d7ff48254b2055a066e29f260414112f/video/out/mac/gl_layer.swift#L45) also use floating point OpenGL buffers for 10-bit playback, which format are most HDR videos in. Further, the official Apple Documentation notes that `itur_2020_PQ` colorspaces are [deprecated](https://developer.apple.com/documentation/coregraphics/cgcolorspace/3524230-itur_2020_pq). 

It turns out that these are all we need to match `QuickTime.app`'s HDR output.

Device: Macbook Pro 14" with M1 pro
Test clip:  [download](https://user-images.githubusercontent.com/30361859/211847796-c2ece693-78a8-49b4-a57e-fff9f40202d2.mp4)

- Left: IINA.app current `develop` branch
- Middle: this PR
- Right: QuickTime.app
IMG_0377

### DIFF
--- a/iina/InspectorWindowController.swift
+++ b/iina/InspectorWindowController.swift
@@ -202,8 +202,17 @@ class InspectorWindowController: NSWindowController, NSTableViewDelegate, NSTabl
       self.setLabelColor(self.vprimariesField, by: sigPeak > 0)
 
       if PlayerCore.lastActive.mainWindow.loaded && controller.fileLoaded {
-        let colorspace = PlayerCore.lastActive.mainWindow.videoView.videoLayer.colorspace?.name;
-        self.vcolorspaceField.stringValue = colorspace == nil ? "Unspecified (SDR)" : String(colorspace!) + " (HDR)"
+        if let colorspace = PlayerCore.lastActive.mainWindow.videoView.videoLayer.colorspace, #available(macOS 10.15, *) {
+          var isHdr: Bool
+          if #available(macOS 11.0, *) {
+            isHdr = CGColorSpaceUsesExtendedRange(colorspace) || CGColorSpaceUsesITUR_2100TF(colorspace)
+          } else {
+            isHdr = colorspace.isHDR()
+          }
+          self.vcolorspaceField.stringValue = "\(colorspace.name!) (\(isHdr ? "H" : "S")DR)"
+        } else {
+          self.vcolorspaceField.stringValue = "Unspecified (SDR)"
+        }
       } else {
         self.vcolorspaceField.stringValue = "N/A"
       }
--- a/iina/InspectorWindowController.swift
+++ b/iina/InspectorWindowController.swift
@@ -202,7 +202,7 @@ class InspectorWindowController: NSWindowController, NSTableViewDelegate, NSTabl
       self.setLabelColor(self.vprimariesField, by: sigPeak > 0)
 
       if PlayerCore.lastActive.mainWindow.loaded && controller.fileLoaded {
-        if let colorspace = PlayerCore.lastActive.mainWindow.videoView.videoLayer.colorspace, #available(macOS 10.15, *) {
+        if #available(macOS 10.15, *), let colorspace = PlayerCore.lastActive.mainWindow.videoView.videoLayer.colorspace {
           var isHdr: Bool
           if #available(macOS 11.0, *) {
             isHdr = CGColorSpaceUsesExtendedRange(colorspace) || CGColorSpaceUsesITUR_2100TF(colorspace)
--- a/iina/VideoView.swift
+++ b/iina/VideoView.swift
@@ -41,6 +41,8 @@ class VideoView: NSView {
 
   lazy var hdrSubsystem = Logger.Subsystem(rawValue: "hdr")
 
+  static let deviceRGBColorspace = CGColorSpaceCreateDeviceRGB()
+
   // MARK: - Attributes
 
   override var mouseDownCanMoveWindow: Bool {
@@ -58,6 +60,7 @@ class VideoView: NSView {
 
     // set up layer
     layer = videoLayer
+    videoLayer.colorspace = VideoView.deviceRGBColorspace
     videoLayer.contentsScale = NSScreen.main!.backingScaleFactor
     wantsLayer = true
 
@@ -283,10 +286,10 @@ class VideoView: NSView {
       }
     }
 
-    if videoLayer.colorspace != nil {
-      Logger.log("Nilling out colorspace", subsystem: hdrSubsystem)
-      videoLayer.colorspace = nil;
+    if videoLayer.colorspace != VideoView.deviceRGBColorspace {
+      Logger.log("Returning to deviceRGB color space", subsystem: hdrSubsystem)
       videoLayer.wantsExtendedDynamicRangeContent = false
+      videoLayer.colorspace = VideoView.deviceRGBColorspace
       player.mpv.setString(MPVOption.GPURendererOptions.targetTrc, "auto")
       player.mpv.setString(MPVOption.GPURendererOptions.targetPrim, "auto")
     }
@@ -319,43 +322,26 @@ extension VideoView {
       Logger.log("HDR primaries and gamma not available", level: .debug, subsystem: hdrSubsystem);
       return false;
     }
+  
+    let peak = mpv.getDouble(MPVProperty.videoParamsSigPeak)
+    Logger.log("HDR gamma=\(gamma), primaries=\(primaries), sig_peak=\(peak)", level: .debug, subsystem: hdrSubsystem)
 
     var name: CFString? = nil;
     switch primaries {
     case "display-p3":
-      switch gamma {
-      case "pq":
-        if #available(macOS 10.15.4, *) {
-          name = CGColorSpace.displayP3_PQ
-        } else {
-          name = CGColorSpace.displayP3_PQ_EOTF
-        }
-      case "hlg":
-        name = CGColorSpace.displayP3_HLG
-      default:
-        name = CGColorSpace.displayP3
+      if #available(macOS 10.15.4, *) {
+        name = CGColorSpace.displayP3_PQ
+      } else {
+        name = CGColorSpace.displayP3_PQ_EOTF
       }
 
     case "bt.2020":
-      switch gamma {
-      case "pq":
-        if #available(macOS 11.0, *) {
-          name = CGColorSpace.itur_2100_PQ
-        } else if #available(macOS 10.15.4, *) {
-          name = CGColorSpace.itur_2020_PQ
-        } else {
-          name = CGColorSpace.itur_2020_PQ_EOTF
-        }
-      case "hlg":
-        if #available(macOS 11.0, *) {
-          name = CGColorSpace.itur_2100_HLG
-        } else if #available(macOS 10.15.6, *) {
-          name = CGColorSpace.itur_2020_HLG
-        } else {
-          fallthrough
-        }
-      default:
-        name = CGColorSpace.itur_2020
+      if #available(macOS 11.0, *) {
+        name = CGColorSpace.itur_2100_PQ
+      } else if #available(macOS 10.15.4, *) {
+        name = CGColorSpace.itur_2020_PQ
+      } else {
+        name = CGColorSpace.itur_2020_PQ_EOTF
       }
 
     case "bt.709":
@@ -383,7 +369,7 @@ extension VideoView {
     videoLayer.wantsExtendedDynamicRangeContent = true
     videoLayer.colorspace = CGColorSpace(name: name!)
     mpv.setString(MPVOption.GPURendererOptions.iccProfile, "")
-    mpv.setString(MPVOption.GPURendererOptions.targetTrc, gamma)
+    mpv.setString(MPVOption.GPURendererOptions.targetTrc, "pq")
     mpv.setString(MPVOption.GPURendererOptions.targetPrim, primaries)
     return true;
   }
--- a/iina/ViewLayer.swift
+++ b/iina/ViewLayer.swift
@@ -47,6 +47,8 @@ class ViewLayer: CAOpenGLLayer {
     var attributeList: [CGLPixelFormatAttribute] = [
       kCGLPFADoubleBuffer,
       kCGLPFAAllowOfflineRenderers,
+      kCGLPFAColorFloat,
+      kCGLPFAColorSize, CGLPixelFormatAttribute(64),
       kCGLPFAOpenGLProfile, CGLPixelFormatAttribute(kCGLOGLPVersion_3_2_Core.rawValue),
       kCGLPFAAccelerated,
     ]


ashen0xa:
Added a nasty hack to workaround a weird behavior about set `colorspace` to `nil`. 

Basically once `colorspace` was set to extended linear, setting it back to nil would cause falling back to `kCGColorSpaceExtendedLinearSRGB` which is definitely not what we want. Without manually setting it to a SDR colorspace the color would be completely broken.
Is it a bug with macOS?

svobs:
+1! 

I pulled this PR and tested with some of my own HDR videos and will confirm they look 10x better, especially on my MacBook's built-in display, but also on my HDR display. Hopefully the maintainers can take a look at it soon.

CC @CarterLi

iina/VideoView.swift
```diff
@@ -285,8 +285,12 @@ class VideoView: NSView {
 
     if videoLayer.colorspace != nil {
       Logger.log("Nilling out colorspace", subsystem: hdrSubsystem)
-      videoLayer.colorspace = nil;
       videoLayer.wantsExtendedDynamicRangeContent = false
+      // It seems that once there is a layer with colorspace set to something extended linear,
+      // then setting colorspace to nil will cause falling back to ExtendedLinearsRGB
+      videoLayer.colorspace = CGColorSpaceCreateDeviceRGB()
```
CarterLi: colorspace is nil by default. Disabling HDR mode should revert all changes to the default value as HDR was never enabled. In addition, since we are testing `videoLayer.colorspace != nil` in [Line 286](https://github.com/iina/iina/pull/4174/files#diff-0da7dedcb527d7cb82822500e08f5a55ee5fb2a7fb99b2bf039dbad1b6787546L286), setting `videoLayer.colorspace` to non-nil makes the resetting code be executed every time when setICCProfile being called

iina/VideoView.swift
```diff
@@ -285,8 +285,12 @@ class VideoView: NSView {
...
+      player.mpv.setString(MPVOption.GPURendererOptions.toneMapping, "bt.2390")
```
CarterLi: Again, according to mpv document, [the default value of `--tone-mapping` is `auto`](https://[mpv.io](https://mpv.io/manual/master/#options-tone-mapping)/manual/master/#options-tone-mapping). Unless we set these options to the values you specified by default, enabling and disabling HDR may make SDR videos look different.

iina/VideoView.swift
```diff
@@ -383,7 +360,9 @@ extension VideoView {
     videoLayer.wantsExtendedDynamicRangeContent = true
     videoLayer.colorspace = CGColorSpace(name: name!)
     mpv.setString(MPVOption.GPURendererOptions.iccProfile, "")
-    mpv.setString(MPVOption.GPURendererOptions.targetTrc, gamma)
+    mpv.setString(MPVOption.GPURendererOptions.toneMapping, "linear")
+    mpv.setDouble(MPVOption.GPURendererOptions.toneMappingParam, peak)
```
CarterLi: Are you sure what you do is correct?

CarterLi:
`itur_2020_PQ` was deprecated ( and unsupported by newer macOS ) because it was replaced by `itur_2100_PQ`. `itur_2100_PQ` was not deprecated.

iina/VideoView.swift
ashen0xa: The bundled `libmpv.dylib` version is `0.34.1`, where default value of `--tone-mapping` is `bt.2390`. 

`--tone-mapping=auto` was added in  mpv-player/mpv@a9cb2e2821adb40e7548e8233390e79706104041 and only presented on newer `0.35.0` version.

iina/VideoView.swift
ashen0xa: The problem was, as I mentioned in the commit, once `videoLayer.colorspace` was set to `CGColorspace.extendedLinear*` before, then setting it to `nil` would *not* clear it, instead, it would fallback to `kCGColorspaceExtendedLinearSRGB` and completely throw off SDR color. I'm not sure whether it is the intended behavior.

Setting it to `CGColorspaceCreateDeviceSRGB()` should also turn off color management on the buffer like `nil`. I do agree condition should be updated, though.

Attached is the screenshot trying to turn off HDR without the last commit d9438770d7c625b37bcc7e37bbf3adf5f1ac165b. 
Linear

iina/VideoView.swift
CarterLi: We will eventually upgrade mpv. Does setting the value to an empty string do the trick?

iina/VideoView.swift
ashen0xa: With `extendedLinear` colorspace the output should be linear value with SDR reference white point at `1.0`. 

What these parameters do is:

- `target-trc=linear` makes mpv linearize input signal with SDR reference white at `1.0` ([see video_shaders.c line 860](https://github.com/mpv-player/mpv/blob/v0.34.1/video/out/gpu/video_shaders.c#L860))
- but because mpv believes `linear` should only be SDR range 0~1.0, it will then try to tone map ([see line 872, `dst.sig_peak` would be 1.0](https://github.com/mpv-player/mpv/blob/v0.34.1/video/out/gpu/video_shaders.c#L872))
- `tone-mapping=linear` makes tone mapping a linear scaling, and `tone-mapping-param=peak` makes the value input peak signal (`> 1.0` SDR reference white) scale to itself, thereby preserving SDR reference white at `1.0` ([see line 766](https://github.com/mpv-player/mpv/blob/v0.34.1/video/out/gpu/video_shaders.c#L766))

After checking shader used by mpv to render video in `mpv.log`, it does behave like what I want.

<details><summary>output shader in mpv.log</summary>

```
[  11.683][d][libmpv_render] fragment shader source:
[  11.684][d][libmpv_render] [  1] #version 410
[  11.684][d][libmpv_render] [  2] #define tex1D texture
[  11.684][d][libmpv_render] [  3] #define tex3D texture
[  11.684][d][libmpv_render] [  4] #define LUT_POS(x, lut_size) mix(0.5 / (lut_size), 1.0 - 0.5 / (lut_size), (x))
[  11.684][d][libmpv_render] [  5] out vec4 out_color;
[  11.684][d][libmpv_render] [  6] in vec2 texcoord0;
[  11.684][d][libmpv_render] [  7] in vec2 texcoord1;
[  11.684][d][libmpv_render] [  8] uniform mat3 colormatrix;
[  11.684][d][libmpv_render] [  9] uniform vec3 colormatrix_c;
[  11.684][d][libmpv_render] [ 10] uniform vec3 src_luma;
[  11.684][d][libmpv_render] [ 11] uniform vec3 dst_luma;
[  11.684][d][libmpv_render] [ 12] uniform sampler2DRect texture0;
[  11.684][d][libmpv_render] [ 13] uniform vec2 texture_size0;
[  11.684][d][libmpv_render] [ 14] uniform mat2 texture_rot0;
[  11.684][d][libmpv_render] [ 15] uniform vec2 texture_off0;
[  11.684][d][libmpv_render] [ 16] uniform vec2 pixel_size0;
[  11.684][d][libmpv_render] [ 17] uniform sampler2DRect texture1;
[  11.684][d][libmpv_render] [ 18] uniform vec2 texture_size1;
[  11.684][d][libmpv_render] [ 19] uniform mat2 texture_rot1;
[  11.684][d][libmpv_render] [ 20] uniform vec2 texture_off1;
[  11.684][d][libmpv_render] [ 21] uniform vec2 pixel_size1;
[  11.684][d][libmpv_render] [ 22] void main() {
[  11.684][d][libmpv_render] [ 23] vec4 color = vec4(0.0, 0.0, 0.0, 1.0);
[  11.684][d][libmpv_render] [ 24] color.r = 1.003906 * vec4(texture(texture0, texcoord0)).r;
[  11.684][d][libmpv_render] [ 25] color.gb = 1.003906 * vec4(texture(texture1, texcoord1)).rg;
[  11.684][d][libmpv_render] [ 26] color = color.rgbr;
[  11.684][d][libmpv_render] [ 27] color.rgb = mat3(colormatrix) * color.rgb + colormatrix_c;
[  11.684][d][libmpv_render] [ 28] color.a = 1.0;
[  11.684][d][libmpv_render] [ 29] // color mapping
[  11.684][d][libmpv_render] [ 30] // linearize

// HLG inverse EOTF function
[  11.684][d][libmpv_render] [ 31] color.rgb = clamp(color.rgb, 0.0, 1.0);
[  11.684][d][libmpv_render] [ 32] color.rgb = mix(vec3(4.0) * color.rgb * color.rgb,
[  11.684][d][libmpv_render] [ 33]                 exp((color.rgb - vec3(0.559911)) * vec3(1.0/0.178833)) + vec3(0.284669),
[  11.684][d][libmpv_render] [ 34]                 bvec3(lessThan(vec3(0.5), color.rgb)));
[  11.684][d][libmpv_render] [ 35] color.rgb *= vec3(1.0/3.179550);
[  11.684][d][libmpv_render] [ 36] color.rgb *= vec3(1.0/3.774119);
[  11.684][d][libmpv_render] [ 37] color.rgb *= vec3(3.774119);

// HLG reference OOTF function
[  11.684][d][libmpv_render] [ 38] // apply ootf
[  11.684][d][libmpv_render] [ 39] color.rgb *= vec3(1.000749 * pow(dot(src_luma, color.rgb), 0.200000));

[  11.684][d][libmpv_render] [ 40] // HDR tone mapping
[  11.684][d][libmpv_render] [ 41] int sig_idx = 0;
[  11.684][d][libmpv_render] [ 42] if (color[1] > color[sig_idx]) sig_idx = 1;
[  11.684][d][libmpv_render] [ 43] if (color[2] > color[sig_idx]) sig_idx = 2;
[  11.684][d][libmpv_render] [ 44] float sig_max = color[sig_idx];
[  11.684][d][libmpv_render] [ 45] float sig_peak = 4.926108;
[  11.684][d][libmpv_render] [ 46] float sig_avg = 0.250000;
[  11.684][d][libmpv_render] [ 47] vec3 sig = min(color.rgb, sig_peak);
[  11.684][d][libmpv_render] [ 48] float sig_orig = sig[sig_idx];
[  11.684][d][libmpv_render] [ 49] float slope = min(1.000000, 0.250000 / sig_avg);
[  11.684][d][libmpv_render] [ 50] sig *= slope;
[  11.684][d][libmpv_render] [ 51] sig_peak *= slope;

// tone-mapping=linear and tone-mapping-param=peak
[  11.684][d][libmpv_render] [ 52] sig = min(4.926108 / sig_peak, 1.0) * sig;
[  11.684][d][libmpv_render] [ 53] vec3 sig_lin = color.rgb * (sig[sig_idx] / sig_orig);
[  11.684][d][libmpv_render] [ 54] float coeff = max(sig[sig_idx] - 0.180000, 1e-6) /               max(sig[sig_idx], 1.0);
[  11.684][d][libmpv_render] [ 55] coeff = 0.750000 * pow(coeff, 1.500000);
[  11.684][d][libmpv_render] [ 56] color.rgb = mix(sig_lin, 1.000000 * sig, coeff);
[  11.684][d][libmpv_render] [ 57] color.rgb *= vec3(1.000000);
[  11.684][d][libmpv_render] [ 58] out_color = color;
[  11.684][d][libmpv_render] [ 59] }
```

</details>

iina/VideoView.swift
ashen0xa: Unfortunately it does not work.

```
[  37.562][v][cplayer] Set property: tone-mapping="" -> -4
```

and the option value was not changed

ashen0xa:
> `itur_2020_PQ` was deprecated ( and unsupported by newer macOS ) because it was replaced by `itur_2100_PQ`. `itur_2100_PQ` was not deprecated.

Yes, that's right. So current code *should* work, but it just *doesn't* (so we have #3772 and probably #4075). In that term I agree with @low-batt and believe switch to MoltenVK and Metal would be the long-term solution. But now at least we can try to make current result a little better.

iina/VideoView.swift
CarterLi: It turns out that it is `kCGLPFAColorFloat, kCGLPFAColorSize, CGLPixelFormatAttribute(64),` that makes nil fallback to `kCGColorSpaceExtendedLinearSRGB`. You may remove these code and try again

CarterLi: `static let SRGB = CGColorSpaceCreateDeviceRGB()`

`videoLayer.colorspace = VideoView.SRGB`

```
if videoLayer.colorspace != VideoView.SRGB {
    videoLayer.colorspace = VideoView.SRGB
    ...
```

In addition, `videoLayer.colorspace = VideoView.SRGB` should be set in `init()`

CarterLi:
Tested. While I didn't see much difference for PQ, HLG was improved greatly. However there are several issues should be addressed.

iina/VideoView.swift
ashen0xa: > It turns out that it is `kCGLPFAColorFloat, kCGLPFAColorSize, CGLPixelFormatAttribute(64),` that makes nil fallback to `kCGColorSpaceExtendedLinearSRGB`. You may remove these code and try again

The problem is output of 10-bit color depth videos would be incorrect without them. These ensure pixel format of OpenGL buffer be 16-bit float-point instead of default 8-bit.

The following test pattern `Mehanik HDR10 test patterns/03. Grayscale/04. 10bit test/01. test_10-bit_23.976.mp4` from https://www.avsforum.com/threads/hdr10-test-patterns-set.2943380/ will demonstrate the problem. Unfortunately, because all screenshots I could take would be clamped to 8-bit color depth, I cannot directly provide a picture here.
Direct download link: [test-10bit.mp4](https://user-images.githubusercontent.com/30361859/212476493-489a114e-06d5-43d1-a2e4-43f9063db855.mp4)

CarterLi: I have a lot of HDR test videos.

I meant that it was not `once videoLayer.colorspace was set to CGColorspace.extendedLinear* before` caused the problem

ashen0xa: You're right indeed, the comment and problematic commit message were removed.

low-batt:
Compare IINA 1.3.1 to Quicktime playing [New York 8k | 4k HDR](https://vimeo.com/196168570) specifically looking at the clock in the beginning. Playing that video IINA's image of the clock face looks pretty close to that displayed by Quicktime. When that video is played with IINA built with the changes in this PR the video is darker than before and definitely looks worse than IINA 1.3.1.

When I play the video shown at the start of this PR with IINA 1.3.1 I see the blown out highlights shown in the screenshots. Playing that video with IINA built with the changes in this PR the video is significantly improved. But it does not match Quicktime. The picture is now slightly dimmer than Quicktime. This is especially visible when comparing the light bulb in the lamp. From the comments in the PR I was expecting the picture to match?

I'm testing on a MacBookPro18,2 with the XDR display running macOS 13.1.

I believe @CarterLi has added this fix to [IINA+](https://github.com/iina-plus/iina), so I also tested with IINA+. I had to build it myself since builds are failing reporting:
> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

With IINA+ the test video is darker still. Possibly the difference is that IINA+ is setting `target-peak`?:

```text
18:23:07.250 [hdr][d] Will enable tone mapping target-peak=1600
```

CarterLi:
Can we apply this patch only to HLG videos and keep PQ videos as it was? So that PQ videos can be displayed without any conversion.

ashen0xa:
> Can we apply this patch only to HLG videos and keep PQ videos as it was? So that PQ videos can be displayed without any conversion.

That's a great idea. I did some experiments and found that actually mapping everything to PQ seeming to also solve HLG problem. And this approach can avoid the mess for getting mpv output linear values.

low-batt:
Comparing IINA built with the latest version of this PR to Quicktime when playing [New York 8k | 4k HDR](https://vimeo.com/196168570) the results are pretty close. If stop the video at the beginning and look at the shutters above the clock you can see the picture displayed by Quicktime is slightly brighter.

When I play the video shown at the start of this PR the results are again pretty close. It also looks to me that Quicktime is just slightly brighter.

There are open mpv issues with trying to match Quicktime behavior. We may be up against those issues. This is definitely an improvement.

@CarterLi do you approve of these changes or still have concerns?

CarterLi:
LGTM

CarterLi:
I suggest that iina should adopt iina-plus's manual tone-mapping setting feature, for low max-brightness monitors ( < 1000 nit )

https://github.com/iina-plus/iina/commit/91938e25e00e5cf1ac5985cf235ce077b390dff3

low-batt:
On adopting iina-plus's manual tone-mapping setting, I've been hoping we'd figure out a way to automatically determine a display's brightness.

By the way, the IINA+ builds are still failing. The latest one seems to be stuck in the queued state.

low-batt (approved):
Pulled PR locally, built and tested. Definitely an improvement over the current behavior with the HDR videos I tested.

CarterLi:
FYI: https://github.com/iina/iina/issues/4075#issuecomment-1449710790

iina/InspectorWindowController.swift
```diff
@@ -202,8 +202,17 @@ class InspectorWindowController: NSWindowController, NSTableViewDelegate, NSTabl
       self.setLabelColor(self.vprimariesField, by: sigPeak > 0)
 
       if PlayerCore.lastActive.mainWindow.loaded && controller.fileLoaded {
-        let colorspace = PlayerCore.lastActive.mainWindow.videoView.videoLayer.colorspace?.name;
-        self.vcolorspaceField.stringValue = colorspace == nil ? "Unspecified (SDR)" : String(colorspace!) + " (HDR)"
+        if let colorspace = PlayerCore.lastActive.mainWindow.videoView.videoLayer.colorspace, #available(macOS 10.15, *) {
```
uiryuu:
```suggestion
        if #available(macOS 10.15, *), let colorspace = PlayerCore.lastActive.mainWindow.videoView.videoLayer.colorspace {
```

uiryuu (changes requested):
Just a small change, other parts LGTM

Issue #4156, opened 2023-01-02
With seek bar/controller moved to top left, parts of seek bar doesn't respond.


haakonstorm:
**System and IINA version:**

- macOS 13.1 (on macbook air m1)
- IINA 1.3.1 (via homebrew)

**Expected behavior:**

Clicking and dragging to seek in the video seeks to expected part of video.

**Actual behavior:**

Clicking on seek bar doesn't work if seek bar/controller is moved to top left of screen. The seek bar doesn't work in the exact area as the "notification" telling which timecode of the video you were seeking to. The notification itself overlays the seek bar.

**Steps to reproduce:**

1. Move video controller to the top left (the controller with the seek bar, play, pause etc) while iina is minimized.
2. Try to use the seek bar to seek to an area close to the beginning of the video. 
3. Iina will not seek, because clicking in this region apparently the clicks are sent to the "seek-to-notification" view in the top left, instead of to the seek bar.

<img width="666" alt="Screenshot of IINA (02-01-2023, 14-31-00)" src="https://user-images.githubusercontent.com/193781/210238703-7be3bd2f-d551-4586-abfb-103b23e03d69.png">

**How often does this happen?**

Consistently, reproducible over many versions of iina.

svobs:
Reproduced.

Was also able to fix this easily. When I was working on #4180 I found that the Z order of each view depends on its order in Interface Builder's outline view (the left sidebar). Views lower in the list cast shadows on views higher in the list.

So the solution is just to drag the OSD view up high enough so that it precedes all the OSCs and anything else which can be clicked. Like this:
![SCR-20230120-sn0](https://user-images.githubusercontent.com/2213815/213843702-3ee760ab-14c0-440d-a28b-e16701892146.png)

low-batt:
I have tested PR #4180 and confirmed it fixes this issue.

low-batt:
The fix for this issue has been merged into the IINA `develop` branch. GitHub [automatically closed](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) the linked issue in reaction to the merge. I am reopening this issue until the fix is available in an official release of IINA so that users intending to report this problem can easily locate this existing report. Once the fix for this issue has been included in an official release this issue will be closed.

low-batt:
This was fixed in IINA 1.3.5.

Issue #3838, opened 2022-06-15
Make The Playback Controller Follow The Movement Of The Quick Settings Panel


scrutinizer11:
**What you want IINA to do:**

 Make the controller follow the movement of the Quick Settings panel and change its position as the Quick Settings panel moves out to retain visibility of the controller as suggested in the bottom picture.

**What IINA does currently:**

When the Quick Settings panel _slides out_ with the playback controller in the area adjoining the right edge of the window, a part of it gets obscured by the panel as shown in the top picture.

**Why you think this should be added:**

- Easier to locate the controller on the screen
- Clicking the controls and moving the slider are hampered when the controller gets behind the Quick Settings panel.

![iina-1](https://user-images.githubusercontent.com/45983734/173891205-e158c143-9bab-4c42-a3ef-b0ae3ddc9ce2.png)
![iina-2](https://user-images.githubusercontent.com/45983734/173891219-c572ddb2-3273-43e5-a876-da61d1044b33.png)

low-batt:
And of course if the IINA UI `On Screen Controller` `Layout` preference is set to `Top` or `Bottom` instead of `Floating` as shown here, then the OSC can't be moved. Instead the width of the OSC would have to be reduced to avoid overlaping.

svobs:
My PR (#4180) was designed to handle the `Top` and `Bottom` cases, but the functional requirements for `Floating` don't seem as straightforward to me. If the side panel comes out and pushes the OSC to the left, should that just be its new permanent position, even after the side panel closes? Or should the OSC's previous position be restored when the panel closes (unless of course, the user invalidates the previous position by moving the OSC again manually after it was pushed)?

low-batt:
I would expect that if you opened the panel and closed the panel the OSC would be where it was before you opened the panel, unless in between you moved the OSC.

I mentioned this issue in the PR only because from a high level this is one feature, don't allow the panel to overlap the OSC. Of course a different implementation is required for the floating OSC.

svobs:
> I would expect that if you opened the panel and closed the panel the OSC would be where it was before you opened the panel, unless in between you moved the OSC.

Sounds good to me. You can assign this one to me if you like.

lhc70000:
With the new UI design, this is supported when edge-to-edge video is disabled. It will be available in IINA 1.5.0.

PR #4180, opened 2023-01-18
Refactor MainWindow/OSC layout handling to fix bugs, improve consistency

svobs:
  - #4156
  - #3863

---

**Description:**

Submitting for consideration. This patch changes the following:

1. Bottom OSC shrinks to fit the remaining space when the sidebar opens
2. Top OSC shrinks to fit the remaining space when the sidebar opens
3. Changes sidebar to overlap title bar and adds shadow so that top OSC is consistent with appearance of bottom OSC
4. Uncomments code so that when top OSC is showing, the 1px black horizontal line below it is hidden, so that it is consistent with appearance of bottom OSC
5. Sets `downShift` of sidebar to a constant rather than try to increase it when top OSC is showing (also avoids a bug where `downShift` is inconsistent; to reproduce: (1) show sidebar, then (2) change OSC's `Layout` from bottom to top)

---

Existing:

https://user-images.githubusercontent.com/2213815/213150322-1c1cd032-335b-420d-85b0-0ec770ad67d3.mov

Updated:

https://user-images.githubusercontent.com/2213815/213150350-4819a41f-421f-465b-8a18-bd6f17876113.mov


iina/MainWindowController.swift
```diff
@@ -728,9 +728,9 @@ class MainWindowController: PlayerWindowController {
         titleBarHeightConstraint.constant = TitleBarHeightWithOSC
       }
       // Remove this if it's acceptable in 10.13-
-      // titleBarBottomBorder.isHidden = true
+      titleBarBottomBorder.isHidden = true
```
low-batt: I'm not understanding how this change is related.

iina/MainWindowController.swift
```diff
@@ -1895,7 +1895,7 @@ class MainWindowController: PlayerWindowController {
     NSLayoutConstraint.activate(constraintsH)
     NSLayoutConstraint.activate(constraintsV)
     var viewController = viewController
-    viewController.downShift = titleBarView.frame.height
+    viewController.downShift = TitleBarHeightNormal - 14
```
low-batt: I've been trying to remove magic numbers from MainWindowController. Can the 14 be replaced with a reference to a component's property?

low-batt (commented):
Pulled the PR locally, built IINA and tested. All worked fine.
See my questions inline.
Note that issue #3838 is related in that it is about moving the floating OSC to prevent side panels from covering it. That can be handled as a separate issue, but I thought I'd mention it.

iina/MainWindowController.swift
svobs: Wasn't sure what the intent was there. The bottom sidebar doesn't have a border, and I did think the top sidebar looked better without a border, especially when the side panel was open. But this is not the hill I want to die on. Backed out the change in the new push.

svobs:
Pushed an update that:

1. Backed out change to the bottom border of the top OSC
2. Removed/simplified some code relating to the downshift of the side panel which was no longer used now that the downshift is constant
3. Made an attempt to minimize the "magic numbers" at the top of the MainWindowController, or at least make them more intuitive. For the most part I wasn't successful. They're basically glue code for the layout of the top & bottom OSC. Trying to replace them with a formula and properties wouldn't be trivial or practical because most of them are defined in the XIB, and they do really have some arbitrary numbers in there which are chosen more for aesthetic reasons than some pure equation.
4. I did manage to find a way to get the height of the title bar from NSWindow, although indirectly. Looks good on Ventura, but would be good to get confirmation that it also works on older versions of MacOS.

More about that last one: I'm also not super eager to work too hard on that right now, because I envision coming back at a later date to focus a bit on the OSC. In particular I've been envisioning adding support for scaling the OSC and maybe other customizations. A lot of the existing layout already supports it, but there are a few bits that need some rework - like the vertical offsets & alignment for which those magic numbers are needed. But it probably will need its own effort. My focus right now (if I actually have one) is more on improving player window geometry and making the side panels more accessible.

iina/MainWindowController.swift
svobs: Tried... See comment below.

low-batt:
My normal process for testing a pull request locally is to do this:
```bash
git fetch origin pull/4180/head:pr4180
git checkout pr4180
```

And if the PR is updated I do this to pull the changes: 
```bash
git pull origin pull/4180/head
```

This does not work if the PR is changed by amending the original commit instead of pushing another one. Do you happen to know the correct `git` commands to handle amends?

iina/MainWindowController.swift
```diff
@@ -20,16 +20,21 @@ fileprivate let isMacOS11: Bool = {
 }()
 
 fileprivate let TitleBarHeightNormal: CGFloat = {
-  if #available(macOS 10.16, *) {
-    return 28
-  } else {
-    return 22
-  }
+  // `NSWindow` doesn't provide title bar height directly, but we can derive it by asking `NSWindow` for
```
low-batt: My understanding is that Markdown only works in swift documentation comments, ///, like at the start of a function. I don't think the tick marks around NSWindow in a normal comment will be rendered by Xcode.

iina/MainWindowController.swift
```diff
@@ -20,16 +20,21 @@ fileprivate let isMacOS11: Bool = {
...
+  // the dimensions of a prototypical window with titlebar, then subtracting the height of its `contentView`.
+  // Note that we can't use this trick to get it from our window instance directly, because our window has the
+  // "full-sized content view" style and so its `frameRect` does not include any extra space for its title bar.
+
+  // Probably doesn't matter what dimensions we pick for the dummy contentRect, but to be safe let's make then nonzero.
```
low-batt: typo
make them nonzero

low-batt (commented):
Pulled latest commit, built IINA, tested under both macOS 13.2.1 and macOS 10.15.7. Worked fine.

Should fix that one comment typo.

iina/MainWindowController.swift
svobs: Right. I'm so used to looking at Markdown in its raw form, it's just an ingrained habit at this point. I changed to documentation-style (I assume) blocks starting with `/**`, and looks like they are highlighted now.

svobs:
So I force-pushed another commit.

> This does not work if the PR is changed by amending the original commit instead of pushing another one. Do you happen to know the correct git commands to handle amends?

I do not... My understanding is the the preference for the IINA project is to always stick to a single commit per PR. I had been doing a soft-reset with the changes, and then re-committing, but that required a force-push to get through...  I tried `git commit --amend` like this but it gets an identical error when I try to push:
```bash
msvoboda|21:02:00|~/LocalHome/Dev/IINA/iina-2$ git commit --amend
[pr-resize-osc-with-sidebar ff8a4368] Shrink OSC when it is at top or bottom of window + the sidebar is shown.
 Date: Tue Feb 21 22:56:55 2023 -0800
 5 files changed, 42 insertions(+), 38 deletions(-)
msvoboda|21:02:16|~/LocalHome/Dev/IINA/iina-2$ git push
To github.com:svobs/iina.git
 ! [rejected]          pr-resize-osc-with-sidebar -> pr-resize-osc-with-sidebar (non-fast-forward)
error: failed to push some refs to 'github.com:svobs/iina.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

Let me know if you know or can find a better solution...

---

In this push I also made some small changes to the vertical offsets of the top OSC to get it to closer to identical with the bottom OSC. In fullscreen, the top OSC was 34 pts. Assuming more space is preferred, I changed this to 40 pts to match the bottom OSC. And also adjusted the top OSC in windowed mode to match, since it had excess bottom space in comparison. Hopefully this meets with approval.

Layout-OSC-Bottom
_Bottom OSC_

Layout-OSC-Top-FS
_Top OSC (fullscreen)_

Layout-OSC-Top-with-TitleBar
_Top OSC + titlebar_

---

> Note that issue #3838 is related in that it is about moving the floating OSC to prevent side panels from covering it. That can be handled as a separate issue, but I thought I'd mention it.

I see your comments on that one. I'd rather just do a separate PR rather than drag this one out further. 

There is also issue #4156, which I didn't see until after my first submission of this PR. I'd really like to fix it because its impact is not minor, yet the fix is trivial: just a drag & drop to move one of the elements in `MainWindowController.xib` (in fact as I did for one of the elements in this PR). But since it's the same XIB, I'll have to wait until after this one is either merged or rejected before submitting anything. Wanted to note that here so it won't get forgotten about.

svobs:
Huh...somehow the `titleBarBottomBorder.isHidden` changes got put back in with the last push...Pushed again to correct that.

svobs:
Not sure why the previous CI job failed. Pushed again to fix it.

low-batt:
On the "one commit in PR". I too am used to that being the rule, so a while back I asked the development team about it and was told I do not need to do that. I believe the new convention is to use GiHub's [Squash and merge](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/about-merge-methods-on-github#squashing-your-merge-commits) feature to pack the changes into one commit right before merging. Adding an additional commit when addressing review comments avoids force pushing which is disruptive to anyone who pulled the original commit. It also makes it easier for reviewers to see the changes that were made to address the review comments.

I agree, handle the floating OSC separately.

I've been hoping we'd merge PR  #3529 before making more OSC changes. I've had to fix merge conflicts in that PR quite a few times now. Fine to go ahead with the tweaks in this PR. I'll resolve any conflicts with #3529, again. Before doing a full cleanup of the OSC it would be nice to get #3529 merged. We definitely need to reduce the number of outstanding PRs. Lots of good fixes that should be merged.

The issue/PR I mumbled about cleaning up magic numbers was issue #3911. That issue also shows that for anything involving a `NSSlider` we have to account for changes Apple made in Big Sur. We need to test such changes under Catalina as well as Ventura. Currently I still have Macs running Catalina and can do such testing.

The OSC tweaks in this PR look good to me under Ventura, but not so much in window mode under Catalina. As can been see here the play button is awfully close to the bottom of the OSC in window mode:
pr-4180-changes

With the changes under Catalina, OSC on bottom:
pr-4180-bottom

IINA 1.3.1 under Catalina, OSC on top:
pr-4180-1 3 1

Thoughts?

svobs:
> On the "one commit in PR". I too am used to that being the rule, so a while back I asked the development team about it and was told I do not need to do that. I believe the new convention is to use GiHub's [Squash and merge](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/about-merge-methods-on-github#squashing-your-merge-commits) feature to pack the changes into one commit right before merging.

Got it. I will just go back to adding additional commits from now on. Only downside of squashing when merging is that the original author of the commits is lost (I think).

> Fine to go ahead with the tweaks in this PR. I'll resolve any conflicts with https://github.com/iina/iina/pull/3529, again. Before doing a full cleanup of the OSC it would be nice to get https://github.com/iina/iina/pull/3529 merged. We definitely need to reduce the number of outstanding PRs. Lots of good fixes that should be merged.

I definitely agree. In fact, I don't think it's acceptable to slow down further development due to an unmerged PR which has a greater-than-not chance of being accepted. I took a peek at #3529 and it looks like, fortunately, the XIB changes there can be reduced to a single line, and overall it looks like there's not too much overlap between that one and this one. If there were more overlap, I'd be inclined to just branch from those changes. You might have noticed I already did so for a few other PRs. It might not the polite way of doing things but it's more practical, since (as I see it) any changes to the upstream PR(s) are likely to be minor and more easily sent downstream, which should be a lot easier than having to do a manual rebase of all the changes each time.

svobs:
> The issue/PR I mumbled about cleaning up magic numbers was issue https://github.com/iina/iina/issues/3911. That issue also shows that for anything involving a NSSlider we have to account for changes Apple made in Big Sur. We need to test such changes under Catalina as well as Ventura. Currently I still have Macs running Catalina and can do such testing.

> The OSC tweaks in this PR look good to me under Ventura, but not so much in window mode under Catalina. As can been see here the play button is awfully close to the bottom of the OSC in window mode:

> Thoughts?

I see.... I never bothered to read through that issue completely. So the problem is that the slider size will be different pre/post Big Sur, huh? That helps tremendously.

I'll see if I can refactor a bit so that those magic numbers can be replaced with constraints. Seems like the most appropriate solution will be to vary the OSC's height a few pixels based on OS version, but will have to fiddle with it to see.

low-batt:
> Only downside of squashing when merging is that the original author of the commits is lost (I think).

I too am unsure, but I've not seen that happen with my PRs. Looking over commits I see that GitHub keeps track of both a committer and an author.

I found the post [macOS Big Sur changes for developers](https://wiki.lazarus.freepascal.org/macOS_Big_Sur_changes_for_developers) on the Free Pascal and Lazarus Wiki to be very helpful. I was unable to find this information on Apple's site. Maybe it is somewhere in WWDC videos.

See that post. Apple tweaked a lot in Big Sur.

svobs:
Take a look at the latest update. I put a lot of effort into reworking the OSC's layout calculations, and I think it's pretty slick. One of the key things I did was to create 2 separate Custom Views to hold the top OSC and space for the title bar, respectively, instead of trying to change the top offset of the top OSC's Stack View. Stack Views aren't really first-class objects like other views and can't always be relied on to push other components around as they resize. Rather than messing with arcane gibberish like "content compression resistance priority" I've found it's easier just to package a Stack View inside a Custom View which has better-defined behavior.

So in this latest code, I did manage to reduce the number of constants for the most part, but there are still a few notable ones. Most importantly, the height of the top & bottom OSCs are now mostly dictated by a height constraint on the `fragSliderView` with `height = 44`. To set the vertical space when both title bar + top OSC are showing (i.e. not in full screen), the height of title bar & top OSC are added together, and then a small amount of vertical space (equal to the distance between the top of the window and and the top of its close button is subtracted. The thought is that this should be consistent between MacOS versions +/- a couple of pixels at most. Though I'm less confident about the vertical layout of the items in the "floating" OSC, so please let me know how that looks for pre-Big Sur.

svobs:
One last bit of cleanup. Hopefully won't need any further changes. Also, since I was modifying the code affected by #4156 I went ahead and made sure that was fixed too. I made a note as such in the description.

svobs:
Something in the last couple of commits caused the floating OSC to be slightly too wide. Committed a fix...

low-batt:
I have confirmed the changes work and look good under macOS Ventura 13.2.1 and macOS Catalina 10.15.7.

As this has turned into a significant refactoring of the OSC code there is another OSC defect that is not fixed. That is issue  #3863. I got as far as tracking down what is causing it. See that issue. As the code for the OSC when it is in the top layout was coded differently I checked in with the team as to whether they knew of a reason for including the OSC in the title bar view. I think we are free to come up with a fix. Normally it would be better to keep these changes separate. I'm mentioning it since the issue lands in the code that is being changed by this PR.

svobs:
@low-batt take a look at my latest commit. Assuming I understand #3863 correctly, it should now be fixed.

low-batt:
I did a full test of the PR under macOS Ventura and Catalina. Issue #3863 is fixed, however I found a regression that needs to be fixed. The quick settings panel is now in front of the title bar as can been seen in this screen shot:

pr-4180-title-bar

svobs:
> I found a regression that needs to be fixed. The quick settings panel is now in front of the title bar as can been seen in this screen shot:

I will concede, that was the design I settled on in my original commit, due to both technical and aesthetic compromises. So, with OSC @ bottom, I could put the sidebar below the titlebar, but what about with OSC @ top? At the time, the title bar & top OSC were the same element, so I would have had to draw the sidebar below the OSC - which wouldn't have been consistent with the behavior of the bottom OSC.

Though due to my more recent commits, the title bar & top OSC can now be sized separately. So I can have the sidebar come out underneath the title bar while still squishing the OSC. But since they are drawn with no horizontal line between them (i.e., it looks like the top OSC is part of the title bar), won't it look a little strange to have the side panel come out on top of only the bottom half of that panel?

Or maybe I should go entirely in the opposite direction, and just adjust the sidebar so that it comes out below the top panel & above the bottom panel, thus never stealing space from the OSC?

At this point it may not be that hard to parameterize the constraints for these things, so that via a pref key, different layouts can be shown. It's like this thread keeps getting pulled, and pulls in more refinements I had been planning.

low-batt:
Sorry I'm slow to respond. Lots going on!

Some of the quick settings tabs have a lot of information on them. So I prefer not to shrink the height of the panel. I think the squishing of the OSC is the right approach. But the title bar must remain on top.

As for how odd that would look when the OSC is in the top layout, I'd have to see what that looks like. I would definitely prefer that over what is shown above.

svobs:
> As for how odd that would look when the OSC is in the top layout, I'd have to see what that looks like. I would definitely prefer that over what is shown above.

@low-batt It looks like this is not technically feasible (at least the way I think you're envisioning it).

In order to get the sidepanel to cover the top OSC, there are 2 bad options.

1. Split the top panel (currently an `NSVisualEffectsView` which contains 2 subviews, for title bar and top OSC, respectively)  into two sibling `NSVisualEffectsView`s which are aligned into a vertical stack. 
  
   Problem: this does not appear to be handled well by Quartz. Even though the two views have the same blending mode & material, the result for any given coordinate doesn't appear to be well-defined. (Speculation: it looks like when blending is done for an `NSVisualEffectsView` view, the controls inside the view are factored in, but not the controls inside neighboring views...) At times, the difference between the two views creates a distinct "lip" between them (note the top left):
Screenshot 2023-03-16 at 05 24 13

2. The other option is to leave the top panel as a solid piece, but that means that it's not possible to squeeze the sidepanel between its two child views. The best that can be done is to bring down the sidepanel so that it doesn't cover the title bar. This looks good when the title bar is showing, but when it disappears it looks pretty ugly:
Screenshot 2023-03-16 at 06 41 05

Here's a video of the second solution with top OSC:

https://user-images.githubusercontent.com/2213815/225767057-c72b3875-bc48-4282-8d08-4dc9fc22e43e.mov

So I'm tempted to just push the sidepanel back so the title bar and/or OSC always cover it (meaning that the top & bottom OSCs never get resized). Curious to hear your thoughts though.

In other news - I have been doing a lot of work on this window. By playing with constraints and have managed to get some useful behavior out of them! It looks like I was able to get the desired behavior for #3838 just by adding a couple constraints. Also I added support for showing sidepanels on the left (maybe that's more appropriate for another PR though).

https://user-images.githubusercontent.com/2213815/225767677-6a007fb7-438b-48e8-9ada-cec3cf33dd59.mov

low-batt:
On the first screen shot, the title bar color is fading near the buttons. That is not true in the OSC. Could this be due to some sort of difference in transparency between the two views? Something like that? I agree it looks wrong. Seems like something that could be fixed.

On the second screen shot, I agree that is no good. Currently the panel extends to the top of the window. The tabs in the panel are set lower so the title can be set in front of it. Why can't that it be handled that way?

The moving of the floating OSC looks good!

On the panel extending from the left side. I agree, that should be a separate change.

Mumblings about reviewing and merging in preparation for a bug fix release have started. Stay tuned.

svobs:
> Mumblings about reviewing and merging in preparation for a bug fix release have started. Stay tuned.

Yay! Hopefully a lot can be merged 🤞🏻

> On the first screen shot, the title bar color is fading near the buttons. That is not true in the OSC. Could this be due to some sort of difference in transparency between the two views? Something like that? I agree it looks wrong. Seems like something that could be fixed.

There may be something I'm missing with visual effect views. But I still don't think that the reason for doing this (having the sidebars cover the top OSC while having the title bar cover the sidebars) would look very good unless a horizontal separating line (or something) was introduced. I haven't seen any apps like that. The trend seems to be going in the direction of getting rid of or using a reduced title bar, and putting toolbars in it instead. So I'm experimenting with how those might look.

> On the second screen shot, I agree that is no good. Currently the panel extends to the top of the window. The tabs in the panel are set lower so the title can be set in front of it. Why can't that it be handled that way?

Yeah, it can be... I wonder if it might be possible (or desirable) to push all the title bar items away from the sidebars when they open so they get squished along with the OSCs... This seems to be how XCode works when you click the sidebar buttons in the upper left / upper right. Some people have been asking for buttons like that. I might look into that too.

low-batt:
Nothing in concrete yet but it is likely there will be a bug fix release followed by a bigger release that includes new functionality. Not sure if the bug fix release will be tightly focused on just fixing regressions or will include additional changes.

On moving the OSC into the title. Ever since the switch to wide screen monitors I've been in favor of UIs conserving vertical space. So that would be an attractive option. I see that as a new layout type. Definitely something for a separate PR. I suggest proceeding with keeping the title on top and with full width and just squishing the OSC. Then see if the other developers think that is ugly or not.

Just went through issues and added the `feature: osc` to a bunch of them. There are quite a few requests for new OSC features.

uiryuu:
Didn't play around with this PR for a long time but quickly found a bug:
- Use OSC on top
- Open the quick setting panel
- Shrink the window

You will see that the video aspect ratio is not respected when the window is shrunk to the minimum.

svobs:
This PR is still a long distance away from being ready. It expanded way past its original scope and now it's blocked by a few other things. I just converted it to Draft status to indicate that - sorry about that.

I noticed the aspect ratio problem and found a solution which fixed it, but then I noticed that the fix caused problems with window resizing: trying to resize the window would result in the window moving down and off the screen. So then I started to run up against the code which determines the window size and position, which needs some fixes or maybe refactoring. It might be possible to fix it by just setting a larger minimum window size... but I think I'm going to try a different approach.

I've been working on adding a bunch of new features to the "MainWindow" layout, and as I do that I keep refactoring the code to make it more flexible. I think once I get it to a stable state with maximum flexibility, I can just remove features from that branch and put them into PRs. That way it will be a lot easier for me to support those PRs because they will be cleaner and more pluggable; I won't have to keep maintaining several different implementations while I wait for them to get reviewed...if that makes sense.

svobs:
Closing this PR as it's now obsolete

PR #4187, opened 2023-01-22
Allow OSD progress bar to scale for large text sizes

### LINKED ISSUE #3171: Have OSD progress bar width grow with text size changes
**What you want IINA to do:**

When Text Size is increased causing the OSD overlay to get wider, the progress bar should also extend across the new width of the OSD overlay area.

**What IINA does currently:**

If I set the OSD text size to 40 pts the progress bar stops a little less than half way across the OSD area making the full video progress feel like it's only half way.

IINA_OSD_bar

It would feel more correct as a progress bar to extend across the area. Right now it just seems to end early at a glance it's unclear by progress bar how far you are actually through the video.

svobs:
- [x] This implements/fixes issue #3171.

---

**Description:**

Was about to file a bug for this, but found that someone already filed this as an enhancement request.

Not sure why the progress bar was given this fixed width constraint; its width is already implicitly specified by virtue of being inside an `NSStackView`, and if the intent was to allow widths only up to 150 points, it's not the most reliable way to do it. Probably it was an artifact of some earlier development which was forgotten about.

### DIFF
--- a/iina/Base.lproj/MainWindowController.xib
+++ b/iina/Base.lproj/MainWindowController.xib
@@ -338,9 +338,6 @@
                                     </textField>
                                     <progressIndicator wantsLayer="YES" horizontalHuggingPriority="270" misplaced="YES" maxValue="1" controlSize="small" style="bar" translatesAutoresizingMaskIntoConstraints="NO" id="i10-rt-9cb">
                                         <rect key="frame" x="0.0" y="0.0" width="161" height="12"/>
-                                        <constraints>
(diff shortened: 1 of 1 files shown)


low-batt (approved):
Reproduced the problem with the OSD progress bar being short when when a larger text size is configured. This also means the progress bar is too long if a smaller text size is configured.

I then pulled this PR, built IINA and tested. All looks good to me.

uiryuu:
It was probably intentional, but I think there's no reason to keep that fixed 150 point size.

PR #4191, opened 2023-01-23
Workaround cursor visible after button click, #4183

### LINKED ISSUE #4183: Cursor remains visible after resuming playback with the touchpad using secondary click
**System and IINA version:**

- macOS: Version 13.1 (22C65)
- IINA: 1.3.1 Build 133

**Expected behavior:**
After resuming playback of a video file with secondary click (two finger click by default on macbook) the UI & cursor should become invisible after a short period of time.

**Actual behavior:**
After resuming playback of a video file with secondary click the UI becomes invisible but the cursor remains visible.

**Steps to reproduce:**
* Open a video file
* Pause the video
* Wait a couple seconds
* Resume the playback using secondary click on the touchpad

Note: I think you need to wait a couple seconds otherwise the system will make the cursor go invisible based on the action before that…?

I don't know what MPV is so I can't say whether it's a bug of MPV.

Consistently. As long as you wait a couple seconds.

low-batt:
This commit will:
- Add a new method workaroundCursorDefect to MainWindowController that will hide the cursor again if the OSC is hidden
- Change mouseDown and mouseUp to call workaroundCursorDefect
- Add methods otherMouseDown, otherMouseUp, rightMouseDown and rightMouseUp to MainWindowController that apply the workaround
- Add method rightMouseDown to VideoView to pass that mouse event to MainWindowController

These changes workaround a regression in the macOS AppKit method NSCursor.setHiddenUntilMouseMoves introduced in macOS Big Sur that causes the cursor to become visible when a mouse button is clicked even if the mouse does not move.

This problem has been reported to Apple.

- [x] This implements/fixes issue #4183.

---

**Description:**
This is a workaround for a macOS AppKit defect.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -873,10 +873,25 @@ class MainWindowController: PlayerWindowController {
     }
   }
 
+  /// Workaround for issue #4183, Cursor remains visible after resuming playback with the touchpad using secondary click
+  ///
+  /// When IINA hides the OSC it also calls the macOS AppKit method `NSCursor.setHiddenUntilMouseMoves` to hide the
+  /// cursor. In macOS Catalina that method works as documented and keeps the cursor hidden until the mouse moves. Starting with
+  /// macOS Big Sur the cursor becomes visible if mouse buttons are clicked without moving the mouse. To workaround this defect
+  /// call this method again to keep the cursor hidden when the OSC is not visible.
+  ///
+  /// This erroneous behavior has been reported to Apple as: "Regression in NSCursor.setHiddenUntilMouseMoves"
+  /// Feedback number FB11963121
+  private func workaroundCursorDefect() {
+    guard #available(macOS 11, *) else { return }
+    NSCursor.setHiddenUntilMouseMoves(true)
+  }
+
   override func mouseDown(with event: NSEvent) {
     if Logger.enabled && Logger.Level.preferred >= .verbose {
       Logger.log("MainWindow mouseDown @ \(event.locationInWindow)", level: .verbose, subsystem: player.subsystem)
     }
+    workaroundCursorDefect()
     // do nothing if it's related to floating OSC
     guard !controlBarFloating.isDragging else { return }
     // record current mouse pos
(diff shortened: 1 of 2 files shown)


uiryuu:
I think it is possible that Apple intentionally did that. So what if we show the UI as well when pausing and resuming even if the cursor doesn't move?

low-batt:
I would find the behavior of bringing up the OSC even though the mouse did not move irritating and would ask for a preference to disable it. Some of the time I intentionally keep the mouse still and pause and resume using the mouse buttons. I did not notice this defect since the Mac's I do that on are still running Catalina (which does not have this problem) as I wait for Apple to fix some of the many regressions added in Big Sur. Without adding new ones.

I too was worried that Apple intentionally made this horrendous change breaking backward compatibility, turning the name of the API into a lie and then on top of that did not update the documentation to reflect this new unexpected behavior, all for something an application can easily accomplish if they want to while processing the button click. I concluded Apple can not have possibly intentionally made such a change and it must be an accidental regression.

As of today Apple has not responded to the feedback I submitted reporting this regression. My guess is that during triage they looked at the title of my feedback and thought "that can't possibly be broken, must be a user error" and inserted my defect report deep into the massive backlog.

low-batt:
I continued to think about this today. I'm now even more convinced that the correct solution is to restore the IINA behavior seen in Catalina before Apple changed AppKit. To get to the OSC all the user has to do is to move the mouse. But that is not what they are doing. They are intentionally clicking somewhere in the window instead of moving the mouse.

They are doing this to invoke whatever action they have configured IINA to perform in reaction to a mouse button click. That can be configured in the IINA settings `Control` tab. 

In my case I have the `Single click to` setting configured to `Hide OSC`. Showing the OSC would definitely not be the correct reaction. If we thought showing the OSC was a good idea in reaction to a button click we could enhance this section of settings to provide such an option.

The user does receive feedback if they have the button click configured to `Pause / Resume` as IINA will display an OSD message (unless you disabled to OSD).

uiryuu (approved):

Issue #885, opened 2017-07-19
Larger thumbnails


adarshv90:
Make thumbnails larger. Better yet, give the option for custom thumbnail sizes

**System and IINA version:**

- macOS 10.13
- IINA 0.0.12

JohnClay:
This would be very nice.

kgoldmann:
Agreed. For comparison:

Netflix
<img width="1535" alt="screen shot 2018-02-22 at 03 28 25" src="https://user-images.githubusercontent.com/8964638/36516963-a06704c2-1780-11e8-9d23-62c29675c445.png">

IINA
<img width="1439" alt="screen shot 2018-02-22 at 03 29 00" src="https://user-images.githubusercontent.com/8964638/36516968-a4fbedae-1780-11e8-91d2-55a9a62daa09.png">

firstel:
I would just like thumbnails to work over smb.  this command works - but only once.  

defaults write com.colliderli.iina enableThumbnailForRemoteFiles -bool YES

inflation:
@firstel Please open another issue or reply in the relevant issue.

bozorgmehr96:
Do we get larger thumbnails ever?

joeynotjoe:
Adding my vote to this, would love to be able to choose what size we display the thumbnail at, as well as what interval they are generated.

AJeezy9:
Please please please please please please please please 😄

triemilalim:
I almost make a new feedback, turns out someone already asked for this 3 years ago. Please give us the option to make the thumbnail seeker bigger, or even custom thumbnail sizes. Thanks guys!!

anastasiuspernat:
I made a pull request for this one! Hopefully @lhc70000 or other IINA demiurge else will have time to review and integrate it if it works out that way. Thanks!

TitanicKiwi:
I'm adding myself to this request. Please add the option to set our own size for the thumbnail preview.

rogery555:
unbelievable this is still open, thumbnails are barely usable they are so small

Careless:
come on alreaddddddddddddy

FaysalAriss:
Seems to be on the way, but if you're in a hurry download the latest nightly (currently 7bc60e6) and in a terminal run: 
defaults write com.colliderli.iina thumbnailWidth -int 600
Replace 600 with any width you want.

uiryuu:
If you are using the nightly builds, you can try out our new settings window by running this command:
```
[...]

Issue #3972, opened 2022-10-07
Bigger Thumbnails For The Love of God


AJeezy9:
I know this has been mentioned before, but for the love of god, can IINA please allow users to customize the size of videos thumbnail previews (pixel width or preferably adaptive). 

The images are currently just WAY too small to even see when IINA is displayed on a high resolution screens.

hugepower:
I had this need once, but I solved it by modifying and compiling it myself.

PR #3289, opened 2021-01-16
Specify larger/arbitrary thumbnail size from UI

anastasiuspernat:
- [X] This change has been discussed with the author - yes, in the form of this pull request.
- [X] It implements / fixes issue #885.

**Description**

Main description of the problem is here: https://github.com/iina/iina/issues/885

Thumbnails can appear too small on larger screens. This improvement allows to specify arbitrary thumbnail size from the UI. Added the "Width" UI element to preferences in IINA.

Screen Shot 2021-01-15 at 7 08 39 PM
Screen Shot 2021-01-15 at 7 09 35 PM

PR #4195, opened 2023-01-26
Allow user to set size of thumbnails + other thumbnail stuff

### LINKED ISSUE #3408: Thumbnail preview doesn't work for videos played from a network drive
**System and IINA version:**
- macOS 11.3.1
- IINA 1.1.2

**Expected behavior:**
Video thumbnails should show for all videos, including videos played from a network drive

**Actual behavior:**
Video thumbnails only appear for videos played from the local drive

**Steps to reproduce:**
Place a video on a network drive.  (I use a NAS connected via SMB)
Play the video
Notice how thumbnails do no appear

Every time

svobs:
- [x] This implements/fixes issue #885 / #3972
- Also fixes #3408. Sorry, didn't want to waste time breaking this into a separate PR.

---

**Description:**

Note: PR #4107 is a prerequisite for this because it builds on its layout.

SCR-20230126-f4

SCR-20230126-15l

I was mucking around in the `Settings` > `UI` tab for unrelated things, but went into a rabbit hole after I saw commented-out stuff and here's the result :)

Looks PR #3289 (although still open) was fully merged into develop, but more work was needed. This commit:
- Adds a level to the thumbnail cache directory structure so that instead of "`thumbnailCacheURL`/`MD5`", the thumbs go in "`thumbnailCacheURL`/`width`/`MD5`"
- Makes the width setting visible in the UI, but swapped its place with the cache size setting, because it seemed funny the way it was.
- Adds code to `PlayerCore` so that it detects when the pref setting is changed, and will regenerate its thumbnails. But saw that it gets notified immediately for each character the user types! So I added a timer mechanism to it, so that it will only regenerate if the new value doesn't change for 2 seconds.
- Replaces hard-coded widths with `width` variable (previously, thumbs were generated with 240px width, then displayed with 120px width).

I was surprised how fast the thumbnails generate and display! Now I don't know how I ever lived with sub-500-pixel thumbnails!

One more change, because it would be hard to separate the code for each; and I already spend too much of my time on these write-ups. I added a new checkbox, `Enable for remote files also`, which toggles enablement of thumbnails for remote video files. This pref already existed but is disabled by default (see #3408). I really agonized over the name, but am open to changes. Or maybe would be good to add a further explanation below it, like "Not recommended for slow network connections" or "May slow down playback if bandwidth is limited". Of course on my NAS it was still very fast and I wouldn't have noticed the difference. Probably a more ideal solution would be to try to check the state of the playback buffer (if possible) and only generate thumbnails when the buffer is close to full.

### DIFF
--- a/iina/AppData.swift
+++ b/iina/AppData.swift
@@ -37,6 +37,11 @@ struct AppData {
   static let seekAmountMapMouse = [0, 0.5, 1, 2, 4]
   static let volumeMap = [0, 0.25, 0.5, 0.75, 1]
 
+  // Time in seconds to wait before regenerating thumbnails.
+  // Each character the user types into the thumbnailWidth text field triggers a new thumb regen request.
+  // This should help cut down on unnecessary requests.
+  static let thumbnailRegenerationDelay = 2.0
+
   static let encodings = CharEncoding.list
 
   static let userInputConfFolder = "input_conf"
--- a/iina/Base.lproj/PrefAdvancedViewController.xib
+++ b/iina/Base.lproj/PrefAdvancedViewController.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="21225" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="21507" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="21225"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="21507"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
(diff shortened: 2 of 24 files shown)


svobs:
Backed out the portion of changes which "fixed" the number of thumbnails generated. I see now why 100 is more of an aspiration due to technical tradeoffs. There are likely bugs in the way it reports progress, but that doesn't seem to be used for anything at present.

svobs:
Updated as noted in the comments to issue #4250: 
> I already pushed an update to #4195 which uses the long-side of the video instead, and changes the verbage from `Width:` to `Size:` in the UI (the thought being that "size" might be vague, but "width" wouldn't be completely honest). I felt it might be a good idea to write up a defect report to explain why this change was needed.

svobs:
Shelving this for now rather than keep maintaining it.

Scrawling down some ideas for further thumbnail enhancements which popped into my head. Maybe will submit a new PR with some of these in, let's say, mid-2024.

- Add pref to map thumbnail size to % of window (so, "thumbnail size" ➜ "thumbnail native size").
- Do not create thumbnails larger than native video size (or possibly provide a checkbox to always store thumb at video's native size and then always scale it).
- When searching for existing thumbnails, check if larger thumbnails exist; if they do, just load those and scale them down instead of generating new ones.
- Use Auto-Layout constraints to make sure the thumbnail is always the right size and position (easier to specify than the current solution which uses code and doesn't cover all cases). Thumb will not block OSC and will not be clipped by side of window; largest possible size will be == video size).

svobs:
This PR is now obsolete. I've completely rewritten the affected code in [my fork](https://github.com/svobs/iina-advance) using an improved strategy which allows the user to specify both the native size of the thumbnails as a percentage of original video and the displayed size as a percentage of window height. Maybe I will submit another PR 
 in the future if/when it makes sense.

PR #4138, opened 2022-12-22
Chrome extension

lusuee:
- [x] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)
- [x] This implements/fixes issue #2501.

---

**Description:**
upgrade manifest_version to 3;
fixed issue #2501

PR #4196, opened 2023-01-26
Encode apostrophe in URI

### LINKED ISSUE #4186: Open In IINA Fails to launch IINA if links contain an apostrophe
I've noticed that some links fail to launch IINA using OpenInIINA because of a quotation issue in the extension. 
The problem seems to be in [this file](https://github.com/iina/iina/blob/develop/browser/Chrome_Open_In_IINA/common.js) on line 62. 

This is an example of such link:
[https://odysee.com/@TheLinuxExperiment:e/no,-arch-isn't-n%C2%B01,-x-org-is-dying,:9](https://odysee.com/@TheLinuxExperiment:e/no,-arch-isn't-n%C2%B01,-x-org-is-dying,:9)

To reproduce, try using the extension to open the link and observe the console.

--- Edit ---
It seems that the problem is caused by `encodeURIComponent()` on line 62 not escaping the single quote! 
Appending `.replace(/'/g, '%27')` after the encode does the trick.

I don't do pull requests for personal reasons. Don't ask!

MacOS 10.13
Vivaldi 5.6.2867.58
Open In IINA 2.0.0

kelvinc6:
- [x] This implements/fixes issue #4186.

---

**Description:** Add extra encoding for single apostrophe for Open In IINA.

### DIFF
--- a/browser/Chrome_Open_In_IINA/common.js
+++ b/browser/Chrome_Open_In_IINA/common.js
@@ -59,7 +59,7 @@ export function restoreOptions() {
 
 export function openInIINA(tabId, url, options = {}) {
     const baseURL = `iina://open?`;
-    const params = [`url=${encodeURIComponent(url)}`];
+    const params = [`url=${encodeURIComponent(url).replace(/'/g, '%27')}`];
     switch (options.mode) {
         case "fullScreen":
             params.push("full_screen=1"); break;


low-batt:
Once a problem has been identified it is always a good idea to give some thought as to whether this is a symptom of a larger problem. In the case at hand the question to ask is whether there are other characters besides single quote that also need to be encoded.

The Mozilla documentation for [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) contains a section [Encoding for RFC3986](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent#encoding_for_rfc3986) which says:
> The more recent [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986) reserves !, ', (, ), and *, even though these characters have no formalized URI delimiting uses.

And goes on to give this code example:
```JavaScript
function encodeRFC3986URIComponent(str) {
  return encodeURIComponent(str)
    .replace(
      /[!'()*]/g,
      (c) => `%${c.charCodeAt(0).toString(16).toUpperCase()}`
    );
}
```

Do we need a more comprehensive fix?

This is outside my area of expertise. See what the other developers say.

NikkSaan:
@low-batt 
Very thoughtful comment, low-batt... what you suggest is indeed a more robust approach. But the real issue is caused by a string containing an apostrophe in conjunction with single quotes in the string template later in the code. So the current solution (which I've been daily driving) is sufficient. Plus, In manifest v3 this bug will fix itself because the template string will not be needed anymore and we can probably go back to stock encodeURIComponent().

kelvinc6:
I see. Thanks for providing additional insight @NikkSaan @low-batt. I suppose this can be closed if no longer needed.

low-batt:
@kelvinc6 Thanks for giving me credit, but my concern was based on a lack of understanding and proved to be unwarranted.

As for closing the issue, I assume that is a reaction to this?:
> In manifest v3 this bug will fix itself

I'm operating way outside of my area of expertise, so @NikkSaan may have to correct my understanding (again), but this has to do with the Chrome extension platform moving to [Manifest V3](https://developer.chrome.com/docs/extensions/mv3/intro/). As the current IINA Chrome extension is coded against [V2](https://github.com/iina/iina/blob/da1e06cf41c32390356e2b69120cc969bcd31612/browser/Chrome_Open_In_IINA/manifest.json#L2) the fix is needed. The issue should remain open.

Updating IINA's extension to V3 is a separate issue that IINA should be tracking as Chrome has plans to eventually phase out support for V2.

I will be adding this PR to the list of changes to review and merge for the next release of IINA.

NikkSaan:
@kelvinc6 No, it should _not_ be closed. It's still needed. Without this fix half of the videos from Odysee can't be opened through the extension! Thank you for putting in the work to upsteaming it!

Your understanding is correct, @low-batt. From looking at the pull request #4138 for implementing manifest v3 it seems that the apostrophe won't be causing un issue after the merge. But I haven't tested the changes for v3 so I can't guarantee anything. I'll keep an eye on it.

PR #4222, opened 2023-02-17
Fix bug: menu items with variable titles and values failed to reset to their defaults after their key bindings were removed

### LINKED ISSUE #4221: Numbers in menu items do not change back to defaults when their bindings are removed
**System and IINA version:**

- macOS 13.1
- IINA 1.3.1 (latest develop: da1e06cf41c32390356e2b69120cc969bcd31612)

Discovered this while investigating #4217. 

For some mpv key bindings, IINA sees if they fall within an expected range, and if so, may alter some of its menu items so that they can match. For example, if `seek` has a value between 5s and 60s, IINA will change its `Playback` > `Step Forward` menu item to match its value, and will display its key binding there. But if the user removes the key binding, or changes it in a way that causes it to no longer qualify for the menu item, then the menu item should fall back to its previous value.

**Steps to reproduce:**

I performed these steps:

1. Open any video using the latest build of IINA. Pause the video near its beginning.
2. Go to `Settings` > `Key Bindings`, and duplicate the `IINA Default` configuration, and make it the active configuration. Make sure `Display raw values` is checked.
3. Locate the entry for the `RIGHT` key. Its action should be `seek 5`.
4. Double-click on its action to edit using the inline editor, and change its action to `seek 30`. Press `RETURN` to accept it.
5. Check the `Playback` menu and confirm that `Step Forward 30s` is present and has the `⇾` shortcut.
6. Double-click the same action again to edit it, and change it to `seek 3`. Press `RETURN` to accept it.
7. Check the Playback menu. You will see that `Step Forward 30s` is still present but without the `⇾` shortcut. This is incorrect. It should fall back to Step Forward 5s. (If you were to close & reopen IINA at this point, you would see that it will correctly fall back to this value.)

**Expected behavior:**

In the Playback` menu, the first item in the second section down should be `Step Forward 5s` and have no shortcut. When clicked on, it should step the video about 5 seconds forward. 

**Actual behavior:**
The item in that location is still called `Step Forward 30s`, and when clicked on, it steps the video about 30 seconds forward. 

Always

svobs:
- [x] This implements/fixes issue #4221.

---

**Description:**
If technical debt could be measured, `MenuController` should be declaring bankruptcy. Had to fight my natural tendency to try to go beyond and try to refactor it. But I think I found a solution which works fairly well, even though it's also kind of ugly.

There are 16 menu items (which I found at least) in which `MenuController` will change the their title and associated data based on their value parameter. IINA has been deriving these based on their defaults in `bindMenuItems()` at program start. It has also been deriving these from the user's active key bindings in `updateKeyEquivalentsFrom()` - but only if they matched one of the user's bindings.

Unfortunately, `bindMenuItems()` doesn't cache the default values it generates, and there's no easy way to modify that method without it being a lot of work. However, the `settings` array inside the `updateKeyEquivalentsFrom()` method contains all the information needed in order to re-derive those default values. It's just necessary to supply the default values twice in order to be able to derive it via the existing code:

```Swift
let (_, value) = sameKeyAction(actions, actions, normalizeLastNum, numRange)
```

Finally, this PR removes some code in `bindMenuItems()` which sets the `representedObject` for those 16 fields at program load. They are completely unnecessary, because the updated code will get the same result when it calls `updateKeyEquivalentsFrom()` at program start. Following the DRY principle, it's important to remove them as a potential source of error, in case their values ever needed to be changed in the future.

### DIFF
--- a/iina/MenuController.swift
+++ b/iina/MenuController.swift
@@ -211,10 +211,6 @@ class MenuController: NSObject, NSMenuDelegate {
     jumpTo.action = #selector(MainMenuActionHandler.menuJumpTo(_:))
 
     // -- speed
-    (speedUp.representedObject,
-     speedUpSlightly.representedObject,
-     speedDown.representedObject,
-     speedDownSlightly.representedObject) = (2.0, 1.1, 0.5, 0.9)
     [speedUp, speedDown, speedUpSlightly, speedDownSlightly, speedReset].forEach { item in
       item?.action = #selector(MainMenuActionHandler.menuChangeSpeed(_:))
     }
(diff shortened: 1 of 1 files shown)


low-batt (approved):
Pulled PR locally, built IINA and tested. Looks good to me.

uiryuu:
Thanks

PR #4224, opened 2023-02-18
Include muted status in Audio menu

### LINKED ISSUE #4201: Audio menu should indicate when audio is muted
SCR-20230128-ty7

_↑ Screenshot taken while the video is muted_

**What you want IINA to do:**

When audio is muted, have the Audio menu's volume indicator show that. Changing it to `Volume: 100 (Muted)` would be good.

Even better, the `Mute` menu item should change to `Unmute` if the audio is already muted.

**What IINA does currently:**

When the media is muted, the menu shows the volume level (0-100), but no indication that it's muted.

It's very misleading if there's no audio playing but the Audio menu just shows 100 volume.

svobs:
- [x] This implements/fixes issue #4201.

---

**Description:**

Adds checkmark to `Muted` menu item, and includes "(Muted)" in the volume indicator menu item, when the audio is muted, as discussed in the above issue.

### DIFF
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -109,6 +109,7 @@
 "preference.enable_adv_settings" = "Enable advanced settings";
 
 "menu.volume" = "Volume: %d";
+"menu.volume_muted" = "Volume: %d (Muted)";
 "menu.audio_delay" = "Audio Delay: %.2fs";
 "menu.sub_delay" = "Subtitle Delay: %.2fs";
 "menu.fullscreen" = "Enter Full Screen";
--- a/iina/MenuController.swift
+++ b/iina/MenuController.swift
@@ -489,7 +489,15 @@ class MenuController: NSObject, NSMenuDelegate {
           player.mainWindow.quickSettingView.currentTab == .audio
     quickSettingsAudio?.title = isDisplayingSettings ? Constants.String.hideAudioPanel :
         Constants.String.audioPanel
-    volumeIndicator.title = String(format: NSLocalizedString("menu.volume", comment: "Volume:"), Int(player.info.volume))
+    let volFmtString: String
+    if player.info.isMuted {
+      volFmtString = NSLocalizedString("menu.volume_muted", comment: "Volume: (Muted)")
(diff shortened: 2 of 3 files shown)


svobs:
This added a new string in Localizable.strings when muted. Still a bit fuzzy about i18n - please let me know if this wasn't the right approach.

low-batt:
Thanks to @uiryuu the contributing document now contains instructions for [localization](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md#localizations):
> When there is a UI text change, only add/update the strings in Base.lproj and en.lproj.

This PR contains changes to `en.lproj/Localizable.strings` that need to be added to `Base.lproj/Localizable.strings` as well.

Other than the need to update the `Base.lproj` file this PR is good to go.

I pulled the changes locally and tested. I tried the various ways of muting, OSC button, menu, key shortcut. All worked fine. I also confirmed that `Mute` is correctly checked if audio is muted due to previous state being restored by the mpv watch later feature.

It should be noted that `Mute` will not be checked if audio is muted at the macOS level. To me that behavior is expected and correct. It would be wrong to display a checkmark that reflected the macOS setting. If we think that is confusing to users then the solution I would propose would be to use a `(Muted in macOS)` suffix on the `Volume` menu item. I view this as a separate issue. I would not add any such changes in this PR. See what others say.

svobs:
Thanks! I've updated accordingly.

> It should be noted that Mute will not be checked if audio is muted at the macOS level.

This is an interesting idea which I hadn't even considered. I don't think I've ever seen such behavior, but it sounds useful and might be worth thinking about. There are certainly too many occasions where we find we have no sound and are forced to troubleshoot several possibilities, and it would be nice to cut away at that problem.

low-batt (approved):
Pulled PR locally, built IINA and confirmed change is working.

Looks good to me.

low-batt:
What to do about audio being muted at the macOS level is definitely a separate issue. For example it could be that the volume is set to zero at the macOS level. Definitely need to think a good bit about the implications of including anything to do with the macOS settings.

uiryuu:
Thanks!

PR #4226, opened 2023-02-19
Fix fullscreen only works for first video, #4202

### LINKED ISSUE #4202: Fullscreen mode only works for first video
**System and IINA version:**
- macOS 13.2
- IINA 1.3.1 Build 133, mpv 0.34.1

**Expected behavior:**
I have set "General / When Media is Opened / Enter Fullscreen", "Always open media in new window".
When clicking on a video file, I want IINA to play it back in fullscreen mode. Afterwards IINA should close the window, but the application should not quit.
When clicking on another video (or the same one again), IINA should play it back in fullscreen mode, close the window at the end and stay open.
Repeat.

**Actual behavior:**
Clicking on the initial video IINA opens and plays the video in fullscreen mode. The window closes at the end. IINA stays open.
Clicking on another/the same video again, IINA starts playing back the video in windowed mode.

I was not able to find a way to playback each video in fullscreen mode (unless "quit after all windows are closed" is active).
Is this the intended behavior?

Consecutive executions of "`mpv /path/to/video.mp4`" always play back in fullscreen mode.

Every time.
[iina.log](https://github.com/iina/iina/files/10536027/iina.log)
[mpv.log](https://github.com/iina/iina/files/10536028/mpv.log)

low-batt:
This commit will:
- Remove a check in `PlayerCore.fileLoaded` that only put the player into full screen mode for the first file
- Remove the property `justLaunched` from `PlaybackInfo`

This causes the "Enter fullscreen" setting under the "When media is opened" settings to when enabled put the player into full screen mode for all videos, not just the first video played.

- [x] This implements/fixes issue #4202.

---

**Description:**

### DIFF
--- a/iina/PlaybackInfo.swift
+++ b/iina/PlaybackInfo.swift
@@ -79,7 +79,6 @@ class PlaybackInfo {
     }
   }
 
-  var justLaunched: Bool = true
   var justStartedFile: Bool = false
   var justOpenedFile: Bool = false
   var shouldAutoLoadFiles: Bool = false
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -1423,12 +1423,8 @@ class PlayerCore: NSObject {
         miniPlayer.defaultAlbumArt.isHidden = self.info.vid != 0
       }
     }
-    // set initial properties for the first file
-    if info.justLaunched {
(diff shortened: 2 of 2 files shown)


uiryuu:
I think this should be a safe fix, based on the current code.

PR #4232, opened 2023-02-20
Improve reliability of click detection in player window by adding minimum threshold for drag

### LINKED ISSUE #4231: Drag detection in main window is way too sensitive, causing some taps to be "ignored"
**System and IINA version:**

- macOS 13.1
- IINA 1.3.1 (latest develop: da1e06cf41c32390356e2b69120cc969bcd31612)

I've been frustrated by this issue for quite a while now, and although I couldn't find anything via a quick search, I'm pretty sure I've come across at least a couple user reports that noted there was "no way to close the side panel" or that clicking to close didn't work reliably. I finally committed a big chunk of time to tracking it down, and I think I've found the culprit.

To get a better view of how clicks were being registered, I placed these logging statements at the start of these methods in `MainWindowController`:
```Swift
  override func mouseDown(with event: NSEvent) {
    Logger.log("MainWindow mouseDown \(event.locationInWindow)", level: .verbose, subsystem: player.subsystem)
...

  override func mouseDragged(with event: NSEvent) {
    Logger.log("MainWindow mouseDrag \(event.locationInWindow)", level: .verbose, subsystem: player.subsystem)
...

  override func mouseUp(with event: NSEvent) {
  Logger.log("MainWindow mouseUp! isDragging: \(isDragging), isResizingSidebar: \(isResizingSidebar), clickCount: \(event.clickCount)", level: .verbose, subsystem: player.subsystem)
...
```

When I ran IINA with my test logging, I saw messages in the log like these:

```
01:06:22.638 [player0][v] MainWindow: mouseDown: (247.8671875, 417.46875)
01:06:22.640 [player0][v] MainWindow: mouseDragged: (247.8671875, 417.5859375)
01:06:22.767 [player0][v] MainWindow mouseUp! isDragging: true, isResizingSidebar: false, clickCount=1
```

This declared my tap was a "drag" (and thus not a click) even though my click lasted for 127ms and the drag distance was only 0.1pt in screen coordinates!

I should point out that I'm using the trackpad on a 16" MacBook Pro M1 Pro. Unlike most developers, I never use a mouse. (It's one of the main reasons I use Macs instead of other operating systems: multi-touch on on Linux/Windows has been almost unusable). And I happen to wear a glove size XL - I've got large hands and fingers. So I can see how this problem may not be noticeable for a mouse user, or someone who has smaller fingers. But for me, Apple's trackpad is so sensitive that it will send a "drag" notification even if my finger does a tiny rolling motion during a tap, or if my motion isn't straight down and up again. This is an unreasonable amount of precision to expect, and it doesn't align with other apps' behavior.

**Expected behavior:**

If the user does what a reasonable person considers to be a "tap", IINA should respond to it as a tap. It should only think the user is dragging once they drag some very small minimum distance, or hold it down longer than some reasonable time threshold, or both.

**Actual behavior:**

Sometimes users' taps or clicks are basically ignored because IINA incorrectly concluded that they wanted to quickly move the window by a single pixel or less.

**Steps to reproduce:**

See above.

When trying to close a side panel by tapping the video, there's about a 50-75% chance that my first tap will fail and a second attempt is needed. Probably the frequency varies widely between different users based on their usage patterns, and maybe trackpad users are more affected than mouse users.

svobs:
- [x] This implements/fixes issue #4231.

---

**Description:**

This updates the click detection in `MainWindowController` and adds a requirement that after mouseDown the user must drag the cursor at least 2pts distance for drag to start.

A given `mouseDown`/`mouseUp` pair is processed as either a click or a drag, but not both. The out-of-the-box behavior in Cocoa is unfortunately not very useful to take on its own: the `mouseDragged` callback is called with very high sensitivity (particularly with trackpad actions); `mouseDragged` will get called for drag distances of just a fraction of a pixel. This is both not very useful as a an actual drag distance, and actively harmful because it results in clicks/taps appearing to be ignored from the user's standpoint.

After some testing and tweaking, I concluded that a drag start threshold requirement of 2pts is all that's needed to ensure that clicks are registered reliably *and* there is no noticeable burden on actual drag attempts. Note that 2pts is all that's needed to *start* the drag, so if the user's goal is to move their player window by only 1pt, they will have to first drag it by 2 pts to register the drag, then drag it back by 1 pt to position it. I believe this behavior is similar to how CAD & layout software usually works, but as a video player it's expected to be very rare that any user will be making such fine adjustments anyway.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -50,7 +50,7 @@ fileprivate extension NSStackView.VisibilityPriority {
 }
 
 // The minimum distance that the user must drag before their click or tap gesture is interpreted as a drag gesture:
-fileprivate let minimumInitialDragDistance: CGFloat = 4.0
+fileprivate let minimumInitialDragDistance: CGFloat = 3.0
 
 class MainWindowController: PlayerWindowController {
 
@@ -843,7 +843,7 @@ class MainWindowController: PlayerWindowController {
 
   override func mouseDown(with event: NSEvent) {
     if Logger.enabled && Logger.Level.preferred >= .verbose {
-      Logger.log("MainWindow mouseDown \(event.locationInWindow)", level: .verbose, subsystem: player.subsystem)
+      Logger.log("MainWindow mouseDown @ \(event.locationInWindow)", level: .verbose, subsystem: player.subsystem)
     }
     // do nothing if it's related to floating OSC
     guard !controlBarFloating.isDragging else { return }
@@ -859,9 +859,6 @@ class MainWindowController: PlayerWindowController {
   }
 
   override func mouseDragged(with event: NSEvent) {
-    if Logger.enabled && Logger.Level.preferred >= .verbose {
-      Logger.log("MainWindow mouseDrag \(event.locationInWindow)", level: .verbose, subsystem: player.subsystem)
-    }
     if isResizingSidebar {
       // resize sidebar
       let currentLocation = event.locationInWindow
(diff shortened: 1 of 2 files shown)


low-batt:
See my long comment in the issue. I'm thinking the user should be able to control this, but wait to see what others think.

iina/MainWindowController.swift
```diff
@@ -49,6 +49,8 @@ fileprivate extension NSStackView.VisibilityPriority {
   static let detachEarliest = NSStackView.VisibilityPriority(rawValue: 750)
 }
 
+// The minimum distance that the user must drag before their click or tap gesture is interpreted as a drag gesture:
+fileprivate let minimumInitialDragDistance: CGFloat = 4.0
```
uiryuu: Please add some comments to clarify that the intended distance is 2 instead of 4 since we are using spared distance.

uiryuu (changes requested):
LGTM

svobs:
Made another commit. The whole "squared" thing was due to copying an answer from StackOverflow, but it's silly to focus on avoiding a square root in this case because it won't affect performance in any meaningful way. Tweaked the drag distance so it is 3.0 pts instead of 4.0, but shouldn't make much difference.

uiryuu (approved):

uiryuu:
Thanks!

PR #4256, opened 2023-03-10
Playback History Window: add confirmation before delete from keypress, add DEL key trigger, remove {modifier}+BS triggers

### LINKED ISSUE #4255: Playback History window: no confirmation when Backspace key used
**System and IINA version:**

- macOS 13.1
- IINA 1.3.1 Build 133

**Expected behavior:**

Because right-click on an entry in this window puts up a confirmation dialog before deleting, one would expect the same if using the Backspace key.

**Actual behavior:**
No confirmation. Item(s) immediately deleted.

Also, using `Ctrl`+`Backspace`, or `Shift`+`Backspace`, `Option`+`Backspace` (i.e. any modifier other than `Command`) also deletes the entry, when they probably shouldn't. Also one would expect the `Delete` key to work also, but it does not.

Always

svobs:
- [X] This implements/fixes issue #4255.

---

**Description:**
Uses KeyCodeHelper to translate like in the welcome window.

### DIFF
--- a/iina/HistoryWindowController.swift
+++ b/iina/HistoryWindowController.swift
@@ -108,12 +108,12 @@ class HistoryWindowController: NSWindowController, NSOutlineViewDelegate, NSOutl
     let flags = event.modifierFlags.intersection(.deviceIndependentFlagsMask)
     if flags == .command  {
       switch event.charactersIgnoringModifiers! {
-        case "f":
-          window!.makeFirstResponder(historySearchField)
-        case "a":
-          outlineView.selectAll(nil)
-        default:
-          break
+      case "f":
+        window!.makeFirstResponder(historySearchField)
+      case "a":
+        outlineView.selectAll(nil)
+      default:
+        break
       }
     } else {
       let key = KeyCodeHelper.mpvKeyCode(from: event)
--- a/iina/HistoryWindowController.swift
+++ b/iina/HistoryWindowController.swift
@@ -95,22 +95,32 @@ class HistoryWindowController: NSWindowController, NSOutlineViewDelegate, NSOutl
     historyData[key]!.append(entry)
   }
 
+  private func removeAfterConfirmation(_ entries: [PlaybackHistory]) {
+    Utility.quickAskPanel("delete_history", sheetWindow: window) { respond in
+      guard respond == .alertFirstButtonReturn else { return }
+      HistoryController.shared.remove(self.selectedEntries)
+    }
+  }
+
   // MARK: Key event
 
   override func keyDown(with event: NSEvent) {
-    let commandKey = NSEvent.ModifierFlags.command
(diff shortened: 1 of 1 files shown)


low-batt:
Failed my testing.

Sometimes the history entry is deleted when I confirm the deletion and sometimes not. When the entry is not deleted the array `selectedEntries` is empty even though the entry is highlighted in blue. A quick look at `HistoryWindowController` shows that array is only updated in `menuNeedsUpdate`. Likely that needs to change.

Glad you noticed this problem. Definitely needs to be fixed.

svobs:
Good catch. I pushed an update.

low-batt (approved):
Pulled latest commit, built IINA and tested under macOS 13.2.1. All seems to be working. Looks good to me.

iina/HistoryWindowController.swift
```diff
@@ -95,22 +95,32 @@ class HistoryWindowController: NSWindowController, NSOutlineViewDelegate, NSOutl
...
+    let flags = event.modifierFlags.intersection(.deviceIndependentFlagsMask)
+    if flags == .command  {
       switch event.charactersIgnoringModifiers! {
-      case "f":
-        window!.makeFirstResponder(historySearchField)
-      case "a":
-        outlineView.selectAll(nil)
-      default:
-        break
+        case "f":
```
uiryuu: Let's try to keep the code consistency -- align `case` with `switch`

iina/HistoryWindowController.swift
```diff
@@ -95,22 +95,32 @@ class HistoryWindowController: NSWindowController, NSOutlineViewDelegate, NSOutl
...
+          window!.makeFirstResponder(historySearchField)
+        case "a":
+          outlineView.selectAll(nil)
+        default:
+          break
+      }
+    } else {
+      let key = KeyCodeHelper.mpvKeyCode(from: event)
```
uiryuu, line 119: Using `mpvKeyCode` here is a little weird to me, although its not unacceptable, but don't we have other elegant way to do this?

uiryuu (changes requested):

iina/HistoryWindowController.swift
svobs: Oh I didn't even notice this! Thanks.

iina/HistoryWindowController.swift
svobs: Well - I don't think so....It looks like MacOS's original design was never made more friendly in this regard, so I think we either have to use hexcodes or use our own constants. I was following the pattern set [in the welcome window.](https://github.com/iina/iina/pull/3739#issuecomment-1139288915).

uiryuu: Maybe later on we can refactor this part, maybe to abstract another helper function...

Issue #4262, opened 2023-03-11
Order of filter parameters is random


low-batt:
**System and IINA version:**

- macOS 13.2.1
- IINA 1.3.1

**Expected behavior:**
The order of UI elements representing filter parameters is consistent.

**Actual behavior:**
The order of UI elements representing filter parameters is random as seen in these two screen shots:
![param-order-1](https://user-images.githubusercontent.com/86170219/224511921-996f56e9-20b7-4722-8e05-f313a975b657.jpg)

![param-order-2](https://user-images.githubusercontent.com/86170219/224512044-1ccb8130-4e2e-4dea-9006-384a7e3ea830.jpg)

**Steps to reproduce:**
- Play a video
- Click `Video filters…` under the `Video` menu
- In the `Video Filters` window click `+`
- In the pop-up click `Blur`
- Note the order of the UI controls representing filter parameters
- Restart IINA
- Repeat
- Do this enough times and the postion of the parameters will change

- [x] MPV does not have this problem.

This is an IINA issue.

**How often does this happen?**
You many have to run IINA multiple times to see the parameters change position.

low-batt:
The fix for this issue has been merged into the IINA `develop` branch. GitHub [automatically closed](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) the linked issue in reaction to the merge. I am reopening this issue until the fix is available in an official release of IINA so that users intending to report this problem can easily locate this existing report. Once the fix for this issue has been included in an official release this issue will be closed.

Should you wish to confirm the fix is working and do not have the ability to build IINA from the sources in the `develop` branch, at some point a new _unsigned_ [nightly build](https://nightly.iina.io/) will be available for download. If you encounter a defect in a nightly build that has not been reported please enter an issue for it and be sure to indicate the problem was found in a nightly build.

The ability to disable this GitHub automatic closing action has been requested by GitHub users (discussions [#23476](https://github.com/orgs/community/discussions/23476) and [#17308](https://github.com/orgs/community/discussions/17308) for example), but so far GitHub has not implemented a way to control this behavior.

low-batt:
This is fixed in the released version of IINA.

PR #4263, opened 2023-03-11
Fix order of filter parameters is random, #4262

low-batt:
This commit will:
- Add a new struct FilterParameters that holds parameters in a KeyValuePairs object
- Change FilterPreset to use the new struct to preserve the order of parameters specified in the filter definition

This insures the order of the UI controls displayed to the user representing the filter parameters remains consistent.

- [x] This implements/fixes issue #4262.

---

**Description:**

### DIFF
--- a/iina/FilterPresets.swift
+++ b/iina/FilterPresets.swift
@@ -21,7 +21,7 @@ class FilterPreset {
   }
 
   var name: String
-  var params: [String: FilterParameter]
+  var params: FilterParameters
   var paramOrder: [String]?
   /** Given an instance, create the corresponding `MPVFilter`. */
   var transformer: Transformer
@@ -31,11 +31,11 @@ class FilterPreset {
   }
 
   init(_ name: String,
-       params: [String: FilterParameter],
+       params: KeyValuePairs<String, FilterParameter>,
        paramOrder: String? = nil,
        transformer: @escaping Transformer = FilterPreset.defaultTransformer) {
     self.name = name
-    self.params = params
+    self.params = FilterParameters(params)
     self.paramOrder = paramOrder?.components(separatedBy: ":")
     self.transformer = transformer
   }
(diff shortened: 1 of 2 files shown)


iina/FilterPresets.swift
```diff
@@ -115,6 +115,30 @@ class FilterParameter {
   }
 }
 
+/// An ordered collection of name to `FilterParameter` pairs.
+///
+/// This structure preserves the order of filter parameters specified in the `FilterPreset` definitions by using [KeyValuePairs](https://developer.apple.com/documentation/swift/keyvaluepairs)
+/// instead of a Swift dictionary to hold the name to `FilterParameter` mapping. The order of the parameters dictates the order
+/// of the UI controls displayed to the user. A dictionary **must not** be used as that would cause the UI controls representing the
+/// parameters to appear in random order.
+struct FilterParameters {
```
uiryuu, line 124: Why should we create a new struct for this? I think we can directly use `KeyValuePairs<String, FilterParameter>` for `FilterPreset.params`. For the `.get()` method, we can implement in the extension of `KeyValuePairs`

uiryuu (changes requested):

iina/FilterPresets.swift
low-batt: I was trying to make it really clear how filter parameters must be contained to avoid people using dictionaries. I'll go ahead and simplify and use `KeyValuePairs` inline.

low-batt: Not possible to implement the `get` method as an extension to `KeyValuePairs` because the `Key` generic parameter is not defined as being equatable. No matter. The approach I used is totally and completely wrong. I missed that there is already a `paramOrder`. But is it optional and missing from several presets. I'm closing this PR and opening a new one that uses the existing system for ordering parameters.

low-batt:
Replaced by PR #4407.

PR #4264, opened 2023-03-13
Migrate color prefs away from NSArchiver

svobs:

---

**Description:**

`NSArchiver` & `NSUnarchiver` are deprecated, and XCode is showing warnings. But currently IINA relies on them to serialize `NSColor` to binary format when storing the following prefs:

1. `subTextColor`
2. `subBgColor`
3. `subBorderColor`
4. `subShadowColor`

But if storing these in binary format was ever necessary, it's not now. There is already code to write the color into a string, in the format that mpv expects (search the existing code for `mpvColorString`). It wasn't too hard to change the code to store a `String` in the prefs. I did have to write a new `ValueTransformer` for it to make the XIB happy but it wasn't hard.

But changing the datatype of the 4 prefs is probably not the best idea, because doing that will cause legacy versions of IINA to show error messages when they try to read the prefs and don't get the type they were expecting.

So the code here will create 4 new pref entries with the same names but ending in `String`:
1. `subTextColorString`
2. `subBgColorString`
3. `subBorderColorString`
4. `subShadowColorString`

And all references in code are changed to use these values.

It also adds some logic at the start of IINA when the prefs are loaded:

- For the 4 prefs, it first checks to see if there is a new-style `String`-format entry.
- If `String` entry found, returns it.
- If `String` entry not found, it looks for an legacy-style `NSArchiver`-style entry.
  - If that is found, it infers that the user has just upgraded their version of IINA. So it:
    - Reads in the legacy entry,
    - Copies it into the new entry in the new `String` format
    - Returns it.
   
See `convertFromLegacyColor()` for more details.

This way, older versions of IINA will not be affected by the change to the new format, but users will be given a path away from `NSArchiver` without losing any data. Maybe after a couple versions of IINA (or when `NSArchiver` is killed by Apple), all references to the legacy entries can be dropped.  Unfortunately, until then, there will still be 1 `NSArchiver` warning (down from 4), but this seems unavoidable unless we are willing to drop the users' existing prefs for these 4 colors.

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -215,6 +215,9 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
     // register for url event
     NSAppleEventManager.shared().setEventHandler(self, andSelector: #selector(self.handleURLEvent(event:withReplyEvent:)), forEventClass: AEEventClass(kInternetEventClass), andEventID: AEEventID(kAEGetURL))
 
+    // Check for legacy pref entries and migrate them to their modern equivalents
+    LegacyMigration.shared.migrateLegacyPreferences()
+
     // guide window
     if FirstRunManager.isFirstRun(for: .init("firstLaunchAfter\(version)")) {
       guideWindow.show(pages: [.highlights])
--- a/iina/LegacyMigration.swift
+++ b/iina/LegacyMigration.swift
@@ -1,0 +1,81 @@
+//
+//  LegacyMigration.swift
+//  iina
+//
+//  Created by Matt Svoboda on 11/27/23.
+//  Copyright © 2023 lhc. All rights reserved.
+//
+
+import Foundation
+
+class LegacyMigration {
+
+  /// Map of [modern pref key → legacy key].
+  ///
+  /// - Important: Do not reference legacy keys outside this file.
+  fileprivate static let legacyColorPrefKeyMap: [Preference.Key: Preference.Key] = [
+    Preference.Key.subTextColorString: Preference.Key("subTextColor"),
+    Preference.Key.subBgColorString: Preference.Key("subBgColor"),
+    Preference.Key.subBorderColorString: Preference.Key("subBorderColor"),
+    Preference.Key.subShadowColorString: Preference.Key("subShadowColor"),
+  ]
+
+  /// The `LegacyMigration` singleton object.
+  static var shared = LegacyMigration()
+
+  /**
+   Loops over the set of legacy preference keys. If a value is found for a given legacy key, but no value is found for its modern equivalent key,
+   the legacy value is migrated & stored under the modern key.
+
+   Older versions of IINA serialized mpv color data into NSObject binary using the now-deprecated `NSUnarchiver` class.
+   This method will transition to the new format which consists of the color components written to a `String`.
+   To do this in a way which does not corrupt the values for older versions of IINA, we'll store the new format under a new `Preference.Key`,
+   and leave the legacy pref entry as-is.
+
+   This method will be executed on each of the affected prefs when IINA starts up. It will first check if there is already an entry for the new
+   pref key. If it finds one, then it will assume that the migration has already occurred, and will just return that.
+   Otherwise it will look for an entry for the legacy pref key. If it finds that, if will convert its value into the new format and store it under
+   the new pref key, and then return that.
+
+   This will have the effect of automatically migrating older versions of IINA into the new format with no loss of data.
+   However, it is worth noting that this migration will only happen once, and afterwards newer versions of IINA will not look at the old pref entry.
+   And since older versions of IINA will only use the old pref entry, users who mix old and new versions of IINA may experience different values
+   for these keys.
+   */
+  func migrateLegacyPreferences() {
+    Logger.log("Looking for legacy color prefs to migrate")
+
+    let appID = InfoDictionary.shared.bundleIdentifier
+    guard let persistedPrefKeys = UserDefaults.standard.persistentDomain(forName: appID)?.keys else {
+      Logger.log("Aborting legacy color prefs migration: failed to find prefs domain for \"\(appID)\"!", level: .error)
+      return
+    }
+
+    var unmigratedEntriesFoundCount: Int = 0
+    var entriesMigratedCount: Int = 0
+    for (modernKey, legacyKey) in LegacyMigration.legacyColorPrefKeyMap {
+      // Migrate pref only if there is a legacy entry with but no corresponding modern entry
+      guard persistedPrefKeys.contains(legacyKey.rawValue),
+            !persistedPrefKeys.contains(modernKey.rawValue) else { continue }
+      unmigratedEntriesFoundCount += 1
+
+      // Deserialize & convert legacy pref value to modern string format:
+      guard let legacyData = Preference.data(for: legacyKey) else { continue }
+      guard let color = NSUnarchiver.unarchiveObject(with: legacyData) as? NSColor,
+            let mpvColorString = color.usingColorSpace(.deviceRGB)?.mpvColorString else {
+        Logger.log("Failed to convert color value from legacy pref \(legacyKey.rawValue)", level: .error)
+        continue
+      }
+      // Store string under modern pref key:
+      Preference.set(mpvColorString, for: modernKey)
+      Logger.log("Converted color value from legacy pref \(legacyKey.rawValue) and stored in pref \(modernKey.rawValue)")
+      entriesMigratedCount += 1
+    }
+    if unmigratedEntriesFoundCount == 0 {
+      Logger.log("No unmigrated legacy color prefs found")
+    } else {
+      Logger.log("Migrated \(entriesMigratedCount) of \(unmigratedEntriesFoundCount) legacy color prefs")
+    }
+  }
+
+}
--- a/iina/Preference.swift
+++ b/iina/Preference.swift
@@ -180,16 +180,16 @@ struct Preference {
     static let subOverrideLevel = Key("subOverrideLevel")
     static let subTextFont = Key("subTextFont")
     static let subTextSize = Key("subTextSize")
-    static let subTextColor = Key("subTextColor")
-    static let subBgColor = Key("subBgColor")
+    static let subTextColorString = Key("subTextColorString")
+    static let subBgColorString = Key("subBgColorString")
     static let subBold = Key("subBold")
     static let subItalic = Key("subItalic")
     static let subBlur = Key("subBlur")
     static let subSpacing = Key("subSpacing")
     static let subBorderSize = Key("subBorderSize")
-    static let subBorderColor = Key("subBorderColor")
+    static let subBorderColorString = Key("subBorderColorString")
     static let subShadowSize = Key("subShadowSize")
-    static let subShadowColor = Key("subShadowColor")
+    static let subShadowColorString = Key("subShadowColorString")
     static let subAlignX = Key("subAlignX")
     static let subAlignY = Key("subAlignY")
     static let subMarginX = Key("subMarginX")
@@ -800,16 +800,16 @@ struct Preference {
     .subOverrideLevel: SubOverrideLevel.strip.rawValue,
     .subTextFont: "sans-serif",
     .subTextSize: Float(55),
-    .subTextColor: NSArchiver.archivedData(withRootObject: NSColor.white),
-    .subBgColor: NSArchiver.archivedData(withRootObject: NSColor.clear),
+    .subTextColorString: NSColor.white.usingColorSpace(.deviceRGB)!.mpvColorString,
+    .subBgColorString: NSColor.clear.usingColorSpace(.deviceRGB)!.mpvColorString,
     .subBold: false,
     .subItalic: false,
     .subBlur: Float(0),
     .subSpacing: Float(0),
     .subBorderSize: Float(3),
-    .subBorderColor: NSArchiver.archivedData(withRootObject: NSColor.black),
+    .subBorderColorString: NSColor.black.usingColorSpace(.deviceRGB)!.mpvColorString,
     .subShadowSize: Float(0),
-    .subShadowColor: NSArchiver.archivedData(withRootObject: NSColor.clear),
+    .subShadowColorString: NSColor.clear.usingColorSpace(.deviceRGB)!.mpvColorString,
     .subAlignX: SubAlign.center.rawValue,
     .subAlignY: SubAlign.bottom.rawValue,
     .subMarginX: Float(25),
@@ -932,10 +932,6 @@ struct Preference {
 
   static func value(for key: Key) -> Any? {
     return ud.value(forKey: key.rawValue)
-  }
-
-  static func mpvColor(for key: Key) -> String? {
-    return ud.mpvColor(forKey: key.rawValue)
   }
 
   static func set(_ value: Bool, for key: Key) {
--- a/iina.xcodeproj/project.pbxproj
+++ b/iina.xcodeproj/project.pbxproj
@@ -172,6 +172,7 @@
 		B4E446F725CB54EF0069F06E /* Mustache in Frameworks */ = {isa = PBXBuildFile; productRef = B4E446F625CB54EF0069F06E /* Mustache */; };
 		B4E4470125CE3F930069F06E /* Sparkle in Frameworks */ = {isa = PBXBuildFile; productRef = B4E4470025CE3F930069F06E /* Sparkle */; };
 		C789872F1E34EF170005769F /* InfoPlist.strings in Resources */ = {isa = PBXBuildFile; fileRef = C78987311E34EF170005769F /* InfoPlist.strings */; };
+		D1A4D98A2B1495270009AB4E /* LegacyMigration.swift in Sources */ = {isa = PBXBuildFile; fileRef = D1A4D9892B1495270009AB4E /* LegacyMigration.swift */; };
 		D1B4E24E2A3AFC9100E36F1D /* MiniPlayerWindow.swift in Sources */ = {isa = PBXBuildFile; fileRef = D1B4E24D2A3AFC9100E36F1D /* MiniPlayerWindow.swift */; };
 		E301EFDA21312AB300BC8588 /* KeychainAccess.swift in Sources */ = {isa = PBXBuildFile; fileRef = E301EFD921312AB300BC8588 /* KeychainAccess.swift */; };
 		E30D2EBD21F5FD2600E1FF0D /* PluginOverlayView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E30D2EBC21F5FD2600E1FF0D /* PluginOverlayView.swift */; };
@@ -1556,6 +1557,7 @@
 		C7DBA5EB1F07C5FD00C2B416 /* ko */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = ko; path = ko.lproj/InitialWindowController.strings; sourceTree = "<group>"; };
 		C7DC79CE1EC63821002DE23B /* ko */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = ko; path = ko.lproj/HistoryWindowController.strings; sourceTree = "<group>"; };
 		C7E90B522087EC5700A58B6B /* ko */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = ko; path = ko.lproj/SubChooseViewController.strings; sourceTree = "<group>"; };
+		D1A4D9892B1495270009AB4E /* LegacyMigration.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = LegacyMigration.swift; sourceTree = "<group>"; };
 		D1B4E24D2A3AFC9100E36F1D /* MiniPlayerWindow.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MiniPlayerWindow.swift; sourceTree = "<group>"; };
 		D27556FC1EC6E1C300CAB2A4 /* zh-Hant */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = "zh-Hant"; path = "zh-Hant.lproj/HistoryWindowController.strings"; sourceTree = "<group>"; };
 		D27E35CE1E379B6D0064BE57 /* zh-Hant */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = "zh-Hant"; path = "zh-Hant.lproj/MainMenu.strings"; sourceTree = "<group>"; };
@@ -2058,6 +2060,7 @@
 				E3513AFA20F120F600F8C347 /* PreferenceViewController.swift */,
 				E374160B20F138A900B4F7F9 /* CollapseView.swift */,
 				519872FE26879B9B00F84BCC /* AccessibilityPreferences.swift */,
+				D1A4D9892B1495270009AB4E /* LegacyMigration.swift */,
 			);
 			name = Preference;
 			sourceTree = "<group>";
@@ -2896,6 +2899,7 @@
 				846352581EEEE11A0043F0CC /* ThumbnailPeekView.swift in Sources */,
 				84E745D61DFDD4FD00588DED /* KeyCodeHelper.swift in Sources */,
 				840D479B1DFEF649000D9A64 /* KeyRecordViewController.swift in Sources */,
+				D1A4D98A2B1495270009AB4E /* LegacyMigration.swift in Sources */,
 				E3965E1A2487C99900607EB4 /* JavascriptPluginSubtitle.swift in Sources */,
 				84F5D4951E44D5230060A838 /* KeyBindingCriterion.swift in Sources */,
 				846121BD1F35FCA500ABB39C /* DraggingDetect.swift in Sources */,
--- a/iina/Base.lproj/PrefSubViewController.xib
+++ b/iina/Base.lproj/PrefSubViewController.xib
@@ -509,9 +509,9 @@
                                 </constraints>
                                 <color key="color" red="0.05813049898" green="0.055541899059999997" blue="1" alpha="1" colorSpace="calibratedRGB"/>
                                 <connections>
-                                    <binding destination="5Up-Ab-aAm" name="value" keyPath="values.subTextColor" id="57K-5E-1e6">
+                                    <binding destination="5Up-Ab-aAm" name="value" keyPath="values.subTextColorString" id="add-O6-WUQ">
                                         <dictionary key="options">
-                                            <string key="NSValueTransformerName">NSUnarchiveFromData</string>
+                                            <string key="NSValueTransformerName">MPVColorStringTransformer</string>
                                         </dictionary>
                                     </binding>
                                 </connections>
@@ -538,9 +538,9 @@
                                 <rect key="frame" x="189" y="8" width="44" height="23"/>
                                 <color key="color" red="0.05813049898" green="0.055541899059999997" blue="1" alpha="1" colorSpace="calibratedRGB"/>
                                 <connections>
-                                    <binding destination="5Up-Ab-aAm" name="value" keyPath="values.subBgColor" id="BDd-oz-ARs">
+                                    <binding destination="5Up-Ab-aAm" name="value" keyPath="values.subBgColorString" id="2V9-yF-uLN">
                                         <dictionary key="options">
-                                            <string key="NSValueTransformerName">NSUnarchiveFromData</string>
+                                            <string key="NSValueTransformerName">MPVColorStringTransformer</string>
                                         </dictionary>
                                     </binding>
                                 </connections>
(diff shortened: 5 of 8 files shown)


low-batt:
I have labeled this a high priority because this is not merely a problem with using a deprecated class. The [NSUnarchiver](https://developer.apple.com/documentation/foundation/nsunarchiver) does not provide the methods required to support [NSSecureCoding](https://developer.apple.com/documentation/foundation/nssecurecoding). That requires methods such as [unarchivedObject(ofClass:from:)](https://developer.apple.com/documentation/foundation/nskeyedunarchiver/2983380-unarchivedobject) provided by [NSKeyedUnarchiver](https://developer.apple.com/documentation/foundation/nskeyedunarchiver) where the expected class of the root object can be specified. This is about resisting object substitution attacks. Thus removing use of [NSArchiver](https://developer.apple.com/documentation/foundation/nsarchiver) is important from a security perspective. Switching to the `mpv` string based format for color (specified [here](https://mpv.io/manual/stable/#options-sub-color)) as the changes in the PR does avoids the possibility of object substitution.

I used this command to check the saved settings:
```text
defaults read com.colliderli.iina | grep 'sub.*Color.*'
```

I used these commands to force migration to repeat:
```text
defaults delete com.colliderli.iina subBgColorString
defaults delete com.colliderli.iina subBorderColorString
defaults delete com.colliderli.iina subShadowColorString
defaults delete com.colliderli.iina subTextColorString
```

iina/Preference.swift
```diff
@@ -755,16 +763,16 @@ struct Preference {
     .subOverrideLevel: SubOverrideLevel.strip.rawValue,
     .subTextFont: "sans-serif",
     .subTextSize: Float(55),
-    .subTextColor: NSArchiver.archivedData(withRootObject: NSColor.white),
-    .subBgColor: NSArchiver.archivedData(withRootObject: NSColor.clear),
+    .subTextColorString: convertFromLegacyColor(.subTextColorString) ?? NSColor.white.usingColorSpace(.deviceRGB)!.mpvColorString,
```
low-batt: When I initially read this code it confused me. This is the `defaultPreference` property. It is supposed to contain the default value for this setting. Why would that be anything other than white in `mpv` color string form? This property is used by `AppDelegate` in a call to [register(defaults:)](https://developer.apple.com/documentation/foundation/userdefaults/1417065-register). This is establishing the _default_ values for the setting. But `convertFromLegacyColor` may return a value extracted from the original setting. Confusing. I don't think this causes a problem, but it does not match up with how [UserDefaults](https://developer.apple.com/documentation/foundation/userdefaults) should be used.

Another concern I have about this change has to do with layering and dependencies. `Preference` is one of the lowest level classes. It is even lower than `Logger` as that class must check IINA's settings to see if advanced preferences and logging are enabled. Previously there was no logging in  `Preference`. These changes add logging so now they depend upon each other. Because logging is used while forming the value for `defaultPreference` it means `Logger` is calling into `Preference` before default values have been established as can be seen in the debugger:

<details>
<summary>Stack Trace:</summary>

```text
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 3.1
  * frame #0: 0x000000010246edcc IINA`static Preference.bool(key=(rawValue = "enableAdvancedSettings"), self=IINA.Preference) at Preference.swift:879:12
    frame #1: 0x00000001026058f4 IINA`one-time initialization function for enabled at Logger.swift:56:35
    frame #2: 0x000000010653699c libdispatch.dylib`_dispatch_client_callout + 20
    frame #3: 0x00000001065388bc libdispatch.dylib`_dispatch_once_callout + 160
    frame #4: 0x00000001026059c8 IINA`Logger.enabled.unsafeMutableAddressor at Logger.swift:56:14
    frame #5: 0x00000001026085c4 IINA`static Logger.log(message="Converted color value from legacyKey subTextColor and stored in key subTextColorString", level=debug, subsystem=(rawValue = "iina"), appendNewlineAtTheEnd=true, self=IINA.Logger) at Logger.swift:154:11
    frame #6: 0x000000010246e08c IINA`static Preference.convertFromLegacyColor(key=(rawValue = "subTextColorString"), self=IINA.Preference) at Preference.swift:960:12
    frame #7: 0x000000010246be3c IINA`one-time initialization function for defaultPreference at Preference.swift:766:26
    frame #8: 0x000000010653699c libdispatch.dylib`_dispatch_client_callout + 20
    frame #9: 0x00000001065388bc libdispatch.dylib`_dispatch_once_callout + 160
    frame #10: 0x000000010246e334 IINA`Preference.defaultPreference.unsafeMutableAddressor at Preference.swift:679:14
    frame #11: 0x0000000102650fb8 IINA`AppDelegate.registerUserDefaultValues(self=0x00006000010f04d0) at AppDelegate.swift:755:93
    frame #12: 0x00000001026441e8 IINA`AppDelegate.applicationWillFinishLaunching(notification=Foundation.Notification @ 0x000000016d9d2220, self=0x00006000010f04d0) at AppDelegate.swift:136:5
    frame #13: 0x00000001026464e0 IINA`@objc AppDelegate.applicationWillFinishLaunching(_:) at <compiler-generated>:0
    frame #14: 0x00000001968f30f0 CoreFoundation`__CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 148
    frame #15: 0x000000019698ee7c CoreFoundation`___CFXRegistrationPost_block_invoke + 88
    frame #16: 0x000000019698edc4 CoreFoundation`_CFXRegistrationPost + 440
    frame #17: 0x00000001968c443c CoreFoundation`_CFXNotificationPost + 704
    frame #18: 0x00000001978228a4 Foundation`-[NSNotificationCenter postNotificationName:object:userInfo:] + 88
    frame #19: 0x0000000199b171bc AppKit`-[NSApplication finishLaunching] + 248
    frame #20: 0x0000000199b16ea8 AppKit`-[NSApplication run] + 252
    frame #21: 0x0000000199aee3cc AppKit`NSApplicationMain + 880
    frame #22: 0x00000001026540a0 IINA`main at AppDelegate.swift:24:7
    frame #23: 0x00000001964c7f28 dyld`start + 2236
(lldb)
```
</details>

Another side effect of this is that these new log messages are emitted before `AppDelegate` has had a chance to emit the log file header messages:
```text
14:06:34.834 [iina][d] Converted color value from legacyKey subTextColor and stored in key subTextColorString
14:06:34.836 [iina][d] Converted color value from legacyKey subBgColor and stored in key subBgColorString
14:06:34.837 [iina][d] Converted color value from legacyKey subBorderColor and stored in key subBorderColorString
14:06:34.840 [iina][d] IINA 1.3.1 Build 133
14:06:34.840 [iina][d] Copyright © 2017-2022
14:06:34.840 [iina][d] Collider LI, et al.
14:06:34.840 [iina][d] Released under GPLv3.
```

I'm thinking the calls to `convertFromLegacyColor` should be removed from the initialization of `defaultPreference` and into a `migrateSettings` function that `AppDelegate` would then call in `applicationWillFinishLaunching` right before emitting the "App will launch" log message.

iina/Preference.swift
```diff
@@ -923,4 +927,38 @@ struct Preference {
...
+   This method will transition to the new format which consists of the color components written to a `String`.
+   To do this in a way which does not corrupt the values for older versions of IINA, we'll store the new format under a new `Preference.Key`,
+   and leave the legacy pref entry as-is.
+
+   This method will be executed on each of the affected prefs when IINA starts up. It will first check if there is already an entry for the new
+   pref key. If it finds one, then it will assume that the migration has already occurred, and will just return that.
+   Otherwise it will look for an entry for the legacy pref key. If it finds that, if will convert its value into the new format and store it under
+   the new pref key, and then return that.
+
+   This wil have the effect of automatically migrating older versions of IINA into the new format with no loss of data.
```
low-batt: Typo "wil"

iina/Preference.swift
```diff
@@ -923,4 +927,38 @@ struct Preference {
...
+   However, it is worth noting that this migration will only happen once, and afterwards newer versions of IINA will not look at the old pref entry.
+   And since older verions of IINA will only use the old pref entry, users who mix old and new versions of IINA may experience different values
```
low-batt: Typo "verions"

iina/Preference.swift
```diff
@@ -923,4 +927,38 @@ struct Preference {
     return T.init(key: key) ?? T.defaultValue
   }
 
+  /**
+   Older versions of IINA converted mpv color data into NSObject binary using the now-deprecated `NSUnarchiver` class.
```
low-batt: The documentation convention is to start with a very short description of the function, followed by a blank line and then go into further discussion about the details. You can see that in Xcode if you bring up the quick help for this function. You will see that Xcode puts this entire paragraph at the top and the other paragraphs later on under a `Discussion` header.

iina/Preference.swift
```diff
@@ -923,4 +927,38 @@ struct Preference {
...
+   pref key. If it finds one, then it will assume that the migration has already occurred, and will just return that.
+   Otherwise it will look for an entry for the legacy pref key. If it finds that, if will convert its value into the new format and store it under
+   the new pref key, and then return that.
+
+   This wil have the effect of automatically migrating older versions of IINA into the new format with no loss of data.
+   However, it is worth noting that this migration will only happen once, and afterwards newer versions of IINA will not look at the old pref entry.
+   And since older verions of IINA will only use the old pref entry, users who mix old and new versions of IINA may experience different values
+   for these keys.
+   */
+  private static func convertFromLegacyColor(_ key: Key) -> String? {
```
low-batt: Normally "convert" methods don't have side effects. I'm thinking "migrate" is a better term for what this function does.

iina/Preference.swift
```diff
@@ -923,4 +927,38 @@ struct Preference {
...
+    // Look for legacy pref:
+    guard let data = ud.data(forKey: legacyKey.rawValue) else { return nil }
+    guard let color = NSUnarchiver.unarchiveObject(with: data) as? NSColor else { return nil }
+    guard let mpvColorString = color.usingColorSpace(.deviceRGB)?.mpvColorString else {
+      Logger.log("Failed to convert color value from legacy key \(legacyKey.rawValue)", level: .error)
+      return nil
+    }
+    // Store converted value in current format for future use:
+    set(mpvColorString, for: key)
+    Logger.log("Converted color value from legacyKey \(legacyKey.rawValue) and stored in key \(key.rawValue)")
```
low-batt: Same issue with the word "Converted". I think users would understand what this method is doing if the log message said "Migrated".

low-batt: Might be better to say "legacy key" instead of using the variable name.

svobs:
Some good points. I tweaked the comments and fixed the typos, and refactored to move the code out of Preference.swift.

I don't remember the exact sequence of events which caused me to call the migration code while setting the defaults, but I agree that it's not a good design (I was probably just being hasty/sloppy). But I found I still have to call the migration code before the defaults have been set, because otherwise I don't have a reliable way to distinguish between the default values and genuine stored user data. Or at least I haven't found a way to do that. Let me know if you have any suggestions, but my workaround for dealing with any logging chicken & egg issues was to just remove the logging from the migration code.

low-batt:
I'm liking the new changes. Because migration of user data is a critical operation that must be solid, I am going to provide a super picky review. Please push back if you disagree with my feedback. On some of this we might want to wait and see what the other reviewers think before making changes. I will post feedback shortly.

I used this command to check the saved settings:
```text
defaults read com.colliderli.iina | grep 'sub.*Color.*\|didMigrateColorFromNSArchiver'
```

I used these commands to force migration to repeat:
```text
defaults delete com.colliderli.iina subBgColorString
defaults delete com.colliderli.iina subBorderColorString
defaults delete com.colliderli.iina subShadowColorString
defaults delete com.colliderli.iina subTextColorString
defaults delete com.colliderli.iina didMigrateColorFromNSArchiver
```

The log messages emitted at startup:
```text
18:46:01.301 [iina][d] App will launch
18:46:01.301 [iina][d] Looking for legacy color pref entries to migrate
18:46:01.302 [iina][d] Converted color value from legacy key subBgColor and stored in key subBgColorString
18:46:01.303 [iina][d] Converted color value from legacy key subBorderColor and stored in key subBorderColorString
18:46:01.304 [iina][d] Converted color value from legacy key subTextColor and stored in key subTextColorString
18:46:01.304 [iina][d] Migrated 3 of 3 legacy color pref entries
```

iina/AppDelegate.swift
```diff
@@ -198,6 +198,9 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
     // register for url event
     NSAppleEventManager.shared().setEventHandler(self, andSelector: #selector(self.handleURLEvent(event:withReplyEvent:)), forEventClass: AEEventClass(kInternetEventClass), andEventID: AEEventID(kAEGetURL))
 
+    // Check for legacy pref entries and migrate them to their modern equivalents
```
low-batt, line 218: Following Apple's renaming of preferences:
// Check for legacy settings and migrate them to their modern equivalents.

iina/LegacyMigration.swift
```diff
@@ -0,0 +1,79 @@
...
+//  LegacyMigration.swift
+//  iina
+//
+//  Created by Matt Svoboda on 11/27/23.
+//  Copyright © 2023 lhc. All rights reserved.
+//
+
+import Foundation
+
+class LegacyMigration {
```
low-batt, line 11: In Xcode this file was added under `Utils`. I'm thinking it should be under `Preference`.

iina/LegacyMigration.swift
```diff
@@ -0,0 +1,79 @@
...
+   Otherwise it will look for an entry for the legacy pref key. If it finds that, if will convert its value into the new format and store it under
+   the new pref key, and then return that.
+
+   This will have the effect of automatically migrating older versions of IINA into the new format with no loss of data.
+   However, it is worth noting that this migration will only happen once, and afterwards newer versions of IINA will not look at the old pref entry.
+   And since older versions of IINA will only use the old pref entry, users who mix old and new versions of IINA may experience different values
+   for these keys.
+   */
+  static func migrateLegacyPreferences() {
+    let didMigrateColor = Preference.bool(for: .didMigrateColorFromNSArchiver)
```
low-batt: When I first thought about how to do migration I thought about adding something to the settings to indicate whether settings had been migrated or not. I didn't like that because it seemed like an abuse of the intention of UserDefaults. If we are going to do this, should it be for this specific migration, or should it represent a "settings version" that we set to 2?

One change in behavior is that with the old proposed solution the migration could happen a second time if a setting was not set the first time migration occurred. If then the user ran the old version of IINA again and changed that particular color from the default, the next time the new IINA is run that setting would get migrated. This could be viewed as being better or confusing.

I think on this one we need to hear the opinion of other reviewers. At the moment I'm thinking this is fine as is.

iina/LegacyMigration.swift
```diff
@@ -0,0 +1,79 @@
...
+    guard !didMigrateColor else {
+      Logger.log("Already migrated color prefs away from NSArchiver; no migration needed", level: .verbose)
```
low-batt:
```suggestion
      Logger.log("Already migrated color settings away from NSArchiver; no migration needed", level: .verbose)
```

iina/LegacyMigration.swift
```diff
@@ -0,0 +1,79 @@
...
+//
+//  Created by Matt Svoboda on 11/27/23.
+//  Copyright © 2023 lhc. All rights reserved.
+//
+
+import Foundation
+
+class LegacyMigration {
+
+  // Map of [modern pref key → legacy key]. Do not reference legacy keys outside this file.
```
low-batt:
```suggestion
  /// Map of [modern settings key → legacy key].
  ///
  /// - Important: Do not reference legacy keys outside this file.
```

iina/LegacyMigration.swift
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   Loops over the set of legacy preference keys. If a value is found for a given legacy key, but no value is found for its modern equivalent key,
+   the legacy value is migrated & stored under the modern key.
+
+   Older versions of IINA serialized mpv color data into NSObject binary using the now-deprecated `NSUnarchiver` class.
+   This method will transition to the new format which consists of the color components written to a `String`.
+   To do this in a way which does not corrupt the values for older versions of IINA, we'll store the new format under a new `Preference.Key`,
+   and leave the legacy pref entry as-is.
+
+   This method will be executed on each of the affected prefs when IINA starts up. It will first check if there is already an entry for the new
```
low-batt, line 35: This reflects the original PR code and does not match what the code is doing now.

iina/LegacyMigration.swift
```diff
@@ -0,0 +1,79 @@
...
+   pref key. If it finds one, then it will assume that the migration has already occurred, and will just return that.
+   Otherwise it will look for an entry for the legacy pref key. If it finds that, if will convert its value into the new format and store it under
+   the new pref key, and then return that.
+
+   This will have the effect of automatically migrating older versions of IINA into the new format with no loss of data.
+   However, it is worth noting that this migration will only happen once, and afterwards newer versions of IINA will not look at the old pref entry.
+   And since older versions of IINA will only use the old pref entry, users who mix old and new versions of IINA may experience different values
+   for these keys.
+   */
+  static func migrateLegacyPreferences() {
```
low-batt: migrateLegacySettings following Apple's new naming scheme.

iina/LegacyMigration.swift
```diff
@@ -0,0 +1,79 @@
...
+    let didMigrateColor = Preference.bool(for: .didMigrateColorFromNSArchiver)
+    guard !didMigrateColor else {
+      Logger.log("Already migrated color prefs away from NSArchiver; no migration needed", level: .verbose)
+      return
+    }
+
+    Logger.log("Looking for legacy color pref entries to migrate")
```
low-batt:
```suggestion
    Logger.log("Looking for legacy color settings to migrate")
```

iina/LegacyMigration.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    var entriesFoundCount: Int = 0
+    var entriesMigratedCount: Int = 0
+    for (modernKey, legacyKey) in legacyColorPrefKeyMap {
+      // Look for legacy pref:
+      guard let data = UserDefaults.standard.data(forKey: legacyKey.rawValue) else { continue }
```
low-batt:
```suggestion
      guard let data = Preference.data(for: legacyKey) else { continue }
```

iina/LegacyMigration.swift
```diff
@@ -0,0 +1,79 @@
...
+        continue
+      }
+
+      guard let color = NSUnarchiver.unarchiveObject(with: data) as? NSColor,
+            let mpvColorString = color.usingColorSpace(.deviceRGB)?.mpvColorString else {
+        Logger.log("Failed to convert color value from legacy key \(legacyKey.rawValue)", level: .error)
+        return  // Data error == fail migration
+      }
+      // Store migrated value in modern string format under modern pref key:
+      UserDefaults.standard.set(mpvColorString, forKey: modernKey.rawValue)
```
low-batt:
```suggestion
      Preference.set(mpvColorString, for: modernKey)
```

iina/LegacyMigration.swift
```diff
@@ -0,0 +1,79 @@
...
+      Logger.log("Converted color value from legacy key \(legacyKey.rawValue) and stored in key \(modernKey.rawValue)")
+      entriesMigratedCount += 1
+    }
+    Logger.log("Migrated \(entriesMigratedCount) of \(entriesFoundCount) legacy color pref entries")
```
low-batt:
```suggestion
    Logger.log("Migrated \(entriesMigratedCount) of \(entriesFoundCount) legacy color settings")
```

iina/LegacyMigration.swift
```diff
@@ -0,0 +1,79 @@
...
+    for (modernKey, legacyKey) in legacyColorPrefKeyMap {
+      // Look for legacy pref:
+      guard let data = UserDefaults.standard.data(forKey: legacyKey.rawValue) else { continue }
+      entriesFoundCount += 1
+
+      // To be super safe, check the value for the modern pref key.
+      // A pref which has not been set looks identical to the default value.
+      let defaultValue = Preference.defaultPreference[modernKey] as? String
+      let currentValue = Preference.value(for: modernKey) as? String
+      guard currentValue == nil || currentValue == defaultValue else {
```
low-batt:
```suggestion
      let currentValue = Preference.string(for: modernKey)!
      guard currentValue == defaultValue else {
```

low-batt (commented):
Looking good. Some minor tweaks.

iina/AppDelegate.swift
svobs: I understand the need to use "Setting" for user-facing text, but I think it's better to continue using "Preference" internally. Mainly to keep consistency, because Preference is already being used to refer to UserDefaults entries. But I also personally like using it because "Setting" is such an overloaded term that it's not immediately clear what it's referring to without more context. A web developer could refer to cookies, local storage, or server state as "settings". And often people use the word "setting" to refer to the _UI of the setting_ instead, like with the Settings sidebar.

iina/LegacyMigration.swift
svobs: Good point, will change this.

iina/LegacyMigration.swift
svobs: > If we are going to do this, should it be for this specific migration, or should it represent a "settings version" that we set to 2?

I too have an aversion to littering the saved state with bits like this. So much so that I didn't even consider it at first. But I don't think there's a clearly better way to record the state.

Using an incrementing version number would be a better solution if IINA was more of a closed-source / centralized project, but I think it might be limiting when considering IINA open-source nature which (at least in principle) can include multiple forks with different nonlinear offshoots which could get merged back and forth. In my mind I see parallels with how older SCM solutions like SVN used incrementing revisions, which became painful with decentralized projects. Then Git came along with its use of hashes, which turned out to be much more flexible.

There is a another possible way to go about this which I almost hesitate to mention because it would mean another round of rewrites to the PR. Instead of doing a one-time migration with a flag which gets set, we could do a two-phased migration. In the next upcoming version of IINA (and maybe some versions after, until end-of-life of `NSArchiver`) strives to keep both legacy & modern values in sync by (1) always writing to both the legacy entry and the modern entry when the user changes data, and (2) checking at startup to see if any pair of legacy-modern entries disagrees, and if so, overwrites the modern with the legacy. This would cover the situation where the user is using an older version of IINA to update their color settings, and it would always honor their most recent change. So this system of dual-entry writes would be the first phase of the migration. The second phase would occur when `NSArchiver` can no longer be used and thus the legacy entries can no longer be read, in which case that future version of IINA would read & write only to the modern entries.

But to implement that system I'd have to see how to write to both entries using Interface Builder or come up with some workaround in code. And tbh it seems like most projects don't put much effort into migrating legacy settings properly - I use Linux sometimes and I regularly see settings get reset back to defaults when upgrading, or buggy behavior that only gets solved by deleting a config file. MacOS has a poor track record with this too, and if you ever upgraded the OS and found that something stopped working properly but going into System Settings and toggling the relevant thing off and on again fixes it, you've seen a corrupted preference getting fixed. But just putting this all on the table.

iina/AppDelegate.swift
low-batt: Sounds fine. I was rather dismayed by Apple changing this term. Imagine how much programmer time was spent on this little tweak when you add up all the time to change this in all the affected apps.

iina/LegacyMigration.swift
low-batt: I have a strong opinion on keeping the settings in sync. Normally it would be good to continue support for the old setting, but not in this case due to security. Due to the security issue with `NSArchiver` we want to remove use of the class as soon as practical. So I do not think we should add code to keep the old key in sync. We'd like to eliminate the migration code sooner than later to eliminate all use of `NSArchiver`.

Why not go back to the original scheme where migration occurs when a non-default stored setting for the old key is found and the new key does not have a persisted value? With that scheme no "migrated" flag is needed.

Might want to wait and see what the other reviewers think about this.

iina/LegacyMigration.swift
low-batt: This was not updated and still reflects the original code which did not use `didMigrateColorFromNSArchiver`. Of course I'm thinking maybe we should go back to that method. If we aren't doing that and sticking with keying off `didMigrateColorFromNSArchiver` then this description of what the method does needs to be updated:
```text
   This method will be executed on each of the affected prefs when IINA starts up. It will first check if there is already an entry for the new
   pref key. If it finds one, then it will assume that the migration has already occurred, and will just return that.
   Otherwise it will look for an entry for the legacy pref key. If it finds that, if will convert its value into the new format and store it under
   the new pref key, and then return that.

File: iina/LegacyMigration.swift (hunk 9 above)
Comment (svobs) [reply iina/LegacyMigration.swift]: > Why not go back to the original scheme where migration occurs when a non-default stored setting for the old key is found and the new key does not have a persisted value? With that scheme no "migrated" flag is needed.

The only problem with the previous solution is that the check needs happen before the default values are set, and thus logging is not available. I wouldn't want to wait until after the defaults are set because that would introduce a blind spot for the default value: the user may end up in a state where they want to use the default value, but the migration code will see that as a non-migration and will see their legacy value and will keep re-migrating it every time they launch.

I think the migration flag is a decent solution. The flag can double as a "receipt" for the migration, and while receipts can be annoying, they can sometimes turn out to be pretty useful.

Comment (low-batt) [reply iina/LegacyMigration.swift]: On this:
> The only problem with the previous solution is that the check needs happen before the default values are set

That is only required when detecting if a key is persisted based on whether a value is present. Add the following to `Preference` and have `LegacyMigration` use it to check if the modern key is present in the property list.

```swift
  static func keyHasBeenPersisted(key: Key) -> Bool {
    let identifier = InfoDictionary.shared.bundleIdentifier
    guard let persisted = ud.persistentDomain(forName: identifier) else { return false }
    return persisted.keys.contains(key.rawValue)
  }
```

svobs:
> That is only required when detecting if a key is persisted based on whether a value is present. Add the following to Preference and have LegacyMigration use it to check if the modern key is present in the property list.

Great tip - thanks. I went ahead and implemented this and removed the `didMigrateColorFromNSArchiver` flag, and it's much cleaner.

iina/LegacyMigration.swift
```diff
@@ -0,0 +1,72 @@
...
+   Otherwise it will look for an entry for the legacy pref key. If it finds that, if will convert its value into the new format and store it under
+   the new pref key, and then return that.
+
+   This will have the effect of automatically migrating older versions of IINA into the new format with no loss of data.
+   However, it is worth noting that this migration will only happen once, and afterwards newer versions of IINA will not look at the old pref entry.
+   And since older versions of IINA will only use the old pref entry, users who mix old and new versions of IINA may experience different values
+   for these keys.
+   */
+  static func migrateLegacyPreferences() {
+    Logger.log("Looking for legacy color prefs to migrate", level: .verbose)
```
uiryuu: I think we should make this the same log level as other logs, there's no reason for only setting this verbose.

iina/LegacyMigration.swift
```diff
@@ -0,0 +1,72 @@
...
+      // Deserialize & convert legacy pref value to modern string format:
+      guard let legacyData = Preference.data(for: legacyKey) else { continue }
+      guard let color = NSUnarchiver.unarchiveObject(with: legacyData) as? NSColor,
+            let mpvColorString = color.usingColorSpace(.deviceRGB)?.mpvColorString else {
+        Logger.log("Failed to convert color value from legacy pref \(legacyKey.rawValue)", level: .error)
+        continue
+      }
+      // Store string under modern pref key:
+      Preference.set(mpvColorString, for: modernKey)
+      Logger.log("Converted color value from legacy pref \(legacyKey.rawValue) and stored in pref \(modernKey.rawValue)")
```
uiryuu, line 71: Should we remove the legacy entry after migration?

iina/Preference.swift
```diff
@@ -959,4 +955,9 @@ struct Preference {
     return T.init(key: key) ?? T.defaultValue
   }
 
+  static func keyHasBeenPersisted(_ key: Key) -> Bool {
+    let identifier = InfoDictionary.shared.bundleIdentifier
+    guard let persisted = ud.persistentDomain(forName: identifier) else { return false }
```
uiryuu: I suggest make it to be
``` swift
static let persisted = ud.persistentDomain(forName: identifier)
guard let _ = persisted else { return false }
```
to prevent retrieving the whole UserDefault multiple times.

uiryuu (changes requested):
Generally looks good, here're some suggestions

iina/LegacyMigration.swift
low-batt: Not now. Especially not in a beta. With a beta release it is much more likely users will go back and forth between releases.

iina/Preference.swift
lhc70000: If so, `persisted` will only be updated when IINA starts. If someone find this function in the future and try to use it, they may encounter problems if they updated the preference when IINA is running. I actually suggest moving this part to `LegacyMigration` where it's being used, as an inner function or block.

svobs:
Rebased and refactored based on above feedback.

uiryuu (approved):
Changes look good to me

PR #4271, opened 2023-03-16
Fix failure to write '#' to key bindings file

### LINKED ISSUE #4267: Key Binding to "#" (Shift+3) fail upon IINA restarts
- macOS Monterey 12.6.1
- IINA v1.3.1

**Expected behavior:**
Keybinds to "#" (+ any combination of modifiers) for any valid function, inserted either via IINA's Key Binding GUI or via input.conf editing, should properly save and thereby continue to function after restarting IINA.

**Actual behavior:**
Functions keybind to "#" fail to work after IINA is restarted. 
Keybinds for "#" are removed from IINA's Key Binding GUI list.

<details>

```
```
</details>

<br>

<details>

```
```
</details>

<br>

**Steps to reproduce:**
(Example)
Open Preferences, goto "Key Binding"
With a custom key bind set/file, use the "+" button to create a new Key Bind.
Record inputs "Shift+3"; "#" should appear in the recording.
Select action "Cycle", "Pause"

Test keybind with content.

Quit IINA.

Re-open IINA, resume playback.

Test "#" keybind again. Failure.
Open Preferences, goto "Key Binding"
Notice previously created keybind is gone.

Note, MPV documentation for input.conf instructs users to use "SHARP" when creating keybinds for "#". Editing IINA's input.conf, attempting to use "SHARP" to create keybinds also fails.

Every time, highly reproducable.

**Thoughts**
It's interesting that when creating Key Bindings to "#" using the GUI, they do immediately work. Even checking the corresponding input.conf, a seemingly valid line for the newly created key bind does show up; the above example creates "# cycle pause". However, I do think there is an underlying issue parsing "#".

I suggest implementing MPV's workaround, and using "SHARP" to create robust input.conf files.

svobs:
- [X] This implements/fixes issue #4267.

---

**Description:**

- `#` and ` ` could previously be entered into the Key Bindings editor by the user, but they would get written out to the `.conf` file as-is, so when they were read in again on the next launch, they would be ignored. Even more confusing for the user, the values in memory would still work, but only until the next launch.
- Since there is no other legitimate reason for a user to enter these into the editor, this patch alters the logic so that `#` is written out as `SHARP` and ` ` is written out as `SPACE`.

### DIFF
--- a/iina/KeyCodeHelper.swift
+++ b/iina/KeyCodeHelper.swift
@@ -17,11 +17,11 @@ class KeyCodeHelper {
   static let SHIFT_KEY = "Shift"
   static let META_KEY = "Meta"
 
-  fileprivate static let modifierOrder: [String: Int] = [
-    CTRL_KEY: 0,
-    ALT_KEY: 1,
-    SHIFT_KEY: 2,
-    META_KEY: 3
+  fileprivate static let modifiersInOrder: [String] = [
+    CTRL_KEY,
+    ALT_KEY,
+    SHIFT_KEY,
+    META_KEY,
   ]
 
   fileprivate static let modifierSymbols: [(NSEvent.ModifierFlags, String)] = [(.control, "⌃"), (.option, "⌥"), (.shift, "⇧"), (.command, "⌘")]
@@ -50,7 +50,7 @@ class KeyCodeHelper {
     0x15: ("4", "$"),
     0x16: ("6", "^"),
     0x17: ("5", "%"),
-    0x18: ("=", "+"),
+    0x18: ("=", "PLUS"),
     0x19: ("9", "("),
     0x1A: ("7", "&"),
     0x1B: ("-", "_"),
@@ -64,10 +64,10 @@ class KeyCodeHelper {
     0x23: ("p", "P"),
     0x25: ("l", "L"),
     0x26: ("j", "J"),
-    0x27: ("'", "\"\"\""),
+    0x27: ("'", "\""),
     0x28: ("k", "K"),
     0x29: (";", ":"),
-    0x2A: ("\"\\\"", "|"),
+    0x2A: ("\\", "|"),
     0x2B: (",", "<"),
     0x2C: ("/", "?"),
     0x2D: ("n", "N"),
@@ -197,7 +197,7 @@ class KeyCodeHelper {
 
   }()
 
-  static let mpvSymbolToKeyName: [String: String] = [
+  private static let mpvSymbolToPrettyMacKey: [String: String] = [
     META_KEY: "⌘",
     SHIFT_KEY: "⇧",
     ALT_KEY: "⌥",
@@ -238,13 +238,25 @@ class KeyCodeHelper {
     "KP9": "9",
   ]
 
-  static var reversedKeyMapForShift: [String: String] = keyMap.reduce([:]) { partial, keyMap in
+  private static var reversedKeyMapForShift: [String: String] = keyMap.reduce([:]) { partial, keyMap in
     var partial = partial
     if let value = keyMap.value.1 {
       partial[value] = keyMap.value.0
     }
     return partial
   }
+
+  private static var lowerToUpperKeyMap: [String: String] = keyMap.reduce([:]) { partial, keyValuePair in
+    var partial = partial
+    let (upper, lower): (String, String?) = keyValuePair.value
+    if let lower = lower {
+      partial[upper] = lower
+    }
+    return partial
+  }
+
+  /** Also includes symbols (e.g., `uppercaseMpvKeySet["4"] == "$"`)  */
+  private static var uppercaseMpvKeySet: Set<String> = Set(lowerToUpperKeyMap.values)
 
   static func canBeModifiedByShift(_ key: UInt16) -> Bool {
     return key != 0x24 && (key <= 0x2F || key == 0x32)
@@ -274,13 +286,13 @@ class KeyCodeHelper {
     var modifiers = event.modifierFlags
 
     if let char = event.charactersIgnoringModifiers, isPrintable(char) {
+      // Is a classic ASCII printable char.
       keyChar = char
-      let (_, rawKeyChar) = event.readableKeyDescription
-      if rawKeyChar != char {
-        modifiers.remove(.shift)
-      }
+      // The char in `charactersIgnoringModifiers` will be either uppercase or lowercase,
+      // so remove the redundant modifier flag so we don't print an extra "SHIFT+"
+      modifiers.remove(.shift)
     } else {
-      // find the key from key code
+      // Is probably an unprintable char such as KP_ENTER.
       guard let keyName = KeyCodeHelper.keyMap[keyCode] else {
         Logger.log("Undefined key code?", level: .warning)
         return ""
@@ -288,7 +300,7 @@ class KeyCodeHelper {
       keyChar = keyName.0
     }
     // modifiers
-    // the same order as `KeyCodeHelper.modifierOrder`
+    // the same order as `KeyCodeHelper.modifiersInOrder`
     if modifiers.contains(.control) {
       keyString += "\(CTRL_KEY)+"
     }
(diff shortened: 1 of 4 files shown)


iina/PrefKeyBindingViewController.swift
```diff
@@ -325,11 +326,14 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
   func saveToConfFile(_ sender: Notification) {
     let predicate = mappingController.filterPredicate
     mappingController.filterPredicate = nil
-    let keyMapping = mappingController.arrangedObjects as! [KeyMapping]
+    let keyMappings = mappingController.arrangedObjects as! [KeyMapping]
+    for mapping in keyMappings {
+      mapping.rawKey = KeyCodeHelper.escapeReservedMpvKeys(mapping.rawKey)
+    }
```
low-batt, line 332: No need to change, but could use `forEach` instead of `for`:
```swift
keyMappings.forEach { $0.rawKey = KeyCodeHelper.escapeReservedMpvKeys($0.rawKey) }

```

low-batt: Almost worked for me.

With these changes the `.conf` file now contains:
```text
SHARP quit
SPACE quit
```
The key bindings all look correct in settings. Looks as expected when I toggle `Display raw values`. When I press space IINA quits. **BUT** when I press # IINA responds with the unrecognized key sound and does not quit.

svobs: > No need to change, but could use `forEach` instead of `for`:
> 
> ```swift
> keyMappings.forEach { $0.rawKey = KeyCodeHelper.escapeReservedMpvKeys($0.rawKey) }
> ```

I actually prefer to use a traditional `for` loop instead of a `forEach`, unless there's a really good reason to use it. I like having named variables and being able to introduce `break`, and maybe there's some jadedness from using debuggers which can't handle closures as well and make it harder to debug. I get the sense that some developers tell themselves they're making their code more efficient by reducing the number of lines written, but I think usually they're just increasing the bug density.

> BUT when I press # IINA responds with the unrecognized key sound and does not quit.

Well...yikes. I dug into this and found a whole bunch of bugs.... will comment more below.

svobs:
Well...when I dug into your feedback, I found a bunch of issues with parsing. I suppose I could pack these changes into a separate PR...

Background refresher:

- There are multiple ways to represent a key sequence in the spec, but IINA does a string comparison to match the user's typed key(s) to a given key binding. Thus, I added a bunch of code to "normalize" it into a common format when doing the comparison (a lot of which used a bunch of code that was already there in `KeyCodeHelper` but tied it together).

1: SHARP & PLUS

There were several problems with the code which deals with special characters. The fix which allowed the `SHARP` key to be written to file broke the code which recognized whether that key was pressed. This underlying issue was that `SHARP` and `PLUS` are special token characters used by mpv, so they need to always been normalized to their whole-word representations instead of their literal `#` and `+` versions. But they weren't being normalized at all.

For these special symbols, the mpv `.conf` file spec allows for either literal (e.g., `+`)  or whole-word identifier (`PLUS`). When MacOS sends a key event to IINA, it is always translated to the literal version by default, so that's what it uses for the comparison by default. That means that anything specified in the whole-word version wouldn't be recognized when typed.

2: Uppercase handling was broken for non-alpha uppercase chars

There some code which I never dug into or tested, but with testing is clearly broken. The check for `key.lowercased() != key.uppercased()` doesn't do what we want to detect keys like `>` (we wanted it to detect it as a shifted `.` key). Likewise, the `event.readableKeyDescription` code is not returning valid data.

So I re-purposed the `keyMap` data structure in `KeyCodeHelper`, which largely wasn't being used and only needed a few fixes, to add a check for these special chars.

3: Case insensitive comparison broken

Most embarrassingly, `localizedCompare()` was being used instead of `localizedCaseInsensitiveCompare()`, which meant that `equalsIgnoreCase()` was actually not ignoring case... So this showed up when I tried `Shift+shift++` as a test case.

I put a few examples in the code comments which illustrate what was breaking. It sounds like a lot but most of these will probably only be noticed by people coming to IINA from mpv.

iina/PrefKeyBindingViewController.swift
low-batt: Yes, that is why I mentioned no need to change. I definitely will use a traditional `for` if the situation makes it useful to be able to use `break` or `continue`.

iina/KeyCodeHelper.swift
```diff
@@ -255,28 +267,40 @@ class KeyCodeHelper {
...
     var modifiers = event.modifierFlags
 
     if let char = event.charactersIgnoringModifiers, isPrintable(char) {
+      // Is a classic ASCII printable char.
       keyChar = char
-      let (_, rawKeyChar) = event.readableKeyDescription
-      if rawKeyChar != char {
-        modifiers.remove(.shift)
-      }
+      // The char in `charactersIgnoringModifiers` will be either uppercase or lowercase,
```
low-batt: From the Swift DocC section [Add a Basic Description for Each Symbol](https://www.swift.org/documentation/docc/writing-symbol-documentation-in-your-source-files#Add-a-Basic-Description-for-Each-Symbol):
> If a symbol already has a source comment that begins with two forward slashes (//), insert an additional forward slash (/) to convert it to a documentation comment.

I'm pretty sure [Markdown formatting](https://www.swift.org/documentation/docc/formatting-your-documentation-content) can only be used in swift documentation comments. I'm having trouble finding a reference, but I believe documentation comments will only be recognized before declarations, not inlined in a code block.

iina/KeyCodeHelper.swift
```diff
@@ -294,40 +318,58 @@ class KeyCodeHelper {
...
-  // Normalizes a single "press" of possibly multiple keys (as joined with '+')
+  /// Normalizes a single "press" of possibly multiple keys (as joined with '+')
+  /// _Examples:
+  /// Ex01: "Shift+-" → "_"
+  /// Ex02: "Shift+=" → "PLUS"
+  /// Ex03: "+-+" → "PLUS-PLUS"
+  /// Ex04: "meta+shift+k" → "Meta+K"
+  /// Ex05: "esc" → "ESC"
+  /// Ex06: "CTRL+SHIFT+SHIFT+SHIFT+x" → "Ctrl+X"
+  /// _
```
low-batt: In Xcode, clicking on `normalizeSingleMpvKeystroke` and then on the `?` button Xcode will show you what the rendered documentation looks like. 
```text
Normalizes a single “press” of possibly multiple keys (as joined with ‘+’) Examples: Ex01: “Shift+-” → “” Ex02: “Shift+=” → “PLUS” Ex03: “+-+” → “PLUS-PLUS” Ex04: “meta+shift+k” → “Meta+K” Ex05: “esc” → “ESC” Ex06: “CTRL+SHIFT+SHIFT+SHIFT+x” → “Ctrl+X” _
```

Markdown squishes it into one line. Maybe make use of a Markdown numbered list?

With a quick change to:
```swift
/// Normalizes a single "press" of possibly multiple keys (as joined with '+')
///
/// _Examples:_
/// 1. "Shift+-" → "_"
/// 2. "Shift+=" → "PLUS"
/// 3. "+-+" → "PLUS-PLUS"
/// 4. "meta+shift+k" → "Meta+K"
/// 5. "esc" → "ESC"
/// 6. "CTRL+SHIFT+SHIFT+SHIFT+x" → "Ctrl+X"
```

Xcode shows:

issue-4267


Could still be improved. I have the [Basic Syntax](https://www.markdownguide.org/basic-syntax/) and [Extended Syntax](https://www.markdownguide.org/extended-syntax/)  pages from the [Markdown Guide](https://www.markdownguide.org/) site bookmarked.

There are various apps that support editing Markdown. I've been using the open source [Joplin](https://joplinapp.org/) application. Sources can be found on GitHub [here](https://github.com/laurent22/joplin).

low-batt:
Something it still not quite right.

As a test I created some bindings using the IINA UI that were stored like this by IINA in the `conf` file:
```text
SHARP quit
SPACE quit
+ quit
! quit
@ quit
$ quit
% quit
TAB quit
BS quit
ESC quit
{ quit
} quit
` quit
F2 quit
ENTER quit
```

A few of these bindings still did not work for me:

| Key | Results | Notes |
| --- |--- | --- |
| SHARP | FAILED | Made the unrecognized key sound |
| SPACE | Passed | |
| + | FAILED | Made the unrecognized key sound |
| ! | Passed | |
| @ | Passed | |
| $ | Passed | |
| % | Passed | |
| TAB | FAILED |**Did not** make the unrecognized key sound |
| BS | Passed | |
| ESC | Passed | |
| { | Passed | |
| } | Passed | |
| ` | Passed | |
| F2 | Passed | |
| ENTER | Passed | |

svobs:
Thanks for the tips on XCode/Markdown formatting. It's one of those things I sort of care about but haven't spent any real time studying. I updated the comments in question so it looks nicer. 

On the use of caret symbols: even though I wasn't looking at the formatted version, I had fallen into using them as a practical way to indicate that the things inside identifiers and/or code snippets. It looks less ambiguous to me than using quotes or no punctuation at all. And XCode renders them in bold inside code blocks. Though I understand how some people might find them ugly or off-putting, so I'll try to avoid using them outside of declarations in the future.

I've been working in a few branches and I didn't notice the call to normalize the key input wasn't included (oops). Fixed in the latest commit.

But I need to look more at the issue of `TAB` not working. It's working in my big binding PR (#4160), but I made lots of fixes there which smooth out some weird stuff that is being done to the responder chain in the `develop` branch. Odd that the arrow keys work, but `TAB` doesn't...

low-batt (approved):
Pulled PR, built IINA, confirmed # key binding is working. Looks good to me.

svobs:
Oops. I think I tried renaming some branches a couple of weeks ago, and this got automatically deleted when I did! Will see if this happened to any other active PRs.

uiryuu:
Thanks!

PR #4274, opened 2023-03-18
Add Ogg video containers to recognized file ext list

HarukaMa:

---

**Description:**

Reported in Telegram group by Joonas Kokko. `ogm` and `ogv` are actually registered as file types by IINA, but they were not in `supportedFileExt` list. Should be safe to add them in as there should be nothing else that is using those extensions.


joonaskokko:
Thank you for adding these!

uiryuu:
Thanks

PR #4276, opened 2023-03-20
Fix lag when scrolling horizontally to seek, #4153

### LINKED ISSUE #4153: Lag when scrolling horizontally to seek in build 1.3.0 and 1.3.1
IINA used to work so well when scrolling horizontally , with the latest build it doesn't work the same anymore.

- Model Identifier:	iMac19,2
- macOS v12.6.1 (21G217)
- IINA v1.3.1 

Expected behavior: "scroll-to-seek" smooth scrolling as previous version (IINA 1.2.0)

Actual behavior: it looks super laggy now , it doesn't work well as before.

Every time

low-batt:
This commit will:
- Add methods displayActive and displayIdle to VideoView
- Add use of a Timer to delay stopping the display link
- Change MPVController to ensure display link is running while seeking
- Change PlayerCore to ensure display link is running while stepping
- Change MainWindowController to ensure display link is running when entering and leaving full screen mode

- [x] This implements/fixes issue #4153.

---

**Description:**

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -677,9 +677,6 @@ class PlayerCore: NSObject {
     } else {
       mpv.command(.frameStep)
     }
-    if info.isPaused {
-      mainWindow.videoView.displayIdle()
-    }
   }
 
   func screenshot() {
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -668,10 +668,17 @@ class PlayerCore: NSObject {
   }
 
   func frameStep(backwards: Bool) {
+    // When playback is paused the display link is stopped in order to avoid wasting energy on
+    // It must be running when stepping to avoid slowdowns caused by mpv waiting for IINA to call
+    // mpv_render_report_swap.
+    mainWindow.videoView.displayActive()
     if backwards {
       mpv.command(.frameBackStep)
     } else {
       mpv.command(.frameStep)
+    }
+    if info.isPaused {
+      mainWindow.videoView.displayIdle()
     }
   }
 
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -870,6 +870,12 @@ class MPVController: NSObject {
 
     case MPV_EVENT_SEEK:
       player.info.isSeeking = true
+      DispatchQueue.main.sync {
+        // When playback is paused the display link may be shutdown in order to not waste energy.
+        // It must be running when seeking to avoid slowdowns caused by mpv waiting for IINA to call
+        // mpv_render_report_swap.
+        player.mainWindow.videoView.displayActive()
+      }
       if needRecordSeekTime {
         recordedSeekStartTime = CACurrentMediaTime()
       }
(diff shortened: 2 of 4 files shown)


low-batt:
See the issue for an analysis of the problem.

There one case the proposed fix does not cover. In order to properly support the user sending commands directly to mpv through the IPC interface, control of the display link should be based on events from mpv. This works for seeking:
```bash
echo '{ "command": ["seek", "5"] }' | socat - /tmp/mpv-socket | jq
```

It does not work for stepping:
```bash
echo '{ "command": ["frame-step"] }' | socat - /tmp/mpv-socket | jq
```

Directly commanding mpv to step through the IPC interface works, but it does cause mpv to report "mpv_render_report_swap() not being called". This is because the code that insures the display link is working for stepping is in `PlayerCore.frameStep` and not in `MPVController` in response to a mpv event. That is because I was unable to find a mpv event that IINA can key off of for stepping.

low-batt:
Rebased with develop and corrected merge conflicts.

iina/PlayerCore.swift
```diff
@@ -668,11 +668,18 @@ class PlayerCore: NSObject {
...
+    // When playback is paused the display link is stopped in order to avoid wasting energy on
+    // It must be running when stepping to avoid slowdowns caused by mpv waiting for IINA to call
+    // mpv_render_report_swap.
+    mainWindow.videoView.displayActive()
     if backwards {
       mpv.command(.frameBackStep)
     } else {
       mpv.command(.frameStep)
     }
+    if info.isPaused {
```
uiryuu: After you send `.frameStep` or `.frameBackStep` to mpv, the playback will always stop, regardless the playback status before the command.

iina/PlayerCore.swift
```diff
@@ -668,11 +668,18 @@ class PlayerCore: NSObject {
   }
 
   func frameStep(backwards: Bool) {
+    // When playback is paused the display link is stopped in order to avoid wasting energy on
+    // It must be running when stepping to avoid slowdowns caused by mpv waiting for IINA to call
+    // mpv_render_report_swap.
+    mainWindow.videoView.displayActive()
```
uiryuu, line 674: It seems even if the display link is not running, frame step commands can run correctly. Do we have to activate the display link manually?

uiryuu (changes requested):

iina/PlayerCore.swift
low-batt: Stepping, seeking, all currently work because mpv will timeout and continue on. When that happens mpv will complain:
```text
[  14.169][v][vo/libmpv] mpv_render_report_swap() not being called.
```

We can tolerate some of this happening, but IINA should always try and avoid it. We don't want to see lots of these warnings from mpv when IINA is working correctly. For one thing that could hide other problems that trigger this warning that should not be occurring.

Because mpv does not give us an event to key off of (at least not that I could find) we have to tolerate this situation if the user sends a step command behind the back of IINA directly to mpv using the IPC interface. Unfortunate, but I don't see a way to deal with that situation.

uiryuu (approved):

svobs:
Very good fix! Looks like it has secondary improvements too like snappier & more reliable OSD updating.

Issue #4273, opened 2023-03-17
Cannot apply video filters to multiple windows


kaieberl:
macOS 13.2.1, IINA 1.3.1

**Behavior:**

When I open a video and apply some video filter to it, then open another video in a new window and try to apply any video filters to it, nothing happens in the current window and instead, the filters get applied to the first window.

**Steps to reproduce:**
Open some video, apply e.g. a crop filter crop=720:::
Then open another video in a new window and try to apply the same or any other filter, e.g. crop=1080:::

low-batt:
Reproduced for me. Sometimes it works. But I was able to trigger incorrect behavior with multiple windows.

low-batt:
### Analysis

The primary problem can been seen in this line from `FilterWindowController`: 
```swift
func reloadTable() {
  filters = PlayerCore.active.mpv.getFilters(filterType)
```

The `PlayerCore.active` property looks like this:
```swift
static var active: PlayerCore {
  if let wc = NSApp.mainWindow?.windowController as? MainWindowController {
    return wc.player
  } else {
    return first
  }
}
```

The problem with  `FilterWindowController` using `PlayerCore.active` is that the filter window is an independent window. Clicking on the filter window will change `NSApp.mainWindow` to be the filter window in which will cause the `first` player core to be returned, which may or may not be appropriate. Other window controllers such as `InspectorWindowController` use `PlayerCore.lastActive` to avoid this issue.

There is another way filters can be incorrectly applied. Stopping a mpv player core does not clear filters. IINA caches `PlayerCore` objects along with their mpv core. If a cached `PlayerCore`  is reused left over filters from the previous video will be applied.

And there is a race condition can that cause a crash if a filter window is open when IINA is quitting.  When IINA is terminating player windows are closed, which causes the iinaMainWindowChanged notification to be posted and that results in an observer establisted in `FilterWindowController`  being called which then attempts to refresh the filter window. This can happen after IINA has commanded mpv to shutdown. Once mpv has been told to shutdown mpv APIs must not be called as it can trigger a crash in mpv as seen below.

<details>
<summary><b>Crash Report:</b></summary>

```text
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               IINA [12379]
Path:                  /Applications/IINA.app/Contents/MacOS/IINA
Identifier:            com.colliderli.iina
Version:               1.3.1 (133)
Code Type:             ARM-64 (Native)
Parent Process:        launchd [1]
User ID:               503

[...]

Issue #4287, opened 2023-03-23
Subtitle delay is applied to the next video


low-batt:
**System and IINA version:**

- macOS 13.2.1
- IINA 1.3.1

**Expected behavior:**
If using the quick settings panel a subtitle delay is applied, it should only be applied to the current video.

**Actual behavior:**
_Sometimes_ the subtitle delay is applied to the next video.

**Steps to reproduce:**
Open IINA settings and:
- Confirm `Quit after all windows are closed` is **not** checked
- Confirm `Resume last playback position` is **not** checked
- Start playing a video that has subtitles
- Click on `Show Subtitles Panel` under the `Subtitles` menu
- The quick settings panel appears with the `SUBTITLES` tab selected
- In the panel move the knob of the `Subtitles delay` slider
- The OSD shows the subtitle delay
- Click on the red window button to close the playback window
- IINA continues to run
- Open another video that has subtitles
- Click on `Show Subtitles Panel` under the `Subtitles` menu
- Notice that the previously set subtitle delay has been applied to this video

If the `Resume last playback position` is checked and the last time the second video was played a different subtitle delay was set and playback of that video was stopped early so a watch later file was saved then the subtitle delay set when playing the first video will not be applied to the second video. Instead the subtitle delay set when previously playing the second video will be restored.

If instead the second video is opened before the window playing the first video is closed no subtitle delay will be applied to the second video. This random behavior is a side effect of the `PlayerCore` cache. Opening the second video before closing the window playing the first video forces IINA to create a second fresh `PlayerCore` and associated mpv core. In the first scenario the  `PlayerCore` and associated mpv core are reused. The mpv core still retains some of the settings from playing the previous video and they are then applied to the second video.

The first question that must be answered is what is the intended behavior of the quick settings panel with regards to what happens when you start playing another video? I'm worried some users may be used to certain settings being carried over and expect that is the intended behavior. As that behavior is **random** the way the code works today must be correct.

A similar issue can be seen in #4273 where video filters are _sometimes_ applied to another video.

- [ ] MPV does not have this problem.

[...]

PR #4288, opened 2023-03-23
Fix cannot apply filters to multiple windows, #4273

low-batt:
This commit will:
- Change `FilterWindowController` to use the active `PlayerCore`
- Change `FilterWindowController.reloadTable` to check if the active `PlayerCore` is shutting down
- Disable the ability to add a filter if the active core is stopping
- Change `PlayerCore.playbackStopped` to clear all audio and video filters

- [x] This implements/fixes issue #4273.

---

**Description:**

### DIFF
--- a/iina/FilterWindowController.swift
+++ b/iina/FilterWindowController.swift
@@ -34,6 +34,7 @@ class FilterWindowController: NSWindowController, NSWindowDelegate {
   @IBOutlet weak var editFilterKeyRecordView: KeyRecordView!
   @IBOutlet weak var editFilterKeyRecordViewLabel: NSTextField!
   @IBOutlet weak var removeButton: NSButton!
+  @IBOutlet weak var addButton: NSButton!
 
   var loaded = false
 
@@ -60,7 +61,7 @@ class FilterWindowController: NSWindowController, NSWindowDelegate {
     splitView.setPosition(splitView.frame.height - 140, ofDividerAt: 0)
 
     savedFilters = (Preference.array(for: filterType == MPVProperty.af ? .savedAudioFilters : .savedVideoFilters) ?? []).compactMap(SavedFilter.init(dict:))
-    filters = PlayerCore.active.mpv.getFilters(filterType)
+    filters = PlayerCore.lastActive.mpv.getFilters(filterType)
     currentFiltersTableView.reloadData()
     savedFiltersTableView.reloadData()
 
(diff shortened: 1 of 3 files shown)


low-batt:
**NOTE** the change to clear filters in `PlayerCore.playbackStopped` raises the question as to what users expect and what other setting IINA is failing to reset when reusing a `PlayerCore` object along with its mpv core. See issue #4287 for an example of random behavior due to sometimes inheriting previous settings when reusing a mpv core.

low-batt:
The mpv player carries over many settings when playing the next video. This behavior is useful when playing a series of videos such as the episodes of a show. However IINA differs from mpv in that if the preference `Always open media in a new window` is enabled then there can be multiple videos open and playable at the same time each with their own mpv core. Each mpv core has its own set of settings. Once there is more than one player/mpv core in the cache will appear random to the user as to which settings are applied when another video is opened. This means we want to preserve the mpv behavior of applying settings to the next video when `Always open media in a new window` is not enabled. However if this preference is enabled then we want to clear certain settings to avoid random behavior. The latest commit takes this attitude and clears audio and video filters if `Always open media in a new window` is enabled.

Trouble is `Always open media in a new window` is enabled by default. Users used to settings carrying over when IINA is just playing a playlist in a single window would not like this. I'm thinking maybe we are going to have to add a new preference.

Thoughts?

I can pull the change to clear filters for now and we can discuss how to handle the random behavior when there are multiple mpv cores in the next release.

iina/FilterWindowController.swift
```diff
@@ -97,10 +104,17 @@ class FilterWindowController: NSWindowController, NSWindowDelegate {
     }
     currentFiltersTableView.reloadData()
     savedFiltersTableView.reloadData()
+    // If the preference "Always open media in a new window" is enabled then once a player window is
+    // being closed the user must not be allowed to add a filter to the active player core if it is
+    // being stopped and about to be added to the player core cache. With that preference enabled
+    // there can be multiple player/mpv cores cached which would result in random behavior as to
+    // what filters would be applied to the next video.
+    addButton.isEnabled = !Preference.bool(for: .alwaysOpenInNewWindow)
```
uiryuu, line 112: Please add parentheses for better readability.

uiryuu (changes requested):
You are assuming that if "Always open media in a new window" is disabled, then there must be one player core at a time, which is not true. One can open multiple window, and then uncheck that option. So logically I think we shouldn't implement the behavior based on that preference.

svobs:
Examined this one a bit more, and I think this isn't quite a good fit for `Always open media in a new window`. With this existing proposal, it's still possible to open a new window in other ways, in which case the filters won't be cleared. And the description of that checkbox doesn't suggest any special behavior besides just opening a new window.

A better solution would be to always clear the filters for the mpv core when the window state goes from closed to opened. From the description in the issue, the user expected to have to re-enable the filter for every window already.

> Trouble is Always open media in a new window is enabled by default. Users used to settings carrying over when IINA is just playing a playlist in a single window would not like this.

If the user is switching between files in a playlist, then `Always open media in a new window` doesn't apply - right? The same window is used. So it should always reuse the same mpv core, and it makes sense to retain the filter settings in this case. But you won't have to add extra logic for that if you only clear filters for windows which are changing state from "closed" -> "opened".

low-batt:
Yes, I regretted pushing that proposed change after making it. That change is going into the toilet due to the problems you pointed out as well as concerns from @uiryuu.

I will pull the changes for clearing filters. We will take up the issue of mpv settings when there are multiple cores in the feature release.

By the way while testing this I think I found another error. I need to reproduce it again and dig into it so I've not filed an issue yet. Retained filters can fail when you open a new video. From the UI you don't know this has happened. I noticed this in mpv log messages.

low-batt:
This PR has been replaced by PR #4316 which does not attempt to deal with the issue of filters remaining configured in the mpv core and being applied to the next video.

PR #4290, opened 2023-03-23
Fix inspector displays "Error" for filename, #4289

### LINKED ISSUE #4289: Inspector displays "Error" for filename
**System and IINA version:**

- macOS 13.2.1
- IINA 

**Expected behavior:**
The filename displayed in the inspector window `File` tab matches the value shown in the `Status` tab if the mpv property `filename` is being watched in the `Watch` section of the inspector window.

**Actual behavior:**

The filename is displayed correctly in the `File` tab:

inspector-file

However the filename shows as "\<error>" in the `Status` tab:

inspector-status

**Steps to reproduce:**
I have not tracked down the exact sequence required to reproduce this. Bringing up the settings panel is part of it. I noticed the problem while code reading and just wanted to demonstrate that it can result in a problem.

The offending code is this line from `InspectorWindowController.tableView`:
```swift
return PlayerCore.active.mpv.getString(property) ?? "<Error>"
```

The rest of the  `InspectorWindowController` code is correctly using `PlayerCore.lastActive` instead of `active`.

I will post a PR with a fix shortly.

This is an IINA problem.

It requires a certain sequence to reproduce, but it is reproducible.

low-batt:
This commit will change InspectorWindowController.tableView to use PlayerCore.lastActive to match up with the rest of the code in the class.

- [x] This implements/fixes issue #4289.

---

**Description:**

### DIFF
--- a/iina/InspectorWindowController.swift
+++ b/iina/InspectorWindowController.swift
@@ -260,7 +260,7 @@ class InspectorWindowController: NSWindowController, NSTableViewDelegate, NSTabl
     if identifier == .key {
       return property
     } else if identifier == .value {
-      return PlayerCore.active.mpv.getString(property) ?? "<Error>"
+      return PlayerCore.lastActive.mpv.getString(property) ?? "<Error>"
     }
     return ""
   }


uiryuu:
Thanks!

PR #4297, opened 2023-03-26
Suppress warning from Sparkle 2.4.0

uiryuu:

---

**Description:**
- See https://sparkle-project.org/documentation/api-reference/Classes/SPUUpdater.html#/c:objc(cs)SPUUpdater(im)clearFeedURLFromUserDefaults
- And previous update from Sparkle 1.x to 2.x: #4047

@low-batt Also I was shocked that SwiftPM's `Package.resolved` is not tracked by git (since it deeply resides in the `xcodeproj` `xcworkspace` and `xcshareddata` folders). That's why I didn't get the warning, and since you start from git clone, all packages are checked out to the latest version. I think we should consider adding `Package.resolved` into the version control. Discussion about this: https://forums.swift.org/t/package-resolved-should-go-in-the-gitignore/14699

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -96,6 +96,8 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
 
   // MARK: - SPUUpdaterDelegate
 
+  @IBOutlet weak var updaterController: SPUStandardUpdaterController!
+
   func feedURLString(for updater: SPUUpdater) -> String? {
     return Preference.bool(for: .receiveBetaUpdate) ? AppData.appcastBetaLink : AppData.appcastLink
   }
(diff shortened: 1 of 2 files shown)


low-batt (approved):
Reproduced the warning and then pulled this PR and confirmed the warning is no longer emitted.

If we want to ensure everyone is using the same version of the swift packages then shouldn't we just change the Xcode settings to specify an exact version to use? Xcode does allow you to set the version of the package to use to `Exact`, but we have it set to `Up to Next Major`:

pr-4297

uiryuu:
So shall we use "Up to Next Minor" for all packages to avoid behavior changes?

low-batt:
I believe the advantage of `Up to Next Minor` over `Exact` is that sometimes packages will publish a security update and pull the offending version. In that case using `Exact` breaks the build if the version we are using has disappeared from the cloud. It would be possible to avoid that by grabbing the packages and changing Xcode to load them as local packages. That is what you might do if you are trying to get to a fully deterministic repeatable build.

I'm thinking that is overkill. I'm good with using `Up to Next Minor`.

Issue #4301, opened 2023-03-28
Crop animation does not respect reduce motion


low-batt:
**System and IINA version:**

- macOS 13.2.1
- IINA 1.3.1

**Expected behavior:**
IINA reduces animations if the user has enabled the macOS [Reduce motion](https://support.apple.com/guide/mac-help/reduce-screen-motion-mchlc03f57a1/mac) setting. From Apple's Human Interface Guidelines section [Color and effects](https://developer.apple.com/design/human-interface-guidelines/foundations/accessibility#color-and-effects):
> **Play tightened animations when Reduce Motion is on.** People can turn on Reduce Motion if they tend to get distracted or experience dizziness or nausea when viewing animations that include effects such as zooming, scaling, spinning, or peripheral motion. In response to this setting, you need to turn off or reduce animations that are known to cause problems (to learn more, see [Responsive Design for Motion](https://webkit.org/blog/7551/responsive-design-for-motion/)).

**Actual behavior:**
Entering and leaving the custom crop view uses a zooming animation.

**Steps to reproduce:**
- Click on `System Settings…` under the  menu
- Click on the `Accessibility` tab
- Click on `Display`
- Enable the `Reduce motion` setting
- Start IINA playing a video.
- Click on `Crop` > `Custom…` under the `Video` menu
- The crop view zooms into place
- Click on `Cancel`
- The view zooms away

- [x] MPV does not have this problem.

**How often does this happen?**
Every time.

low-batt:
IINA 1.3.2 contains the fix for this issue.

Issue #3437, opened 2021-06-28
Support macOS Reduce motion accessibility preference


low-batt:
**What you want IINA to do:**

Support the Apple Human Interface Guidelines on handling animation when the macOS System Preference Reduce motion is enabled.

Proper handling of the Reduce motion preference setting is covered in the [Apple Human Interface Guidelines under Appearance Effects and Motion](https://developer.apple.com/design/human-interface-guidelines/accessibility/overview/appearance-effects/).

The Reduce motion preference was added in iOS after new animations in iOS 7 induced cybersickness in a significant number of people. The preference was added to macOS in Sierra.

In the case of IINA sliding and zooming animations should be suppressed if Reduce motion has been enabled.

**What IINA does currently:**

IINA currently does not support this preference.

**Why you think this should be added:**

It is part of the Apple Human Interface Guidelines.

**Examples of other projects that have something similar:**

This support article for Discord discusses their support for this preference:
[Reduced Motion Setting](https://support.discord.com/hc/en-us/articles/360040613412-Reduced-Motion-Setting)

low-batt:
The changes in the pull request suppress the zoom animation when transitioning to and from full screen mode and the sliding animation when showing and hiding the side bar when Reduce motion has been enabled.

I encountered one problem in implementing these changes. With animation suppressed `toggleFullScreen` did not restore the window to its previous size. Could this behavior be related to the "Loss of window size when switching to full screen and back" issue #3144? I added code to the `windowDidExitFullScreen` method to explicitly set the frame to `priorWindowedFrame` when animation is suppressed.

Thoughts? Critical reviews including criticism of style and naming are welcome.

abh:
Thank you for doing this, @low-batt ! I'm looking forward to when it's released. I tried IINA again today and quickly realized why it was I never got into using it.

I know it seems weird for a video player to need this (videos are pretty universally full of motion!) but for my vestibular disorder most movies are fine, but just a moment of an animation like this in software or on websites can make me nauseous for hours.

low-batt:
Thanks for the feedback. Nice to know someone appreciates this feature.

The change only addresses the two animations listed above. You will still see animations for transitioning to PiP and back for example. At the moment I need to stay focused on helping to get this next release out. Maybe in the future I will take another look at this and see if I can include some of the other animations.

For some reason the software industry pretty much ignores this issue. Quite a while back now a major Mac app removed their preference for disabling UI animations and then were rude and dismissive to users that asked for it to be restored.

[...]

PR #4303, opened 2023-03-28
Fix crop animation does not respect reduce motion, #4301

low-batt:
The commit will change the `MainWindowController` methods `enterInteractiveMode` and `exitInteractiveMode` to use the method `AccessibilityPreferences.adjustedDuration` to suppress the zooming animation if reduce motion is enabled.

- [x] This implements/fixes issue #4301.

---

**Description:**
Minor change that makes use of the code added for issue #3437.


uiryuu:
Good point!

Issue #4299, opened 2023-03-27
Custom crop view resize issues


yuxiqian:
**System and IINA version:**

- macOS Ventura 13.3
- IINA 1.3.1

Start IINA and play a video, open "Quick Settings" panel and select Crop - "Custom...". Drag the right edge over the left boundary and release.

**Expected behavior:**

Drag the left edge of the box, should be able to resize the box again.

**Actual behavior:**

Unable to resize by drag the left edge, and its hit-box stays on the right side.

<img width="389" alt="screenshot" src="https://user-images.githubusercontent.com/34335406/227893506-25132460-556c-4f9b-b2aa-de8ea50330ee.png">

</details>

<br>

</details>

<br>

**Steps to reproduce:**

https://user-images.githubusercontent.com/34335406/227890268-c57467a0-158d-4b8f-b46c-81bb3dd3020f.mov

- [x] MPV does not have this problem.

**How often does this happen?**

Every time.

low-batt:
This is fixed in the released version of IINA.

PR #4302, opened 2023-03-28
Fix: cursor lost when dragging clip view edges

yuxiqian:
Start IINA and play a video, open "Quick Settings" panel and select Crop - "Custom...". Drag the right edge over the left boundary and release. The cursor that should appear on the left edge will be lost.

- [x] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)
- [x] This implements/fixes issue #4299 .

---

**Description:**

In `CropBoxView.swift`, it uses a `boxRect: NSRect` to store currently selected clip region, and drew cursor rects around it to capture dragging events. `boxRect.size` can be negative, indicating that it expands southwest from its origin point.

However, `NSRect.width` and `NSRect.height` are always positive, so the cursor rects (colored in red) are misplaced.

A screenshot of misplaced cursor rects

Using `NSRect.size.width` and `NSRect.size.height` should fix this problem.

PR #4304, opened 2023-03-28
Fix resizing issue in interactive mode when the crop box width or height is negative

uiryuu:
Standardize crop box upon `mouseUp`. Use `minX` and `minY` instead of `origin` when updating cursor rects to correct the behavior.

- [x] This implements/fixes issue #4299.


uiryuu:
Close in favor of #4302

Issue #4268, opened 2023-03-13
Black window if video is paused and toggled out of PiP


svobs:
**System and IINA version:**

- macOS 13.1
- IINA 131 build 133.
  - Also confirmed in HEAD of develop 21992670821f40c58c46ac322f815a2c03ac8409
  - Also confirmed with fix #4249 applied. Same problem, and with an additional issue: in step 3 below the video window flickers (seems to hide) momentarily when switching into PiP

**Steps to reproduce:**

1. Enable `UI` > `Picture-in-Picture` > `Toggle Picture-in-Picture by minimizing/un-minimizing the window`. The other PiP settings don't seem to make a difference.
2. Open a video and start playing. Pause.
3. Enter Picture-in-Picture.
4. Wait a second to make sure it's settled.
5. Click the Toggle PiP button inside the PiP window to toggle out of PiP.

**Expected behavior:**

Assuming `Accessibility` > `Reduce motion` is disabled in System Settings, smooth scaling animation into and out of PiP, displaying the content of the video with no breaks.

**Actual behavior:**

There is some flickering when coming out of PiP and the main window is displayed a bit too early, and when animation is done, the main video window is black (specifically, the VideoView appears to be hidden).

Bonus issue: `Reduce motion` does not appear to be honored when enabled.

- [X] MPV does not have this problem.

**How often does this happen?**

Always

low-batt:
The change to shutdown the display link while the video is paused in order to not waste energy has caused a number of problems. I'm working on that issue right now. That might be the root cause of this issue. I will look into it.

On `Reduce motion` not being honored, I'll double check, but I'm pretty certain the animation is coming from Apple's code.

svobs:
> On Reduce motion not being honored, I'll double check, but I'm pretty certain the animation is coming from Apple's code.

I've found that you sometimes need to check for `AccessibilityPreferences.motionReductionEnabled`, and there is a way to make a call to disable implicit animations. But...I'm unable to figure out where the actual resize is occurring. It looks like it might be inside `PIPViewController.presentAsPicture(inPicture: pipVideo)`? Is that in Objective-C code somewhere...?

low-batt:
The class `AccessibilityPreferences` was added by me. There are definitely still some places in IINA where AppKit provides an animate parameter that should be keying off the `Reduce motion` setting.

I would expect we need to probe what `MainWindowController.enterPIP` is doing. I vaguely remember being told the PiP stuff is tricky and used undocumented methods.

[...]

PR #4309, opened 2023-03-30
Fix black window after exiting PiP, #4268

low-batt:
This commit will:
- Add a VideoPIPViewController class that extends PIPViewController
- This class will force drawing when entering and exiting PiP
- Change MainWindowController to use VideoPIPViewController for the PiP controller
- Remove the pendingRedrawsAfterEnteringPIP property from VideoView
- Remove the layout method from VideoView
- Remove code that set pendingRedrawsAfterEnteringPIP from the MainWindowController.enterPIP method

- [x] This implements/fixes issue #4268.

### DIFF
--- a/iina/VideoPIPViewController.swift
+++ b/iina/VideoPIPViewController.swift
@@ -1,0 +1,35 @@
+//
+//  VideoPIPViewController.swift
+//  iina
+//
+//  Created by low-batt on 3/29/23.
+//  Copyright © 2023 lhc. All rights reserved.
+//
+
+import Cocoa
+
+class VideoPIPViewController: PIPViewController {
+
+  /// Force a draw after entering PiP.
+  ///
+  /// If playback is paused then after entering PiP mode the PiP window will sometimes be white. Force a draw to ensure this does not
+  /// happen and a frame is displayed. See PR #3973 for details.
+  ///
+  /// Any changes in this area must be tested on multiple macOS versions. Under 10.15, `presentAsPictureInPicture` behaves
+  /// asynchronously.
+  override func viewDidLayout() {
+    super.viewDidLayout()
+    guard let controller = delegate as? MainWindowController else { return }
+    controller.videoView.videoLayer.draw(forced: true)
+  }
+
+  /// Force a draw after exiting PiP.
+  ///
+  /// If playback is paused then after exiting PiP mode the main window will sometimes be black. Force a draw to ensure this does not
+  /// happen and a frame is displayed. See issue #4268 and PR #4286 for details.
+  override func viewDidDisappear() {
+    super.viewDidDisappear()
+    guard let controller = delegate as? MainWindowController else { return }
+    controller.videoView.videoLayer.draw(forced: true)
+  }
+}
--- a/iina/VideoPIPViewController.swift
+++ b/iina/VideoPIPViewController.swift
@@ -10,6 +10,16 @@ import Cocoa
 
 class VideoPIPViewController: PIPViewController {
 
+  /// Force a draw, if needed.
+  ///
+  /// If the image is changing there is no need to force a draw. However if playback is paused, or if playback is in progress but the video
+  /// track is an album art still image then drawing is required.
+  private func forceDraw() {
+    guard let controller = delegate as? MainWindowController, controller.player.info.isPaused
+            || controller.player.info.currentTrack(.video)?.isAlbumart ?? false else { return }
+    controller.videoView.videoLayer.draw(forced: true)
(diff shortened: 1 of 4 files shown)


iina/VideoPIPViewController.swift
```diff
@@ -0,0 +1,35 @@
...
+  /// Force a draw after entering PiP.
+  ///
+  /// If playback is paused then after entering PiP mode the PiP window will sometimes be white. Force a draw to ensure this does not
+  /// happen and a frame is displayed. See PR #3973 for details.
+  ///
+  /// Any changes in this area must be tested on multiple macOS versions. Under 10.15, `presentAsPictureInPicture` behaves
+  /// asynchronously.
+  override func viewDidLayout() {
+    super.viewDidLayout()
+    guard let controller = delegate as? MainWindowController else { return }
```
uiryuu: What about encapsulating these two lines into a private function?

uiryuu (changes requested):
The solution LGTM, tho it need a lot of testing, especially on older systems.

iina/VideoPIPViewController.swift
uiryuu: Also please bring back the checks for pause/album art detection

low-batt:
I added a shared private function that only forces drawing when needed.

I tested the new commit under macOS 13.3 and 10.15.7 with a video file and a MP3 file with embedded album art.

HarukaMa:
https://user-images.githubusercontent.com/861659/229276145-2720f99d-0fa5-45c1-8d13-f8e59b172d54.mp4

When exiting PIP while the video is paused, there is still a small amount of delay (much better already as it was a blank window before). Also there is a slight flicker when the video is playing, but not sure if it's easy to fix.

low-batt:
This:
> Also there is a slight flicker when the video is playing, but not sure if it's easy to fix.

Is very concerning. Does this happen with IINA 1.2.0? Could this be related to #4285? What kind of Mac? Which macOS version?

I'm testing on two Macs. 

On my MacBookAir8,2 running macOS Catalina 10.15.7 I see the slight delay when exiting PiP. The picture is displayed in the main window and then a moment later it looks like the picture is painted again, this time showing the UI. Testing IINA 1.2.0 I see the same behavior except initially the main window is still showing the "This video is playing in picture in picture" screen that after a pause is refreshed with the still frame.

On my MacBookPro18,2 running macOS Ventura 13.3 I don't see exactly that same behavior with the fix or with IINA 1.2.0. PiP animates directly into the window. Some of the time I can see the main window turn black before it is refreshed with a still frame. Sometimes not.

On both machines if you use the `x` button to exit PiP it is easier to see the main window turning black before it is refreshed. This is seen with IINA 1.2.0 as well.

I've not seen the PiP flicker when playing so far.

svobs:
> Also there is a slight flicker when the video is playing, but not sure if it's easy to fix.

@HarukaMa To clarify, you mean "when exiting PIP while the video is playing", right?

@low-batt From a high-level perspective, I can see why there would be a noticeable flicker and it might be very difficult to prevent, because the `VideoView` is being moved from the "main window" to the "mini player window". In order to attach it to the new window, it has to first detach it from the old one. Not just that, but it has to open a new `NSWindow` in the process, which is a relatively heavyweight object compared to a view and has to hook into the greater OS in lots of ways. It's not hard to imagine that a number of things are happening in the background which are beyond our control and which could potentially cause hiccups.

I don't know why the decision was made to use a separate window for PiP. The main window already uses a `fullSizeContentView` style mask, so it could have just made modifications to the existing window and avoided a lot of issues. In fact, the code for "interactive mode" does a lot more work than PiP so that it can reuse the window, and it seems to work flawlessly by comparison [so well that it actually fixes things which are broken, like HDR! Allegedly, hah]. Unfortunately, like @HarukaMa noted, it's not an easy change.

HarukaMa:
> To clarify, you mean "when exiting PIP while the video is playing", right?

Yes, that's right. Not when it's playing and not being touched. I omitted that part as that's the same as the previous sentence, but I guess that still left some ambiguous around.

low-batt:
Ah. Thanks for clarifying. I was confused because I had recently read issue #4285 which does seem to be about video stuttering while playing in the PiP window.

It does seem like it might be possible to improve the current behavior, but the purpose of this PR is to make sure the current sources are working as well as 1.2.0 when entering and exiting PiP.

Do we think something needs to be addressed or are these changes approved?

uiryuu (approved):
I'm approving the changes since it works as good as 1.2.0

PR #4311, opened 2023-04-01
Use `addAttributes` instead of `setAttribute`

uiryuu:

---

**Description:**
This commit:
- Removes CSS used to set attributes for `NSAttributedString`
- Sets the font and color all together
- Fixes an issue where the font becomes incorrect when trying to select the text using mouse

Before:
image

After:
image

You can try this by commenting out this line:
https://github.com/iina/iina/blob/763700dea8afcff89af02868a0621f6d0097cb4e/iina/InitialWindowController.swift#L176

### DIFF
--- a/iina/Base.lproj/InitialWindowController.xib
+++ b/iina/Base.lproj/InitialWindowController.xib
@@ -377,7 +377,7 @@
                     <rect key="frame" x="10" y="96" width="298" height="42"/>
                     <textFieldCell key="cell" controlSize="small" selectable="YES" sendsActionOnEndEditing="YES" allowsEditingTextAttributes="YES" id="3aN-Hg-GkT">
                         <font key="font" metaFont="menu" size="11"/>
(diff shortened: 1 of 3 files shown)


uiryuu:
There's a minor issue with the underline color of the links. When the mouse select the text field, the underline color will become `secondaryLabelColor`, which is the same as the text color. Even though there is `.linkColor` in `NSAttributeString.Key`, I failed to find a way to control it effectively.

low-batt (approved):
Pulled PR, built and tested. Looks good to me.

I did notice some additional problems.

`Preferences - General` should use `Settings` on Ventura.

When I hover over the links I get the  I-beam cursor, not the expected pointing-hand cursor. In fact while I move the cursor around in the pop-up I see it is changing the selection highlighting behind the pop-up.

uiryuu:
> When I hover over the links I get the I-beam cursor, not the expected pointing-hand cursor. In fact while I move the cursor around in the pop-up I see it is changing the selection highlighting behind the pop-up.

This is a weird bug. Let's try to solve it in another issue.

low-batt:
Agree, that cursor behavior is a separate issue.

PR #4319, opened 2023-04-04
Fix overlay loadfile thread

xjbeta:

---

**Description:**
fix `Main Thread Checker: UI API called on a background thread: -[NSWindowController isWindowLoaded]`


svobs:
This looks like a legit bug, but you could instead fix it in one line by changing `mainWindow.isWindowLoaded` to `mainWindow.loaded`.

PR #4323, opened 2023-04-06
Use expanded colorwell

### LINKED ISSUE #4322: Unable to adjust alpha with new NSColorWell style
**System and IINA version:**

- macOS 13.3
- IINA develop branch
- Xcode 14.3

PR #4039 brought a new `.minimal` look for `NSColorWell` which looks nice. However it doesn't support colors with alpha channel.

**Expected behavior:**

Clone and build IINA locally, open "IINA Settings" -> "Subtitle" and click any color well.

User should be able to select the first "transparent" option (it's common to set subtitle background transparent).

Click "Show Colors..." button should open an `NSColorPanel` with alpha slider.

**Actual behavior:**

The "transparent" color button in the pop-up view is disabled.

color pop-up

"Show Colors..." window doesn't have alpha slider, either.

color panel without alpha slider

Every time

uiryuu:
- [x] This implements/fixes issue #4322 .

---

**Description:**
It seems that we cannot adjust alpha value using the `.minimal` NSColorWell. The `.expanded` variant contains all functions of `.minimal`, plus it can adjust alpha. It is questionable from an aesthetics view, but we seems to have no choice.

Before:
image

After:
image


low-batt (approved):
Pulled PR, built and tested. Looks good to me.

PR #4324, opened 2023-04-07
Fix `setHTMLValue` encoding issue

yuxiqian:

---

**Description:**

PR #4311 updates `setHTMLValue` extension method but seems it caused an encoding issue in "About - Translators" text field.

IINA nightly build [79345ab](https://github.com/iina/iina/commit/79345ab):

image

Adding `.textEncodingName` option when constructing `NSMutableAttributedString` should fix this.


uiryuu:
Thanks!

PR #4360, opened 2023-04-18
Close add HDR tone mapping settings, #4358

### LINKED ISSUE #4358: Add HDR tone mapping settings
**What you want IINA to do:**
Provide a setting that allows the user to configure the [target-peak](https://mpv.io/manual/stable/#options-target-peak) option along with a way to request that IINA determine the appropriate value for this setting.

Provide a setting that allows the user to configure the mpv [tone-mapping](https://mpv.io/manual/stable/#options-tone-mapping) algorithm.

Display the pixel format in the inspector window.

These are improvements to HDR support that have already been implemented by @CarterLi. This issue is for tracking the need to merge this support.

These are mpv options.

low-batt:
This commit will:
- Add the new settings `enableToneMapping`, `toneMappingTargetPeak` and `toneMappingAlgorithm`
- Update the `Audio/Video` settings panel to allow the user to control these settings
- Update `VideoView.refreshEdrMode` to support these new settings
- Add the `CoreDisplay` framework to the Xcode project
- Add support to `VideoView.refreshEdrMode` to automatically determine the target peak setting based on the peak brightness of the display using the `CoreDisplay` framework
- Add reporting of the pixel format to the Inspector window
- Enhance HDR detection in `InspectorWindowController.updateInfo`
- Correct the path to the PiP framework in the Xcode project

These change provide the ability to mpv's tone mapping when playing HDR videos.

The main author of these changes is Carter Li. This is a merge of changes that have been tested in his IINA fork. A few minor changes were made during the merge.

- [x] This implements/fixes issue #4358.

---

**Description:**

### DIFF
--- a/iina/PrefCodecViewController.swift
+++ b/iina/PrefCodecViewController.swift
@@ -108,4 +108,16 @@ class PrefCodecViewController: PreferenceViewController, PreferenceWindowEmbedda
     toneMappingTargetPeakTextField.isEnabled = sender.state == .on;
     toneMappingAlgorithmPopUpBtn.isEnabled = sender.state == .on;
   }
+
+  @IBAction func toneMappingHelpAction(_ sender: Any) {
+    NSWorkspace.shared.open(URL(string: AppData.toneMappingHelpLink)!)
+  }
+
+  @IBAction func targetPeakHelpAction(_ sender: Any) {
+    NSWorkspace.shared.open(URL(string: AppData.targetPeakHelpLink)!)
+  }
+
+  @IBAction func algorithmHelpAction(_ sender: Any) {
+    NSWorkspace.shared.open(URL(string: AppData.algorithmHelpLink)!)
+  }
 }
--- a/iina/PrefCodecViewController.swift
+++ b/iina/PrefCodecViewController.swift
@@ -46,7 +46,6 @@ class PrefCodecViewController: PreferenceViewController, PreferenceWindowEmbedda
     super.viewDidLoad()
     audioLangTokenField.commaSeparatedValues = Preference.string(for: .audioLanguage) ?? ""
     updateHwdecDescription()
-    updateEnableToneMapping(enableToneMappingBtn)
   }
 
   override func viewWillAppear() {
@@ -102,11 +101,6 @@ class PrefCodecViewController: PreferenceViewController, PreferenceWindowEmbedda
   private func updateHwdecDescription() {
     let hwdec: Preference.HardwareDecoderOption = Preference.enum(for: .hardwareDecoder)
     hwdecDescriptionTextField.stringValue = hwdec.localizedDescription
-  }
-
-  @IBAction func updateEnableToneMapping(_ sender: NSButton) {
-    toneMappingTargetPeakTextField.isEnabled = sender.state == .on;
-    toneMappingAlgorithmPopUpBtn.isEnabled = sender.state == .on;
   }
 
   @IBAction func toneMappingHelpAction(_ sender: Any) {
--- a/iina/PrefCodecViewController.swift
+++ b/iina/PrefCodecViewController.swift
@@ -38,11 +38,15 @@ class PrefCodecViewController: PreferenceViewController, PreferenceWindowEmbedda
 
   @IBOutlet weak var audioDevicePopUp: NSPopUpButton!
 
+  @IBOutlet weak var enableToneMappingBtn: NSButton!
+  @IBOutlet weak var toneMappingTargetPeakTextField: NSTextField!
+  @IBOutlet weak var toneMappingAlgorithmPopUpBtn: NSPopUpButton!
 
   override func viewDidLoad() {
     super.viewDidLoad()
     audioLangTokenField.commaSeparatedValues = Preference.string(for: .audioLanguage) ?? ""
     updateHwdecDescription()
+    updateEnableToneMapping(enableToneMappingBtn)
   }
 
   override func viewWillAppear() {
@@ -98,5 +102,10 @@ class PrefCodecViewController: PreferenceViewController, PreferenceWindowEmbedda
   private func updateHwdecDescription() {
     let hwdec: Preference.HardwareDecoderOption = Preference.enum(for: .hardwareDecoder)
     hwdecDescriptionTextField.stringValue = hwdec.localizedDescription
+  }
+
+  @IBAction func updateEnableToneMapping(_ sender: NSButton) {
+    toneMappingTargetPeakTextField.isEnabled = sender.state == .on;
+    toneMappingAlgorithmPopUpBtn.isEnabled = sender.state == .on;
   }
 }
--- a/iina/VideoView.swift
+++ b/iina/VideoView.swift
@@ -221,7 +221,7 @@ class VideoView: NSView {
     CVDisplayLinkSetCurrentCGDisplay(link, displayId)
     let actualData = CVDisplayLinkGetActualOutputVideoRefreshPeriod(link)
     let nominalData = CVDisplayLinkGetNominalOutputVideoRefreshPeriod(link)
-    var actualFps: Double = 0;
+    var actualFps: Double = 0
 
     if (nominalData.flags & Int32(CVTimeFlags.isIndefinite.rawValue)) < 1 {
       let nominalFps = Double(nominalData.timeScale) / Double(nominalData.timeValue)
@@ -232,11 +232,11 @@ class VideoView: NSView {
 
       if abs(actualFps - nominalFps) > 1 {
         Logger.log("Falling back to nominal display refresh rate: \(nominalFps) from \(actualFps)")
-        actualFps = nominalFps;
+        actualFps = nominalFps
       }
     } else {
       Logger.log("Falling back to standard display refresh rate: 60 from \(actualFps)")
-      actualFps = 60;
+      actualFps = 60
     }
     player.mpv.setDouble(MPVOption.Video.overrideDisplayFps, actualFps)
 
@@ -311,14 +311,13 @@ class VideoView: NSView {
     }
 
     if videoLayer.colorspace != VideoView.SRGB {
-      videoLayer.colorspace = VideoView.SRGB;
+      videoLayer.colorspace = VideoView.SRGB
       videoLayer.wantsExtendedDynamicRangeContent = false
       player.mpv.setString(MPVOption.GPURendererOptions.targetTrc, "auto")
       player.mpv.setString(MPVOption.GPURendererOptions.targetPrim, "auto")
       player.mpv.setString(MPVOption.GPURendererOptions.targetPeak, "auto")
       player.mpv.setString(MPVOption.GPURendererOptions.toneMapping, "auto")
       player.mpv.setString(MPVOption.GPURendererOptions.toneMappingParam, "default")
-      player.mpv.setString(MPVOption.GPURendererOptions.toneMapping, "")
       player.mpv.setFlag(MPVOption.Screenshot.screenshotTagColorspace, false)
     }
   }
@@ -331,7 +330,7 @@ extension VideoView {
   func refreshEdrMode() {
     guard player.mainWindow.loaded else { return }
     guard player.mpv.fileLoaded else { return }
-    guard let displayId = currentDisplay else { return };
+    guard let displayId = currentDisplay else { return }
     if let screen = self.window?.screen {
       NSScreen.log("Refreshing HDR for \(player.subsystem.rawValue) @ display\(displayId)", screen)
     }
@@ -347,14 +346,14 @@ extension VideoView {
     guard let mpv = player.mpv else { return false }
 
     guard let primaries = mpv.getString(MPVProperty.videoParamsPrimaries), let gamma = mpv.getString(MPVProperty.videoParamsGamma) else {
-      Logger.log("HDR primaries and gamma not available", level: .debug, subsystem: hdrSubsystem);
-      return false;
+      Logger.log("HDR primaries and gamma not available", level: .debug, subsystem: hdrSubsystem)
+      return false
     }
   
     let peak = mpv.getDouble(MPVProperty.videoParamsSigPeak)
     Logger.log("HDR gamma=\(gamma), primaries=\(primaries), sig_peak=\(peak)", level: .debug, subsystem: hdrSubsystem)
 
-    var name: CFString? = nil;
+    var name: CFString? = nil
     switch primaries {
     case "display-p3":
       if #available(macOS 10.15.4, *) {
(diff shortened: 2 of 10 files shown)


low-batt:
### Notes for Reviewers

A few things of note…
- The changes add use of the CoreDisplay framework
- Display information returned by macOS is different in older versions
- The changes require the planned mpv upgrade as they will try and set `tone-mapping` to `auto` which is not available in mpv 0.34.1. Nothing drastically bad happens, you just need to change the setting to avoid mpv complaining.
- The `Target peak` setting only accepts a number, it does not accept `auto` as the mpv [target-peak](https://mpv.io/manual/stable/#options-target-peak) does

### Changes Made During Merge

I made some changes during the merge. Hopefully they seem reasonable, if not object!

#### Audio/Video Settings Panel

I indented the controls for `Target peak` and `Algorithm` as they are disabled when enable tone mapping is disabled. That makes them a part of a "hierarchy of settings". The HIG indicates indentation should be used to communicate that. Other similar IINA settings use indentation.

I greatly simplified the descriptions of the new settings to match the existing "minimal" style. Of note the original descriptions contain URLs pointing to the relevant section of the mpv manual:
```text
Maximum luminance supported by your screen
if set to 0, IINA will try to detect the maximum luminance of current monitor
Ref: https://mpv.io/manual/master/#options-target-peak
```

That is an interesting idea. Of course these should be clickable links and we'd have to discuss whether they should point to a version specific manual rather than master. Since this is not done for other settings I thought it best to consider such a change for the feature release and update the descriptions of many settings.

I this particular case referencing the mpv manual might make you think you could use the value `auto`, but as noted above that is not accepted.

I also tried to correct constraint warnings reported by Xcode. @svobs, If you have time please take a look and see if I'm adding regressions with respects to layout constraints.

#### Logging

To avoid the large log message that results from logging the full display info dictionary I changed the logging in `VideoView.requestEdrMode` to only log the dictionary if we are unable to find either of the two keys.

#### Preference Handling

I changed the code in `VideoView.requestEdrMode` that obtains the algorithm from the preference setting to fall back to the default value if the preference value could not be converted into a mpv string. This matches up with how other code uses the `HardwareDecoderOption` preference enumeration. This prevents a crash if a user does something like this:
```bash
defaults write com.colliderli.iina.plist toneMappingAlgorithm 99
```

In other words even though users should not be messing with the preference file behind IINA's back we treat it as external input and provide some protection against invalid values.

#### Screen Shots

With the changes in this PR:
issue-4358-merged

The original code with detailed descriptions of the settings:
issue-4358-orig

iina/VideoView.swift
```diff
@@ -393,8 +397,45 @@ extension VideoView {
     videoLayer.wantsExtendedDynamicRangeContent = true
     videoLayer.colorspace = CGColorSpace(name: name!)
     mpv.setString(MPVOption.GPURendererOptions.iccProfile, "")
-    mpv.setString(MPVOption.GPURendererOptions.targetTrc, "pq")
     mpv.setString(MPVOption.GPURendererOptions.targetPrim, primaries)
+    // PQ videos will be display as it was, HLG videos will be converted to PQ
+    mpv.setString(MPVOption.GPURendererOptions.targetTrc, "pq")
+    mpv.setFlag(MPVOption.Screenshot.screenshotTagColorspace, true)
```
CarterLi, line 402: This was used for screenshot in [jxl format](https://github.com/CarterLi/iina/blob/develop/iina/Preference.swift#L538).

Since [(newer) mpv actually doesn't support ppm / pgm ...](https://[mpv.io](https://mpv.io/manual/master/#options-screenshot-format)/manual/master/#options-screenshot-format), I suggest merging my screenshot code too.

CarterLi:
I like the style change in preferences window. Good work!

However, as tone-mapping algorithm is hard to understand for most people, I still prefer keeping the link to MPV document.

iina/VideoView.swift
low-batt: I had not noticed the old screenshot formats. They were removed from the UI in commit 994d6228202221a8ffccc74c29aa3530ec5366f5 back in 2017. I guess they did not remove the values from the enum to preserve compatibility with old plists. Probably safe to remove them now. I found the IINA+ commit that adds the new formats. I will prepare a separate PR for that change. Thanks for pointing this out.

low-batt:
I do like the idea of linking to the mpv manual. If the other reviewers approve I can make that change. I think they need to be clickable links. I'm thinking they should link to the stable manual and not master. Thoughts reviewers?

uiryuu:
@low-batt Can you rebase this PR? I'm having a hard time building this PR since we've already updated our deps.

We should align the new options with previous options. I understand you are trying to align the texts, but this seems off to me. The original design looks better to me in this sense.
image

As for link to the mpv manual, we can definitely do that, but adding a clickable button to the link of mpv manual is better, just like we previously did:
image

low-batt:
I will rebase and fix the merge conflicts. I will add `?` links to mpv manual.

Late here. I will get this done tomorrow.

low-batt:
I rebased with `develop` and corrected the merge conflicts.

I added help buttons that link to:
- https://en.wikipedia.org/wiki/Tone_mapping
- https://mpv.io/manual/stable/#options-target-peak
- https://mpv.io/manual/stable/#options-tone-mapping

The panel now looks like:

issue-4358-help

iina/PrefCodecViewController.swift
```diff
@@ -99,4 +103,21 @@ class PrefCodecViewController: PreferenceViewController, PreferenceWindowEmbedda
     let hwdec: Preference.HardwareDecoderOption = Preference.enum(for: .hardwareDecoder)
     hwdecDescriptionTextField.stringValue = hwdec.localizedDescription
   }
+
+  @IBAction func updateEnableToneMapping(_ sender: NSButton) {
```
uiryuu: Can you do this in xib by bindings?

iina/VideoView.swift
```diff
@@ -310,12 +310,16 @@ class VideoView: NSView {
...
+    if videoLayer.colorspace != VideoView.SRGB {
+      videoLayer.colorspace = VideoView.SRGB;
       videoLayer.wantsExtendedDynamicRangeContent = false
-      videoLayer.colorspace = VideoView.deviceRGBColorspace
       player.mpv.setString(MPVOption.GPURendererOptions.targetTrc, "auto")
       player.mpv.setString(MPVOption.GPURendererOptions.targetPrim, "auto")
+      player.mpv.setString(MPVOption.GPURendererOptions.targetPeak, "auto")
+      player.mpv.setString(MPVOption.GPURendererOptions.toneMapping, "auto")
+      player.mpv.setString(MPVOption.GPURendererOptions.toneMappingParam, "default")
+      player.mpv.setString(MPVOption.GPURendererOptions.toneMapping, "")
```
uiryuu: Why did you set this option twice? On line 319
```
player.mpv.setString(MPVOption.GPURendererOptions.toneMapping, "auto")
```

iina/VideoView.swift
```diff
@@ -393,8 +397,45 @@ extension VideoView {
...
+      if targetPeak == 0 {
+        if let displayInfo = CoreDisplay_DisplayCreateInfoDictionary(currentDisplay!)?.takeRetainedValue() as? [String: AnyObject] {
+          Logger.log("Successfully obtained information about the display", subsystem: hdrSubsystem)
+          // Prefer ReferencePeakHDRLuminance, which is reported by newer macOS versions.
+          if let hdrLuminance = displayInfo["ReferencePeakHDRLuminance"] as? Int {
+            Logger.log("Found ReferencePeakHDRLuminance: \(hdrLuminance)", subsystem: hdrSubsystem);
+            targetPeak = hdrLuminance
+          } else if let hdrLuminance = displayInfo["DisplayBacklight"] as? Int {
+            // We know macOS Catalina uses this key.
+            Logger.log("Found DisplayBacklight: \(hdrLuminance)", subsystem: hdrSubsystem);
```
uiryuu: Remove ;

iina/VideoView.swift
```diff
@@ -393,8 +397,45 @@ extension VideoView {
...
+    if Preference.bool(for: .enableToneMapping) {
+      var targetPeak = Preference.integer(for: .toneMappingTargetPeak)
+      // If the target peak is set to zero then IINA attempts to determine peak brightness of the
+      // display.
+      if targetPeak == 0 {
+        if let displayInfo = CoreDisplay_DisplayCreateInfoDictionary(currentDisplay!)?.takeRetainedValue() as? [String: AnyObject] {
+          Logger.log("Successfully obtained information about the display", subsystem: hdrSubsystem)
+          // Prefer ReferencePeakHDRLuminance, which is reported by newer macOS versions.
+          if let hdrLuminance = displayInfo["ReferencePeakHDRLuminance"] as? Int {
+            Logger.log("Found ReferencePeakHDRLuminance: \(hdrLuminance)", subsystem: hdrSubsystem);
```
uiryuu: Remove ;

iina/VideoView.swift
```diff
@@ -393,8 +397,45 @@ extension VideoView {
...
+            targetPeak = hdrLuminance
+          } else if let hdrLuminance = displayInfo["DisplayBacklight"] as? Int {
+            // We know macOS Catalina uses this key.
+            Logger.log("Found DisplayBacklight: \(hdrLuminance)", subsystem: hdrSubsystem);
+            targetPeak = hdrLuminance
+          } else {
+            Logger.log("Didn't find ReferencePeakHDRLuminance or DisplayBacklight, assuming HDR400", subsystem: hdrSubsystem);
```
uiryuu: Remove ;

iina/VideoView.swift
```diff
@@ -393,8 +397,45 @@ extension VideoView {
...
+            Logger.log("Display info dictionary: \(displayInfo)", subsystem: hdrSubsystem)
+            targetPeak = 400
+          }
+        } else {
+          Logger.log("Unable to obtain display information", level: .warning, subsystem: hdrSubsystem);
```
uiryuu: Remove ;

iina/VideoView.swift
```diff
@@ -393,8 +397,45 @@ extension VideoView {
...
+        }
+      }
+      let algorithm = Preference.ToneMappingAlgorithmOption(rawValue: Preference.integer(for: .toneMappingAlgorithm))?.mpvString
+        ?? Preference.ToneMappingAlgorithmOption.defaultValue.mpvString
+
+      Logger.log("Will enable tone mapping target-peak=\(targetPeak) algorithm=\(algorithm)", subsystem: hdrSubsystem);
```
uiryuu: Remove ;

iina/VideoView.swift
uiryuu: We should also fallback the target peak to 400 if the display info cannot be obtained. Otherwise if we cannot obtain  the display info, `MPVOption.GPURendererOptions.targetPeak` will be set to 0 later.

uiryuu (changes requested):
Should we call `refreshEdrMode` when the display is changed? A quick search shows that we are not doing that currently, but I might have missed that.

From a UI point of view, I think the text field under "Enable tone mapping" and the text field under the algorithm drop down do not provide useful information, given that we already provide the help link to the users. If they are confused by the settings, these text fields would not help.

For the text field under "Target peak", should we also inform the users that if we cannot detect automatically, the nits value will fallback to 400 nits?

iina/PrefCodecViewController.swift
low-batt: Yes, I'm changing this to binding enabling of the two controls to the checkbox in the XIB.

iina/VideoView.swift
low-batt: Because it is that way in the code in the other repository and I failed to notice that when preparing the merge. GOOD CATCH!

The "auto" for toneMapping was a recent add by mpv, so maybe the empty string setting was supposed to be removed after upgrading to 035.1?

@CarterLi Which is the correct value to use?

I'm going to guess "auto" and remove the setting to the empty string.

iina/VideoView.swift
low-batt: I did a search in this source and removed a number of semicolons.

iina/VideoView.swift
low-batt: I've updated this to also default to 400.

low-batt:
This sequence shows the window starting out on my external LG display which does not support HDR. Then moving to the internal XDR display which does support HDR. Then moving back. IINA enabled and disabled HDR appropriately:
<details>
<summary><b>IINA log:</b></summary>

```text
21:01:01.361 [iina][d] Refreshing HDR for player0 @ display3: "LG UltraFine" visible frame (0.0, 0.0, 2560.0, 1415.0) EDR: {supports=false maxPotential=1.0 maxCurrent=1.0}
21:01:01.361 [hdr][d] HDR gamma=pq, primaries=bt.2020, sig_peak=4.926108360290527
21:01:01.361 [hdr][d] HDR video was found but the display does not support EDR mode
21:01:01.361 [hdr][d] Loading ICC profile
21:01:05.189 [iina][d] Falling back to nominal display refresh rate: 120.0 from 59.9990747912984
21:01:05.189 [iina][d] Refreshing HDR for player0 @ display1: "Built-in Retina Display" visible frame (458.0, -1056.0, 1728.0, 1018.0) EDR: {supports=true maxPotential=16.0 maxCurrent=1.0}
21:01:05.189 [hdr][d] HDR gamma=pq, primaries=bt.2020, sig_peak=4.926108360290527
21:01:05.190 [hdr][d] Will activate HDR color space instead of using ICC profile
21:01:05.191 [hdr][d] Successfully obtained information about the display
21:01:05.191 [hdr][d] Found ReferencePeakHDRLuminance: 1000
21:01:05.191 [hdr][d] Will enable tone mapping target-peak=1000 algorithm=auto
21:01:21.269 [iina][d] Falling back to nominal display refresh rate: 59.99925000937488 from 120.25120635180502
21:01:21.270 [iina][d] Refreshing HDR for player0 @ display3: "LG UltraFine" visible frame (0.0, 0.0, 2560.0, 1415.0) EDR: {supports=false maxPotential=1.0 maxCurrent=1.0}
21:01:21.270 [hdr][d] HDR gamma=pq, primaries=bt.2020, sig_peak=4.926108360290527
21:01:21.270 [hdr][d] HDR video was found but the display does not support EDR mode
21:01:21.270 [hdr][d] Loading ICC profile
```
</details>

I removed the two unhelpful text fields and changed the remaining one to indicate the fall back to 400. The panel now looks like:
issue-4358-review

iina/VideoView.swift
CarterLi: Just use the default value of MPV

low-batt: The default is "auto", so I removed this [line](https://github.com/CarterLi/iina/blob/a7e5cb961b8f4b209d3532e8653ef73f0d88415c/iina/VideoView.swift#L292) that is in IINA+

PR #4367, opened 2023-04-20
Close JPEG-XL decoding/screenshot support, #3852

### LINKED ISSUE #3852: JPEG-XL decoding/screenshot support?
**What you want IINA to do:**
Preferably, be built to enable full decoding and screenshot encoding support for JXL, as such has now been fully implemented in ffmpeg and mpv.

**What IINA does currently:**

mpv/IINA should be used as the ultimate image viewer. JXL also supports arbitrary bit depths, and thats very important for not storing 16b pngs of 10b content.

low-batt:
This commit will:
- Remove `ppm`, `pgm`, `pgmyuv` and `tga` from the `Preferences.ScreenshotFormat` enum
- Add `webp` and `jxl` to the `Preferences.ScreenshotFormat` enum
- Add `WebP (.webp)` and `JPEG XL (.jxl)` as choices for the `Format` setting in the `Screenshots` section of `General` settings

NOTE that removing enumeration cases breaks backward compatibility as the new cases reuse the tag values previously assigned to `ppm` and `pgm`. These four screenshot formats were dropped by mpv and removed from IINA's setting panel back in 2017. Anyone using these old formats would have had to update their settings therefore it should be acceptable to reuse the tag values.

The main author of these changes is Carter Li. This is a merge of changes that have been tested in his IINA fork.

- [x] This implements/fixes issue #3852.

---

**Description:**

### DIFF
--- a/iina/FFmpegController.h
+++ b/iina/FFmpegController.h
@@ -38,6 +38,21 @@
 
 @property(nonatomic) NSInteger thumbnailCount;
 
+/// Initializes and returns an image object with the contents of the specified URL
+///
+/// At this time, the normal [NSImage](https://developer.apple.com/documentation/appkit/nsimage/1519907-init)
+/// initializer will fail to create an image object if the image file was encoded in [JPEG XL](https://jpeg.org/jpegxl/) format.
+/// In older versions of macOS this will also occur if the image file was encoded in [WebP](https://en.wikipedia.org/wiki/WebP/)
+/// format. As these are supported formats for screenshots it is desirable to have an alternative method for creating image objects that
+/// supports these formats so that a screenshot preview can be displayed as is done for other screenshot formats. This method is
+/// intended to be used when the normal AppKit supplied initializer is unable to create an image.
+/// - Attention: This method only supports decoding JPEG XL and WebP encoded images. It is not intended to replace the normal
+///       AppKit supplied initializer.
+/// - Parameter url: The URL identifying the image.
+/// - Returns: An initialized NSImage object or null if the method cannot create an image representation from the contents of the
+///       specified URL.
++ (nullable NSImage *)createNSImageWithContentsOf:(nonnull NSURL *)url;
+
 - (void)generateThumbnailForFile:(nonnull NSString *)file
                       thumbWidth:(int)thumbWidth;
 
--- a/iina/FFmpegController.h
+++ b/iina/FFmpegController.h
@@ -51,7 +51,7 @@
 /// - Parameter url: The URL identifying the image.
 /// - Returns: An initialized NSImage object or null if the method cannot create an image representation from the contents of the
 ///       specified URL.
-+ (nullable NSImage *)createNSImageWithContentsOf:(nonnull NSURL *)url;
++ (nullable NSImage *)createNSImageWithContentsOfURL:(nonnull NSURL *)url;
 
 - (void)generateThumbnailForFile:(nonnull NSString *)file
                       thumbWidth:(int)thumbWidth;
--- a/iina/FFmpegController.m
+++ b/iina/FFmpegController.m
@@ -20,24 +20,24 @@
 
 #import "IINA-Swift.h"
 
-#define ERROR(msg, ...) [FFmpegLogger error:([NSString stringWithFormat:(msg), ##__VA_ARGS__])];
-#define LOG(msg, ...) [FFmpegLogger debug:([NSString stringWithFormat:(msg), ##__VA_ARGS__])];
-#define WARN(msg, ...) [FFmpegLogger warn:([NSString stringWithFormat:(msg), ##__VA_ARGS__])];
+#define LOG_DEBUG(msg, ...) [FFmpegLogger debug:([NSString stringWithFormat:(msg), ##__VA_ARGS__])];
+#define LOG_ERROR(msg, ...) [FFmpegLogger error:([NSString stringWithFormat:(msg), ##__VA_ARGS__])];
+#define LOG_WARN(msg, ...) [FFmpegLogger warn:([NSString stringWithFormat:(msg), ##__VA_ARGS__])];
 
 #define THUMB_COUNT_DEFAULT 100
 
 #define CHECK_NOTNULL(ptr,msg) if (ptr == NULL) {\
-ERROR(@"Error when getting thumbnails: %@", msg);\
+LOG_ERROR(@"Error when getting thumbnails: %@", msg);\
 return -1;\
 }
 
 #define CHECK_SUCCESS(ret,msg) if (ret < 0) {\
-ERROR(@"Error when getting thumbnails: %@ (%d)", msg, ret);\
+LOG_ERROR(@"Error when getting thumbnails: %@ (%d)", msg, ret);\
 return -1;\
 }
 
 #define CHECK(ret,msg) if (!(ret)) {\
-ERROR(@"Error when getting thumbnails: %@", msg);\
+LOG_ERROR(@"Error when getting thumbnails: %@", msg);\
 return -1;\
 }
 
@@ -138,7 +138,7 @@ return -1;\
 
   // Check whether the denominator (AVRational.den) is zero to prevent division-by-zero
   if (videoAvgFrameRate.den == 0 || av_q2d(videoAvgFrameRate) == 0) {
-    LOG(@"Avg frame rate = 0, ignore");
+    LOG_DEBUG(@"Avg frame rate = 0, ignore");
     return -1;
   }
 
@@ -156,7 +156,7 @@ return -1;\
   if (pCodecCtx->pix_fmt < 0 || pCodecCtx->pix_fmt >= AV_PIX_FMT_NB) {
     avcodec_free_context(&pCodecCtx);
     avformat_close_input(&pFormatCtx);
-    ERROR(@"Error when getting thumbnails: Pixel format is null");
+    LOG_ERROR(@"Error when getting thumbnails: Pixel format is null");
     return -1;
   }
 
@@ -290,7 +290,7 @@ return -1;\
   // Close the video file
   avformat_close_input(&pFormatCtx);
 
-  // LOG(@"Thumbnails generated.");
+  // LOG_DEBUG(@"Thumbnails generated.");
   return 0;
 }
 
@@ -355,7 +355,7 @@ return -1;\
   ret = avformat_open_input(&pFormatCtx, cFilename, NULL, NULL);
   free(cFilename);
   if (ret < 0) {
-    ERROR(@"Error when opening file %@ to obtain info: %s (%d)", file, av_err2str(ret), ret);
+    LOG_ERROR(@"Error when opening file %@ to obtain info: %s (%d)", file, av_err2str(ret), ret);
     return NULL;
   }
 
@@ -363,7 +363,7 @@ return -1;\
   if (duration <= 0) {
     ret = avformat_find_stream_info(pFormatCtx, NULL);
     if (ret < 0) {
-      ERROR(@"Error when probing %@ to obtain info: %s (%d)", file, av_err2str(ret), ret);
+      LOG_ERROR(@"Error when probing %@ to obtain info: %s (%d)", file, av_err2str(ret), ret);
       duration = -1;
     } else
       duration = pFormatCtx->duration;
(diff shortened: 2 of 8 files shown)


low-batt:
I marked this PR as a draft because taking a JPEG XL screenshot while playing a HDR video did did not generate a screenshot that matched the video.

Also there is a second issue in that with `Show previews after taking screenshots` enabled the OSD only shows `Screenshot Captured`. This is because `NSImage` is unable to create an image from a JPEG XL file as macOS does not support this kind of file. The code in `PlayerCore.screenshotCallback` has a `guard` statement to handle this. _However_ not only does the OSD not display the screenshot, it does not provide the `DELETE`, `EDIT` and `REVEAL` buttons. The `DELETE` and `REVEAL` buttons are definitely still applicable. It is desirable to at least provide the `REVEAL` button.

To test JPEG XL I played a downloaded copy of the YouTube video [The World in HDR in 4K (ULTRA HD)](https://www.youtube.com/watch?v=tO01J-M3g0U) using [IINA+](https://github.com/CarterLi/iina) build 170 and took a screenshot. I then repeated this with IINA

I tested IINA with the changes in this PR applied to the changes in branch `close-4358` which sets `screenshot-tag-colorspace` to `true` using mpv 0.35.1 and FFmpeg 6.

The screenshot generated by IINA+ uses this color space:
```text
Color space: RGB, D65, Rec.2100 primaries, PQ transfer 
```

When displayed in IINA or IINA+ that screenshot displays in HDR and looks like the video.

The screenshot generated by IINA uses this color space:
```text
Color space: RGB, D65, sRGB primaries, sRGB transfer
```

When viewed  in IINA or IINA+ that screenshot is not displayed in HDR and is washed out.

<details>
<summary><b>Full IINA+ Screenshot Info:</b></summary>

```bash
low-batt@gag Screenshots$ ls -la The\ World\ in\ HDR\ in\ 4K\ \(ULTRA\ HD\)-0001.jxl 
-rw-r--r--@ 1 low-batt  staff  456600 Apr 19 19:17 The World in HDR in 4K (ULTRA HD)-0001.jxl
low-batt@gag Screenshots$ jxlinfo -v The\ World\ in\ HDR\ in\ 4K\ \(ULTRA\ HD\)-0001.jxl 
JPEG XL image, 3840x2160, lossy, 16-bit RGB
num_color_channels: 3
num_extra_channels: 0
intensity_target: 10000.000000 nits
min_nits: 0.000000
relative_to_max_display: 0
linear_below: 0.000000
have_preview: 0
have_animation: 0
Intrinsic dimensions: 3840x2160
Orientation: 1 (Normal)
Color space: RGB, D65, Rec.2100 primaries, PQ transfer function, rendering intent: Relative
low-batt@gag Screenshots$ 
```
</details>

<details>
<summary><b>Full IINA Screenshot Info:</b></summary>

```text
low-batt@gag Screenshots$ ls -la The\ World\ in\ HDR\ in\ 4K\ \(ULTRA\ HD\)-0002.jxl 
-rw-r--r--@ 1 low-batt  staff  415701 Apr 19 19:25 The World in HDR in 4K (ULTRA HD)-0002.jxl
low-batt@gag Screenshots$ jxlinfo -v The\ World\ in\ HDR\ in\ 4K\ \(ULTRA\ HD\)-0002.jxl 
JPEG XL image, 3840x2160, lossy, 16-bit RGB
num_color_channels: 3
num_extra_channels: 0
have_preview: 0
have_animation: 0
Intrinsic dimensions: 3840x2160
Orientation: 1 (Normal)
Color space: RGB, D65, sRGB primaries, sRGB transfer function, rendering intent: Relative
low-batt@gag Screenshots$ 
```
</details>

@CarterLi Any guess as to the cause of the difference between IINA and IINA+? IINA+ was built with mpv master whereas IINA is using the mpv 0.35.1 release, so a difference in mpv is one possibility.

low-batt:
HELP!!

I'm requesting help on this one as I lack experience in this area.

@CarterLi, I'm hoping for some advice on the HDR aspects.

First, on the issue with mpv 0.35.1 and JPEG XL screenshots of HDR videos. I believe that is due to the mpv issue https://github.com/mpv-player/mpv/issues/10988. The one line fix for that issue was merged right after the 0.35.1 release was cut. I will prepare a patch to mpv for that issue.

With a working mpv and the screenshot format set to JPEG XL you can take a screenshot of a HDR video that looks like the video when viewed using IINA. The screenshot preview looks like EDR is working, but I'm unsure if the presentation is fully correct.

What could be improved is the screenshot preview you see for a HDR video if HDR is turned off. It is clipped instead of tone mapped. Anyone know how to fix that?

I was unable to get FFmpeg to generate what I believe is the bitmap format required by Apple. I added a workaround for that. Is there another approach I should be using?

I included support for the display P3 color spaces, following what VideoView does, but I was unable to find a video that tests that.

Anyway, this PR could use an intense mean detailed critical review.

iina/FFmpegController.m
```diff
@@ -372,4 +381,324 @@ + (NSDictionary *)probeVideoInfoForFile:(nonnull NSString *)file
...
+      case AV_PIX_FMT_ARGB: // WebP with screenshot-webp-lossless mpv option enabled.
+      case AV_PIX_FMT_RGB24: // JPEG XL SDR video.
+      case AV_PIX_FMT_YUV420P: // WebP default.
+        pFrameRGB->format = AV_PIX_FMT_RGBA;
+        bitmapInfo = (CGBitmapInfo)kCGImageAlphaPremultipliedLast;
+        break;
+      case AV_PIX_FMT_RGB48LE: // JPEG XL HDR video.
+        // Workaround missing FFmpeg 6.0 scalar capabilities. As per Apple EDR requires using 16 bit
+        // floating point components in the image bit map. Therefore we want the scalar to convert
+        // the frame to the AV_PIX_FMT_RGBAF16LE pixel formation. However when that was specified
```
low-batt: Opps. I see I need to correct the "formation" typo.

CarterLi:
If I were you, I would use MPV (VideoView) to display the image, so that I could reuse most of the existing code, and would not have HDR issues.

low-batt:
An interesting idea. I initially discounted using mpv due to issues with integrating with the OSD and the heavy overhead of dealing with a mpv core. I will give it more thought, however I believe it is going to be even harder to get that to work in the other use case which I did not mention. The other HDR issue we have not addressed is thumbnails for HDR videos. We could generate JPEG XL thumbnails that would show HDR thumbnails. The thumbnails all live in one file in a proprietary IINA format. Not sure how we could get mpv to read that. I suppose we could generate a video whose frames are the thumbnails. Something like that maybe. I will think on this.

low-batt:
I played around with FFmpeg today from the command line and was able to generate from a 10 minute video at tiny video with only 65 frames, all of them keyframes. Something like that should allow us to use mpv for displaying HDR thumbnails.

One reason I was focused on creating a `NSImage` object due to this code from `PlayerCore.screenshotCallback`:
```swift
NSPasteboard.general.writeObjects([image])
```

If we use mpv to read the screenshot, we'd need a way to put the decoded screenshot into the pasteboard. I still need to figure that out.

The case for using mpv is that this is the best way to ensure that these images match the video. That is a really compelling argument.

To do this cleanly is going to require a lot of refactoring. It would be inappropriate to use `PlayerCore` objects for this. We would not want to use a `VideoView` with a `CVDisplayLink` for a still image. I'm currently thinking it would not be a good idea to add a lot of checks to classes to cover still images. Possibly we should introduce protocols, abstract base class with separate video and image implementations. Something like:
- MPVController (protocol)
- MPVControllerCommon
- MPVControllerImage
- MPVControllerVideo

I have a some other issues I need to turn my attention to at the moment, but when I have time I will hack up an ugly test that uses mpv to make sure there isn't some sort of blocker to this plan. 

I'm thinking for now we continue to review this PR with the understanding that it is temporary code likely to be replaced later.

iina/FFmpegController.h
```diff
@@ -38,6 +38,21 @@
...
+/// In older versions of macOS this will also occur if the image file was encoded in [WebP](https://en.wikipedia.org/wiki/WebP/)
+/// format. As these are supported formats for screenshots it is desirable to have an alternative method for creating image objects that
+/// supports these formats so that a screenshot preview can be displayed as is done for other screenshot formats. This method is
+/// intended to be used when the normal AppKit supplied initializer is unable to create an image.
+/// - Attention: This method only supports decoding JPEG XL and WebP encoded images. It is not intended to replace the normal
+///       AppKit supplied initializer.
+/// - Parameter url: The URL identifying the image.
+/// - Returns: An initialized NSImage object or null if the method cannot create an image representation from the contents of the
+///       specified URL.
++ (nullable NSImage *)createNSImageWithContentsOf:(nonnull NSURL *)url;
```
saagarjha:
```suggestion
+ (nullable NSImage *)createNSImageWithContentsOfURL:(nonnull NSURL *)url;
```

iina/FFmpegController.m
```diff
@@ -18,20 +18,26 @@
 #import <libavutil/mastering_display_metadata.h>
 #pragma clang diagnostic pop
 
+#import "IINA-Swift.h"
+
+#define ERROR(msg, ...) [FFmpegLogger error:([NSString stringWithFormat:(msg), ##__VA_ARGS__])];
+#define LOG(msg, ...) [FFmpegLogger debug:([NSString stringWithFormat:(msg), ##__VA_ARGS__])];
```
saagarjha: I would call this INFO or similar personally, "LOG" isn't really a log level

iina/FFmpegController.m
```diff
@@ -372,4 +381,324 @@ + (NSDictionary *)probeVideoInfoForFile:(nonnull NSString *)file
...
+  CGColorSpaceRef cgColorSpace = NULL;
+  CGContextRef cgContext = NULL;
+  CGImageRef cgImage = NULL;
+
+  @try {
+    char *cFilename = strdup(url.fileSystemRepresentation);
+
+#if DEBUG
+    LOG(@"Creating image with contents of file: %s", cFilename)
+#endif
```
saagarjha: Consider moving this up so the strdup/free pair are closer together. I think technically the copy isn't required anyways

iina/FFmpegController.m
```diff
@@ -372,4 +381,324 @@ + (NSDictionary *)probeVideoInfoForFile:(nonnull NSString *)file
...
+      return NULL;
+    }
+
+    NSImage *image = [[NSImage alloc] initWithCGImage:cgImage size: NSZeroSize];
+    if (!image) {
+      ERROR(@"Cannot create image");
+    }
+    return image;
+  }
+  @finally {
```
saagarjha, line 665: cute

iina/FFmpegController.h
low-batt: Fixed. I was trying to pattern this after [this](https://developer.apple.com/documentation/appkit/nsimage/1519907-init) `NSImage` initializer. But I could not get the Swift name translation to use `contentsOf` for the parameter name. I had to add "With".

iina/FFmpegController.m
low-batt: The problem was the correct name would be `DEBUG` and that name is taken. I added a `LOG_` prefix to these to allow uniform naming.

iina/FFmpegController.m
low-batt: Good catch. I patterned this after `getPeeksForFile` figuring it must be correct, however using `fileSystemRepresentation` directly works fine.

iina/FFmpegController.m
low-batt: I thought this try-finally pattern was better than trying to cleanup in each of those error returns. If this was Swift I'd use `defer` after each allocation.

iina/FFmpegController.m
low-batt: Typo fixed.

low-batt:

lhc70000:
> NOTE that removing enumeration cases breaks backward compatibility as the new cases reuse the tag values previously assigned to ppm and pgm. 

Why not 
```swift
  enum ScreenshotFormat: Int, InitializingFromKey {
    case png = 0
    case jpg
    case jpeg
    case webp = 7
    case jxl
}
```

low-batt:
Back when this was first coded we decided that as these settings had been all dropped back in 2017 it was very unlikely that reusing the values would be a problem. If we want to avoid reusing then we can change the enum like shown. We'd also have to override the normal tag values in the XIB for the pull down menu items so the tags match up with the enum values. All very doable. Should I make that change?

lhc70000:
OK, it should be fine. If we also need to change the XIB then it's probably not worth the hassle.

PR #4369, opened 2023-04-20
Update MPVOption for mpv 0.35.1

low-batt:
This commit will:
- Update `MPVOption` to represent mpv 0.35.1
- Correct `MPVController.mpvInit` to work with new `MPVOption`
- Remove reference to missing header file `vaapi.h` from Xcode project
- Add many missing header files to Xcode project

---

**Description:**


uiryuu:
Can we only reference the headers that we need? We can also use .gitkeep to only track the needed headers and ignore all other headers.

low-batt:
I noticed we still had a dangling red reference to the header file `vaapi.h`, so I changed the Xcode project to align it with the files we have committed to our git repository.

I was going to ask why we are putting all these headers into git. Is this because during an update we are copying all the headers from FFmpeg for the associated library and then adding them to git? Unless there is some reason we need headers that are not needed by our build, then we should not add these to the repository.

Maybe we need a bash script for updating this that only copies the headers we need? It could also run `parse_doc.rb` so we don't forget to update our generated mpv swift sources.

I'm reading that `.gitkeep` is not a git feature, merely a file name convention for a dummy file that you put in an otherwise empty directory to keep git from discarding the empty directory.

We can create a whitelist using [.gitignore](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files) Something along the lines of:
```text
*
!.gitignore
!avcodec.h
!version.h
```

Of course if we do that and have to add a new header during an update, git will not remind us we need to commit it to the repository.

And `.gitignore` has no effect on files that are already in the repository. From the GitHub doc:

> If you want to ignore a file that is already checked in, you must untrack the file before you add a rule to ignore it. From your terminal, untrack the file.
>```bash
>$ git rm --cached FILENAME
>```

I'm happy to clean this up. Let me know how to proceed.

uiryuu:
> We can create a whitelist using [.gitignore](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files) Something along the lines of:

Please create the gitignore file for headers and remove the unused headers.

low-batt:
Roger that. I will create the needed gitignore whitelists, update the Xcode project accordingly and resolve the latest merge conflicts.

low-batt:
I believe all of the remaining headers are required for the build.

PR #4372, opened 2023-04-21
Added option to always show OSC, #4366

### LINKED ISSUE #4366: Option request
**What you want IINA to do:** I want permanent osd controller option

**What IINA does currently:**When mouse goes out of inna app window, osd controller disappears.

**Why you think this should be added:**I am doing a job that i have to see the time always. when osd controller disappear, my job gets harder to handle

smishralm10:

---

**Description:**
This commit adds the requested feature #4366. Changes:

- Option added to always show OSC in Settings > UI > On Screen Controller > Always show OSC (disabled by default)
Screenshot 2023-04-21 at 1 58 36 PM
- when this option is enabled "Auto hide after" textfield will be disabled (shown in the image below)
Screenshot 2023-04-21 at 1 58 50 PM

### DIFF
--- a/iina/Base.lproj/PrefUIViewController.xib
+++ b/iina/Base.lproj/PrefUIViewController.xib
@@ -574,16 +574,8 @@
                         <binding destination="lH7-Vv-0M1" name="selectedTag" keyPath="values.oscPosition" id="CEs-HC-aAO"/>
                     </connections>
                 </popUpButton>
-                <textField horizontalHuggingPriority="251" verticalHuggingPriority="750" allowsCharacterPickerTouchBarItem="YES" translatesAutoresizingMaskIntoConstraints="NO" id="6ml-Xj-nVX">
-                    <rect key="frame" x="118" y="112" width="98" height="16"/>
-                    <textFieldCell key="cell" scrollable="YES" lineBreakMode="clipping" sendsActionOnEndEditing="YES" title="Auto hide after:" id="82A-XQ-BvS">
-                        <font key="font" metaFont="system"/>
-                        <color key="textColor" name="labelColor" catalog="System" colorSpace="catalog"/>
-                        <color key="backgroundColor" name="controlColor" catalog="System" colorSpace="catalog"/>
-                    </textFieldCell>
-                </textField>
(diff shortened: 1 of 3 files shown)


low-batt:
@smishralm10 Thanks for looking into this.

This should be simplified. This should not be about adding a new setting to keep the OSC always visible. This should be about making the existing setting to hide the OSC optional. Instead of adding `Always show OSC`, a checkbox should be added to the existing `Auto hide after` setting. When that is unchecked the `TextField` for entering the number of seconds should be disabled. The preference becomes `enableControlBarAutoHide`.

smishralm10:
Thanks for feedback. I'll implement the required changes

smishralm10:
This commit adds a checkbox to the existing `Auto hide after` option and removes `Always show OSC` option. Disabling `Auto hide after` option will disable the `TextField` beside it, as shown in the below images.

Screenshot 2023-04-22 at 12 10 37 AM

Screenshot 2023-04-22 at 12 10 46 AM

svobs:
Pulled and ran this PR. Wanted to note a couple of wrinkles.

Biggest concern: I noticed this also disables the "hide when cursor exits the window" functionality. I've been meaning to post a PR at some point which breaks that out into its own checkbox. Doesn't think seem like a separate piece of functionality?

More of a side note: this PR also means that the title bar also won't go away, even though it's covering the video. I realize this is hard to avoid from a functional standpoint, because the title bar & OSC are currently so coupled. At least there's still one way to hide them both, which is to go to `Settings` > `Control` and configure one of the mouse actions to hide the OSC.

low-batt (approved):
I pulled the PR locally, built IINA and confirmed the new setting worked.

low-batt:
@smishralm10 Thanks for making the changes!

@svobs I tried this:
> At least there's still one way to hide them both, which is to go to Settings > Control and configure one of the mouse actions to hide the OSC.

But could not get the OSC to hide.

More work is needed in this area. There are outstanding requests for other OSC options. One is to add a `Enable OSC` like we have for the the OSD to be able to disable showing of the OSC. For example this is useful for people that are controlling IINA using an Apple TV remote.

uiryuu:
Thanks!

PR #4385, opened 2023-04-27
Close enhance identification of IINA build, #4383

### LINKED ISSUE #4383: Enhance identification of IINA build
**What you want IINA to do:**
Make it very easy to determine what kind of IINA build you are running.

**What IINA does currently:**
There are four kinds of IINA builds:
- Beta
- Debug
- Nightly
- Release

Currently Beta builds can be identified by the orange `BETA` button on the welcome window:

welcome-window-beta

However that window will not be shown unless the `At launch` setting is set to `Show welcome window`. In the `develop` branch the about window has been enhanced to show the build date and the git branch from which the build was generated. This is also logged:

```text
15:24:23.697 [iina][d] Built Apr 26, 2023 at 12:22:21 AM from branch develop, commit bea84bb41b43daa4d84587a6b3ff5def6fe214d5
```

However this information does not explicitly identify the kind of build.

To avoid any confusion over what kind of build is being run. This is especially important when reporting problems.

The GNOME Human Interface Guidlines for [App Icons](https://developer.gnome.org/hig/guidelines/app-icons.html) says:
> **Nightly Variant**
> If you provide a nightly or beta build of your app, it is recommended to create a specific version of the app icon for it so it is possible to tell them apart.

That convention can be seen on [this](https://www.mozilla.org/en-US/firefox/channel/desktop/) page which shows the Beta, Developer and Nightly icons used by Firefox. IINA should adopt this convention of using different icons to clearly distinguish between the different kinds of builds.

As there already is an IINA convention of identifying a Beta build on the welcome window, that should be extended to identify Nightly and Debug builds. Might want to make these buttons like is done for BETA. The NIGHTLY button would identify the build date a provide a link to the commit the build was generated from. Not as critical for DEBUG but might as well do the same for that kind of build.

The about window should be enhanced as follows:
- Beta: Instead of the branch name show the orange `BETA` label
- Debug: Show branch name and clickable commit `fix-4279 bea84bb`
- Nightly: Instead of the branch name show Nightly and clickable commit `NIGHTLY bea84bb`
- Release: Similar to welcome window, don't call out that this is the release build, suppress showing the branch name

At the start of the log file append BETA, DEBUG or NIGHTLY after the build number:
```text
00:25:13.792 [iina][d] IINA 1.3.1 Build 133 NIGHTLY
```
As the build date, branch and commit are already logged, that the only change needed to the log file.

low-batt:
This commit will:
- Add Beta and Nightly Xcode build configurations to the Xcode project
- Add Beta.xcconfig and Nightly.xcconfig build configuration files for the new build configuration
- Add two new icons, AppIconBeta and AppIconNightly
- Use ASSETCATALOG_COMPILER_APPICON_NAME to select the correct application icon
- Add a key com.colliderli.iina.build.configuration to Info.plist to allow IINA to determine at runtime the build configuration used to generate the executable
- Add a property buildConfiguration to InfoDictionary that exposes the raw value in Info.plist
- Add a BuildType enumeration that corresponds to the build configurations in the Xcode project
- Add buildType, buildTypeIdentifier and isRelease properties to InfoDictionary to make it easy for other code to access the build configuration

This commit is only a start on the changes needed to implement #4383. This commit only contains the changes to Xcode needed to use different application icons for Beta and Nightly builds and to allow IINA to know what type of build it is running. The new application icons have been populated with document icons for testing. These must be replaced with the appropriate icons once they have been generated. The commit does not contain any changes to the welcome window or about window. Those changes still have to be implemented.

- [x] This implements/fixes issue #4383.

---

**Description:**


low-batt:
This draft PR only contains the changes to Xcode needed to use different application icons for Beta and Nightly builds and to allow IINA to know what type of build it is running.

No changes have been made to the welcome and about windows. They still need to be updated to indicate the type of build.

Two new application icon sets, `AppIconBeta` and `AppIconNightly` were added as assets. They have been populated with document icons for testing. These must be replaced with the appropriate icons once they have been generated.

The log file has been updated to indicate the build type:
```text
17:50:49.168 [iina][d] IINA 1.3.1 Build 133 DEBUG
```

Details on the Xcode changes follow…

Two new configurations, `Beta` and `Nightly` were added by opening the IINA project, selecting the `Info` tab, and using the `+` button in the `Configurations` section and selecting `Duplicate "Release" Configuration` in the popup. The instructions in the Apple article [Adding a build configuration file to your project](https://developer.apple.com/documentation/xcode/adding-a-build-configuration-file-to-your-project) were then followed to add `Beta.xcconfig` and `Nightly.xcconfig` configuration files.

The configuration files make use of the following standard Xcode [Build settings](https://developer.apple.com/documentation/xcode/build-settings-reference):

- [Primary App Icon Set Name](https://developer.apple.com/documentation/xcode/build-settings-reference#Primary-App-Icon-Set-Name)
    > **Setting name:** ASSETCATALOG_COMPILER_APPICON_NAME
    > 
    > Name of an app icon set for the target’s default app icon. The contents will be merged into the Info.plist.

- [CONFIGURATION](https://developer.apple.com/documentation/xcode/build-settings-reference#CONFIGURATION)
    >**Setting name:** CONFIGURATION
    >
    > Identifies the build configuration, such as Debug or Release, that the target uses to generate the product.

A new entry was then added to the `Info.plist` file following the instructions in [Edit property lists](https://help.apple.com/xcode/mac/current/#/dev3f399a2a6). Prefixing custom keys with the bundle identifier is recommended by Apple so the new key is named `com.colliderli.iina.build.configuration`. It is assigned the value of `$(CONFIGURATION)`. This enables IINA at runtime to determine how it was built.

Properties were added to `InfoDictionary` to make it easy for other code to determine the build type.

iina/Base.lproj/Localizable.strings
```diff
@@ -21,6 +21,13 @@
 "guide.basic_settings" = "Basic Settings";
 "guide.faq" = "Frequently Asked Questions";
 
+// Initial Window
+"initial.announcement" = "You are using a %@ version of IINA";
+"initial.beta.desc" = "Uncheck \"receive beta updates\" from Settings > General, and install a stable release from our <a href=\"https://iina.io\">official website</a>.";
+"initial.nightly.desc" = "Nightly is not recommanded to be used as daily drive. Currently nightly builds do not support auto update. Please manually download newer builds from our <a href=\"https://nightly.iina.io\">nightly website</a>.";
+"initial.debug.desc" = "Debug builds are usually not optimized for performance, so it is not recommanded to be used as daily drive. Debug builds do not support auto update.";
```
low-batt: typo recommanded should be recommended

iina/Base.lproj/Localizable.strings
```diff
@@ -21,6 +21,13 @@
...
+"initial.bug_report" = "Report to us via <a href=\"https://github.com/iina/iina/issues\">GitHub issues</a>.";
```
low-batt: Could shorten this to just "Report via"

iina/InitialWindowController.swift
```diff
@@ -64,6 +64,20 @@ fileprivate extension NSColor {
       return NSColor(calibratedRed: 1, green: 0.6, blue: 0.2, alpha: 1)
     }
   }()
+  static let initialWindowNightlyLabel: NSColor = {
+    if #available(macOS 10.14, *) {
+      return NSColor(named: .initialWindowNightlyLabel)!
```
low-batt, line 69: I was unable to test Nightly as a run blows up here with a nil.

iina/InitialWindowController.swift
```diff
@@ -64,6 +64,20 @@ fileprivate extension NSColor {
...
+    } else {
+      return NSColor(calibratedRed: 0, green: 1, blue: 1, alpha: 1)
+    }
+  }()
+  static let initialWindowDebugLabel: NSColor = {
+    if #available(macOS 10.14, *) {
+      return NSColor(named: .initialWindowDebugLabel)!
```
low-batt, line 76: I was unable to test Debug as a run blows up here with a nil.

iina/en.lproj/Localizable.strings
```diff
@@ -21,6 +21,13 @@
 "guide.basic_settings" = "Basic Settings";
 "guide.faq" = "Frequently Asked Questions";
 
+// Initial Window
+"initial.announcement" = "You are using a %@ version of IINA";
+"initial.beta.desc" = "Uncheck \"receive beta updates\" from Settings > General, and install a stable release from our <a href=\"https://iina.io\">official website</a>.";
+"initial.nightly.desc" = "Nightly is not recommanded to be used as daily drive. Currently nightly builds do not support auto update. Please manually download newer builds from our <a href=\"https://nightly.iina.io\">nightly website</a>.";
```
low-batt: recommended

iina/InitialWindowController.swift
```diff
@@ -453,16 +482,29 @@ class InitialWindowViewActionButton: NSView {
...
+      self.layer?.backgroundColor = NSColor.initialWindowDebugLabel.cgColor
+    case .beta:
+      self.layer?.backgroundColor = NSColor.initialWindowBetaLabel.cgColor
+    case .nightly:
+      self.layer?.backgroundColor = NSColor.initialWindowNightlyLabel.cgColor
+    default:
+      break
+    }
+
+    announcementLabel.stringValue = String(format: NSLocalizedString("initial.announcement", comment: "Version annoucnement"), buildType.rawValue)
```
low-batt: typo "Version annoucnement" should be "Version announcement"

iina/InitialWindowController.swift
uiryuu: My fault, I unstaged too many changes.

Configs/Debug.xcconfig
```diff
@@ -8,8 +8,8 @@
 
 #include "Shared.xcconfig"
 
-ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon
-CONFIGURATION = Debug
+ASSETCATALOG_COMPILER_APPICON_NAME = AppIconNightly
+CONFIGURATION = Nightly
```
low-batt: I am so not understanding this change. We want the nightly build to be a release build, yes? The debug builds are pretty much for developers. Developers need to be able to distinguish them from nightly builds.

iina/InfoDictionary.swift
```diff
@@ -83,7 +87,6 @@ struct InfoDictionary {
   /// Enum corresponding to the build configurations in IINA's Xcode project.
   enum BuildType: String, CustomStringConvertible {
     case beta = "Beta"
-    case debug = "Debug"
     case nightly = "Nightly"
```
low-batt, line 90: This change makes the above comment incorrect. Xcode has a Debug build configuration, so you would expect it to be here. In other words this is no longer a direct mapping to Xcode configurations as the comment would have you believe. This is confusing and definitely needs a comment to explain the thinking. Possibly whatever we are trying to achieve is being done at too low of a level. In other words this low level code should do the simple thing and map to the Xcode configurations. That is easy and straight forward for people to understand. Higher level code can then treat the debug configuration differently, like mapping it to nightly.

low-batt:
I was not able to test and see what the latest changes look like because something about them has really broken Xcode. It seems to only be building `iina-cli` so when it goes to start IINA this is all the Xcode console shows:
```text
Cannot find IINA binary. This command line tool only works in IINA.app bundle.
Program ended with exit code: 1
```

low-batt:
I figured out that the changes switched the scheme that was being built on me to `iina-cli`. If you bring up `Manage Schemes…` it is possible to change the order of the schemes. I think the IINA schemes should come first.

I was able to build and run `beta` and `nightly`. Something has caused the configuration to disappear from the log file:
```text
14:40:58.631 [iina][d] IINA 1.3.1 Build 133
```

I was unable to build `debug`:

Screenshot 2023-04-29 at 2 39 17 PM



Lots of these errors:
```text
info.plist found in '/Users/low-batt/Library/Developer/Xcode/DerivedData/iina-gwhhvjgxybcchtdcwvmukizbfeip/SourcePackages/artifacts/sparkle/Sparkle/sparkle.app/Contents' but failed to parse: keyNotFound(CodingKeys(stringValue: "AvailableLibraries", intValue: nil), Swift.DecodingError.Context(codingPath: [], debugDescription: "No value associated with key CodingKeys(stringValue: \"AvailableLibraries\", intValue: nil) (\"AvailableLibraries\").", underlyingError: nil))
```

saagarjha:
@low-batt has mentioned some of this already but this is the hierarchy I expect:

* Debug configuration
	* Debug.xcconfig
		* iina target (local build)
* Release configuration
	* Release.xcconfig
		* iina target (local build)
* Beta configuration
	* Release.xcconfig
		* iina target (beta build)
* Nightly configuration
	* Release.xcconfig
		* iina target (nightly build)
* Production configuration
	* Release.xcconfig
		* iina target (production build)

Debug and Release exist for local development and profiling. They currently use what is called "iina.xcconfig". Beta, Nightly, and Production are identical from a build layout perspective; they only exist to ship to users and exist with Release.xcconfig applied. The target-level xcconfig is basically iina.xcconfig, with a specific app icon set. How you want to do this is up to you. One options is this:

### iina.xcconfig (not used by any targets)
```
[Current contents of iina.xcconfig save for the app icon]
```

### iina-local.xcconfig (used by Debug, Release)
```
#include "iina.xcconfig"

ASSETCATALOG_COMPILER_APPICON_NAME = LocalAppIcon
```

### iina-beta.xcconfig (used by Beta)
```
#include "iina.xcconfig"

ASSETCATALOG_COMPILER_APPICON_NAME = BetaAppIcon
```

…and so on for the Nightly and Production builds.

uiryuu:
@saagarjha Any updates? Or should I continue to implement this PR?

saagarjha:
Oh I just self-assigned myself to review the updated version before you merged it

low-batt:
With the latest commit, building and running `iina` crashes out while trying to force unwrap the debug label:
```swift
  static let initialWindowDebugLabel: NSColor = {
    if #available(macOS 10.14, *) {
      return NSColor(named: .initialWindowDebugLabel)!
```

uiryuu:
@low-batt I didn't show the "BETA" and "NIGHLY" label in the about window because we already have the icon which already clearly shows which version we are on, so showing another label looks redundant.

low-batt:
I tested all the different kinds of builds.

Debug builds are properly identified in the log file:
```text
11:54:33.622 [iina][d] IINA 1.3.1 Build 133 DEBUG
```

Nightly and Beta builds are no longer identified:
```text
11:56:33.547 [iina][d] IINA 1.3.1 Build 133
```
This is because isRelease is true for Nightly and Beta builds.
It used to key off the buildType:
```swift
var isRelease: Bool { buildType == .release }
```

But was changed to:
```swift
  #if DEBUG
  let isRelease = false
  #else
  let isRelease = true
  #endif
```

The commit number on the welcome window still needs to be changed to be a link to GitHub.
Same for the about window

Otherwise looking good!

uiryuu:
@low-batt I've updated to make logger pickup the release type.

> The commit number on the welcome window still needs to be changed to be a link to GitHub. Same for the about window

The commit number in the about window is clickable. I don't think it's necessary to make the commit number in the welcome window clickable. If we really want to, that need a better design (other than just make it a NSButton). So I'd leave it as is for this PR.

iina/InfoDictionary.swift
```diff
@@ -44,14 +50,57 @@ struct InfoDictionary {
...
+
   var bundleIdentifier: String { dictionary["CFBundleIdentifier"] as! String }
 
   var copyright: String { dictionary["NSHumanReadableCopyright"] as! String }
 
   let dictionary = Bundle.main.infoDictionary!
 
+  /// A Boolean value that indicates whether this executable was an optimized (not debug) build.
+  #if DEBUG
+  let isRelease = false
```
low-batt: I'm thinking this should be renamed `isDebug`. This is tied to the `Preprocessor Macros` Xcode setting. It is indicating the Debug Xcode build configuration was used.

iina/InfoDictionary.swift
```diff
@@ -44,14 +50,57 @@ struct InfoDictionary {
     bundleIdentifier + ".build"
   }
 
+  /// The type of build used to generate this IINA executable.
+  ///
+  /// This corresponds to the Xcode build configuration.
+  var buildType: BuildType {
+    guard let buildConfiguration = buildConfiguration else { return .nightly }
```
low-batt, line 57: Will other developers be able to explain why this defaults to `.nightly`?

low-batt (commented):
I tested each build configuration and confirmed the logging issue is fixed. However, there is one issue left that I'm thinking needs to be addressed. I completely missed that the commit in the about window was a link because it does not follow the underscored blue text convention for links. As the other links in the about window do adhere to that convention this seems like something that should be fixed. Only the commit number should be linked, not the branch.

Fine to not use a link in the initial window. The about window is the primary source of information.

My two other comments are minor and not critical to address.

PR #4391, opened 2023-05-03
Load external subtitle option opens directory of current video (fixes #4382)

### LINKED ISSUE #4382: Loading subtitles menu action should open containing directory of current video file instead of last opened directory
**What you want IINA to do:**

I would like IINA "load external subtitles..." to open the directory containing the current video file
image

Maybe even auto-detect subtitle files by extension or by subfolder(subs folder for example).

**What IINA does currently:**
Opens the last opened directory opened with this action. For me it just means it opens a different movie folder, sometimes in the "subs" directory of that directory. Before I knew this was the behavior I opened the english subtitles in the to only have loaded the subs for a different video.

Less annoying and confusion.

VLC

smishralm10:

---

**Description:**
This commit will allow `NSOpenPanel` to open in current playing video's directory when loading subtitle from `Load External Subtitle...` option from `Subtitles` menu. This behaviour is already present when loading subtitle from `Quick Settings > Subtitles > External Subtitles:`

### DIFF
--- a/iina/MainMenuActions.swift
+++ b/iina/MainMenuActions.swift
@@ -316,7 +316,8 @@ extension MainMenuActionHandler {
 
 extension MainMenuActionHandler {
   @objc func menuLoadExternalSub(_ sender: NSMenuItem) {
-    Utility.quickOpenPanel(title: "Load external subtitle file", chooseDir: false) { url in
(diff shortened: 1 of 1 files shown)


uiryuu:
Thanks!

PR #4398, opened 2023-05-06
Fix an issue where stop command doesn't close mini player

uiryuu:

---

**Description:**
When a stop command is sent to a main window, the window will close. However, if the stop command is sent to a mini player window, the playback will stop, but the mini player window will still remain.

Steps to reproduce:
- Set a keybinding to the mpv "stop" command
- Open two windows: one of which is in mini player
- Try to press the key to send stop command to two windows and observe the different behavior

This commit closes the mini player on responding to idle-active if in mini player.


low-batt (approved):
Pulled PR, built and tested.
Looks good to me.

Issue #4399, opened 2023-05-07
Window location descends with each invocation


low-batt:
**System and IINA version:**

- macOS 13.3.1 (a)
- IINA 1.3.1

**Expected behavior:**
Initial window position does not descend with each invocation.

**Actual behavior:**
In music mode with cover art hidden and the playlist visible, each time IINA is restarted the initial window screen position is lower on the screen until the OSC portion of the IINA window is at the bottom of the screen and the playlist is not visible as it is below the screen.

**Steps to reproduce:**
- Start IINA playing music
- Hide cover art
- Show playlist
- Quit IINA
- Start IINA playing music again
- Note window screen position is lower
- Quit IINA
- Repeat
- Window will end up starting at the bottom of the screen

- [x] MPV does not have this problem.

**How often does this happen?**
Every time when the instructions are followed.

PR #4409, opened 2023-05-12
Bump the mini player window on opening

uiryuu:
- [x] This implements/fixes issue #4399 .

---

**Description:**
Manually bump up the mini player window position on opening when the video view is hidden by the height of the video view.

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -580,6 +580,9 @@ class PlayerCore: NSObject {
     if needRestoreLayout {
       if !Preference.bool(for: .musicModeShowAlbumArt) {
         miniPlayer.toggleVideoView(self)
+        if let origin = miniPlayer.window?.frame.origin {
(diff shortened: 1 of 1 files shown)


low-batt (approved):
I pulled the PR and tested a lot. This definitely greatly improves the behavior.

I'm approving it based on that testing. However…

This code seems like a workaround for bad code in `MiniPlayerWindowController`? I investigated a tiny bit when I was entering the issue. From what I recall the method `normalWindowHeight` was being called and `isVideoVisible` was still set to its default value of `true`. It seemed like the root cause of the problem had to do with that property not being set to its proper value soon enough. 

The other issue here is proper separation of concerns. IINA needs to be slowly refactoring code into the appropriate classes. I think `PlayerCore` has to direct the transition between the main and mini windows. But I would expect a window's frame to be fully under the control of the associated `NSWindowController`.

uiryuu:
I've tried to set `normalWindowHeight` according to `Preference.bool(for: .musicModeShowAlbumArt)` but that doesn't solve the problem. I agree that this seems like a workaround; if you think this is too workaround to be merged, we can close it and find a better solution.

PR #4417, opened 2023-05-16
chore: Improved a download_libs script

semty:

---

**Description:**

- The script doesn't assume that it is run from a project's root. It will try to find the root folder by itself and stop execution if it is not successful (it means that it's okay now to run it from its own folder, for example)
- A new optional `--arch` option that can take an argument of `universal` | `arm64` | `x86_64` (default: `universal`)
- A new optional `--yt-dlp-src` option that can take an argument of `github` | `iina` (default: `github`)
- A new optional `-h` | `--help` option that shows a short usage help for the script execution


lhc70000:
Thanks!

PR #4425, opened 2023-05-19
chore: Stripped dead code

semty:

---

**Description:**

Stripped dead code throughout the project. 
It's a good practice to perform from time to time, leaving only the code that is actually in use. 

**Note:**

I'm not sure whether this project still accepts pull requests from not core team members. 
@low-batt, it seems to me that you're the most active participant lately, could you, please, help me with that?


low-batt:
@Semty, you definitely do not need to be one of IINA's core developers to propose changes.

IINA is slow at reviewing and merging PRs as the senior developers are very busy. Usually it occurs in spurts when those developers have time to focus on IINA.

I am a newbie junior contributor. I don't merge, but I do review PRs. No promises, but I will try and find time to take a look at this one this weekend. This really needs a review by the senior developers as it covers areas of IINA I am unfamiliar with.

At the moment the senior developers are focused on getting a release out. The release branch as already been cut. The hold up is the need to patch some problems in the updated dependencies.

Commits going into `develop` now will be going into the following release. Which is why this is sort of a good time to clear out dead code. One concern is that there is resistance to refactoring/cleanup with so many PRs outstanding over the fear of introducing lots of merge conflicts.

While on the topic of contributing I should point out that although [CONTRIBUTING](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md) indicates the latest version of Xcode must be used, at the moment we are building with Xcode 13 as well as Xcode 14 in order to support macOS 10.11 and 10.12, which Xcode 14 dropped.

semty:
> @Semty, you definitely do not need to be one of IINA's core developers to propose changes.
> 
> 
> 
> IINA is slow at reviewing and merging PRs as the senior developers are very busy. Usually it occurs in spurts when those developers have time to focus on IINA.
> 
> 
> 
> I am a newbie junior contributor. I don't merge, but I do review PRs. No promises, but I will try and find time to take a look at this one this weekend. This really needs a review by the senior developers as it covers areas of IINA I am unfamiliar with.
> 
> 
> 
> At the moment the senior developers are focused on getting a release out. The release branch as already been cut. The hold up is the need to patch some problems in the updated dependencies.
> 
> 
> 
> Commits going into `develop` now will be going into the following release. Which is why this is sort of a good time to clear out dead code. One concern is that there is resistance to refactoring/cleanup with so many PRs outstanding over the fear of introducing lots of merge conflicts.
> 
> 
> 
> While on the topic of contributing I should point out that although [CONTRIBUTING](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md) indicates the latest version of Xcode must be used, at the moment we are building with Xcode 13 as well as Xcode 14 in order to support macOS 10.11 and 10.12, which Xcode 14 dropped.

@low-batt hey! Got it, thank you so much for your thorough response 👌. 
Let me know if you find anything that should not be removed, I'm open to fix it if needed

iina/MPVCommand.swift
```diff
@@ -10,12 +10,8 @@ struct MPVCommand: RawRepresentable {
 
   init?(rawValue: RawValue) { self.rawValue = rawValue }
 
-  /** ignore */
-  static let ignore = MPVCommand("ignore")
   /** seek <target> [<flags>] */
```
low-batt: My rule is that generated files should never be checked into source control. They should always be regenerated as a part of the build. Too often I have encountered generated files where developers modified the generated files instead of correcting the true source or the generator that created the generated source. I have even countered the situation where the true source and the generator have been lost.

IINA is violating this rule. The swift classes `MPVCommand`, `MPVOption` and `MPVProperty` are all generated and must not be modified by developers. These files are generated from mpv documentation by the script `other/parse_doc.rb` and then checked into git. This is done whenever IINA upgrades to a newer version of the mpv library.

Another of my own rules is being violated, which is that all source code generator must start the generated code with a BIG comment that tries to make it clear to developers that this is a generated file that must not be directly modified by developers. Possibly that comment should contain details about the generation. In this case it probably should indicate the mpv version the source was generated from.

Within Xcode these sources aren't even in their own "Generated" group. The [Current structure](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md#current-structure) section of the [Contributing](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md) document does not warn contributors about these generated sources.

I just entered issue #4429 that suggests improvements in this area.

iina/MPVController.swift
```diff
@@ -1394,7 +1360,7 @@ class MPVController: NSObject {
   }
 }
 
-fileprivate func mpvGetOpenGLFunc(_ ctx: UnsafeMutableRawPointer?, _ name: UnsafePointer<Int8>?) -> UnsafeMutableRawPointer? {
+fileprivate func mpvGetOpenGLFunc(_: UnsafeMutableRawPointer?, _ name: UnsafePointer<Int8>?) -> UnsafeMutableRawPointer? {
```
low-batt, line 1363: Will have to see what the senior developers think, but I'm strongly against removing the name of parameters, even if they are not used. After this change a developer reading this code can't tell what it is that is being passed and ignored.

iina/MPVOption.swift
```diff
@@ -6,462 +6,102 @@ struct MPVOption {
     static let alang = "alang"
     /** --slang=<languagecode[ */
     static let slang = "slang"
-    /** --vlang=<...> */
-    static let vlang = "vlang"
     /** --aid=<ID|auto|no> */
```
low-batt: Generated source, must not be modified.

iina/MPVProperty.swift
```diff
@@ -1,174 +1,62 @@
...
-  static let audioSpeedCorrection = "audio-speed-correction"
-  /** video-speed-correction */
-  static let videoSpeedCorrection = "video-speed-correction"
-  /** display-sync-active */
-  static let displaySyncActive = "display-sync-active"
-  /** filename */
-  static let filename = "filename"
-  /** filename/no-ext */
-  static let filenameNoExt = "filename/no-ext"
   /** file-size */
```
low-batt: Generated source, must not be modified.

iina/ShooterSubtitle.swift
```diff
@@ -148,7 +140,3 @@ class Shooter {
     }
   }
 }
-
-extension Logger.Sub {
-  static let shooter = Logger.makeSubsystem("sub.shooter")
-}
```
low-batt: This was added in anticipation of log messages being added by someone who knows this provider.

There are problems with not being able to tell what is going wrong when trying to find and download subtitles. Some refactoring was done to share logging and reporting in `OnlineSubtitle`. However as can be seen in `OpenSubSubtitle` some logging should be done by the specific provider class.

iina/VideoView.swift
```diff
@@ -454,10 +452,10 @@ extension VideoView {
...
 fileprivate func displayLinkCallback(
-  _ displayLink: CVDisplayLink, _ inNow: UnsafePointer<CVTimeStamp>,
-  _ inOutputTime: UnsafePointer<CVTimeStamp>,
-  _ flagsIn: CVOptionFlags,
-  _ flagsOut: UnsafeMutablePointer<CVOptionFlags>,
+  _: CVDisplayLink, _: UnsafePointer<CVTimeStamp>,
+  _: UnsafePointer<CVTimeStamp>,
+  _: CVOptionFlags,
+  _: UnsafeMutablePointer<CVOptionFlags>,
   _ context: UnsafeMutableRawPointer?) -> CVReturn {
```
low-batt, line 459: Although unused with respects to the code, these parameter names effectively provide documentation for developers reading the code. See what the senior developers say, but I am against removing parameter names.

iina/ViewLayer.swift
```diff
@@ -196,39 +196,6 @@ class ViewLayer: CAOpenGLLayer {
 
   // MARK: Utils
 
-  /** Check OpenGL error (for debug only). */
```
low-batt, line 199: As the comments indicates this is a debugging tool for developers and must not be removed. I'm thinking the correct change may be to use a Swift [Conditional Compilation Block](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements/#Conditional-Compilation-Block) to avoid including dead code. See what the senior developers think.

low-batt (changes requested):
I pulled the PR locally, built IINA and confirmed IINA started and was able to play a video.

I'm unfamiliar with a lot of this code, so the real review must be done by the senior developers. I had planned on not requesting changes and leaving it to those developers to request changes as I am unsure as to whether they would agree with some of my comments. However…

I'm certain this PR in its current state will be rejected due to including changes to generated sources. This is an existing pitfall for new contributors. I have entered issue #4429 for the failure to properly identify generated sources. See that issue and my inline comment.

On my other comments, feel free to wait and see what the other developers think.

Thank you for proposing these changes to IINA!

iina/MPVCommand.swift
semty: @low-batt hey, thanks for your review! I didn't really know that these classes are auto-generated, I agree with your take here. 
Fixed in ce4a8bca1c16f0f7077a75b0fbe75b010cc44c95

iina/MPVOption.swift
semty: Fixed in ce4a8bca1c16f0f7077a75b0fbe75b010cc44c95

iina/MPVProperty.swift
semty: Fixed in ce4a8bca1c16f0f7077a75b0fbe75b010cc44c95

iina/VideoView.swift
semty: Hmm, I tend to disagree here. My two points:
1. In my opinion, it actually better to omit some stuff that you don't need to perform your logic, you have less things to grasp and see only the important ones
2. If you want to change/debug this piece of code, it's better if you're familiar with it a bit more than just knowing the parameter names. It can be done really easily by checking where this function is used (`CVDisplayLinkSetOutputCallback`) and locating the parameter (`CVDisplayLinkOutputCallback`). After that you'll see the complete documentation with everything you may need

But, of course, I understand that it can be a hot topic, so I'm fully okay to return parameter names in every function even if it's not used

iina/ViewLayer.swift
semty: @low-batt for the time being, I just returned the old code, fixed here: 
- c6e3bbc32ef70daf08139d84cfb6828ce61aebaa

iina/ShooterSubtitle.swift
semty: @low-batt Fixed: 6b202ba4703d6a8885379596bb79c0a96e4fa2e2

iina/VideoView.swift
low-batt: On this one lets see what the senior developers think.

iina/ViewLayer.swift
low-batt: Sounds good.

iina/AppDelegate.swift
```diff
@@ -14,8 +14,6 @@ let IINA_ENABLE_PLUGIN_SYSTEM = Preference.bool(for: .iinaEnablePluginSystem)
 
 /** Max time interval for repeated `application(_:openFile:)` calls. */
 fileprivate let OpenFileRepeatTime = TimeInterval(0.2)
-/** Tags for "Open File/URL" menu item when "Always open file in new windows" is off. Vice versa. */
-fileprivate let NormalMenuItemTag = 0
```
saagarjha, line 18: As far as I can tell this was never used, perhaps worth adjusting the code to reflect that?

iina/Aspect.swift
```diff
@@ -12,24 +12,6 @@ class Aspect: NSObject {
...
-
-  var height: CGFloat {
-    get {
-      return size.height
-    }
-    set {
-      size.height = newValue
-    }
-  }
-
```
saagarjha: We should probably keep this even if it's unused

iina/CollapseView.swift
```diff
@@ -55,7 +55,7 @@ class CollapseView: NSStackView {
     updateContentView(animated: animated)
   }
 
-  @objc private func triggerAction(_ sender: NSControl) {
+  @objc private func triggerAction() {
```
saagarjha, line 58: You can't remove this parameter, see https://developer.apple.com/library/archive/documentation/General/Conceptual/Devpedia-CocoaApp/TargetAction.html#//apple_ref/doc/uid/TP40009071-CH3-SW2

iina/Extensions.swift
```diff
@@ -154,11 +130,6 @@ extension NSSize {
   func multiply(_ multiplier: CGFloat) -> NSSize {
     return NSSize(width: width * multiplier, height: height * multiplier)
   }
-
-  func add(_ multiplier: CGFloat) -> NSSize {
```
saagarjha, line 158: Probably worth keeping this but with the right parameter name

iina/InfoDictionary.swift
```diff
@@ -70,13 +70,6 @@ struct InfoDictionary {
 
   let dictionary = Bundle.main.infoDictionary!
 
-  /// A Boolean value that indicates whether this executable was an optimized (not debug) build.
-  #if DEBUG
-  let isDebug = true
-  #else
-  let isDebug = false
-  #endif
-
```
saagarjha, line 79: Pretty sure this was just added recently, @low-batt you should probably look at this and see if you needed it

iina/JavascriptPlugin.swift
```diff
@@ -334,10 +333,6 @@ class JavascriptPlugin: NSObject {
     self.helpPage = jsonDict["helpPage"] as? String
     self.domainList = (jsonDict["allowedDomains"] as? [String]) ?? []
     self.subProviders = jsonDict["subtitleProviders"] as? [[String: String]]
-    
-    if externalURL != nil {
-      self.isExternal = true
-    }
```
saagarjha, line 340: cc @lhc70000 did you intend to do something with this

saagarjha (changes requested):
I appreciate konmari efforts. Lots of this needs to go. Deciding which parts, though, is complicated. In general, these things should probably not be removed:

* Sets of related properties or methods that exist for completeness: e.g. if we have setString, setInt, setDouble, … it doesn't make sense to remove one of these because it's not used. People expect the full set of APIs to be available.
* Things that were just added and intend to be used in the near future. Determining this will probably be hard. cc @lhc70000 @uiryuu since they should sign off on this.

Marking for changes because some of these definitely need to be rolled back–see the comments for the ones I spotted.

iina/CollapseView.swift
semty: @saagarjha hey! thank you for your review! 
> You can't remove this parameter, see https://developer.apple.com/library/archive/documentation/General/Conceptual/Devpedia-CocoaApp/TargetAction.html#//apple_ref/doc/uid/TP40009071-CH3-SW2

are you sure? it works on iOS to omit the sender

iina/Aspect.swift
semty: @saagarjha fixed: https://github.com/iina/iina/pull/4425/commits/1d610bdd56709994411939c29b895d90177539c1

iina/Extensions.swift
semty: @saagarjha fixed: https://github.com/iina/iina/pull/4425/commits/676b2847047bc273c39582aa00f1308e45ef1373

iina/InfoDictionary.swift
low-batt: Yes, this was recently added as a part of issue #4383 / PR #4385. However my role in that issue was to just get it started and hand it off to @uiryuu. Will have to ask @uiryuu about this change.

iina/CollapseView.swift
saagarjha: It's [legal on iOS](https://developer.apple.com/documentation/uikit/uicontrol#1658488), but not on macOS. Things will mostly still work because the argument register will probably not get touched but it's good to do it correctly.

iina/Extensions.swift
```diff
@@ -154,11 +130,10 @@ extension NSSize {
...
-    return NSSize(width: width + multiplier, height: height + multiplier)
+
```
uiryuu, line 133: Remove spaces

iina/InfoDictionary.swift
uiryuu: Let keep this, although there isn't any code using this. We might want to use this in the future.

iina/MPVController.swift
uiryuu: +1 for not changing

iina/ViewLayer.swift
```diff
@@ -195,7 +195,7 @@ class ViewLayer: CAOpenGLLayer {
   }
 
   // MARK: Utils
-
+
```
uiryuu, line 198: Space

uiryuu (changes requested):

iina/MPVController.swift
```diff
@@ -554,21 +541,6 @@ class MPVController: NSObject {
...
-  func setIntAsync(_ name: String, _ value: Int) {
-    var data = Int64(value)
-    mpv_set_property_async(mpv, 0, name, MPV_FORMAT_INT64, &data)
-  }
-
-  func setDoubleAsync(_ name: String, _ value: Double) {
-    var data = value
-    mpv_set_property_async(mpv, 0, name, MPV_FORMAT_DOUBLE, &data)
-  }
-
```
uiryuu, line 571: These functions were added intentionally, in the future we might want to use them. Please convert this back.

iina/Aspect.swift
```diff
@@ -11,7 +11,7 @@ import Cocoa
 class Aspect: NSObject {
 
   private var size: NSSize!
-
```
uiryuu, line 14: Space

iina/Aspect.swift
```diff
@@ -20,7 +20,7 @@ class Aspect: NSObject {
       size.width = newValue
     }
   }
-
```
uiryuu, line 23: Space

iina/VideoView.swift
uiryuu: I'd like to keep it as it is.

uiryuu (changes requested):
Just drop some comments here, will take a deeper look another day. Merge conflicts look scary but are surprisingly trivial.

uiryuu:
Merged via https://github.com/iina/iina/commit/f121558adb89d8b9ac55e313bbb65cf71898d2d9. Thanks @Semty!

PR #4439, opened 2023-06-01
Send directory arg to mpv instead of processing in IINA (fixes #4434)

### LINKED ISSUE #4434: iina-cli --mpv-shuffle does not shuffle playlist
**System and IINA version:**

- macOS 13.4
- IINA 1.3.1

**Expected behavior:**
Executing `iina-cli --mpv-shuffle ~/Movies`  plays files in random order as per the documented behavior of the mpv [shuffle](https://mpv.io/manual/stable/#options-shuffle) option.

**Actual behavior:**
The files in the directory are always played in the same order.

From a quick look at the code, it appears IINA is processing the directory itself and adding the files to the playlist one by one. Possibly if IINA passed the directory to mpv it would load the files and shuffle the playlist. If that is not feasible then IINA needs to recognize and implement this mpv option.

<br>

**Steps to reproduce:**
- Put multiple video files into a directory
- Start IINA using `iina-cli`, specifying `--mpv-shuffle` and the directory containing the video files
- Click on `Show Playlist Panel` under the `Playback` menu
- Take note of the order of the files in the playlist
- Repeat
- The same file starts playing and the playlist order has not changed

Every time.

svobs:
- [x] This implements/fixes issue #4434.

---

**Description:**

This simple change checks if there is only one path arg given, and if that path arg is a local directory. If true, sends it to mpv to process instead of processing it directly via IINA's code. This allows `iina-cli --mpv-shuffle` to work properly, at least for the case where there is only one path arg given. This should provide a workable solution for most users. (If one were to try implementing this for multiple path args and allow for fully compatibility with mpv while not breaking IINA's directory loading features, it looks like larger structural changes would be needed).

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -294,12 +294,18 @@ class PlayerCore: NSObject {
     guard !urls.isEmpty else { return 0 }
     let urls = Utility.resolveURLs(urls)
 
-    // handle BD folders and m3u / m3u8 files first
-    if urls.count == 1 && (isBDFolder(urls[0]) ||
-      Utility.playlistFileExt.contains(urls[0].absoluteString.lowercasedPathExtension)) {
-      info.shouldAutoLoadFiles = false
-      open(urls[0])
-      return nil
+    // Handle folder URL (to support mpv shuffle, etc), BD folders and m3u / m3u8 files first.
+    // For these cases, mpv will load/build the playlist and notify IINA when it can be retrieved.
+    if urls.count == 1 {
+      let url = urls[0]
+
+      if url.isExistingDirectory
+          || isBDFolder(url)
+          || Utility.playlistFileExt.contains(url.absoluteString.lowercasedPathExtension) {
+        info.shouldAutoLoadFiles = false
+        open(url)
+        return nil
+      }
     }
 
     let playableFiles = getPlayableFiles(in: urls)


low-batt:
Changes look good, but I think there is another defect lurking. The `iina-cli` does not require use of the `--mpv-` prefix:
```text
mpv Option:
Raw mpv options without --mpv- prefix. All mpv options are supported here.
Example: --volume=20 --no-resume-playback
```

But for some reason only `--mpv-shuffle` is working for me. When I tested `--shuffle` the playlist was not shuffled. Possibly something is wrong with argument processing in `AppDelegate`.

svobs:
Well...this is interesting. It seems that `CommandLineStatus` was putting args which don't start with `--mpv-` into a separate array called `iinaArguments`, and then...doing nothing with that array ever after. The end result is that mpv options without the `--mpv-` prefix are being ignored.

This behavior doesn't make sense. Very odd that this wasn't noticed before. I checked some previous commits, going as far as the `v1.3.1` tag, but it seems this behavior is not new. I worked on this code before but somehow missed it.

Maybe it's is a bit scope creepy, but I made another commit which removes the unused array and instead passes the previously unhandled arguments to the mpv core. This will allow `--shuffle` (and other mpv options) to be used without needing the prefix.

low-batt:
Yes, odd that this is so broken. The `iina-cli` help is very clear that this should work.

Testing with:
```bash
low-batt@gag MacOS$ ./iina-cli --shuffle --pip  ~/Movies/
```

The mpv log shows:
```text
[   0.040][v][cplayer] Set property: shuffle="yes" -> 1
[   0.040][v][cplayer] Set property: pip="yes" -> -3
```

Shuffle is working now. I'm thinking IINA should be filtering out the known `iina-cli` options such as `pip`.

svobs:
OK. I see now I'm getting pulled deeper into this. Rabbit holes everywhere! 🤣

> I'm thinking IINA should be filtering out the known iina-cli options such as pip.

Yes, you're right. The IINA args should be processed before the mpv args, and should not be passed to mpv. I see that een with my fix, they are not.

I think I should back out my `CommandLineStatus` changes and put them into a new issue. It is a completely separate problem than the one this PR was meant to address, and will require more new lines of code than the original problem did.

low-batt:
Agree. The problem with mpv options can be a separate issue.

low-batt (approved):
Tested and `iina-cli --mpv-shuffle` is working now.

uiryuu:
Thanks!

PR #4442, opened 2023-06-01
Creating plugins using iina-cli

lhc70000:
**Description:**

- Added plugin-related commands to `iina-cli`.
  - `create`: See https://github.com/iina/iina-plugin-template for more description.
  - `pack`: Create a zip file with the plugin folder's content.
  - `link`: Create a symlink in IINA's plugin folder.
  - `unlink`: Remove the above symlink.
- Fixed a crash when installing local `.iinaplgz` plugins.

Recommended steps for testing:

- Make sure that `iina-cli` can create plugins using various options.
- If building is required for the generated plugin (`iina-cli` will print instructions), make sure `npm install` and `npm run build` finish successfully.
- Make sure that `link` and `unlink` work for the generated plugin folder, i.e. IINA can load the folder as a plugin correctly.
- Make sure IINA can load the zip archive generated by `pack`.

PR #4451, opened 2023-06-04
Cmd line fixes: pass raw mpv args to mpv (fixes #4450) + clean up broken code

### LINKED ISSUE #4450: IINA command line mpv args without "--mpv" prefix are ignored
**System and IINA version:**

- macOS 13.3.1
- IINA develop branch, rev c0c859520c656a9308de11ab3d74d058eb26c3e3

In the discussion for PR #4439, @low-batt discovered that raw mpv args do not appear to be working.

The IINA usage message includes support for raw mpv options:
```
    mpv Option:
    Raw mpv options without --mpv- prefix. All mpv options are supported here.
    Example: --volume=20 --no-resume-playback
```

**Steps to reproduce:**

1. Open IINA.
2. Set the volume to 100.
3. Quit IINA.
4. Launch IINA with `--volume=20` to set the volume to 20 at start: `./IINA.app/Contents/MacOS/iina-cli --volume=20 test-video.mp4`
5. Check the volume.

**Expected behavior:**

The volume should be set to 20 by mpv.

**Actual behavior:**

The volume is unchanged. Also checking the IINA log, the argument is received by IINA but is not passed to mpv:

```
00:01:10.838 [iina][d] IINA arguments: ["--volume=20"]
[...]
00:01:11.114 [iina][d] Using mpv 0.35.1
00:01:11.114 [iina][d] Setting mpv properties from arguments: []
```

Always

svobs:
- [x] This implements/fixes issue #4450.

---

**Description:**
Attempting to fix #4450 was much easier to do after some refactoring of the command-line parsing logic. It's still spread over 2 files, which is hard to avoid because not all of the code is visible to `main.swift`, but it's been cleaned up to remove some redundant logic and dead code, and it's also been consolidated in places to be more readable.

Summary:
* Arg parsing in `main.swift` has been combined into one loop which builds a new array of args (to send to `AppDelegate`) rather than changing the existing array. The new array also omits args which are no longer needed, so that they will not be inadvertantly sent to mpv.
* Deleted some logic in `main.swift` which looks like it was trying to add `--mpv-` to some args, but appeared to be broken and also unnecessary
* The arg parsing logic in `AppDelegate.swift` has been combined into one loop inside the `CommandLineStatus` class (instead of dividing between it and the`AppDelegate` class). 
* Deleted some redundant variables from `AppDelegate.swift` which were unnecessary.
* Changed the parsing logic in `CommandLineStatus` so that any args starting with `--` and not recognized as IINA args will now be sent to mpv.

### DIFF
--- a/iina-cli/main.swift
+++ b/iina-cli/main.swift
@@ -8,34 +8,7 @@
 
 import Foundation
 
-guard var execURL = Bundle.main.executableURL else {
-  print("Cannot get executable path.")
-  exit(1)
-}
-
-let currentDirURL = URL(fileURLWithPath: FileManager.default.currentDirectoryPath)
-
-execURL.resolveSymlinksInPath()
-
-let processInfo = ProcessInfo.processInfo
-
-let iinaPath = execURL.deletingLastPathComponent().appendingPathComponent("IINA").path
-
-guard FileManager.default.fileExists(atPath: iinaPath) else {
-  print("Cannot find IINA binary. This command line tool only works in IINA.app bundle.")
-  exit(1)
-}
-
-let task = Process()
-task.launchPath = iinaPath
-
-var keepRunning = false
-
-// Check arguments
-
-var userArgs = Array(processInfo.arguments.dropFirst())
-
-if userArgs.contains(where: { $0 == "--help" || $0 == "-h" }) {
+func printHelp() {
   print(
     """
     Usage: iina-cli [arguments] [files] [-- mpv_option [...]]
@@ -66,8 +39,32 @@ if userArgs.contains(where: { $0 == "--help" || $0 == "-h" }) {
     Raw mpv options without --mpv- prefix. All mpv options are supported here.
     Example: --volume=20 --no-resume-playback
     """)
-  exit(0)
 }
+
+guard var execURL = Bundle.main.executableURL else {
+  print("Cannot get executable path.")
+  exit(1)
+}
+
+let currentDirURL = URL(fileURLWithPath: FileManager.default.currentDirectoryPath)
+
+execURL.resolveSymlinksInPath()
+
+let processInfo = ProcessInfo.processInfo
+
+let iinaPath = execURL.deletingLastPathComponent().appendingPathComponent("IINA").path
+
+guard FileManager.default.fileExists(atPath: iinaPath) else {
+  print("Cannot find IINA binary. This command line tool only works in IINA.app bundle.")
+  exit(1)
+}
+
+let task = Process()
+task.launchPath = iinaPath
+
+// Check arguments
+
+let userArgs = Array(processInfo.arguments.dropFirst())
 
 if userArgs.contains("--music-mode"), userArgs.contains("--pip") {
   // Music mode does not support Picture-in-Picture. Combining these options is not permitted.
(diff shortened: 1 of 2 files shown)


svobs:
It looks like I keep angering the CI gods 😖

My best guess is there is a problem when there is a new commit to `develop`, causing my branch to be out of date in relation to the `develop` branch. Then when the CI build is triggered, something is missing which causes a linker error. But it looks like I can't fix the CI build once it's been broken, even if I bring it up to date...?

low-batt:
This PR is incorrect and it is my fault. It seems I have led you astray. There are multiple defects here, but not what I thought. The command line shown in issue #4450:
```text
iina-cli --volume=20 test-video.mp4
```

Should have elicited this response from `iina-cli`:
```text
iina-cli: unrecognized option '--volume=20'
```

If it had, I might have recognized my error. So this is the first change that is needed, better error reporting. This is the correct way to pass a raw mpv option without the `--mpv-` prefix:
```text
iina-cli ~/Movies/Airplaneski.mp4 -- --volume=20
```

The end of the `iina-cli` help message says:
```text
    mpv Option:
    Raw mpv options without --mpv- prefix. All mpv options are supported here.
    Example: --volume=20 --no-resume-playback
```

Another improvement is to change the example to show the complete command line:
```text
iina-cli movie.mp4 -- --volume=20 --no-resume-playback
```

The "mpv Option" is referring to "mpv_option" in the start of the help message:
```text
Usage: iina-cli [arguments] [files] [-- mpv_option [...]]
```

The `[arguments]` should be changed to use normal bash terminology, `[options]`. Maybe something more like:

```text
Usage: iina-cli [options] [files] [-- raw_mpv_option [...]]
```

The help is trying to indicate that raw mpv options without the `--mpv-` prefix can be used after the bash end of options delimiter (`--`). When invoked with the command structure `iina-cli` expects, `--shuffle` works correctly.

Sorry! I totally missed what the `iina-cli` help was trying to convey.

PS: The check builds are failing due to changes to dependencies being made as a part of trying to get the release out the door. They need to update the Xcode project. That should happen soon.

svobs:
I see. That puts some of the code into more perspective. I should have paid more attention to the first line of the usage message.

I agree that the help could use some improvement. I think there should also be 1-2 lines detailing the special handling for mpv args starting with `--no-`, e.g.,
```
iina-cli movie.mp4 -- --volume=20 --no-resume-playback
will send these args to mpv: --volume=20 --resume-playback=no
```

Definitely agree that it should complain about unrecognized options. In order to do that, both `main.swift` and `AppDelegate.swift` will need to know about all the arguments. This will require rewriting all the code I've written. And since this turned out to not be an urgent defect, it may be a better use of time to just trash what I've written and go on to other things.

However...I'm a sucker for sunk costs, and I already thought up a more elegant solution, which involves pulling out the `CommandLineStatus` class into its own file which can be put under the `Utils` group, and using it exclusively for parsing command-line options. Then the file could be added to the `iina-cli` project so it can also be used in `main.swift`. I'll see if I get around to it, but that will be a different PR.

low-batt:
Sorry about missing how this works and pointing you in the wrong direction.

At the moment work on `iina-cli` can run into merge conflicts with PR #4442. You will see one of my review comments on that one suggests a separate tool is needed. If that is followed then `CommandLineStatus` may be broken out as a part of that work.

PR #4453, opened 2023-06-05
Fix heap buffer overflow in levDistance, #4449

### LINKED ISSUE #4449: Heap buffer overflow in wcslen called from levDistance
**System and IINA version:**

- macOS 13.4
- IINA `develop` a3442b0bec90e79517aaedda67a2dd3ae3fde463

**Expected behavior:**
IINA running with the address sanitizer enabled does not trigger a crash.

**Actual behavior:**
The address sanitizer is reporting a heap buffer overflow on the last line of this code from `ObjUtils.levDistance`:
```objc
const wchar_t *cstr0 = (const wchar_t *)[str0 cStringUsingEncoding:NSUTF32LittleEndianStringEncoding];
const wchar_t *cstr1 = (const wchar_t *)[str1 cStringUsingEncoding:NSUTF32LittleEndianStringEncoding];
size_t len0 = wcslen(cstr0);
```

The Apple documentation for [cStringUsingEncoding](https://developer.apple.com/documentation/foundation/nsstring/1408489-cstringusingencoding#1679638) contains this section:
> ## Special Considerations
>
>UTF-16 and UTF-32 are not considered to be C string encodings, and should not be used with this method—the results of passing NSUTF16StringEncoding, NSUTF32StringEncoding, or any of their variants are undefined.

That seems to indicate it is invalid to use  `NSUTF32LittleEndianStringEncoding` with this method.

At the point of the failure `str0` looks fine, but `cstr0` looks like the `null` terminator is missing:
```text
AddressSanitizer report breakpoint hit. Use 'thread info -s' to get extended information about the report.
(lldb) frame variable
(const Class) self = ObjcUtils
(SEL) _cmd = "levDistance:and:"
(__NSCFString *) str0 = 0x0000000119d09990 @" 01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789"
(__NSCFString *) str1 = 0x0000000119d099d0 @" [1]Airplane.1980.BluRay.720p.DTS.x264-CHD.eng"
(int) i = 0
(int) j = 45
(const wchar_t *) cstr0 = 0x000000010d17a880 L" 01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd\U0000fffd"
(const wchar_t *) cstr1 = 0x000000012e144c70 L" [1]Airplane.1980.BluRay.720p.DTS.x264-CHD.eng"
(size_t) len0 = 1
(size_t) len1 = 1
(int *) _dist = 0x000000010baf63d0
(int (*)[][]) dist = 0x000000010baf63d0
(int) result = 0
(lldb) 
```

<br>

**Steps to reproduce:**
This problem is elusive and hard to reproduce.

Create a directory and populate it with 3 files named exactly as seen here:
```text
low-batt@gag issue-heap-overflow$ ls -l
total 1827392
-rwxrwxrwx@ 1 low-batt  staff     373485 Jul 29  2021 01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789.mp4
-rwxrwxrwx@ 1 low-batt  staff  935151008 Apr  9  2022 Big Buck Bunny 60fps 4K - Official Blender Foundation Short Film [aqz-KE-bpKQ].mp4
-rw-r--r--  1 low-batt  staff      91241 Dec 13  2021 [1]Airplane.1980.BluRay.720p.DTS.x264-CHD.eng.srt
low-batt@gag issue-heap-overflow$ 
```

In Xcode edit the scheme and enable `Address Sanitizer`.

In IINA settings make sure `Auto load` is set to `Subtitles containing media filename` in the `Subtitles` tab.

Play `Big Buck Bunny` from that directory. Hopefully the address santizier will report the problem.

Every time for me with the specific input described above.

low-batt:
This commit will:
- Remove use of wcslen from levDistance
- Correct off by one indexing in levDistance

This corrects two separate issues with reading memory outside of bounds that were detected by the address sanitizer.

- [x] This implements/fixes issue #4449.

---

**Description:**

### DIFF
--- a/iina/ObjcUtils.m
+++ b/iina/ObjcUtils.m
@@ -48,19 +48,22 @@ static inline int min(int a, int b, int c) {
 + (NSUInteger)levDistance:(NSString *)str0 and:(NSString *)str1 {
   int i, j;
   
-  str0 = [@" " stringByAppendingString:str0];
-  str1 = [@" " stringByAppendingString:str1];
+  // Prepend a blank space and add a null terminator. The leading space is required so the strings
+  // can be accessed using the one based indexes in the last loop below. The string must be null
+  // terminated to be able to use the wcslen method and because the loop expects to be able to
+  // access a null at the end of the string.
+  str0 = [@" " stringByAppendingFormat:@"%@\0", str0];
+  str1 = [@" " stringByAppendingFormat:@"%@\0", str1];
 
   // Convert from variable length character encoding to fixed length UTF-32 to make it easy to
   // access individual characters.
   const NSData *str0AsUTF32 = [str0 dataUsingEncoding:NSUTF32LittleEndianStringEncoding];
   const NSData *str1AsUTF32 = [str1 dataUsingEncoding:NSUTF32LittleEndianStringEncoding];
-  const NSUInteger len0 = str0AsUTF32.length / sizeof(wchar_t);
-  const NSUInteger len1 = str1AsUTF32.length / sizeof(wchar_t);
 
-  // CAUTION these strings are not null terminated.
   const wchar_t *cstr0 = (const wchar_t *)[str0AsUTF32 bytes];
   const wchar_t *cstr1 = (const wchar_t *)[str1AsUTF32 bytes];;
+  const size_t len0 = wcslen(cstr0);
+  const size_t len1 = wcslen(cstr1);
 
   int *_dist = malloc(sizeof(int) * (len0 + 1) * (len1 + 1));
   int (*dist)[len0 + 1][len1 + 1] = (int (*)[len0 + 1][len1 + 1])_dist;
@@ -77,7 +80,7 @@ static inline int min(int a, int b, int c) {
     for (i = 1; i <= len0; ++i)
       (*dist)[i][j] = min((*dist)[i - 1][j] + INDEL_WEIGHT,
                           (*dist)[i][j - 1] + INDEL_WEIGHT,
-                          (*dist)[i - 1][j - 1] + (cstr0[i - 1] == cstr1[j - 1] ? 0 : SUBSTITUTION_WEIGHT));
+                          (*dist)[i - 1][j - 1] + (cstr0[i] == cstr1[j] ? 0 : SUBSTITUTION_WEIGHT));
   
   int result = (*dist)[len0][len1];
   free(_dist);
--- a/iina/ObjcUtils.m
+++ b/iina/ObjcUtils.m
@@ -50,11 +50,18 @@ static inline int min(int a, int b, int c) {
   
   str0 = [@" " stringByAppendingString:str0];
   str1 = [@" " stringByAppendingString:str1];
-  const wchar_t *cstr0 = (const wchar_t *)[str0 cStringUsingEncoding:NSUTF32LittleEndianStringEncoding];
-  const wchar_t *cstr1 = (const wchar_t *)[str1 cStringUsingEncoding:NSUTF32LittleEndianStringEncoding];
-  size_t len0 = wcslen(cstr0);
-  size_t len1 = wcslen(cstr1);
-  
+
+  // Convert from variable length character encoding to fixed length UTF-32 to make it easy to
+  // access individual characters.
+  const NSData *str0AsUTF32 = [str0 dataUsingEncoding:NSUTF32LittleEndianStringEncoding];
+  const NSData *str1AsUTF32 = [str1 dataUsingEncoding:NSUTF32LittleEndianStringEncoding];
+  const NSUInteger len0 = str0AsUTF32.length / sizeof(wchar_t);
+  const NSUInteger len1 = str1AsUTF32.length / sizeof(wchar_t);
+
+  // CAUTION these strings are not null terminated.
+  const wchar_t *cstr0 = (const wchar_t *)[str0AsUTF32 bytes];
+  const wchar_t *cstr1 = (const wchar_t *)[str1AsUTF32 bytes];;
+
   int *_dist = malloc(sizeof(int) * (len0 + 1) * (len1 + 1));
   int (*dist)[len0 + 1][len1 + 1] = (int (*)[len0 + 1][len1 + 1])_dist;
   for (i = 0; i <= len0; ++i)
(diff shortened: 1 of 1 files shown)


low-batt:
Passing `NSUTF32LittleEndianStringEncoding` to `cStringUsingEncoding` is invalid as discussed in the issue. It seems it is returning a string without the null terminator. A lot of the time the buffer has zeros, so `wcslen` happens to work. But if you enable something like malloc scribble, bad things happen.

In addition there was a separate issue with indexing into the strings with a one based index.

iina/ObjcUtils.m
```diff
@@ -70,7 +77,7 @@ + (NSUInteger)levDistance:(NSString *)str0 and:(NSString *)str1 {
     for (i = 1; i <= len0; ++i)
       (*dist)[i][j] = min((*dist)[i - 1][j] + INDEL_WEIGHT,
                           (*dist)[i][j - 1] + INDEL_WEIGHT,
-                          (*dist)[i - 1][j - 1] + (cstr0[i] == cstr1[j] ? 0 : SUBSTITUTION_WEIGHT));
+                          (*dist)[i - 1][j - 1] + (cstr0[i - 1] == cstr1[j - 1] ? 0 : SUBSTITUTION_WEIGHT));
```
uiryuu: This is not off-by-one error. This is part of the state transition equation (`dist[i][j]` is updated by `dist[i - 1][j - 1]` and the cost at (i, j))

uiryuu (changes requested):

iina/ObjcUtils.m
low-batt: Been too long since I worked on this so I'm still trying to remember clearly, plus a bit burned out tonight, however I believe this is the issue:

`dist` is allocated like so with len + 1:
```swift
  int *_dist = malloc(sizeof(int) * (len0 + 1) * (len1 + 1));
```

The indexes start at 1 and go to len:
```swift
  for (j = 1; j <= len1; ++j)
    for (i = 1; i <= len0; ++i)
```

This all good for accessing `dist` with `[i - 1][j - 1]`.

But this code is accessing the 0 based strings with a 1 based index:
```swift
(*dist)[i - 1][j - 1] + (cstr0[i] == cstr1[j] ? 0 : SUBSTITUTION_WEIGHT));
```

That means it skips the over the first character in the two strings since `i` and `j` start at 1. That seems very wrong.

And since the for loop goes to `<= len` when the index gets to the length of the string the index is out of bounds, triggering the address sanitizer. 

I think the change is needed and correct? At least there is no question something is wrong here as I just confirmed that the address sanitizer triggers if I back out that change.

I am going to push a rebase with develop to get the CI build to pass.

uiryuu: > But this code is accessing the 0 based strings with a 1 based index:

On line 51 & 52, I added a space before the 2 strings. Then, in the transition equation, we have already covered all 3 initial possibilities (0, 1) (1, 0) (1, 1). There's no reason for (0, 0) since the 0th character is the space.

> And since the for loop goes to <= len when the index gets to the length of the string the index is out of bounds, triggering the address sanitizer.

This is because previously, the strings are null terminated (C-style strings), thus accessing the len'th character won't cause any issue. Since you have changed it to non null-terminated, you should add another character ('\0' is a good candidate) at the end of each strings.

uiryuu (changes requested):

low-batt:
The strings were never null terminated. Previously the code was using [cStringUsingEncoding](https://developer.apple.com/documentation/foundation/nsstring/1408489-cstringusingencoding) with [NSUTF32LittleEndianStringEncoding](https://developer.apple.com/documentation/foundation/nsutf32littleendianstringencoding). The documentation for `cStringUsingEncoding` cautions about strings not being `NULL` terminated in some cases, although for  `NSUTF32LittleEndianStringEncoding` the note in the special considerations section applies, which indicates the results of passing `NSUTF32StringEncoding` variants is undefined.

With the original code the pattern of memory use is such that the memory after the string usually happens to have zero values. But sometimes the call to `wcslen` was returning large values that did not reflect the size of the string. Enabling malloc scribble and the address sanitizer picked up the invalid access by `wcslen` that was being used in the original code.

Once the issue with `wcslen` was fixed the address sanitizer complained about this access. I vaguely remember being very confused about why that initial space was being added. I definitely did not understand that the algorithm was intentionally going beyond the length of the string and looking at the null terminator.

I'll see if I can change it to null terminate the strings.

uiryuu:
I didn't check the documentation and was expecting that `cStringUsingEncoding` will yield a null-terminated C string. But this part of code was too C-like, maybe that's why I did so many hard-to-understand tricks.

low-batt:
I think the fix I just pushed is better. But I'm starting to remember some of my other confusion while working on this. I was unable to get test results to match what I read in posts on this in the Internet. You can see that in this results:
```text
22:01:58.855 [iina][d] Test helo hello should be 1: 1
22:01:58.856 [iina][d] Test kelo hello should be 2: 3
22:01:58.856 [iina][d] Test kel hello should be 3: 4
22:01:58.856 [iina][d] Test kitten sitting should be 3: 5
22:01:58.856 [iina][d] Test GRATE GIRAFFE should be 3: 4
22:01:58.856 [iina][d] Test TALK ALSO should be 4: 4
```
Is this one of the Levenshtein variants?

uiryuu (approved):

low-batt:
I too would have been tripped up by the `cStringUsingEncoding` behavior. Took me a while to figure that out. At least Apple documented the gotcha.

PR #4454, opened 2023-06-07
Ignore unwanted resize when changing between screens (workaround for #4248)

### LINKED ISSUE #4248: Video window incorrectly changes size between identical screens
### System and IINA Version

- macOS 12.6.3
- IINA 1.3.1 Build 133

### Expected Behavior
Moving an IINA video window between multiple screens retains the same visual dimensions like any other window on macOS (e.g. moving a Finder or Preview window between screens).

### Actual Behavior
Moving an IINA video window between identical screens changes dimensions without reason, possibly because another screen with a different resolution and UI scaling is present. I’m using three monitors, two with identical dimensions (5120×2880, looks like 2560×1440) and another (2560×1440, no scaling). Moving the window between the identical screens (from centre to right) makes the window smaller on drop and it returns to its original dimensions when dragged back (from right to centre). More unexpectedly, moving the window between screens with different resolution and UI scaling (centre to left) sometimes *does not* change the window dimensions.

I’d expect the window to always retain the same size as long as it isn’t dragged to a screen too small for its current size.

### Reproduction
Open a regular 1080p video file in IINA and drag the window between screens of identical or differing dimensions. The screens I use are all Dell UltraSharp 27” with native resolutions of either 3840×2160 or 2560×1440.

### Attachments
Unfortunately, I can’t seem to record multiple screens at once so here’s a capture of how the window is dragged out, then dragged back where it has resized and restores its original dimensions, so you can get a grasp of the extent of resizing happening.

https://user-images.githubusercontent.com/7656669/222896934-74f330be-aa42-4301-9079-e5f8b66c70ae.mp4

svobs:
- [x] This implements/fixes issue #4248.

---

**Description:**
This workaround detects when a window is moved to/from a hi-DPI ("Retina") display and a regular display, and sets a flag. This takes advantage of the fact that `windowWillResize()` is called shortly after and (because the flag is set), instructs AppKit to keep the existing size.

This seems to successfully prevent unwanted behavior where AppKit was changing the window size when moving the window between screens.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -149,6 +149,8 @@ class MainWindowController: PlayerWindowController {
 
   var lastMagnification: CGFloat = 0.0
 
+  var denyNextWindowResize = false
+
   /** Views that will show/hide when cursor moving in/out the window. */
   var fadeableViews: [NSView] = []
 
@@ -692,7 +694,8 @@ class MainWindowController: PlayerWindowController {
 
 
   private func addVideoViewToWindow() {
-    guard let cv = window?.contentView else { return }
+    guard let window = window,
+          let cv = window.contentView else { return }
     cv.addSubview(videoView, positioned: .below, relativeTo: nil)
     videoView.translatesAutoresizingMaskIntoConstraints = false
     // add constraints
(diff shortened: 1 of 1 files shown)


svobs:
After more testing, this does not look like a good fix - it has inconsistent side effects. Will continue to search for a good solution.

PR #4462, opened 2023-06-14
Improve Music Mode auto-switch override logic (fixes #4461)

### LINKED ISSUE #4461: "Switch to music mode automatically" stops working after a few times
**System and IINA version:**

- macOS 13.4
- IINA 1.3.1 Build 133

**Expected behavior:**

When `Settings` > `General` > `Switch to "music mode" automatically when playing an audio file` is checked, changing tracks between video and audio should automatically and consistently switch to or from "music mode" appropriately.

**Actual behavior:**

It works a few times and then stops switching.

**Steps to reproduce:**

1. Launch IINA. Make sure `Settings` > `General` > `Switch to "music mode" automatically when playing an audio file` is checked.
2. Create or open a playlist with 2 items: an audio file followed by a video file.
3. Open the audio file. _IINA switches to music mode._
4. Select `Playback` > `Next Media` to go to the next track (the video file). _IINA switches back to video mode._
5. Select `Playback` > `Previous Media` to go back to the audio file.  _IINA switches back to music mode._
5. Select `Playback` > `Next Media` to go back to the video file.  _IINA does not switch modes._

I was able to reproduce this consistently by following the above steps.

svobs:
- [x] This implements/fixes issue #4461.

---

**Description:**

### Rationale
If the user manually enters Music Mode for a non-audio file (or exits Music Mode for an audio-file), they have indicated that they want to override the default behavior, and likely want to stay in that state until they indicate otherwise. They indicate otherwise by either toggling Music Mode again (thus undoing their previous behavior), or by enabling `Switch to "music mode" automatically` in `Settings`.

### Details

This replaces `switchedToMiniPlayerManually` and `switchedBackFromMiniPlayerManually` in `PlayerCore` with a single var, `overrideAutoSwitchToMusicMode`, which fixes inconsistent behavior reported in #4461.

It also cleans up the `autoSwitchToMusicMode` logic in `PlayerCore.trackListChanged()` so:
- It's more understandable.
- Removes the requirement that `audioStatusWasUnkownBefore` is `true`. This is totally redundant because it is covered by the checks which occur right afterwards.

The single state variable, `overrideAutoSwitchToMusicMode`, if true, keeps the current music mode state (off/on), even when the media changes and `autoSwitchToMusicMode` is true in the preferences. Its behavior is simple:
- The first time the user manually toggles Music Mode, it's set to `true`, so automatic switching is disabled, even if `autoSwitchToMusicMode` is true in the preferences.
- The next time the user manually toggles Music Mode, it's set to `false`, so automatic switching goes back to using the value of `autoSwitchToMusicMode` in the preferences.
- The value of `overrideAutoSwitchToMusicMode` is set back to `false` when the player window is closed, or if `autoSwitchToMusicMode` is changed in the preferences. (It doesn't make sense to stay in manual override if the user has just indicated that they want to enable auto-switch).
- The value of `overrideAutoSwitchToMusicMode` is not used if the `autoSwitchToMusicMode` pref is `false`.

### DIFF
--- a/iina/MiniPlayerWindowController.swift
+++ b/iina/MiniPlayerWindowController.swift
@@ -112,11 +112,11 @@ class MiniPlayerWindowController: PlayerWindowController, NSPopoverDelegate {
     // close button
     closeButtonVE.action = #selector(self.close)
     closeButtonBox.action = #selector(self.close)
-    closeButtonView.alphaValue = 0
     closeButtonBackgroundViewVE.roundCorners(withRadius: 8)
-    closeButtonBackgroundViewBox.isHidden = true
 
-    // switching UI
+    // hide controls initially
+    closeButtonBackgroundViewBox.isHidden = true
+    closeButtonView.alphaValue = 0
     controlView.alphaValue = 0
     
     // tool tips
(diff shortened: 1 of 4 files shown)


lhc70000:
Resolved conflicts and merged via 74bb19a.

PR #4464, opened 2023-06-15
Fix TAB & ESC key bindings not working in main player & music mode windows (fixes #4465)

### LINKED ISSUE #4465: TAB key doesn't work in player window
**System and IINA version:**

- macOS 13.4
- IINA 1.3.1 Build 133

**Expected behavior:**

Using a key binding mapped to `TAB` doesn't work in IINA's player windows.

**Actual behavior:**

`TAB` is ignored.

**Steps to reproduce:**

Add a key binding such as this one to the active configuration:
Key: `TAB`
Action: `cycle mute`

Then start playing a video in IINA.
While the player window is in focus, press `TAB`.

Always.

This also fails in Music Mode.

svobs:
- [x] This implements/fixes issue #4465.

---

**Description:**
Uses `NSWindow` subclasses for `MainWindow` and `MiniPlayerWIndow`, and overrides `keyDown()` so that it bypasses MacOS's built-in diversion of the `TAB` & `ESC` keys for keyboard navigation, which IINA does not desire to use for these windows. This allows `TAB` and `ESC` key bindings to work. (Note that `ESC` was working for `MainWindow` due to a workaround, but not `MiniPlayerWIndow`; while `TAB` was not working for either).

I ended up duplicating the code for `keyDown()` in each window class, because I found I needed separate subclasses for each window which inherited directly from `NSWindow`. Originally I tried to create a common base class which inherited from `NSWindow`, and then have both windows inherit from that, but that didn't seem to work.

Removes the workaround for `ESC` from `MainWindow`, because it's no longer needed. Removes the workaround for sending `SPACE` from `PlaylistViewController`, because it now was sending a second unwanted space.

One wrinkle with this solution is that it drew a focus ring around either the Playlist or Chapters tab button because it was detected as the first responder. I was unable to find a way to say "no first responder please", but setting `refusesFirstResponder="YES"` for each of these got rid of the focus rings.

### DIFF
--- a/iina.xcodeproj/project.pbxproj
+++ b/iina.xcodeproj/project.pbxproj
@@ -238,6 +238,7 @@
 		B4E446F725CB54EF0069F06E /* Mustache in Frameworks */ = {isa = PBXBuildFile; productRef = B4E446F625CB54EF0069F06E /* Mustache */; };
 		B4E4470125CE3F930069F06E /* Sparkle in Frameworks */ = {isa = PBXBuildFile; productRef = B4E4470025CE3F930069F06E /* Sparkle */; };
 		C789872F1E34EF170005769F /* InfoPlist.strings in Resources */ = {isa = PBXBuildFile; fileRef = C78987311E34EF170005769F /* InfoPlist.strings */; };
+		D1B4E24E2A3AFC9100E36F1D /* MiniPlayerWindow.swift in Sources */ = {isa = PBXBuildFile; fileRef = D1B4E24D2A3AFC9100E36F1D /* MiniPlayerWindow.swift */; };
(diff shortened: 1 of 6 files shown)


uiryuu:
Tested and worked. Thanks!

lhc70000:
I found that after this commit, any menu item with key equivalent that

1. is not a mpv keybinding, and
2. doesn't contain `Command` (e.g. `D` or `Option+D`, but not `Cmd+Option+D`)

will not work anymore.

If this is a mpv keybinding, IINA can still trigger the action when it's passed to the main window, however there are still some menu items with user-defined non-mpv key shortcuts, for example, custom filters and menus added by plugins.

I'm finding a solution right now.

PR #4467, opened 2023-06-16
Change "Music Mode" menu item to "Enter Music Mode"/"Exit Music Mode" based on status (closes #4463)

### LINKED ISSUE #4463: Prepend "Enter" and "Exit" to "Music Mode" menu item
**What you want IINA to do:**

Change the text of the `Music Mode` menu item:

1. If last active player **IS NOT** in Music Mode, it should display "Enter Music Mode".
2. If last active player **IS** in Music Mode, it should display "Exit Music Mode".

**What IINA does currently:**

It always displays "Music Mode".

SCR-20230613-spvu

This is already the behavior of IINA's "Picture-in-Picture" and "Full Screen" modes. See above screenshot.

svobs:
- [x] This implements/fixes issue #4463.

---

**Description:**
See issue #4463.

### DIFF
--- a/iina/AppData.swift
+++ b/iina/AppData.swift
@@ -87,6 +87,8 @@ struct Constants {
     static let subDelay = "Subtitle Delay"
     static let pip = NSLocalizedString("menu.pip", comment: "Enter Picture-in-Picture")
     static let exitPIP = NSLocalizedString("menu.exit_pip", comment: "Exit Picture-in-Picture")
+    static let miniPlayer = NSLocalizedString("menu.mini_player", comment: "Enter Music Mode")
+    static let exitMiniPlayer = NSLocalizedString("menu.exit_mini_player", comment: "Exit Music Mode")
     static let custom = NSLocalizedString("menu.crop_custom", comment: "Custom crop size")
     static let findOnlineSubtitles = NSLocalizedString("menu.find_online_sub", comment: "Find Online Subtitles")
     static let chaptersPanel = NSLocalizedString("menu.chapters", comment: "Show Chapters Panel")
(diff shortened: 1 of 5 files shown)


low-batt (changes requested):
The somewhat confusing IINA rules on localization require both the base and english localization files be updated. The rest are updated through [Crowdin](https://translate.iina.io/). So `Localizable.strings` under `en.lproj` also needs to be updated as a part of this PR.

I've been confused as to why we have to duplicate strings like this. Seems wrong. This PR got me to look into that. From [Enabling Base Internationalization](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/InternationalizingYourUserInterface/InternationalizingYourUserInterface.html#//apple_ref/doc/uid/10000171i-CH3-SW4):
> Xcode modifies your project folder according to the selections you make in this dialog. Xcode creates a Base.lproj folder in your project folder and adds to it the resource files you select. Xcode creates a language folder for the development language but only adds resources that need translation to the folder. For example, if you select English as the development language, Xcode inserts the resource file in the Base.lproj project folder but not the en.lproj folder because the resource is already in English.

Checking `info.plist` confirms the development language is set to English:
```xml
<key>CFBundleDevelopmentRegion</key>
<string>en</string>
```

From the Apple documentation it sounds like `en.lproj` is not needed.

A separate issue for another day…

svobs:
> The somewhat confusing IINA rules on localization require both the base and english localization files be updated. The rest are updated through [Crowdin](https://translate.iina.io/). So Localizable.strings under en.lproj also needs to be updated as a part of this PR.

And here I was feeling clever that I finally got it right! I made another commit with the additions.

> From the Apple documentation it sounds like en.lproj is not needed.

Interesting! I played with this and you are correct. This file appears to be [worse than] redundant.

> A separate issue for another day…

But looks like a trivial fix. Can't resist this low-hanging fruit. I will file a bug report and PR.

svobs:
Sorry, inadvertently committed extra lines. Force-pushed the correct commit...

low-batt (approved):
Changes look good to me.

uiryuu:
Thanks!

PR #4475, opened 2023-07-05
fix plugin postMessage(name, data) JSON.parse fail when fields in data contains newline

magicdawn:

---

**Description:**

example plugin code

```js
iina.standaloneWindow.postMessage('test', { a: 'b\nc' })
```

example webpage code
```js
iina.onMessage('test', data => {
  console.log(data)
})
```

when `postMessage(name, data)` `data.a` contains newline `\n`, will cause errors show like `unterminated string` in JSON.parse

image

image

----------- 

this fix use ES6 `String.raw`, and it do work for this case

image

image

-----------

## about `String.raw`
- supported since Safari 9 [OS X 10.9]
  - https://caniuse.com/?search=String.raw
  - https://en.wikipedia.org/wiki/Safari_version_history

## other ways to fix

manual escape like https://stackoverflow.com/questions/15843570/objective-c-how-to-convert-nsstring-to-escaped-json-string

### DIFF
--- a/iina/JavascriptMessageHub.swift
+++ b/iina/JavascriptMessageHub.swift
@@ -27,7 +27,7 @@ class JavascriptMessageHub {
         webView.evaluateJavaScript("window.iina._emit(`\(name)`)")
         return
       }
-      webView.evaluateJavaScript("window.iina._emit(`\(name)`, `\(dataString)`)")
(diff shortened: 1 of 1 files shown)


lhc70000:
Thank you very much.

By the way, the plugin documentation is available at https://docs.iina.io. Please feel free to share any suggestions!

PR #4498, opened 2023-07-13
Work around freeze due to VP9 hardware decode, #4486

### LINKED ISSUE #4486: Freeze on Intel based Macs while playing VP9 encoded videos
**System and IINA version:**

- macOS 13.4.1
- IINA 1.3.2

**Expected behavior:**
Playback of YouTube videos does not freeze.

**Actual behavior:**
@AlejandroVolkova has reported that IINA 1.3.2 will freeze while playing some YouTube videos.

This was encountered running on an Intel based Mac:
iMac Retina 4K, 21.5-Inch, 2019
Processor : 3 GHz 6-Core Intel Core i5
Graphics : Radeon Pro 560X 4 GB
Memory : 32 GB 2667 MHz DDR4

**Steps to reproduce:**
Play a YouTube video with VP9 encoding on an Intel based Mac with hardware decoding enabled.

~Has not be tested, but the expectation is that the problem will reproduce with mpv if hardware decoding is enabled.~

This has now been reproduced with mpv. The `playback core` thread is blocked on a call to `VTDecompressionSessionWaitForAsynchronousFrames` as is seen with IINA.

Every time.

low-batt:
This commit will add code to MPVController.mpvInit that changes the value of the mpv option hwdec-codecs, removing VP9, if IINA is not running on a Mac with an Apple Silicon chip. This change works around the FFmpeg defect reported in ticket 9599.

- [x] This implements/fixes issue #4486.

---

**Description:**

### DIFF
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -153,6 +153,66 @@ class MPVController: NSObject {
     removeOptionObservers()
   }
 
+
+  /// Determine if this Mac has an Apple Silicon chip.
+  /// - Returns: `true` if running on a Mac with an Apple Silicon chip, `false` otherwise.
+  private func runningOnAppleSilicon() -> Bool {
+    // Old versions of macOS do not support Apple Silicon.
+    if #unavailable(macOS 11.0) {
+      return false
+    }
+    var sysinfo = utsname()
+    let result = uname(&sysinfo)
+    guard result == EXIT_SUCCESS else {
+      Logger.log("uname failed returning \(result)", level: .error)
+      return false
+    }
+    let data = Data(bytes: &sysinfo.machine, count: Int(_SYS_NAMELEN))
+    guard let machine = String(bytes: data, encoding: .ascii) else {
+      Logger.log("Failed to construct string for sysinfo.machine", level: .error)
+      return false
+    }
+    return machine.replacingOccurrences(of: "\0", with: "") == "arm64"
+  }
+
+  /// Apply a workaround for issue [#4486](https://github.com/iina/iina/issues/4486), if needed.
+  ///
+  /// On Macs with an Intel chip VP9 hardware acceleration is causing a hang in
+  ///[VTDecompressionSessionWaitForAsynchronousFrames](https://developer.apple.com/documentation/videotoolbox/1536066-vtdecompressionsessionwaitforasy).
+  /// This has been reproduced with FFmpeg and has been reported in ticket [9599](https://trac.ffmpeg.org/ticket/9599).
+  ///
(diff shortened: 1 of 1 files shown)


low-batt:
I was at the eye doctor today, so eyes dilated and things are blurry. A critical review is needed. Especially the detection of Apple Silicon. Is there a better way to do that?  So far we have not seen the problem under Apple Silicon. so the workaround is not applied when running on AS and VP9 hardware decoding is still used. Does this make sense, or should we always disable VP9 for now to be safe?

svobs:
> Especially the detection of Apple Silicon. Is there a better way to do that?

Not really. Seems like there should be, but did some searching and yours appears to be the standard solution. Very typical to see Linux shell scripts rely on parsing the `uname` command, so this isn't unexpected.

Have not tried running it yet, but looks like a good solution. +1 for adding a secret pref allowing the workaround to be disabled.

Also, you should know the word "workaround" (with no space) is only used as a noun. It doesn't have an exact equivalent verb in common English, although you could use "work around".

iina/MPVController.swift
```diff
@@ -153,6 +153,66 @@ class MPVController: NSObject {
...
+    guard result == EXIT_SUCCESS else {
+      Logger.log("uname failed returning \(result)", level: .error)
+      return false
+    }
+    let data = Data(bytes: &sysinfo.machine, count: Int(_SYS_NAMELEN))
+    guard let machine = String(bytes: data, encoding: .ascii) else {
+      Logger.log("Failed to construct string for sysinfo.machine", level: .error)
+      return false
+    }
+    return machine.replacingOccurrences(of: "\0", with: "") == "arm64"
```
uiryuu: I suggest use `machine.starts(with: "arm64")`, which is not only more robust but also more elegant.

uiryuu (changes requested):
Other changes LGTM

iina/MPVController.swift
low-batt: Good catch. Uploaded a new commit with this change.

low-batt:
On this:
> +1 for adding a secret pref allowing the workaround to be disabled.

I thought about that, but instead I made sure that if this mpv option is set in `Advanced` settings, the workaround will not override the value configured there. In fact the workaround detects that a value has been configured for the option and does not apply the workaround. I did that so that it was clear the workaround is not being used when you read the log file.

I was wondering if we should have an internal preference that by passes the check for Apple Silicon and always applies the workaround. Just in case we find it also has trouble on Apple Silicon machines. I did not do that as so far I have not been able trigger a problem on my M1 machine.

PR #4525, opened 2023-07-22
Add checking of core video return codes

low-batt:
This commit will add code to `VideoView` that checks that core video methods are returning `kCVReturnSuccess`. If not then a warning will be logged reporting the error code that was returned.

---

**Description:**

### DIFF
--- a/iina/VideoView.swift
+++ b/iina/VideoView.swift
@@ -206,20 +206,22 @@ class VideoView: NSView {
 
   func startDisplayLink() {
     if link == nil {
-      CVDisplayLinkCreateWithActiveCGDisplays(&link)
+      checkResult(CVDisplayLinkCreateWithActiveCGDisplays(&link),
+                  "CVDisplayLinkCreateWithActiveCGDisplays")
     }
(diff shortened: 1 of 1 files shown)


low-batt:
Currently `VideoView` is ignoring [result codes](https://developer.apple.com/documentation/corevideo/1572713-result_codes) returned by core video methods. Issue #4520 reports a case where apparently a call to [CVDisplayLinkCreateWithActiveCGDisplays](https://developer.apple.com/documentation/corevideo/1456863-cvdisplaylinkcreatewithactivecgd) failed. The expectation is that the result code returned should provide some sort of indication as to what went wrong.

The changes in the commit do not fix anything, they merely add code to check the returned `CVResult` and log a warning message if it is not `kCVReturnSuccess`. Simulating a bad result the logged message looks like:

```text
17:12:12.308 [iina][w] Core video method CVDisplayLinkCreateWithActiveCGDisplays returned: The allocation for a buffer or buffer pool failed. Most likely because of lack of resources (-6662)
```

PR #4531, opened 2023-07-24
Inspector window: allow Watch table column resizing + clean & polish layout

### LINKED ISSUE #4521: Inspector pallet: entries in the Status/Watch section are truncated
**What you want IINA to do:**

The _Watch_ list of the _Inspector_ has

(1) Resizable columns (2) lines wrap 

**What IINA does currently:**

In the Watch list, key identifiers and values are truncated without an option to resize the columns by dragging the separator.

Screenshot 2023-07-21 at 21 11 26

svobs:
- [x] This implements/fixes issue #4521.

---

**Description:**

**Window layout changes:**
1. Fix vertical alignment between columns in all 4 tabs, by replacing "top"/"bottom" constraints between `NSTextView`s with "firstBaseline" constraints.
2. Fix constraints which cause the height of the window to change for "General" tab.
3. Tweaked offset constraints to be consistent across all tabs and fields.
4. Re-order views in the XML for consistency, and various other constraint cleanup
5. Adjust minimum widths & heights to be consistent across tabs.

**Watch table changes:**
1. Add `Name` and `Value` column headers to the "Watch" table in the "Status" tab, and made the first tab resizable and saveable, and the second tab set to scale to the size of the window (fixing #4521).
2.  Add code to make the Name column reduce itself in width when the window is resized, so the `Value` column won't get hidden.
3. Refactor Watch table to support translucent headers while working around `NSTableView` bugs: Make table view-based, add custom column headers, reimplement background translucency. Make table expand the window vertically instead of scrolling to avoid overlapping views issue.
4. Add animations to Watch table for row insert & delete (respecting the "Reduce Motion" pref).
5. Change `Error` entries to remove brackets, add italic & grayed text.
6. Add localization entries for 3 new fields; see comments below.

**Final product:**

SCR-20230727-ndhl

### DIFF
--- a/iina/Base.lproj/InspectorWindowController.xib
+++ b/iina/Base.lproj/InspectorWindowController.xib
@@ -1086,7 +1086,7 @@
                                                         <rect key="frame" x="0.0" y="0.0" width="420" height="201"/>
                                                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                                                         <subviews>
-                                                            <tableView focusRingType="none" verticalHuggingPriority="750" allowsExpansionToolTips="YES" columnAutoresizingStyle="lastColumnOnly" tableStyle="fullWidth" columnReordering="NO" multipleSelection="NO" typeSelect="NO" autosaveName="InspectorWatchTable" rowSizeStyle="automatic" headerView="pnK-fx-7et" viewBased="YES" floatsGroupRows="NO" id="oUp-DO-OKl" userLabel="WatchTable Table View">
+                                                            <tableView focusRingType="none" verticalHuggingPriority="750" allowsExpansionToolTips="YES" columnAutoresizingStyle="lastColumnOnly" tableStyle="fullWidth" columnReordering="NO" typeSelect="NO" autosaveName="InspectorWatchTable" rowSizeStyle="automatic" headerView="pnK-fx-7et" viewBased="YES" floatsGroupRows="NO" id="oUp-DO-OKl" userLabel="WatchTable Table View">
                                                                 <rect key="frame" x="0.0" y="0.0" width="420" height="173"/>
                                                                 <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                                                                 <size key="intercellSpacing" width="0.0" height="2"/>
--- a/iina/Base.lproj/InspectorWindowController.xib
+++ b/iina/Base.lproj/InspectorWindowController.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="20037" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="21701" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="20037"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="21701"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <objects>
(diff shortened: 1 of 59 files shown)


svobs:
I think I got it. I had to refactor the Watch table to be view-based instead of cell-based, and various other tweaks. Also, need to localize the two columns. I suppose I/someone could actually just use "Name" and "Value" and copy all the localizations from the`Settings` > `Advanced` table...

SCR-20230725-bnie

_[Edit: saw a color regression in previous commit; pushed another quick update to fix.]_

low-batt (changes requested):
Thanks for taking this one. This is an important issue. To me it is almost a bug that you currently can't resize the columns.

3 issues.

First issue. Us programmers are used to "property key". Not sure the public is. When you add something to the table the IINA prompt asks for `mpv property name`. So I don't think the name of the first column should be `Key`. Since we use `Name` for the column that contains the option in `Advanced` settings, I'm thinking that is the correct choice. The other choice would be `Property`.

Second issue. When the window first came up the column containing the property was small and truncating the names despite all the space we have to work with. When possible the columns should be initially sized appropriately.

Third issue. Resizing the first column introduced a scroll bar, which seems odd given the values in the table are so small compared to the available space:
pr-4531

svobs:
> First issue. Us programmers are used to "property key". Not sure the public is. When you add something to the table the IINA prompt asks for mpv property name. So I don't think the name of the first column should be Key. Since we use Name for the column that contains the option in Advanced settings, I'm thinking that is the correct choice. The other choice would be Property.

Changed to Name.

> Second issue. When the window first came up the column containing the property was small and truncating the names despite all the space we have to work with. When possible the columns should be initially sized appropriately.

I added a bit more default width to the Name column. I'm just not sure too much extra space will be needed in most cases, because all the property names I came up with could easily fit, although I'm not at all an expert in mpv properties and their values. At least the column widths will be remembered across launches.

Also, I think the actual width of the window will vary depending on the length of some of the fields. _Edit: Looks like this can't really be avoided without allowing text to wrap..._

> Third issue. Resizing the first column introduced a scroll bar, which seems odd given the values in the table are so small compared to the available space

I think I fixed it - please confirm.

Also:

1. I found a case where the new code could crash at quit due to a call to `mpv.getString()`. Added the appropriate checks.
2. Added localizations for `Name` and `Value` column headers by copying the strings from the Advanced tab. Also localized the "&lt;Error&gt;" message by copying the strings from `"alert.title_error"` in `Localizable.strings`. But I thought that leaving the brackets around it looked kind of weird when viewed with Asian hieroglyphs, so I swapped the brackets for an italic & grayed style.

SCR-20230725-lnrh

low-batt:
The latest commit seems to have fixed my concerns. For some reason the table seems to be messing up the cursor shown when changing column width. Not sure what is causing that. My guess is the other reviewers will not like the changes to the localization files. The usual issue with disrupting Crowdin automation.

svobs:
> My guess is the other reviewers will not like the changes to the localization files. The usual issue with disrupting Crowdin automation.

I'm fine with backing out the localizations if needed. That took a lot less time than the other parts.

> For some reason the table seems to be messing up the cursor shown when changing column width.

Do you mean that it's a `resizeLeft` cursor instead of `resizeLeftRight`?

SCR-20230725-tkwg

...OK, so I was halfway into a detailed analysis, but then I followed a hunch and discovered that the problem looks like a bug in Apple's new "Full Width" table style. It is fixed by setting the table's style to "Automatic" instead (...although the resize cursor is still not 100% right...it's less obviously wrong. I don't think we can expect too much from `NSTableView`, tbh ✌️)


Posting the remainder for posterity!

---

I've set the `Name` column resizing mode to `User Resizable`, and the `Value` column to `Autoresizes with Table`, and set the table's `columnAutoResizingStyle` to `Last Column Only`. Then I set the `Value` column's width so that its right side is hugging the right side of the table.

When configured like this, the `Value` column "snaps" to the right edge, and the result is that the `Name` column will only resize via the user's drag of item 1 below, while the `Value` column will only resize when the window resizes. You should see the vertical line at the right of the column header (2) move as the window resizes.
SCR-20230725-toxp

If instead I were to make both columns `User Resizable` or `Both`, the second column will need the user to resize it manually:
SCR-20230725-rnkl

For research, I went to see what Apple does with Finder, so I tried playing with its List View and putting the `Size`/`Date Modified` columns up against the right edge of a Finder window. It looks like they are configured as `Both`, and also have maximum sizes. But there seems to be other weird logic that Apple is using, and to be honest I don't understand why I sometimes can resize the second-to-rightmost column and sometimes can't. So that has not been helpful.

svobs:
SCR-20230725-ufyp

With style `Automatic`. Main difference seems to be more horizontal padding in parts of the header. But the cursor is fixed.

svobs:
Argh. I now see visual artifacts with style `Automatic`... Perhaps I should just change the second column to `User Resizable`...?

low-batt:
Mostly I wanted to make sure the oddity with the cursor was AppKit and not something we were doing. Seems to me some AppKit programmer was lazy and did not properly use the resizing cursors.

I have always thought tables should look is shown with `Autoresizes with Table`. That 3<sup>rd</sup> empty column seems super odd to me. But it seems to match up with other IINA tables. I performed the following test twice cause I didn't expect the answer and thought I must have messed up…

I brought this over to my MacBook Air running Catalina and the behavior changed. Under Catalina the table acts and looks just like what is shown above with `Autoresizes with Table`. Isn't that interesting.

On `<Error>`, I always disliked that. I like switching to dimming. I think that is fine for now, but really a proper error message is required. A future task.

Thumbs up from me for the changes as is, BUT I am certain the PR will be rejected due to including the translation files for other languages. They have gotten serious about that.

I have not forgotten about the base/en issue and how the en translation files are not needed. People have been busy with the current release. 1.3.3 just got released. At some point soon attention will move to the next release which is planned to be the feature release that has been talked about for some time.

low-batt (approved):
I pulled the latest commits, build IINA and tested under macOS Ventura and macOS Catalina. Major improvement to support resizing of the table columns.

Leaving it to other reviewers to comment on the inclusion of the language localization files.

svobs:
So, couple things...

@low-batt was hard to tell if you were endorsing the Fulll-Width style or Automatic. I prefer the Full-Width, despite the cursor oddity.

There is another, more serious problem which I discovered, and I made several pushes thinking I had fixed it (and also other things), only to ultimately have to admit defeat and roll back, having run out of time...

There's a problem coming from the translucent background for the table. There are color glitches. It only presents itself if once you try to horizontally scroll. The background of the header suddenly brightens. It looks even worse if you then resize the window:
SCR-20230726-ozmz

If I clear out the table background and show just the header, like this, the problems are gone. But not sure if it is the desired look, and it visually separates the table from the +/- buttons.
SCR-20230726-pblb

low-batt:
I liked the one where the `Value` column "snaps" to the right edge.

For the latest issues you brought up I have to defer to the other developers. I too am unsure what they would prefer.

svobs:
Much soul-draining trial & error later, I have found solutions & workarounds.

I was able to get a bug-free translucent background by enclosing the table inside a container view, and adding a translucent background to that. Hooray.

SCR-20230727-cwnn

I also made some progress on figuring out the arcane way column sizing is done. I was able to make the `Name` column reduce itself in width when the window is resized, so the `Value` column won't get hidden.

Also added animations when inserting & deleting rows (respecting the "Reduce Motion" pref) - was not hard and looks nice.

Lastly I discovered another problem caused by using translucent headers. If there are enough entries and the table starts to scroll down, `NSTableView` will draw rows overlapping the header as noted [here](https://stackoverflow.com/questions/70265009/custom-nstableheaderview-does-not-overlap-cells-when-scrolling) and [here](https://stackoverflow.com/questions/984330/scrolling-table-overlaps-section-header). I think it's safe to say `NSTableView` doesn't support scrolling a table with a translucent header. So I added a workaround which will make the window taller as more rows are added, so that the table will not scroll. Don't expect this will be a problem for the vast majority as I can easily fit 25+ properties just on my MacBook's 16" screen.

uiryuu:
The new look of the TableView is cool! I just tried by rebasing the branch to develop. Just a small thing to add, could you make the TableView multiple selection? It would be convenient to remove multiple entries.

svobs:
> Just a small thing to add, could you make the TableView multiple selection? It would be convenient to remove multiple entries.

Done - take a look at the latest commit.

Issue #4538, opened 2023-07-28
<delete> does not remove files from the playlist


Shimushushushu:
**System and IINA version:**

- macOS Ventura 13.4.1 
- IINA 1.3.3 

**Expected behavior:**

Pressing &lt;delete&gt; removes the selected files in the playlist.

**Actual behavior:**

Nothing happens, except for an alert sound (the same as `printf "\a"`), and no crash occurs.

**Steps to reproduce:**

1. Open several files.
2. Choose arbitrary number of them in the playlist.
3. press &lt;delete&gt;. 

- [x] MPV does not have this problem.

**How often does this happen?**

Always, as long as the steps above are performed.

low-batt:
Reproduced. Thank you for reporting this!
Confirmed IINA 1.3.2 works. This is a regression added in IINA 1.3.3.

From a very quick analysis…

This method was added to the `MainWindow` class in commit 0bc3691cf5805f51ddb491ee249c4d75e056d147:
```swift
  override func keyDown(with event: NSEvent) {
    /// Forward all key events which the window receives to its controller.
    /// This allows `ESC` & `TAB` key bindings to work, instead of getting swallowed by
    /// MacOS keyboard focus navigation (which we don't use).
    if let controller = windowController as? MainWindowController {
       controller.keyDown(with: event)
    }
  }
```

Adding debug logging to see the responder chain shows:
```text
<IINA.MainWindow: 0x10e6f2420>
<IINA.MainMenuActionHandler: 0x600003ea0030>
<IINA.MainWindowController: 0x11e886000>
```

The above code by passes `MainMenuActionHandler`. That prevents this method in `PlaylistViewController` from being called:
```swift
  @objc func delete(_ sender: NSMenuItem) {
    player.playlistRemove(playlistTableView.selectedRowIndexes)
  }
```

At least I think that is what is going on.

If the menu item is invoked directly the entry is removed from the playlist.

@svobs Please look into this one.

svobs:
> The above code by passes MainMenuActionHandler. That prevents this method in PlaylistViewController from being called:
> At least I think that is what is going on.

If only it were that simple.

The `MainMenuActionHandler` class just contains extra actions for `MainWindowController` but doesn't really need to listen for input. It relies on the code in `MenuController` to bind keys from `MainWindowController` to those actions. (It doesn't really need to be its own separate object, really). The problem here is that the `PlaylistViewController` is being left out of the chain.

[...]

PR #4550, opened 2023-08-01
Fix regression: could not use DEL or BS keys to remove from playlist

svobs:
- [x] This implements/fixes issue #4538.

---

**Description:**

Using the easiest route to fix this, which is to add a special check for Backspace and Delete when the playlist is open and row(s) are selected.

Had to duplicate the code for Music Mode.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -855,6 +855,20 @@ class MainWindowController: PlayerWindowController {
 
   // MARK: - Mouse / Trackpad events
 
+  override func keyDown(with event: NSEvent) {
+    if sideBarStatus == .playlist {
+      // Special case for playlist delete
+      let key = KeyCodeHelper.mpvKeyCode(from: event)
+      if key == "DEL" || key == "BS" {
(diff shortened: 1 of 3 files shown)


low-batt (approved):
I pulled the PR, built IINA and confirmed this fixes the regression.

low-batt:
I also confirmed this does not fix the regression reported in issue #4562 which could be related. I'm expecting additional work in this area is required.

svobs:
I made some extra fixes to key bindings in PR #4567, but because there's no overlapping code between the two I've submitted it as a separate PR. Hopefully that was the right way to go.

svobs:
Confirmed that de5fdf50d08a3f1787dfcc7f2b701cfd1a11de1f from PR #4569 fixes this in a superior way. Closing this one.

> I made some extra fixes to key bindings in PR https://github.com/iina/iina/pull/4567, but because there's no overlapping code between the two I've submitted it as a separate PR. Hopefully that was the right way to go.

Looks like that was the right way to go. That fix is still needed.

PR #4560, opened 2023-08-04
Support installing plugins from GitHub releases & directly opening iinaplgz files

lhc70000:

---

**Description:**

This PR implements two critical features for the upcoming release of the plugin system:

- Support installing a `iinaplgz` file by directly opening it
- Previously, when installing from GitHub, IINA could only download the source code, therefore required the built files to be included in the repo. However, this is a terrible choice, especially for a large project with megabytes of transpiled files. Now IINA will try to find an `iinaplgz` file from the latest release and only install from the source when failed.

How to test

- Use the `iina-plugin` CLI to generate `iinaplgz` packages.
- Double click, or drag & drop a `iinaplgz` file to see if IINA can install it correctly.
- Install `iina/plugin-userscript` from GitHub to see if IINA can install from the `iinaplgz` release.

### DIFF
--- a/iina/JavascriptPlugin.swift
+++ b/iina/JavascriptPlugin.swift
@@ -250,6 +250,27 @@ class JavascriptPlugin: NSObject {
         try? FileManager.default.removeItem(at: pluginsRoot.appendingPathComponent(item))
       }
     }
+    
+    // If there is a iinaplgz file inside the latest release, use the plgz file
+    
+    let response = Just.get("https://api.github.com/repos\(url.path)/releases/latest")
+    if let json = response.json as? [String: Any],
+       let assets = json["assets"] as? [[String: Any]],
+       let plgzItem = assets.first(where: { ($0["name"] as? String)?.hasSuffix(".iinaplgz") ?? false }),
+       let dlURL = plgzItem["browser_download_url"] as? String
+    {
+      do {
+        let plgzName = plgzItem["name"] as! String
+        let destURL = Utility.tempDirURL.appendingPathComponent(plgzName)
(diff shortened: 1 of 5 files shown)


iina/JavascriptPlugin.swift
```diff
@@ -250,6 +250,27 @@ class JavascriptPlugin: NSObject {
...
+       let dlURL = plgzItem["browser_download_url"] as? String
+    {
+      do {
+        let plgzName = plgzItem["name"] as! String
+        let destURL = Utility.tempDirURL.appendingPathComponent(plgzName)
+        let downloadResponse = Just.get(dlURL)
+        try downloadResponse.content?.write(to: destURL)
+        return try create(fromPackageURL: destURL)
+      } catch {
+        // ignore any error
```
saagarjha: probably worth logging here

iina/PreferenceWindowController.swift
```diff
@@ -350,6 +353,17 @@ class PreferenceWindowController: NSWindowController {
...
 
+  func performAction(_ action: Action) {
+    switch action {
+    case .installPlugin(url: let url):
+      guard let idx = viewControllers.firstIndex(where: { $0 is PrefPluginViewController }) else {
+        return
+      }
+      loadTab(at: idx)
+      let vc = viewControllers[idx] as! PrefPluginViewController
+      vc.perform(#selector(vc.installPluginAction(localPackageURL:)), with: url, afterDelay: 0.25)
```
saagarjha: I assume this delay is to make the preference window appear? Is there really no better way to do this?

PR #4568, opened 2023-08-08
Add Settings checkbox to disable EOF restart (addresses #3785)

### LINKED ISSUE #3785: Starting in 1.3.0, pressing Play (spacebar) while at the end of a video file starts playing the file again from the beginning
- macOS 10.15.7
- IINA 1.3.0 Build 132

**Expected behavior:**

Whenever IINA is paused/stopped at the end of a file, pressing Play (spacebar) should not do anything since there is nothing left to play

**Actual behavior:**

Since upgrading to 1.3.0, IINA now loops back and starts playing again from the beginning the file. This was not the case in the previous release.

**Steps to reproduce:**

Open any file, go to the end of it and press the keyboard shortcut for Play (space bar)

confirmed on mpv 34.1 that this behvavior is not present.

Everytime.

I even tried setting `loop-file` and `playlist-loop` to `no` just in case but this did not fix the issue.

svobs:
- [x] This implements/fixes issue #3785.

---

**Description:**

~~NOTE: If testing this, **do not use key bindings**! Use either the `Playback` menu or the Play button in the OSC. There are currently a couple of open bugs which are causing this key binding to be unreliable.~~

Trying to push this along. Not in love with the text, but would love suggestions.
SCR-20230808-crzq

~~Also encountered an issue with the Play button. Because it's implemented as a Cocoa toggle button with 2 states, clicking on it will always toggle the icon. Couldn't find an easy way to tell it not to toggle, which presented a problem when at the end and the Play button does nothing. Added a quick & dirty fix by adding code to the `syncUI` mechanism to bring it back in sync with the actual play state, but there may be numerous other ways to do it and seemed like that could be a whole issue by itself.~~

This also changes `playButton` into a push button instead of a toggle button. This means that its image does not automatically change in response to being clicked; it can only be changed via code. (I checked mpv's UI and it appears to handle its play button similarly. The image does not toggle between play & pause at mouseDown.)

Doing this was necessary to prevent the button image from changing from `pause` to `play` when clicked even when it is not possible to play (e.g. at EOF when this PR's new setting is unchecked). It was also necessary to make some minor logic changes to ensure that state always flows down from `info.state == .paused` to `playButton.image`.

### DIFF
--- a/iina/Base.lproj/MainWindowController.xib
+++ b/iina/Base.lproj/MainWindowController.xib
@@ -561,7 +561,7 @@
                 <button translatesAutoresizingMaskIntoConstraints="NO" id="gxw-pJ-Lcg">
                     <rect key="frame" x="62" y="0.0" width="24" height="24"/>
                     <buttonCell key="cell" type="square" bezelStyle="shadowlessSquare" image="play" imagePosition="only" alignment="center" alternateImage="pause" refusesFirstResponder="YES" state="on" imageScaling="proportionallyUpOrDown" inset="2" id="mWQ-R0-GGr">
-                        <behavior key="behavior" pushIn="YES" changeContents="YES" lightByContents="YES"/>
+                        <behavior key="behavior" pushIn="YES" lightByBackground="YES" lightByGray="YES"/>
                         <font key="font" metaFont="system"/>
                     </buttonCell>
                     <constraints>
--- a/iina/Base.lproj/MiniPlayerWindowController.xib
+++ b/iina/Base.lproj/MiniPlayerWindowController.xib
@@ -147,7 +147,7 @@
                                     <button translatesAutoresizingMaskIntoConstraints="NO" id="aC9-OK-a4x">
                                         <rect key="frame" x="138" y="8" width="28" height="28"/>
                                         <buttonCell key="cell" type="square" bezelStyle="shadowlessSquare" image="play" imagePosition="only" alignment="center" alternateImage="pause" refusesFirstResponder="YES" state="on" imageScaling="proportionallyUpOrDown" inset="2" id="F9X-N7-Uv0">
-                                            <behavior key="behavior" pushIn="YES" changeContents="YES" lightByContents="YES"/>
+                                            <behavior key="behavior" pushIn="YES" lightByBackground="YES" lightByGray="YES"/>
                                             <font key="font" metaFont="system"/>
                                         </buttonCell>
                                         <constraints>
--- a/iina/Base.lproj/PrefGeneralViewController.xib
+++ b/iina/Base.lproj/PrefGeneralViewController.xib
@@ -93,6 +93,16 @@
                         <binding destination="IBP-Mz-Maa" name="value" keyPath="values.keepOpenOnFileEnd" id="4qh-CV-UAH"/>
                     </connections>
                 </button>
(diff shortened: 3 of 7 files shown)


low-batt:
I too am struggling with a name for this setting. As there are multiple ways to restart playback I would go generic and avoid using controls names such as "play" and "resume". Also, possibly we can rely upon the context of being a sub-option of `Keep window open after playback finishes`. Maybe: `Restarting playback seeks to the beginning`? I'm not entirely happy with that.

On this:
> Because it's implemented as a Cocoa toggle button with 2 states, clicking on it will always toggle the icon. Couldn't find an easy way to tell it not to toggle, which presented a problem when at the end and the Play button does nothing.

The correct solution is to disable the button once EOF is reached if the user has turned off the setting to restart from the beginning. This also means the `Resume` menu item should be disabled. With the feature of automatically restarting from the beginning disabled you can't start playback while at EOF. The Apple way to indicate controls and menu items are not applicable is to disable them. If the user changes the playback position then these controls would be enabled again.

With this PR not addressing the issues with the spacebar, which is what issue #3785 is reporting we can't say this PR fixes that issue. I'm thinking this needs to be put in draft mode until the key binding problems are sorted out. IINA is too far behind on reviewing and merging. I will see if I can get the attention of the other developers.

svobs:
> The correct solution is to disable the button once EOF is reached if the user has turned off the setting to restart from the beginning. This also means the Resume menu item should be disabled. With the feature of automatically restarting from the beginning disabled you can't start playback while at EOF. The Apple way to indicate controls and menu items are not applicable is to disable them. 

Yeah, definitely a rabbit hole of its own...

So I actually attempted this, and it's harder than it seems. We can't generalize that the play button should be disabled at EOF, because it may still be legitimate to play the next item in the playlist. Or loop back from the beginning of the playlist. Or something else. It will require careful consideration of all the possibilities so that we don't risk introducing a new bug where it worked fine before.

Also, what about the `Playback` > `Resume` menu item? And what about the Next and Prev buttons? I would think they should match the behavior of the `Play` button too.

Considering the effort that would go into that, let's not have perfect be the enemy of the good? It's a pretty minor issue to worry about, considering it's blocking a fix which multiple users have expressed interest in getting resolved.

> Maybe: `Restarting playback seeks to the beginning`? I'm not entirely happy with that.

That could work. Will continue to ponder...

> I'm thinking this needs to be put in draft mode until the key binding problems are sorted out

Looks like that should happen soon, but I'll put in draft for now as per request.

low-batt:
Thanks for putting this into draft mode. I hope the devs will find some time for reviewing and merging soon.

> Also, what about the Playback > Resume menu item?

I mentioned that in my comment. Yes, that menu item also needs to be disabled. I don't think the next and prev operations are tied to this.

Disabling is definitely the correct way to handle this. It is not a case of disabling at EOF. It is disabling only when pressing space does not start playback. Exactly the same conditions the code already has to check.

svobs:
Rebased and fixed conflicts, and made a further refinement. I stumbled onto a way to fix the play button behavior which allows this PR to appear complete. See the updated description.

low-batt:
When `Keep window open after playback finishes` is not checked the new sub-option `Restart media from the end using Play or Resume` should be disabled, similar to how `Not while in "music mode" or only playing audio` behaves.

The new text needs to be added to the English `PrefGeneralViewControler.strings`.

PR #4572, opened 2023-08-10
Don't occupy main thread when emitting logs

uiryuu:

---

**Description:**
- Logs are stored in `Logger.logs` temporarily
- When Log Viewer is shown, the logs will be synced every 0.1s
- The Log Viewer will follow the latest log if previously the latest is shown
- The Log Viewer will not scroll when a new log is fetched if previously the latest log is not shown

### DIFF
--- a/iina/LogWindowController.swift
+++ b/iina/LogWindowController.swift
@@ -39,6 +39,9 @@ class LogWindowController: NSWindowController, NSMenuDelegate {
     }
     levelPopUpButton.selectItem(withTag: Logger.Level.preferred.rawValue)
     subsystemPopUpButton.menu!.delegate = self
+
+    Timer.scheduledTimer(timeInterval: 0.1, target: self, selector: #selector(syncLogs), userInfo: nil, repeats: true)
+    syncLogs()
   }
 
   fileprivate static func indicatorIcon(withColor color: NSColor) -> NSImage {
@@ -116,6 +119,24 @@ class LogWindowController: NSWindowController, NSMenuDelegate {
     let pasteboard = NSPasteboard.general
     pasteboard.clearContents()
     pasteboard.setString(string, forType: .string)
+  }
+
+  // MARK: - Logs
+
+  @objc private func syncLogs()
+  {
+    guard !Logger.logs.isEmpty else { return }
+    var scroll = false
+    let range = logTableView.rows(in: logTableView.visibleRect)
+    if range.location + range.length == logs.count {
+      scroll = true
+    }
+
+    logs.append(contentsOf: Logger.logs)
+    Logger.logs.removeAll()
+    if scroll {
+      logTableView.scrollRowToVisible(logs.count - 1)
+    }
   }
 }
 
--- a/iina/LogWindowController.swift
+++ b/iina/LogWindowController.swift
@@ -40,7 +40,6 @@ class LogWindowController: NSWindowController, NSMenuDelegate {
     levelPopUpButton.selectItem(withTag: Logger.Level.preferred.rawValue)
     subsystemPopUpButton.menu!.delegate = self
 
-    Timer.scheduledTimer(timeInterval: 0.1, target: self, selector: #selector(syncLogs), userInfo: nil, repeats: true)
     syncLogs()
   }
 
@@ -123,19 +122,21 @@ class LogWindowController: NSWindowController, NSMenuDelegate {
 
   // MARK: - Logs
 
-  @objc private func syncLogs()
-  {
-    guard !Logger.logs.isEmpty else { return }
-    var scroll = false
-    let range = logTableView.rows(in: logTableView.visibleRect)
-    if range.location + range.length == logs.count {
-      scroll = true
-    }
+  @objc func syncLogs() {
+    guard isWindowLoaded else { return }
+    Logger.$logs.withLock() { logs in
+      guard !logs.isEmpty else { return }
+      var scroll = false
+      let range = logTableView.rows(in: logTableView.visibleRect)
+      if range.location + range.length >= self.logs.count {
+        scroll = true
+      }
 
-    logs.append(contentsOf: Logger.logs)
-    Logger.logs.removeAll()
-    if scroll {
-      logTableView.scrollRowToVisible(logs.count - 1)
+      self.logs.append(contentsOf: logs)
+      logs.removeAll()
+      if scroll {
+        logTableView.scrollRowToVisible(self.logs.count - 1)
+      }
     }
   }
 }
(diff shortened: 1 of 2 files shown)


iina/LogWindowController.swift
```diff
@@ -39,6 +39,9 @@ class LogWindowController: NSWindowController, NSMenuDelegate {
     }
     levelPopUpButton.selectItem(withTag: Logger.Level.preferred.rawValue)
     subsystemPopUpButton.menu!.delegate = self
+
+    Timer.scheduledTimer(timeInterval: 0.1, target: self, selector: #selector(syncLogs), userInfo: nil, repeats: true)
```
low-batt: This adds a violation of [Best Practices](https://developer.apple.com/library/archive/documentation/Performance/Conceptual/power_efficiency_guidelines_osx/BestPractices.html#//apple_ref/doc/uid/TP40013929-CH38-SW1) in Apple's Energy Efficiency Guide for Mac Apps. CPU samples show this timer continues to run after the log window has been closed. The timer should be invalidated like the changes in PR  #4546 do for the inspector window when it is closed. As we allow the log viewer to be minimized to the dock, that is another situation where the timer really should not be running, but not as critical as handling the case where the log viewer window is closed.

iina/Logger.swift
```diff
@@ -201,9 +203,7 @@ class Logger: NSObject {
     let date = Date()
     let string = formatMessage(message, level, subsystem, true, date)
     let log = Log(subsystem: subsystem.rawValue, level: level.rawValue, message: message, date: dateFormatter.string(from: date), logString: string)
-    DispatchQueue.main.async {
-      (NSApp.delegate as? AppDelegate)?.logWindow.logs.append(log)
-    }
+    logs.append(log)
```
low-batt: No good. Swift arrays are not atomic, as the TSan error below shows. If we are not going to dispatch to the main thread, then locking must be used.

I'm thinking the code pattern used for the `subsystems` array should be followed for the `logs` array as well. Add the `@Atomic` property wrapper and then use the lock that creates for the `logs` property in the same way the `subsystems` lock is used.

<details>
<summary>TSan Swift access race:</summary>

```text
==================
WARNING: ThreadSanitizer: Swift access race (pid=19070)
  Modifying access of Swift variable at 0x0001036178a8 by thread T10:
    #0 static Logger.log(_:level:subsystem:) Logger.swift:206 (IINA:arm64+0x1002ffa58) (BuildId: c3962dc48779319b876b05ec10b3f1e432000000200000000100000000000b00)
    #1 PlayerCore.trackListChanged() PlayerCore.swift:1628 (IINA:arm64+0x1006f46f0) (BuildId: c3962dc48779319b876b05ec10b3f1e432000000200000000100000000000b00)
    #2 MPVController.handlePropertyChange(_:_:) MPVController.swift:1233 (IINA:arm64+0x1007781e4) (BuildId: c3962dc48779319b876b05ec10b3f1e432000000200000000100000000000b00)
    #3 MPVController.handleEvent(_:) MPVController.swift:908 (IINA:arm64+0x10076d88c) (BuildId: c3962dc48779319b876b05ec10b3f1e432000000200000000100000000000b00)
    #4 closure #1 in MPVController.readEvents() MPVController.swift:832 (IINA:arm64+0x10076b850) (BuildId: c3962dc48779319b876b05ec10b3f1e432000000200000000100000000000b00)
    #5 partial apply for closure #1 in MPVController.readEvents() <compiler-generated> (IINA:arm64+0x100787de4) (BuildId: c3962dc48779319b876b05ec10b3f1e432000000200000000100000000000b00)
    #6 thunk for @escaping @callee_guaranteed () -> () <compiler-generated> (IINA:arm64+0x100043130) (BuildId: c3962dc48779319b876b05ec10b3f1e432000000200000000100000000000b00)
    #7 __tsan::invoke_and_release_block(void*) <null>:178268480 (libclang_rt.tsan_osx_dynamic.dylib:arm64+0x87684) (BuildId: b4e4be85d92a32d7a3e9edbe68feb69132000000200000000100000000000b00)
    #8 _dispatch_client_callout <null>:165782528 (libdispatch.dylib:arm64e+0x6998) (BuildId: 6c6be4e9b2013b028e6933df61ff3a4432000000200000000100000000050d00)

  Previous modifying access of Swift variable at 0x0001036178a8 by thread T1:
    #0 static Logger.log(_:level:subsystem:) Logger.swift:206 (IINA:arm64+0x1002ffa58) (BuildId: c3962dc48779319b876b05ec10b3f1e432000000200000000100000000000b00)
    #1 closure #4 in PlayerCore.fileStarted(path:) PlayerCore.swift:1477 (IINA:arm64+0x1006ed750) (BuildId: c3962dc48779319b876b05ec10b3f1e432000000200000000100000000000b00)
    #2 partial apply for closure #4 in PlayerCore.fileStarted(path:) <compiler-generated> (IINA:arm64+0x100711bec) (BuildId: c3962dc48779319b876b05ec10b3f1e432000000200000000100000000000b00)
    #3 thunk for @escaping @callee_guaranteed () -> () <compiler-generated> (IINA:arm64+0x100043130) (BuildId: c3962dc48779319b876b05ec10b3f1e432000000200000000100000000000b00)
    #4 __tsan::invoke_and_release_block(void*) <null>:178268480 (libclang_rt.tsan_osx_dynamic.dylib:arm64+0x87684) (BuildId: b4e4be85d92a32d7a3e9edbe68feb69132000000200000000100000000000b00)
    #5 _dispatch_client_callout <null>:165782528 (libdispatch.dylib:arm64e+0x6998) (BuildId: 6c6be4e9b2013b028e6933df61ff3a4432000000200000000100000000050d00)

  Location is global 'static Logger.logs' at 0x0001036178a8 (IINA+0x100a578a8)

  Thread T10 (tid=3691782, running) is a GCD worker thread

  Thread T1 (tid=3691763, running) is a GCD worker thread

SUMMARY: ThreadSanitizer: Swift access race Logger.swift:206 in static Logger.log(_:level:subsystem:)
==================
ThreadSanitizer report breakpoint hit. Use 'thread info -s' to get extended information about the report.
```
</details>

low-batt (changes requested):
The problem with multiple threads accessing an array must be corrected. The failure to shutdown the timer really should be corrected.

iina/LogWindowController.swift
low-batt: After thinking on this some more, I think a A non-repeating timer should be used instead of a repeating timer. Using a non-repeating timer allows IINA to go idle when logs are not being emitted without the need for extra logic. Something like the following…

When a message is logged:
Add the log to the array.
Check if a non-repeating timer has already been scheduled, if not schedule one.
This would have to be done under a lock, but locking needs to be added anyway to coordinate access to the array as discussed below.

This way IINA goes idle when playback is paused even if the log viewer is open.

low-batt (approved):
Pulled PR, built and tested latest commit. Changes look good.

PR #4575, opened 2023-08-12
Include core video return code in fatal error

low-batt:
This commit will add code to `VideoView` that includes the result code returned by `CVDisplayLinkCreateWithActiveCGDisplays` in the alert posted when a display link can not be created.

---

**Description:**

### DIFF
--- a/iina/VideoView.swift
+++ b/iina/VideoView.swift
@@ -204,14 +204,27 @@ class VideoView: NSView {
 
   // MARK: Display link
 
-  func startDisplayLink() {
-    if link == nil {
-      checkResult(CVDisplayLinkCreateWithActiveCGDisplays(&link),
-                  "CVDisplayLinkCreateWithActiveCGDisplays")
-    }
+  /// Returns a [Core Video](https://developer.apple.com/documentation/corevideo) display link.
+  ///
+  /// If a display link has already been created then that link will be returned, otherwise a display link will be created and returned.
+  ///
+  /// - Note: Issue [#4520](https://github.com/iina/iina/issues/4520) reports a case where it appears the call to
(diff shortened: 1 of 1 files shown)


low-batt:
We have not been able to determine the cause of issue #4520 where it appears the call to [CVDisplayLinkCreateWithActiveCGDisplays](https://developer.apple.com/documentation/corevideo/1456863-cvdisplaylinkcreatewithactivecgd) is failing. The commit in this pull request merely enhances the error reporting by including the return code in the fatal alert posted.

Possibly `link` should be changed to be a lazy stored property. Given there is unexplained trouble in this area I restricted the changes to just improving the error reporting.

The updated alert will look something like this if creation of the display link fails:

display-link-alert

iina/VideoView.swift
```diff
@@ -204,14 +204,27 @@ class VideoView: NSView {
...
+  ///
+  /// If a display link has already been created then that link will be returned, otherwise a display link will be created and returned.
+  ///
+  /// - Note: Issue [#4520](https://github.com/iina/iina/issues/4520) reports a case where it appears the call to
+  ///[CVDisplayLinkCreateWithActiveCGDisplays](https://developer.apple.com/documentation/corevideo/1456863-cvdisplaylinkcreatewithactivecgd) is failing. In case that failure is
+  ///encountered again this method is careful to log any failure and include the [result code](https://developer.apple.com/documentation/corevideo/1572713-result_codes) in the alert displayed
+  /// by `Logger.fatal`.
+  /// - Returns: A [CVDisplayLink](https://developer.apple.com/documentation/corevideo/cvdisplaylink-k0k).
+  private func obtainDisplayLink() -> CVDisplayLink {
+    guard link == nil else { return link! }
```
uiryuu:
```suggestion
    if let link = link { return link }
```

uiryuu (changes requested):

PR #4590, opened 2023-08-17
Make open/save dialogs block window input while open (fixes #4583)

### LINKED ISSUE #4583: Pressing arrow keys to move the cursor in some input fields affects playback
In "save downloaded subtitle" dialog, pressing arrow keys to move the cursor in filename input affects playback.

**System and IINA version:**

- macOS Ventura 13.4
- IINA 1.3.3 Build 138

**Expected behavior:**

Video duration to stay the same

**Actual behavior:**

Video rewinds when pressed left arrow key to move the cursor inside input field, fast forwards when pressed right arrow key, just like they normally do in playback window.

**Steps to reproduce:**

Download an online subtitle, open "save download subtitle" dialog, try to move the cursor in filename field with arrow keys.

All the time

svobs:
- [x] This implements/fixes issue #4583.

---

**Description:**

This fixes the use of arrow key navigation in the following open/save dialogs:
1. "Load external audio file" (Audio Settings sidebar)
2. "Load external subtitle" (Subtitles menu)
3. "Load subtitle..." (Subtitles Settings sidebar)
4. "Save downloaded subtitle" (Subtitles menu)

Some users were complaining that mpv bindings were getting activated in response to arrow keys when they wanted to use them to navigate within the file picker dialog. This ensures that they will work, but it comes at some cost. They now use Apple's tyrannical modal dialog state, which clings onto the window and won't let the user do anything with it until the dialog is dismissed. I'm not a fan, although using it is probably adheres more closely to Apple's HIG than using an independent window.

### DIFF
--- a/iina/MainMenuActions.swift
+++ b/iina/MainMenuActions.swift
@@ -317,7 +317,8 @@ extension MainMenuActionHandler {
 extension MainMenuActionHandler {
   @objc func menuLoadExternalSub(_ sender: NSMenuItem) {
     let currentDir = player.info.currentURL?.deletingLastPathComponent()
-    Utility.quickOpenPanel(title: "Load external subtitle file", chooseDir: false, dir: currentDir) { url in
+    Utility.quickOpenPanel(title: "Load external subtitle file", chooseDir: false, dir: currentDir,
+                           sheetWindow: player.currentWindow) { url in
       self.player.loadExternalSubFile(url, delay: true)
     }
   }
@@ -401,8 +402,8 @@ extension MainMenuActionHandler {
     }
     let subURL = URL(fileURLWithPath: path)
     let subFileName = subURL.lastPathComponent
-    Utility.quickSavePanel(title: NSLocalizedString("alert.sub.save_downloaded.title",
-         comment: "Save Downloaded Subtitle"), filename: subFileName) { (destURL) in
+    let windowTitle = NSLocalizedString("alert.sub.save_downloaded.title", comment: "Save Downloaded Subtitle")
+    Utility.quickSavePanel(title: windowTitle, filename: subFileName, sheetWindow: player.currentWindow) { (destURL) in
       do {
         // The Save panel checks to see if a file already exists and if so asks if it should be
         // replaced. The quickSavePanel would not have called this code if the user canceled, so if
--- a/iina/MainMenuActions.swift
+++ b/iina/MainMenuActions.swift
@@ -29,7 +29,7 @@ class MainMenuActionHandler: NSResponder {
   }
 
   @objc func menuSavePlaylist(_ sender: NSMenuItem) {
-    Utility.quickSavePanel(title: "Save to playlist", types: ["m3u8"]) { (url) in
+    Utility.quickSavePanel(title: "Save to playlist", types: ["m3u8"], sheetWindow: player.currentWindow) { (url) in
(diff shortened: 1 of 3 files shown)


low-batt:
I too am not a fan of sheets, but…

On the question of what the HIG has to say, [Panels](https://developer.apple.com/design/human-interface-guidelines/panels) says:
> In a macOS app, a panel typically floats above other open windows providing supplementary controls, options, or information related to the active window or current selection.

The HIG section on sheets can be found [here](https://developer.apple.com/design/human-interface-guidelines/sheets#macOS).

I did not find the HIG that helpful. Looking at the behavior of `TextEdit` the convention seems to be that you use a floating panel for an operation that is not tied to a specific window, but if the operation is specific to a particular window then a sheet is used.

Applied to IINA that suggests operations such as loading an external subtitle file should be a sheet as it is to be applied to a specific window.

This certainly is a significant change that will be noticed by users.

The `NSOpenPanel` put up by `AppDelegate.openFile` does not exhibit the problem of passing key strokes to the player window. That panel is displayed using [runModal](https://developer.apple.com/documentation/appkit/nssavepanel/1525357-runmodal).

As a side issue, that portion of the HIG also contains the following which IINA is violating:
> **Refer to panels by title in your interface and in help documentation.** In menus, use the panel’s title without including the term panel: for example, “Show Fonts,” “Show Colors,” and “Show Inspector.”

low-batt (changes requested):
The menu item `Save Current Playlist…` under the `File` menu still exhibits the the problem.

svobs:
> The NSOpenPanel put up by AppDelegate.openFile does not exhibit the problem of passing key strokes to the player window. That panel is displayed using [runModal](https://developer.apple.com/documentation/appkit/nssavepanel/1525357-runmodal).

This is very interesting. I suppose it's a special case because its action is attached to `AppDelegate` and thus automagically considered to be an app-wide feature (maybe), rather than an action tied to a player window. However it blocks key events to all the windows in the app, which is even less desirable. I think I'd be happier with sheets if they were easier to resize vertically and move around - seems very unintuitive that you are expected to grab the title bar of the window behind it, the one which is otherwise blocked from all events.

> The menu item Save Current Playlist… under the File menu still exhibits the the problem.

Added a new commit to include this.

> As a side issue, that portion of the HIG also contains the following which IINA is violating:
> 
> > Refer to panels by title in your interface and in help documentation. In menus, use the panel’s title without including the term panel: for example, “Show Fonts,” “Show Colors,” and “Show Inspector.”

Ah, do you mean that `Load Subtitle...`, et al, should be renamed `Open Subtitle...`? Yeah I think I agree with that - it would be more "Mac-like".

low-batt:
No, it was this part of the HIG requirement I was thinking of:
> without including the term panel

Which means the menu item `Show Video Panel` and others are clearly in violation. Bummer. Right now I think we should pretend I did not notice this.

low-batt (approved):
Pulled latest commit and tested. Changes look good.

low-batt:
By the way I'm now using the [GitHub CLI](https://cli.github.com/) to checkout PRs:
```text
gh pr checkout 4590
```

The CLI automatically handles the case where the author has forced pushed changes to the PR.

uiryuu (approved):
I think the proposed the changes are reasonable.

Issue #4559, opened 2023-08-04
Thumbnail visible with OSC hidden


low-batt:
**System and IINA version:**

- macOS 13.5
- IINA 1.3.3

**Expected behavior:**
When the OSC is hidden the thumbnail showed by the OSC is hidden as well.

**Actual behavior:**
As can be seen in this macOS screen shot the OSC disapeared, the video continued playing and the thumbnail continued to be shown:

![issue-thumbnail](https://github.com/iina/iina/assets/86170219/24da861e-5211-41af-adc4-b120cd6a2bdf)

**Steps to reproduce:**
I have no idea how to reproduce this.

- [x] MPV does not have this problem.

**How often does this happen?**
So far once in my entire lifetime.

uiryuu:
@low-batt Can you help to test whether this still exists in the current develop branch?

low-batt:
I played around, but I can't even reproduce this with the released version of IINA. This is one where if we made a change that we think could have addressed it we will need to classify it as "presumed fixed".

low-batt:
And we now have our answer. Not fixed. Reproduces with IINA 1.4.4 and the `develop` branch. See issue #6288.

PR #4637, opened 2023-10-05
Fix #4156 (OSD stealing mouse clicks from OSC even while hidden), #4612: (OSC responds to mouse hover when behind OSD)

### LINKED ISSUE #4156: With seek bar/controller moved to top left, parts of seek bar doesn't respond.

### LINKED ISSUE #4612: OSC responds to mouse when behind the OSD
**System and IINA version:**

- macOS 13.5
- IINA 1.3.3

**Expected behavior:**
OSC does not respond to mouse when behind the OSD.

**Actual behavior:**
As can be seen here even though the OSC was behind the OSD it responded to the mouse position and displayed a thumbnail, blocking a portion of the OSD:
osc-thumbnail

- [x ] MPV does not have this problem.

Every time.

svobs:
  - #4156 
  - #4612.

---

**Description:**

Could not have been a more straightforward fix. When the OSD was hidden, its alpha was set to 0, but `isHidden` was never set to `true` (looks deleted or forgot to include).

Update: added a fix for #4612, since it needed the original fix as a prerequisite.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -2018,6 +2018,7 @@ class MainWindowController: PlayerWindowController {
     }) {
       if self.osdAnimationState == .willHide {
         self.osdAnimationState = .hidden
+        self.osdVisualEffectView.isHidden = true
         self.osdStackView.views(in: .bottom).forEach { self.osdStackView.removeView($0) }
       }
     }
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -1056,8 +1056,7 @@ class MainWindowController: PlayerWindowController {
         timePreviewWhenSeek.isHidden = false
         thumbnailPeekView.isHidden = !player.info.thumbnailsReady
       }
-      let mousePos = playSlider.convert(event.locationInWindow, from: nil)
-      updateTimeLabel(mousePos.x, originalPos: event.locationInWindow)
+      refreshSeekTimeAndThumnail(from: event)
     }
   }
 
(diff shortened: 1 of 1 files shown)


svobs:
Added another commit which fixes #4612.

It also adds an `else` which takes care to hide the thumbnail when it is determined it should not be shown. This may help with #4559.

iina/MainWindowController.swift
```diff
@@ -2265,6 +2262,16 @@ class MainWindowController: PlayerWindowController {
     }
   }
 
+  private func refreshSeekTimeAndThumnail(from event: NSEvent) {
+    let isCoveredByOSD = !osdVisualEffectView.isHidden && isMouseEvent(event, inAnyOf: [osdVisualEffectView])
```
uiryuu, line 2266: I also found that if the sidebar covers the thumbnail, it will also trigger the thumbnail, and I've confirmed that something like this
`let isCoveredBySidebar = !sideBarView.isHidden && isMouseEvent(event, inAnyOf: [sideBarView])`
and checking both `isCoveredByOSD` and `isCoveredBySidebar` is able to fix this problem

uiryuu (changes requested):
Sidebar can also cover the thumbnail, please also fix this issue. Otherwise this pr is good to merge.

iina/MainWindowController.swift
```diff
@@ -2265,6 +2262,16 @@ class MainWindowController: PlayerWindowController {
     }
   }
 
+  private func refreshSeekTimeAndThumnail(from event: NSEvent) {
```
Coeur: Would you mind renaming it without a typo?
```suggestion
  private func refreshSeekTimeAndThumbnail(from event: NSEvent) {
```

svobs: Pushed another commit which should address feedback, thanks.

uiryuu:
Thanks!

PR #4668, opened 2023-10-24
Fix noticeable wobble if video is playing while window is resized (issue #4667)

### LINKED ISSUE #4667: Video wobbles if window is resized while playing
**System and IINA version:**

- macOS 14.0 Sonoma
- IINA 1.3.3

**Expected behavior:**
Smooth animation during trackpad pinch-to-zoom

_Edit: also occurs when resizing window by any means while video is playing._

**Actual behavior:**

https://github.com/iina/iina/assets/2213815/7c2c73a9-e267-44c2-b751-234673783d93

https://github.com/iina/iina/assets/2213815/a4dcdb84-c66d-4dd5-9646-79008f79f55f

<br>

**Steps to reproduce:**
1. Go to IINA's `Settings` > `Control` > set `Pinch to:` to `Adjust window size`
2. Open a video in IINA and begin playing it.
3. Pinch the video. (Seems more prevalent on lower-FPS but high-quality videos. I picked a bad one to show above because it's 60fps) and I was playing it on a 60Hz display.

Usually.

svobs:
- [x] This implements/fixes issue #4667.

---

**Description:**

1. Sets `isAsynchronous` flag on `videoLayer`, which allows video redraw requests to be initiated asynchronously by the view (and altering the DisplayLink interval slightly) when needed.
2. Adds new checks to prevent crash during quit.

For more info, see the comments for issue #4667.

### DIFF
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -518,6 +518,7 @@ not applying FFmpeg 9599 workaround
 
   func shouldRenderUpdateFrame() -> Bool {
     guard let mpvRenderContext = mpvRenderContext else { return false }
+    guard !player.isStopping && !player.isShuttingDown else { return false }
     let flags: UInt64 = mpv_render_context_update(mpvRenderContext)
     return flags & UInt64(MPV_RENDER_UPDATE_FRAME.rawValue) > 0
   }
--- a/iina/VideoView.swift
+++ b/iina/VideoView.swift
@@ -213,7 +213,14 @@ class VideoView: NSView {
       Logger.fatal("Cannot Create display link!")
     }
     guard !CVDisplayLinkIsRunning(link) else { return }
+
+    /// Set this to `true` to enable video redraws to match the timing of the view redraw during animations.
+    /// This fixes a situation where the layer size may not match the size of its superview at each redraw,
+    /// which would cause noticable clipping or wobbling during animations.
+    videoLayer.isAsynchronous = true
+
     updateDisplayLink()
+
     checkResult(CVDisplayLinkSetOutputCallback(link, displayLinkCallback, mutableRawPointerOf(obj: self)),
                 "CVDisplayLinkSetOutputCallback")
     checkResult(CVDisplayLinkStart(link), "CVDisplayLinkStart")
@@ -221,6 +228,11 @@ class VideoView: NSView {
 
   @objc func stopDisplayLink() {
     guard let link = link, CVDisplayLinkIsRunning(link) else { return }
+
+    /// If this is set to `true` while the video is paused, there is some degree of busy-waiting as the
+    /// layer is polled at a high rate about whether it needs to draw. Disable this to save CPU while idle.
+    videoLayer.isAsynchronous = false
+
     checkResult(CVDisplayLinkStop(link), "CVDisplayLinkStop")
   }
 
--- a/iina/ViewLayer.swift
+++ b/iina/ViewLayer.swift
@@ -26,7 +26,7 @@ class ViewLayer: CAOpenGLLayer {
     super.init()
 
     isOpaque = true
-    isAsynchronous = false
+    isAsynchronous = true
 
     autoresizingMask = [.layerWidthSizable, .layerHeightSizable]
   }
--- a/iina/ViewLayer.swift
+++ b/iina/ViewLayer.swift
@@ -26,7 +26,6 @@ class ViewLayer: CAOpenGLLayer {
     super.init()
 
     isOpaque = true
-    isAsynchronous = true
 
     autoresizingMask = [.layerWidthSizable, .layerHeightSizable]
   }
--- a/iina/ViewLayer.swift
+++ b/iina/ViewLayer.swift
@@ -94,6 +94,12 @@ class ViewLayer: CAOpenGLLayer {
 
   override func draw(inCGLContext ctx: CGLContextObj, pixelFormat pf: CGLPixelFormatObj, forLayerTime t: CFTimeInterval, displayTime ts: UnsafePointer<CVTimeStamp>?) {
     let mpv = videoView.player.mpv!
+
+    CGLLockContext(ctx)
+    defer {
+      CGLUnlockContext(ctx)
+    }
+    
     needsMPVRender = false
 
     glClear(GLbitfield(GL_COLOR_BUFFER_BIT))


krackers:
If I understand this PR correctly, it always sets `isAsynchronous` to true, and iina never changes it back to false. Wouldn't this cause unnecessary draws (since the system might potentially request a draw up to once every vsync)? (If libmpv does not have a new frame it will redraw previous frame so it probably is not an issue in video-sync=audio mode, but for video-sync=display-resample where timing constraints are much tighter it might be possible that the asynchronous draw initiated with the system will overlap with the draw requested by libmpv. There is a mutex to guard against multiple threads drawing at the same time, but the wait here might lead to unnecessarily delayed frames).

Perhaps a more conservative solution might be to only set `isAsynchronous` when we are inside a resize?

svobs:
@krackers thanks for bringing up these questions. I should start by saying I'm very new to `CAOpenGLLayer`, how mpv interacts with it, and video timing in general, so I'm trying to learn as I go along. I did some further testing based on your questions. Let me start with this though:

> Perhaps a more conservative solution might be to only set isAsynchronous when we are inside a resize?

This doesn't seem completely possible, unfortunately. The APIs are a little messy in this area, and there are a few different mechanisms which we might all consider a "resize", but not all of them have a complete set of notifications. There is the traditional "live resize" which comes from dragging the window edges, and that can be covered, but for implicit animations which are created from a call to `window.setFrame()` (which IINA mostly uses), or as the result of a constraint update, or one of the mysterious MacOS window manager triggers, only `windowDidResize` is fired. There's no way to always know when the resize is starting. Worse, there's no way to tell when an implicit animation is running (which creates problems due to race conditions in other areas, and one reason `windowDidFailToEnterFullScreen`/`windowDidFailToExitFullScreen` exist).

> If I understand this PR correctly, it always sets `isAsynchronous` to true, and iina never changes it back to false. Wouldn't this cause unnecessary draws (since the system might potentially request a draw up to once every vsync)?

It doesn't cause unnecessary draws. When `isAsynchronous` is true in `ViewLayer`, it results in possible additional draw requests. But it will always call `canDraw()` before calling `draw()`. And unless there is a request to "force" a render (which is a special request by various IINA components and only happens occasionally in very specialized instances), `canDraw()` calls `mpv_render_context_update` and checks for the `MPV_RENDER_UPDATE_FRAME` flag, and otherwise returns `false`.

But there seems to be another interesting effect going on when `isAsynchronous` is set. I ran some basic tests collecting statistics on function calls and printing them to the console once per second, and sampled various conditions. I found:

1. When any video was paused, `canDraw` was called ~120x/sec when the window was on my 120 Hz Macbook display, and ~60x/sec on my external display which was running at 60 Hz. But -
2. On my 120 Hz display, playing a 24 FPS video resulted in ~28 calls to `canDraw` and 24 calls to `draw` per sec.
3. Same display, playing a 30 FPS video resulted in avg 30-31 calls to `canDraw` and 30 calls to `draw` per sec.
4. Same display, playing a 60 FPS video resulted in 60 calls to `canDraw` and 60 calls to `draw` per sec.

So having no visibility into what's going on here, it _looks like_ there is throttling going on here at some level. I read this as there being a 1/30 sec grace period after the last draw before the layer is asked to draw again. That would explain why there are still a few extraneous calls to `canDraw` at 24fps, none at 60fps, and the "spinning" when paused.

> There is a mutex to guard against multiple threads drawing at the same time, but the wait here might lead to unnecessarily delayed frames).

So it's a lot harder for me to quantify how the redraw timing is effected. Still thinking about that. The timing certainly doesn't drift over the long run, but I can't yet say whether each 1/24 sec frame is perfectly evenly spaced in time. Is that even guaranteed by mpv? From the numbers above, I think it _looks like_ the `isAsynchronous`-originated redraw requests are subordinate to the redraws coming from the mpv update callback.  It looks like they are coming from different threads so I may be able to use that.

I pushed an update which sets `isAsynchronous` to `true` only when the DisplayLink is running, and `false` when it is stopped. It's only needed while playing anyway, and I found that the "spinning" mentioned above resulted in a 1% CPU usage if left `true` while idle.

krackers:
>or as the result of a constraint update, or one of the mysterious MacOS window manager triggers, only windowDidResize is fired

I'm not as familiar with appkit APIs here, but does `windowWillResize` do what we want?

>And unless there is a request to "force" a render 

Ah you're right, I forgot about the `canDraw` check. So yeah there should be no additional draws here: since the mutex is around the entire `display()` method (which I assume is what is periodically called by the system when `isAsync = true`) this implicitly includes the call to `canDraw`, so it's ok. Although this assumes that in the case where `canDraw` returns false the system initiated truly is a noop though time-wise (e.g. hopefully order of < 1k µs). I see that we also call `[CATransaction flush]` which from my experience sometimes can take an order magnitude longer; is there any penalty paid for that when nothing has changed (the system implicitly does a flush when it calls us as I understand it, so we'd also be paying the cost of the flush twice). 

>I read this as there being a 1/30 sec grace period after the last draw before the layer is asked to draw again

That's actually really interesting and good to know, it has the net effect that the system's callback frequency basically locks onto the content fps, so it's probably intentional and a clever optimization. I had always assumed that the async mode was basically just triggered off of a  CVDisplayLink. 

>The timing certainly doesn't drift over the long run, but I can't yet say whether each 1/24 sec frame is perfectly evenly spaced in time.

You can read a bit more about this [here](https://github.com/iina/iina/discussions/4652#discussioncomment-7418414) but in the current implementation the libmpv backend actually does not wait for the glFlush before firing the next render callback: after the render callback fires the libmpv backend just waits until the next reported swap before continuing with the render loop (this is not actually a good thing because it means there is no backpressure in the case of a render that takes too long, beyond that given by use of a serial dispatch queue for the rendering which is not necessarily aligned against vsync boundaries). Because the render loop is thus blocked on the dispatch queue, and rendering itself may be slow in practice due to heavy subtitles, the timings between each `glFlush` call won't necessarily be constant even in the current implementation.

But in terms of this PR I don't think the timing should really matter. When mpv has a frame ready for us, either the system callback or the libmpv initiated callback will be the first one to service it. With this change the worst-case time between frame ready and `glFlush()` should not become any worse than it was before (assuming that effective-noop calls to `display()` are "free"). So at least from a theoretical standpoint this should not impact anything.

If you're willing to test a bit, I think quickest way to confirm that this doesn't introduce any issue would be to try the slow-motion 60fps test in https://github.com/iina/iina/discussions/4652#discussioncomment-7418414, using a 60hz monitor.

svobs:
Not a complete reply but a semi-quick update -

> I'm not as familiar with appkit APIs here, but does windowWillResize do what we want?

Unfortunately, often it doesn't. Its [documentation](https://developer.apple.com/documentation/appkit/nswindowdelegate/1419292-windowwillresize) states:
> Tells the delegate that the window is being resized (whether by the user or through one of the setFrame... methods other than [setFrame(_:display:)](https://developer.apple.com/documentation/appkit/nswindow/1419753-setframe)).

This implies some very specific behavior. IINA is using two overloads of `setFrame()` in various places. I was expecting to report that `windowWillResize` was being called for only one of them. But after doing some testing, this documentation does not appear to be correct. I found:
1. `window.setFrame(newFrame, display: true)` triggers only `windowDidResize`
2. `window.setFrame(newFrame, display: true, animate: true)` triggers `windowDidResize`, but also triggers `windowWillStartLiveResize` and `windowDidEndLiveResize`, which is unexpected(!)
3. `window.setFrame(newFrame, display: true, animate: false)` triggers only `windowDidResize`

While it may be possible to craft a workaround which will patch all these spots, seems easier to just prove the viability of this patch, no? ;)

> Ah you're right, I forgot about the canDraw check. So yeah there should be no additional draws here: since the mutex is around the entire display() method (which I assume is what is periodically called by the system when isAsync = true) this implicitly includes the call to canDraw, so it's ok. Although this assumes that in the case where canDraw returns false the system initiated truly is a noop though time-wise (e.g. hopefully order of < 1k µs). I see that we also call [CATransaction flush] which from my experience sometimes can take an order magnitude longer; is there any penalty paid for that when nothing has changed (the system implicitly does a flush when it calls us as I understand it, so we'd also be paying the cost of the flush twice).

There's no extra call to `display()` if `canDraw()` returns `false`. In fact, other than a couple of calls at initial layout, `display()` is not called at all except via the visible call from `draw()`. The call to `super.display()` then calls `canDraw()` and then `draw(inCGLContext...)`. I was able to confirm this via testing.

I don't know exactly why the code needs to call `CATransaction.flush()`.

Or what the "skip render" is supposed to do (in fact it looks like I can make it work fine without it).

> If you're willing to test a bit, I think quickest way to confirm that this doesn't introduce any issue would be to try the slow-motion 60fps test in https://github.com/iina/iina/discussions/4652#discussioncomment-7418414, using a 60hz monitor.

Yeah I saw that and I will give it a go. Trying to juggle my time between different things, but may have some time this weekend.

krackers:
>There's no extra call to display() if canDraw() returns false

Interesting, yeah I looked at the backtrace and it's like

```
[layer drawInCGL...]
[layer canDrawInCGL...]
CAOpenGLLayerDraw
CAOpenGLLayer_timer_callback
```

so `display` is never even called by the system under the regular callback, it just directly jumps into the draw method. However if you set `setNeedsDisplay` then the backtrace that happens is . 

```
[layer display]
Layer::display_if_needed
```

>I don't know exactly why the code needs to call CATransaction.flush().

IIRC this is because it ensures the recent layer updates are queued to be presented as soon as possible, instead of waiting until the end of the runloop for this to happen (there could possibly a long delay between the two).

krackers:
Wait doesn't the fact that the system's call jumps directly to canDraw + draw (bypassing `display`) mean that there's effecitvely no mutex to avoid the situation where both the mpv initiated callback and system initiated callback call `render` at the same time? libmpv does have a mutex internally but that's only to protect internal mpv state. (I'm sorry for all these questions, hopefully I'm not being a nuisance).

Basically the situation that might happen:

* libmpv callback fires -> display (protected by lock) -> canDraw (returns true) -> we're about to enter draw method
* System callback happens -> canDraw (returns true, since we haven't yet called libmpv render to consume the frame) -> we enter draw

Now we have a situation where both libmpv and system are going to call render, then glFlush. libmpv's render method has an internal lock to protect internal libmpv state here (https://github.com/mpv-player/mpv/blob/7480efa62c0a2a1779b4fdaa804a6512aa488400/video/out/vo_libmpv.c#L401) but it does not surround the actual rendering. So it seems what would happen is:

* libmpv triggered render consumes the frame, and enters draw. Assume that we're busy rendering while the below happens
* system triggered render sees that frame is already consumed, and thinks it needs to redraw. But the frame is not fully rendered.  At best case we have a duplicate frame presented, at worst case we present garbage. And there's also the issue that two threads would be trying to access the same opengl context.

Thus do we need surround the draw method with a `CGLLockContext` (or some other mutex mechanism)?

svobs:
> Wait doesn't the fact that the system's call jumps directly to canDraw + draw (bypassing display) mean that there's effecitvely no mutex to avoid the situation where both the mpv initiated callback and system initiated callback call render at the same time? libmpv does have a mutex internally but that's only to protect internal mpv state. (I'm sorry for all these questions, hopefully I'm not being a nuisance).

Not at all! Actually it looks like you are correct. I've updated the PR accordingly. Though I didn't see any blanking or tearing that I could notice. I wonder if it's because each time it's drawing to a separate frame buffer. From some debugging I did see that the different threads were sharing the same context (bad), but `fbo` was alternating between `1`, `2`, and `3`. (I expected 2 buffers. Not sure why there are 3). But even if it works I agree it's not good practice, so I added `CGLLockContext` & `CGLUnlockContext` around the drawing code.

But in the process, I ended up playing with the code a bit more and found I could actually remove the use of the mpv update callback entirely, which also allowed me to remove the `DispatchQueue` and simplify the code, and the result so far seems to be the same. Take a look at PR #4676 which has the resulting changes. I was hesitant to add it here since it feels a bit like mission creep. So in that case, the result relies only on the use of the `CVDisplayLink` and `mpv_render_context_update`. I'm still not clear on how it manages to negotiate the correct frame rate. But it looks like a cleaner solution and removes some ambiguity about which thread has agency over rendering.

krackers:
>but fbo was alternating between 1, 2, and 3

I've actually seem more unique values of `fbo`, from some disassembly iirc it maintains a pool of buffers that it hands out round-robin. It could be using a depth-3 swapchain behind the scenes, not sure. That you only requested a double-buffered context in the opengl parameters really doesn't matter, because with layer-backed views almost everything seems to be a lie anyway.

Edit: I was curious so I just checked again, it will return up to 16 buffers before draw is no longer called. But what I don't know is if all 16 buffers will actually be queued to present. That will require a bit more experimentation to determine.

Edit 2: I was curious enough so I did some tests with the only opengl I know how to write, painting the window a single color. I did a `glClear` with the color chosen based on the frame number. It definitely isn't a 16-depth swap queue, if you do a [catransaction flush] multiple times within a vsync interval only one of the frames actually wins. Now what I don't know is what the limit is for how many frames can actually be in flight at a given time. I painted an r->g->b->w cycle for 16 frames with an 8000us wait in between, and there were dropped frames (there were 3 different fbo used). With a 16000us wait, there were no dropped frames. So based on this I conclude that while caopengllayer will be happy to give you as many as 16 buffers, in practice if you want to avoid dropped frames you only have 2 buffers to work with. (Whether or not I used single, double, or triple buffering in the pixelformat did not make a difference)

>Though I didn't see any blanking or tearing that I could notice

Yes, with layer-backed views each draw is into a separate buffer so there won't be tearing. There could be a chance of dropped frames, but since in the above scenario the two frames being rendered are the same anyway, dropping one doesn't shouldn't really make a difference. Since we have two threads potentially accessing same opengl context there could be a chance for random crash however, and if potentially both frames end up being scheduling for presentation we'd have duplicate frame, or possibly some other random frame gets dropped. I think the reason why no issues were observed was because in practice the libmpv initiated callback services the frame first.

(Tbh I still don't fully understand the precise queuing behavior of layer backed views, and as mentioned in https://github.com/iina/iina/discussions/4652#discussioncomment-7418414 even the existing implementation of when we report the swap causes dropped frames; but that's a discussion for another PR.)

With the addition of locking in `draw` this should at least now be safe, but we still have the possibility for duplicate draws to be queued up, because the `canDraw -> draw` transition is not atomic. I don't really know how bad this is, on one hand the duplicate frame is already rendered so libmpv will skip the rendering part, but on the other hand CAOpenGLLayer starts delaying or dropping frames when you try to present a frame and it isn't "ready" yet (for some definition of ready that's somehow related to a combination of # of swaps in flight and vsync interval). Assuming that no noticeable regressions are observed with the slow-motion ticking clock test (60fps on a 60hz display, preferably filmed at 960fps slow motion), then I'd say this PR looks good.

Edit: I also wonder if adding cgllockcontext around draw would fix https://github.com/mpv-player/mpv/issues/11681

Super late edit: Actually I think CAOpenGLLayer only supports 1 buffer at a time (the IOSurface that the layer is currently displaying). You can draw into a second buffer and do the glFlush, but you can't let CAOpenGLLayer that IOSurface as the backing for the layer yet.

Here you can see 3 different test-cases of displaying R,G,B frames in that order. We start with R, wait a bit to synchronize on vsync, then display G and B. We want to probe the behavior of CAOpenGLLayer to see how it will behave.

In a traditional opengl implementation, you would be able to flush to buffer 1 for g, swapbuffer, then flush to buffer 2 for b. Depending on the implementation it might block after this flush for buffer 2, or it might allow you to queue up additional commands for buffer 1 again but it will _not_ allow you to flush it until buffer 1 is unused again (i.e. it would block _before_ the third flush).

1) flush G, flush B, wait for vsync after the `[CATransaction flush]` so that G is cleared.  This does not work, you see R then B: the G gets clobbered.

```objective-c

@interface MyNSLayer : NSOpenGLLayer
@end


@implementation MyNSLayer

- (BOOL)canDrawInOpenGLContext:(NSOpenGLContext *)context 
pixelFormat:(NSOpenGLPixelFormat *)pixelFormat 
forLayerTime:(CFTimeInterval)t 
displayTime:(const CVTimeStamp *)ts {
    return YES;
}

int i = 0;

_Atomic int pendingFlush = 0;

NSCondition *condition;

- (void)drawInOpenGLContext:(NSOpenGLContext *)context 
pixelFormat:(NSOpenGLPixelFormat *)pixelFormat 
forLayerTime:(CFTimeInterval)t 
displayTime:(const CVTimeStamp *)ts {
    if (i % 4 == 0) {
        glClearColor(1, 0, 0, 1);
    } else if (i % 4 == 1) {
        glClearColor(0, 1, 0, 1);
    } else if (i % 4 == 2) {
        glClearColor(0, 0, 1, 1);
    } else if (i % 4 == 3) {
        glClearColor(1,1, 1, 1);
    }
    glClear(GL_COLOR_BUFFER_BIT);
    i += 1;
       
    int p;
    glGetIntegerv(GL_DRAW_FRAMEBUFFER_BINDING, &p);
    // CAOpenGLLayer has ownership of FBO zero yet can return it to us,
    // so only utilize a newly received FBO ID if it is nonzero.
//  printf("%d\n", p);
    
    // Swap the buffers to display the rendered image


//  printf("FLUSHED\n");
    
    
    glFlush();


//  triggerWait();
}

@end

CVDisplayLinkRef displayLink;


TPPreciseTimer *tp;

CVReturn displayCallback(CVDisplayLinkRef displayLink,
    const CVTimeStamp *inNow, const CVTimeStamp *inOutputTime,
    CVOptionFlags flagsIn, CVOptionFlags *flagsOut,
    void *displayLinkContext)
{
    uint64_t scheduleTime = (((double)inOutputTime->hostTime) * 125.0 / 3.0 / 1e3);
    
    double vsyncInterval = inOutputTime->videoRefreshPeriod/inOutputTime->videoTimeScale;
    

//  [tp scheduleBlock:^{
//      usleep(800);
        [condition lock];
        pendingFlush = MAX(0, pendingFlush - 1);
        [condition signal];
        [condition unlock];
//  } atTime:inOutputTime->hostTime];
//      

    
    

    return kCVReturnSuccess;
}


int main(int argc, const char * argv[]) {
    // Set up the application and window
    @autoreleasepool {
        condition = [[NSCondition alloc] init];
        tp = [[TPPreciseTimer alloc] initWithSpinLock:0 spinLockSleepRatio: 0 highPrecision: YES];
        
        NSApplication *app = [NSApplication sharedApplication];
        NSWindow *window = [[NSWindow alloc] initWithContentRect:NSMakeRect(0, 0, 400, 400)
            styleMask:NSTitledWindowMask | NSClosableWindowMask | NSMiniaturizableWindowMask
            backing:NSBackingStoreBuffered
            defer:NO];
        
        // Enable layer-backing for the window
        [window setBackingType:NSBackingStoreBuffered];
        
        // Create an NSOpenGLPixelFormat object with double buffering enabled
        NSOpenGLPixelFormatAttribute attribs[] = {
            NSOpenGLPFADoubleBuffer,
            NSOpenGLPFAOpenGLProfile, NSOpenGLProfileVersion3_2Core,
            0
        };
        NSOpenGLPixelFormat *pixelFormat = [[NSOpenGLPixelFormat alloc] initWithAttributes:attribs];
        
        // Create an NSOpenGLContext object with the pixel format
        NSOpenGLContext *glContext = [[NSOpenGLContext alloc] initWithFormat:pixelFormat shareContext:nil];
        
        // Set the pixel format and context for the view
        window.contentView = [[NSView alloc] init];
        [window contentView].layer = [[MyNSLayer alloc] init];
        
        CGDirectDisplayID displayID = CGMainDisplayID();
        CVReturn error = CVDisplayLinkCreateWithCGDisplay(displayID, &displayLink);

        
        error = CVDisplayLinkSetOutputCallback(displayLink, displayCallback, NULL);
    
        CVDisplayLinkStart(displayLink);
        
        [window makeKeyAndOrderFront:nil];
        dispatch_after(dispatch_time(DISPATCH_TIME_NOW, 0.5 * NSEC_PER_SEC), dispatch_get_global_queue(QOS_CLASS_USER_INTERACTIVE, 0), ^{
                // start synchronized with vblank
                [condition lock];
                [condition wait];
                [condition unlock];
                
                for (int i = 0; i < 2; i++) {
                    [[window contentView].layer display];
                    [CATransaction flush];
                    
                    [condition lock];
                    pendingFlush += 1;
                    while (pendingFlush > 1) {
                        printf("Waiting, pending flush %d\n", pendingFlush);
                        [condition wait];
                    }
                    [condition unlock];
                }
            

        
//              usleep(8000);
                

//              sleep(2);

    
        });
        [app run];

        
    }
    return 0;
}
```


2) Here we flush for G, flush for B, then block after the flush for B before we exit the draw method. This works reliably, I always see the G frame.

```objective-c

@interface MyNSLayer : NSOpenGLLayer
@end

int waitQ = 0;
void triggerWait() {
    if (waitQ) {
        usleep(8000);
    }
}

@implementation MyNSLayer

- (BOOL)canDrawInOpenGLContext:(NSOpenGLContext *)context 
pixelFormat:(NSOpenGLPixelFormat *)pixelFormat 
forLayerTime:(CFTimeInterval)t 
displayTime:(const CVTimeStamp *)ts {
    return YES;
}

int i = 0;

_Atomic int pendingFlush = 0;

NSCondition *condition;

- (void)drawInOpenGLContext:(NSOpenGLContext *)context 
pixelFormat:(NSOpenGLPixelFormat *)pixelFormat 
forLayerTime:(CFTimeInterval)t 
displayTime:(const CVTimeStamp *)ts {
    if (i % 4 == 0) {
        glClearColor(1, 0, 0, 1);
    } else if (i % 4 == 1) {
        glClearColor(0, 1, 0, 1);
    } else if (i % 4 == 2) {
        glClearColor(0, 0, 1, 1);
    } else if (i % 4 == 3) {
        glClearColor(1,1, 1, 1);
    }
    glClear(GL_COLOR_BUFFER_BIT);
    i += 1;
       
    int p;
    glGetIntegerv(GL_DRAW_FRAMEBUFFER_BINDING, &p);
    // CAOpenGLLayer has ownership of FBO zero yet can return it to us,
    // so only utilize a newly received FBO ID if it is nonzero.
//  printf("%d\n", p);
    
    // Swap the buffers to display the rendered image


//  printf("FLUSHED\n");
    // III (3/10)
    // I (1/10)
    // iiiiii (6/10)
    // i (1/10)
    // iiiii (5/10)
    
    glFlush();
    [condition lock];
    pendingFlush += 1;
    while (pendingFlush > 1) {
        printf("Waiting, pending flush %d\n", pendingFlush);
        [condition wait];
    }
    [condition unlock];

//  triggerWait();
}

@end

CVDisplayLinkRef displayLink;


TPPreciseTimer *tp;

CVReturn displayCallback(CVDisplayLinkRef displayLink,
    const CVTimeStamp *inNow, const CVTimeStamp *inOutputTime,
    CVOptionFlags flagsIn, CVOptionFlags *flagsOut,
    void *displayLinkContext)
{
    uint64_t scheduleTime = (((double)inOutputTime->hostTime) * 125.0 / 3.0 / 1e3);
    
    double vsyncInterval = inOutputTime->videoRefreshPeriod/inOutputTime->videoTimeScale;
    

//  [tp scheduleBlock:^{
//      usleep(800);
        [condition lock];
        pendingFlush = MAX(0, pendingFlush - 1);
        [condition signal];
        [condition unlock];
//  } atTime:inOutputTime->hostTime];
//      

    
    

    return kCVReturnSuccess;
}


int main(int argc, const char * argv[]) {
    // Set up the application and window
    @autoreleasepool {
        condition = [[NSCondition alloc] init];
        tp = [[TPPreciseTimer alloc] initWithSpinLock:0 spinLockSleepRatio: 0 highPrecision: YES];
        
        NSApplication *app = [NSApplication sharedApplication];
        NSWindow *window = [[NSWindow alloc] initWithContentRect:NSMakeRect(0, 0, 400, 400)
            styleMask:NSTitledWindowMask | NSClosableWindowMask | NSMiniaturizableWindowMask
            backing:NSBackingStoreBuffered
            defer:NO];
        
        // Enable layer-backing for the window
        [window setBackingType:NSBackingStoreBuffered];
        
        // Create an NSOpenGLPixelFormat object with double buffering enabled
        NSOpenGLPixelFormatAttribute attribs[] = {
            NSOpenGLPFADoubleBuffer,
            NSOpenGLPFAOpenGLProfile, NSOpenGLProfileVersion3_2Core,
            0
        };
        NSOpenGLPixelFormat *pixelFormat = [[NSOpenGLPixelFormat alloc] initWithAttributes:attribs];
        
        // Create an NSOpenGLContext object with the pixel format
        NSOpenGLContext *glContext = [[NSOpenGLContext alloc] initWithFormat:pixelFormat shareContext:nil];
        
        // Set the pixel format and context for the view
        window.contentView = [[NSView alloc] init];
        [window contentView].layer = [[MyNSLayer alloc] init];
        
        CGDirectDisplayID displayID = CGMainDisplayID();
        CVReturn error = CVDisplayLinkCreateWithCGDisplay(displayID, &displayLink);

        
        error = CVDisplayLinkSetOutputCallback(displayLink, displayCallback, NULL);
    
        CVDisplayLinkStart(displayLink);
        
        [window makeKeyAndOrderFront:nil];
        dispatch_after(dispatch_time(DISPATCH_TIME_NOW, 0.5 * NSEC_PER_SEC), dispatch_get_global_queue(QOS_CLASS_USER_INTERACTIVE, 0), ^{
                // start synchronized with vblank
                [condition lock];
                [condition wait];
                [condition unlock];
                for (int i = 0; i < 2; i++) {
                    [[window contentView].layer display];
                    [CATransaction flush];
                }
            

        
//              usleep(8000);
                

//              sleep(2);

    
        });
        [app run];

        
    }
    return 0;
}
```

3) Here we flush for G, flush for B, then wait. But compared to the previous, we only start counting the flush after the commit. Since this leads to strictly longer wait times than the previous, it works as well. (To see why it leads to longer wait, say you glflush right before a vsync X. In the previous case the pending glflush would be decremented by that vsync X. But if we have a `[CATransaction  flush]` that finishes only after that vsync, the pending count would only be reset by the vsync X+1.

```objective-c

@interface MyNSLayer : NSOpenGLLayer
@end

int waitQ = 0;
void triggerWait() {
    if (waitQ) {
        usleep(8000);
    }
}

@implementation MyNSLayer

- (BOOL)canDrawInOpenGLContext:(NSOpenGLContext *)context 
pixelFormat:(NSOpenGLPixelFormat *)pixelFormat 
forLayerTime:(CFTimeInterval)t 
displayTime:(const CVTimeStamp *)ts {
    return YES;
}

int i = 0;

_Atomic int pendingFlush = 0;

NSCondition *condition;

- (void)drawInOpenGLContext:(NSOpenGLContext *)context 
pixelFormat:(NSOpenGLPixelFormat *)pixelFormat 
forLayerTime:(CFTimeInterval)t 
displayTime:(const CVTimeStamp *)ts {
    if (i % 4 == 0) {
        glClearColor(1, 0, 0, 1);
    } else if (i % 4 == 1) {
        glClearColor(0, 1, 0, 1);
    } else if (i % 4 == 2) {
        glClearColor(0, 0, 1, 1);
    } else if (i % 4 == 3) {
        glClearColor(1,1, 1, 1);
    }
    glClear(GL_COLOR_BUFFER_BIT);
    i += 1;
       
    int p;
    glGetIntegerv(GL_DRAW_FRAMEBUFFER_BINDING, &p);
    // CAOpenGLLayer has ownership of FBO zero yet can return it to us,
    // so only utilize a newly received FBO ID if it is nonzero.
//  printf("%d\n", p);
    
    // Swap the buffers to display the rendered image


//  printf("FLUSHED\n");
    // iiiii (5/10)
    // iiii (4/10)
    // iiiiii (6/10)
    
    
    glFlush();
    [condition lock];
    while (pendingFlush > 0) {
        printf("Waiting, pending flush %d\n", pendingFlush);
        [condition wait];
    }
    [condition unlock];

//  triggerWait();
}

@end

CVDisplayLinkRef displayLink;


TPPreciseTimer *tp;

CVReturn displayCallback(CVDisplayLinkRef displayLink,
    const CVTimeStamp *inNow, const CVTimeStamp *inOutputTime,
    CVOptionFlags flagsIn, CVOptionFlags *flagsOut,
    void *displayLinkContext)
{
    uint64_t scheduleTime = (((double)inOutputTime->hostTime) * 125.0 / 3.0 / 1e3);
    
    double vsyncInterval = inOutputTime->videoRefreshPeriod/inOutputTime->videoTimeScale;
    

//  [tp scheduleBlock:^{
//      usleep(800);
        [condition lock];
        pendingFlush = MAX(0, pendingFlush - 1);
        [condition signal];
        [condition unlock];
//  } atTime:inOutputTime->hostTime];
//      

    
    

    return kCVReturnSuccess;
}


int main(int argc, const char * argv[]) {
    // Set up the application and window
    @autoreleasepool {
        condition = [[NSCondition alloc] init];
        tp = [[TPPreciseTimer alloc] initWithSpinLock:0 spinLockSleepRatio: 0 highPrecision: YES];
        
        NSApplication *app = [NSApplication sharedApplication];
        NSWindow *window = [[NSWindow alloc] initWithContentRect:NSMakeRect(0, 0, 400, 400)
            styleMask:NSTitledWindowMask | NSClosableWindowMask | NSMiniaturizableWindowMask
            backing:NSBackingStoreBuffered
            defer:NO];
        
        // Enable layer-backing for the window
        [window setBackingType:NSBackingStoreBuffered];
        
        // Create an NSOpenGLPixelFormat object with double buffering enabled
        NSOpenGLPixelFormatAttribute attribs[] = {
            NSOpenGLPFADoubleBuffer,
            NSOpenGLPFAOpenGLProfile, NSOpenGLProfileVersion3_2Core,
            0
        };
        NSOpenGLPixelFormat *pixelFormat = [[NSOpenGLPixelFormat alloc] initWithAttributes:attribs];
        
        // Create an NSOpenGLContext object with the pixel format
        NSOpenGLContext *glContext = [[NSOpenGLContext alloc] initWithFormat:pixelFormat shareContext:nil];
        
        // Set the pixel format and context for the view
        window.contentView = [[NSView alloc] init];
        [window contentView].layer = [[MyNSLayer alloc] init];
        
        CGDirectDisplayID displayID = CGMainDisplayID();
        CVReturn error = CVDisplayLinkCreateWithCGDisplay(displayID, &displayLink);

        
        error = CVDisplayLinkSetOutputCallback(displayLink, displayCallback, NULL);
    
        CVDisplayLinkStart(displayLink);
        
        [window makeKeyAndOrderFront:nil];
        dispatch_after(dispatch_time(DISPATCH_TIME_NOW, 0.5 * NSEC_PER_SEC), dispatch_get_global_queue(QOS_CLASS_USER_INTERACTIVE, 0), ^{
                // start synchronized with vblank
                [condition lock];
                [condition wait];
                [condition unlock];
                
                for (int i = 0; i < 2; i++) {
                    [[window contentView].layer display];
                    [CATransaction flush];
                    [condition lock];
                    pendingFlush += 1;
                    [condition unlock];
                }
            

        
//              usleep(8000);
                

//              sleep(2);

    
        });
        [app run];

        
    }
    return 0;
}
```

4) If you take (1) and move the vsync wait before the commit, it works. So it seems commit really is the point at which things are actually "finalized" and any pending IOSurfaces replace the current backing surface.

svobs:
@krackers thanks for your input - this has been educational.

I somehow missed your mention of filming on an iPhone in slow-motion. I never noticed that my old iPhone 13 Pro Max can record at 240 fps! I tried testing with various screen grabbers but couldn't find one which was up to the task. Even Apple's Screenshot only recorded at ~54 fps, for some reason. But I'll see what I can get with my iPhone and will report those results.

Regarding timing and given what I've recently learned, I am fuzzy on why the `CVDisplayLink` is needed, or exactly what `mpv_render_context_report_swap` is accomplishing. If the `CVDisplayLink` is just a timer, why does it need to report a swap every vsync? Shouldn't the vsync be extremely predictable unless the active screen changes? Or if it's actually supposed to be reporting a buffer swap, wouldn't that rely on when `draw()` is called?

I'm also uneasy about the use of a `DispatchQueue` for rendering. For one, I'm not sure any of its QoS levels are really designed for such precise and/or consistent timing. But also because it's a queue which is suspended and resumed, more timing quirks are introduced once the queue has size > 1. The most egregious example of this is when toggling full screen, which takes 0.5s. During this time, the queue is suspended, and then resumed. This results in a noticeable freeze of the video during the toggle, and then a noticeable speed-up afterwards as the queue is drained. It does seem likely that the more or less correct timing is restored once it catches up. But is it really better to play catch-up than to just drop mistimed frames? And what's the point of reporting the timing via the DisplayLink if almost all rendering happens on a completely separate queue?

Could rendering instead be triggered directly by the `CVDisplayLink`? You mentioned that mpv is not really designed for a "pull" based approach. But I wonder if it would be possible to continue to allow libmpv to push an update to IINA and then have the DisplayLink respond to it via its next callback. It would probably need a lock to prevent data races, but that would be better than a work queue, in theory, right?

krackers:
This was educational for me too, I had some mistaken assumptions about CAOpenGLLayer that this helped clear up.


>Even Apple's Screenshot only recorded at ~54 fps,

I don't trust screenshot utilities to properly diagnose rendering issues, because I don't know if they go through the same compositing process. The most foolproof way is the slow-motion capture as I mentioned (I was lucky enough to have a 960fps slow-mo capable which makes frame drops really stand out, but 240fps should be perfectly fine too).

>why the CVDisplayLink is needed, or exactly what mpv_render_context_report_swap is accomplishing

Look at the source for vo_libmpv.c. If it's used, the context_report_swap is used to gate the render loop. It was necessary with old approach because unlike traditional OpenGL, the CAOpenGLLayer provides no backpresure when all backbuffers are tied up (as previously mentioned it is happy to give you 15 different backbuffers, and will silently drop things). So with the old approach (`isAsync = false`) where we used a separate render queue paired with libmpv initiated callback, we need to block on a vsync _somewhere_, which is what the CVDisplayLink reporting accomplished. With the new approach (`isAsync = true`) where we have CAOpenGLLayer call us, you are right, this should not be needed, because the vsync-gating is already done by the timing of the system callbacks. In the libmpv render api, calling `report_swap` is optional, if it's not done then the vo loop is immediately unblocked and the next frame prepared once `context_render` is called.

> If the CVDisplayLink is just a timer

Implementation-wise I believe it's a timer but with some machinery to periodically synchronize itself against the vblank to ensure the discrepancy is bounded.

>  Shouldn't the vsync be extremely predictable unless the active screen changes?

In theory yes, the vsync interval is fixed (e.g. 1/60 sec). However if we are working in theory, then our a-v sync would always be zero because our perfect theoretical clocks would never drift at all :). In the real-world things are analog, and so there can sometimes be a bit of skew, so things need to be periodically resynchronized. Beyond that, another utility of CVDisplayLink is that it actually runs as a fixed priority thread, so the kernel prioritizes it over most other processes.

>if it's actually supposed to be reporting a buffer swap, wouldn't that rely on when draw() is called

Unlike the opengl glx_oml_sync extension (and similar presentation feedback mechanisms) it doesn't report a buffer swap. In fact on modern macOS it doesn't even represent start or end of vertical blanking interval, as seen in that other post I linked you. The only official guarantee about when it is called is that you are called at roughly the same phase in the vsync cycle. I have some speculation about what the precise position corresponds to, but it's probably out of scope for this PR. 

>I'm also uneasy about the use of a DispatchQueue for rendering. 

I don't really have any such unease about it... it's fine because with the exception of subtitles we only just send out opengl commands, which isn't really that much work. Plus the user-interactive QoS is precisely what Apple wants you to render UI on.

> I'm not sure any of its QoS levels are really designed for such precise and/or consistent timing

The dirty secret is that CVDisplayLink is not really that precise either: it's precise down to about 1ms, but you can actually get 0.01ms precision: if you want consistent timing, you set a real-time thread constraint (which apple probably wishes they could remove, since you have to drop down to mach_task api in order to even use it). A CVDisplayLink thread uses fixed-priority scheduling at priority 54. A dispatch queue with USER_INTERACTIVE priority seems to use default timeshare scheduling with a priority of 46. You'd think that the CVDisplayLink is obviously better (it's got higher priority AND fixed scheduling!) but surprisingly I've seen that the CVDisplayLink can perform worse in terms of latency (not by much, maybe 0.3 ms difference). And the fixed-priority scheduling comes with some constraints, if you look at xnu source it only supports round-robin fixed scheduling, not fifo.

If you look at what apple calls your CAOpenGLLayer callback on... it's the main thread, which is precisely just user-interactive QoS. So I really suspect that Apple basically optimizes for this use-case in terms of presentation latency.

>But also because it's a queue which is suspended and resumed, more timing quirks are introduced once the queue has size > 1

The queue size should almost always be bounded at <= 2. Also I assume that GCD is smart enough to not completely tear down the thread, instead it just waits for new work, so there's no real cost here (gcd is open source so this could be verified if really needed, but it's the obvious thing to do.)


>The most egregious example of this is when toggling full screen, which takes 0.5s

Why is the queue suspension needed here, mpv's native mac backend does not need to do this. Also are you sure that suspending a queue is responsible for this delay, I'd think that the implementation should make this basicaly "free". Or is the backing thread actually torn down during a suspend (in which case you could replace the suspend mechanism by a mutex on your side which you check before running a new queue task).

>And what's the point of reporting the timing via the DisplayLink if almost all rendering happens on a completely separate queue?

Yes you're absolutely right, this is precisely the point I made in https://github.com/iina/iina/discussions/4652#discussioncomment-7418414. We should actually be waiting for the next vsync after the `[CATransaction flush]` and not just reporting flips independent of the rendering thread. I tested this, and it fixes the silent dropped frames issue. But actually even waiting until the next vsync after the flush is not quite right, since the flush merely submits things, it does not wait for things to be rendered (`glFinish` would do that, but using it seems to absolutely kill performance in a manner that `glFlush` + wait does not, for reasons I don't really understand). So the sad reality of a CAOpenGLLayer is that there is actually no mechanism to reliably even detect if a frame was dropped in the first place, because unlike a traditional double-buffered opengl context, you will _never_ block if the backbuffer was still busy. The best we can do is make the optimistic guess that it will finish by one vsync, and just accept that if that is violated the system will silently drop a frame for us.

>Could rendering instead be triggered directly by the CVDisplayLink?

Sure. In fact this is what mpv's mac backend used to do before it was switched to a separate render thread. The reason for that switch doesn't make sense to me (something about long renders blocking the cvdisplaylink, but that's actually precisely what you want!). I don't really see how this is much different than your other PR of driving the rendering completely off of the CAOpenGLLayer async callbacks though. In both cases we completely disregard the mpv callbacks, and rely on system callbacks to run things. I'd think that the CAOpenGLLayer callbacks are probably better because it potentially has knowledge of when exactly a buffer is freed up.

Edit: Ah I just read the comments on the other pr, about how rendering on the main thread causes UI lag when resizing. Then yeah in that case rendering triggered via a CVDisplayLink would solve that. But let's take a step back here, the main problem we want to solve is just the wobble when resizing right, which I believe is caused by our inability to detect such resize events and trigger a redraw ourselves. The solution as seen in this PR of setting async = true solves this, by permitting system callbacks in addition to our own callbacks to trigger drawing. Moving the rendering off of our own queue onto a CVDisplayLink would not really change the fact that we still need `async = true`, because only the system seems to knows when we are in the middle of this pinch-zoom resizing. (Maybe there's an undocumented appkit notification we could use, in which case that would actually be the way to go, I'm still not a fan of this async stuff if there's a way to avoid it. But this would require reverse engineering.).

svobs:
Went back and reread some of your old posts. I think I'm starting to get the gist of this, maybe...

> The queue size should almost always be bounded at <= 2

That makes a lot of sense. I've been wondering how best to handle the case when playback becomes CPU constrained. But if I understand right, the current solution will empty the queue as quickly as it can call `display()` for each, which ends up overflowing into more of the 16 buffers, which is basically wasted work because flushing more than 1 buffer per vsync will result in dropped frames anyway.

> Why is the queue suspension needed here, mpv's native mac backend does not need to do this. Also are you sure that suspending a queue is responsible for this delay, I'd think that the implementation should make this basicaly "free". Or is the backing thread actually torn down during a suspend (in which case you could replace the suspend mechanism by a mutex on your side which you check before running a new queue task).

This is an area I've been working on improving. The code in this area is pretty old and has some rough edges. You're right, the built-in or "native" full screen transition can be made to work without suspending playback. I've already gotten it working on [my fork](https://github.com/svobs/iina). But my fix there is part of a big refactoring of the whole player window layout system, so it's not trivial to merge back. Moreover, I'm still ironing out some wrinkles, e.g. getting rid of an unpleasant wobble when the window zooms ;)

But IINA also has an option to use its own custom full screen mode ("legacy full screen") which works by explicitly hiding the Dock & menu bar, removing `.titled` from `window.styleMask`, and resizing it to fit the whole screen (plus a few other things). It needs to make all of these changes as close to at the same time as possible to make for a pleasant user experience. But this has been a moving target. I started this on Ventura, and I had a lot of trouble hiding the Dock and menu bar. Its animation caused a noticeable hiccup for nearly a quarter-second, couldn't be sped up, and caused other animation to hang. Fortunately, after the update to Sonoma, it was no longer an issue.

However, adding or removing `NSWindow.StyleMask.titled` still presents some challenges. It's almost as if the whole window is getting torn down and reconstructed. It seems to be faster at doing this in Sonoma, but if done while zooming to full screen it still causes the video to stutter, so it still seems more pleasant to just freeze it. (I also tried changing the style mask *before* the zoom. That avoided the stutter, but the upfront delay made it feel less responsive. And changing after the zoom? An unwanted title bar during the animation, and briefly, rounded corners in full screen).

> We should actually be waiting for the next vsync after the [CATransaction flush] and not just reporting flips independent of the rendering thread. I tested this, and it fixes the silent dropped frames issue.

You buried this good news pretty deep in your post! At least it sounds like a demonstrable improvement(?)

> But actually even waiting until the next vsync after the flush is not quite right, since the flush merely submits things, it does not wait for things to be rendered [...] So the sad reality of a CAOpenGLLayer is that there is actually no mechanism to reliably even detect if a frame was dropped in the first place [...] The best we can do is make the optimistic guess that it will finish by one vsync, and just accept that if that is violated the system will silently drop a frame for us.

Sorry for yet another newbie question, but is there really a lot of benefit of knowing the frame was dropped, besides collecting statistics? Would it drive any kind of conditional handling? I mean, if it can't render a frame in time, you still need to carry on, right?

> Moving the rendering off of our own queue onto a CVDisplayLink would not really change the fact that we still need async = true, because only the system seems to knows when we are in the middle of this pinch-zoom resizing.

That's what I'm guessing. It would be great if the CAOpenGLLayer could time its redraws from CVDisplayLink's clock, instead of highjacking it. Although we still don't know for sure if that is really the case. I'm going to work on further testing.

krackers:
>will result in dropped frames anyway

Yup you got it. You don't need to worry about the CPU constrained case, the CPU constrained case is actually "good" in the sense that the GPU buffers will be sitting idle, and when we are CPU constrained mpv can detect it and adjust a/v sync accordingly. It's the GPU constrained case (where we don't have a gpu backbuffer ready in time) that bites us. And the sad state of CAOpenGLLayer is that you have no way of knowing if a backbuffer is actually free or not, because unlike traditional opengl contexts, the `glFlush()` will never block. The best you can do is just make the assumption to wait until the next vsync after the `glFlush()`, which is at least better than what the code currently does which just waits for the next vsync after render() was called, instead of waiting for the next vsync after `glFlush()` was called.

>You buried this good news pretty deep in your post

I didn't think it was relevant to this PR, which was focused mainly on benefits of sync vs. async approach! :) But yes I am happy to provide more info if you are interested in porting the same approach to iina, we can perhaps continue discussion on that github discussion page to avoid mixing different PRs. Yes it is a noticeable improvement, on my 960fps slow motion test, it completely eliminates the dropped frames.

>Is there really a lot of benefit of knowing the frame was dropped

In a "proper" conforming opengl implementation what would happen is that if some frame took too long to render on the gpu, the `glSwap()` would block until the backbuffer is free. So we would have the behavior that every frame would be displayed, but we would have a "delayed frame", as mpv calls it in the stats panel. This introduces an a-v sync, which mpv must now handle. If the delay is small enough, in display-resample modes mpv will speed up the audio. If the delay is big enough, then mpv has no choice but to drop some other video frame. With CAOpenGLLayer, it's like the system always decides to drop the video frame for us if it is too late. The result is the same as if mpv did not have the ability to speed up video to compensate, it always decided to drop the next frame to maintain a-v sync. So the "loss" by not properly reporting dropped frames is that we don't get the full benefit of display-resample mode.. If the a-v sync introduced by delayed frame was not too big (maybe it took 1700us instead of 1600), with a "proper" opengl implementation mpv would not need to drop frames at all and could just speed up the audio slightly. 

> CAOpenGLLayer could time its redraws from CVDisplayLink's clock

I just checked and this is not the case, the CAOpenGLLayer callbacks are not tied to vsync in any way. They are roughly the same distance apart (although the accuracy of this is pretty terrible, it fluctuates by as much as 160us). But phase is not constant across runs, unlike a CVDisplayLink. Also obviously it doesn't run on the CVDisplayLink thread. So I suspect it's just using a plain old NSTimer or something. Man the more I investigate CAOpenGLLayer the more I hate it, it's a shame that it's basically the only usable option for opengl on mac now.

Not sure what you mean by the CAOpenGLLayer "hijacking" the redraw times though. We could of course easily have the calls to render happen from the CVDisplayLink thread instead of our current dispatch queue, but to me this is going in the wrong direction: it doesn't really bring any benefits, it prevents us from making use of the (limited)-double buffering that caopengllayer seems to provide, and means that we can never force a render ourselves if we ever had a need to (there's no way to dispatch something onto the CVDisplayLink thread). I guess one benefit is that it might theoretically provide another way to solve the "silent dropped frame" issue since the CVDisplayLink callbacks are timed in a way that the OS tries to guarantee a backbuffer buffer is free at that point (buffer is freed at start of vsync, the callback happens about 60% into the interval), but there's dead time there the CPU could use to send commands for next frame (or render subtitles, or handle VOCTRL). I feel like the "classic" opengl approach gives you better throughput at the expense of 1 frame of latency (which doesn't matter for video player since we can compensate) while the CVDisplayLink approach is intended to try to minimize latency. (But even if the goal is minimzing latency the CVDisplayLink approach of just always calling you 60% into the interval is a bit dumb, it makes assumptions about constant rendering times. Smarter solutions like [NVidia Reflex](https://github.com/klasbo/GamePerfTesting/blob/master/text/02-reflex.md) use previous rendering times to dynamically determine the best time to unblock in the render loop so that the frame is computed as close as possible to the next vsync deadline as possible)

>because only the system seems to knows when we are in the middle of this pinch-zoom resizing.
>that's what I'm guessing

Tbh I still suspect there's some private notification that gets sent out for these sorts of resize events that we could listen to. After all the window server must ultimately communicate this to the app somehow. It used to be the case pre-Monterey that you could listen for all notifications sent to the app via [this technique](https://stackoverflow.com/questions/7954905/how-can-i-listen-for-all-notifications-sent-to-the-ios-nsnotificationcenters-de), if you happen to have a pre-monterey device around, maybe it's worth doing this. But I don't know how you want to weigh the tradeoff of finding and discovering the private api versus the approach of setting the async layer.

Edit: Seems there's private `NSResizeMoveHelper` and `NSMoveHelper` which do this stuff. Seems like it internally does some stuff with a CVDisplayLink callback and repeatedly calls `setFrame` (in `animateStep:`), but my reversing skills are not very good.

svobs:
Closing this PR. It wasn't going anywhere, & the approach in #5013 looks to get the job done with a lot less code.

PR #4672, opened 2023-10-25
fix: Failed to download subtitles from assrt.

### LINKED ISSUE #4671: Failed to download subtitles from assrt due to rate limit.
**System and IINA version:**

- macOS 14.0
- IINA 1.3.3 Build 138

**Steps to reproduce:**

1. Use assrt as source of subtitles
2. Choose a subtitle resource with more than 2 subtitles (which means there are more than 2 subtitles in the `filelist`, usually this happens when the resource contains subtitles for the whole season, rather than one episode)
3. It should pop up a hint to display the download task is failed.

**Cause analysis:**

After validating the API and server ability of assrt, this issue is in the IINA side. When use proxy tools such as Proxyman, it's easy to guess the assrt server limits the concurrency requests.

Quite often

jjaychen1e:
- [x] This implements/fixes issue #4671 .

---

**Description:**

Here I restrict the concurrency as at maximum 2 download tasks at the same time for assrt. In my test, if I change 2 to 3, it will simply return code `503`, but 2 works well.

### DIFF
--- a/iina/AssrtSubtitle.swift
+++ b/iina/AssrtSubtitle.swift
@@ -48,8 +48,12 @@ class Assrt {
     override func download() -> Promise<[URL]> {
       if let fileList = fileList {
         // download from file list
-        return when(fulfilled: fileList.map { file -> Promise<URL> in
-          Promise { resolver in
+        var fileGenerator = fileList.makeIterator()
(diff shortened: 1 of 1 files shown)


lhc70000:
Merged via 79cd70c.

PR #4676, opened 2023-11-04
Remove use of mpv update callback + use videoLayer.isAsynchronous for rendering (alternate fix for #4667) 

svobs:
- [x] This implements/fixes issue #4667.

---

**Description:**
This carries PR #4668 a step further and removes the use of `mpv_render_context_set_update_callback` and `mpvGLQueue`. When `videoLayer.isAsynchronous == true` the DisplayLink will call `canDraw()` as needed, which makes the callback & queuing redundant.

### DIFF
--- a/iina/ViewLayer.swift
+++ b/iina/ViewLayer.swift
@@ -26,7 +26,7 @@ class ViewLayer: CAOpenGLLayer {
     super.init()
 
     isOpaque = true
-    isAsynchronous = false
+    isAsynchronous = true
 
     autoresizingMask = [.layerWidthSizable, .layerHeightSizable]
   }
--- a/iina/ViewLayer.swift
+++ b/iina/ViewLayer.swift
@@ -14,12 +14,8 @@ class ViewLayer: CAOpenGLLayer {
 
   weak var videoView: VideoView!
 
-  let mpvGLQueue = DispatchQueue(label: "com.colliderli.iina.mpvgl", qos: .userInteractive)
-  @Atomic var blocked = false
-
   private var fbo: GLint = 1
 
-  private var needsMPVRender = false
   private var forceRender = false
 
   override init() {
@@ -94,8 +90,8 @@ class ViewLayer: CAOpenGLLayer {
 
   override func draw(inCGLContext ctx: CGLContextObj, pixelFormat pf: CGLPixelFormatObj, forLayerTime t: CFTimeInterval, displayTime ts: UnsafePointer<CVTimeStamp>?) {
     let mpv = videoView.player.mpv!
-    needsMPVRender = false
 
+    CGLLockContext(ctx)
     glClear(GLbitfield(GL_COLOR_BUFFER_BIT))
 
     var i: GLint = 0
@@ -128,17 +124,20 @@ class ViewLayer: CAOpenGLLayer {
       }
     }
     glFlush()
+    CGLUnlockContext(ctx)
   }
 
   func suspend() {
-    blocked = true
-    mpvGLQueue.suspend()
+    DispatchQueue.main.async { [self] in
+      isAsynchronous = false
+    }
   }
 
   func resume() {
-    blocked = false
-    draw(forced: true)
-    mpvGLQueue.resume()
+    DispatchQueue.main.async { [self] in
+      isAsynchronous = true
+      draw(forced: true)
+    }
   }
 
   func draw(forced: Bool = false) {
@@ -146,37 +145,12 @@ class ViewLayer: CAOpenGLLayer {
       // The properties forceRender and needsMPVRender are always accessed while holding isUninited's
       // lock. This avoids the need for separate locks to avoid data races with these flags. No need
       // to check isUninited at this point.
-      needsMPVRender = true
       if forced { forceRender = true }
     }
 
     // Must not call display while holding isUninited's lock as that method will attempt to acquire
     // the lock and our locks do not support recursion.
     display()
-
-    videoView.$isUninited.withLock() { isUninited in
-      guard !isUninited else { return }
-      if forced {
-        forceRender = false
-        return
-      }
-      guard needsMPVRender else { return }
-
-      videoView.player.mpv.lockAndSetOpenGLContext()
-      defer { videoView.player.mpv.unlockOpenGLContext() }
-      // draw(inCGLContext:) is not called, needs a skip render
-      if let renderContext = videoView.player.mpv.mpvRenderContext {
-        var skip: CInt = 1
-        withUnsafeMutablePointer(to: &skip) { skip in
-          var params: [mpv_render_param] = [
-            mpv_render_param(type: MPV_RENDER_PARAM_SKIP_RENDERING, data: .init(skip)),
-            mpv_render_param()
-          ]
-          mpv_render_context_render(renderContext, &params);
-        }
-      }
-      needsMPVRender = false
-    }
   }
 
   override func display() {
--- a/iina/ViewLayer.swift
+++ b/iina/ViewLayer.swift
@@ -26,7 +26,6 @@ class ViewLayer: CAOpenGLLayer {
     super.init()
 
     isOpaque = true
-    isAsynchronous = true
 
     autoresizingMask = [.layerWidthSizable, .layerHeightSizable]
   }
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -518,6 +518,7 @@ not applying FFmpeg 9599 workaround
 
   func shouldRenderUpdateFrame() -> Bool {
     guard let mpvRenderContext = mpvRenderContext else { return false }
+    guard !player.isStopping && !player.isShuttingDown else { return false }
     let flags: UInt64 = mpv_render_context_update(mpvRenderContext)
     return flags & UInt64(MPV_RENDER_UPDATE_FRAME.rawValue) > 0
   }
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -478,7 +478,6 @@ not applying FFmpeg 9599 workaround
       ]
       mpv_render_context_create(&mpvRenderContext, mpv, &params)
       openGLContext = CGLGetCurrentContext()
-      mpv_render_context_set_update_callback(mpvRenderContext!, mpvUpdateCallback, mutableRawPointerOf(obj: player.mainWindow.videoView.videoLayer))
     }
   }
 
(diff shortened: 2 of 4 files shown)


krackers:
To be precise here it's not `DisplayLink` that's calling isDraw but rather the the system's CAOpenGL timer right? (I don't actually know if the two are related in any way. My guess is that they are actually separate things since the CAOpenglLayer callback is on the main thread, and initiated by the runloop, whereas a displaylink runs on a separate fixed priority thread). In IINA it seems we only use displaylink to report the vsyncs (and that too, it's actually off by half a vsync, see [this](https://thume.ca/2017/12/09/cvdisplaylink-doesnt-link-to-your-display/)).

Several things to consider with this approach (I'm just thinking out loud here, don't know all the answers to below):
* Now the system is fully in control of when things are drawn, we no longer use the callback to initiate drawing. This means that there can be a delay between mpv wants a frame to be drawn and when it is actually drawn. Do we have an upper bound on this delay? Obviously it can be no worse than 1/content_fps, but that's actually quite bad so I suspect in practice it's smaller. Can we get a quantitive measure, both in audio-timed and display-resample modes?

* Related to the above, when you control drawing via a cvdisplaylink (or the caopengllayer callback), you seem to lose one of the benefits of double buffering: resilience to temporary fps drops. With double buffering, because you always render one frame ahead, you have some "slack" that allows frames to occasionally exceed the 1/fps deadline.

Visually, let r1 be rendering, s1 be the glFlush(), and p1 be the final present on vsync. With traditional double-buffering as shown below, r2 can begin immediately after s1, even without waiting for the next vsync. However r3 must wait for a buffer to be freed up, so it can happen only after p1. Note that because we render one frame ahead of the present, even in the case where something takes too long to render, like r5 -> s5, it is no problem, we don't have any dropped frames.

```
|r1-----s1--r2-s2-|p1|r3--------s3---|p2|r4-----s4----|p3|r5----------|p4|------s5-r6-----|p5|
```

When you use a displaylink approach now the rendering is constrained to start only after a vsync, which means you effectively don't have any double buffering happening; we're not "one frame ahead" because the rendering is gated to happen after the vsync, so we're not making full use of the second buffer (this is technically good for low-latency input to present situations, but bad for video where latency is not an issue and we want throughput).

* Another thing: what is the interaction between between the timing of the CAOpenGLLayer callbacks, and CVDisplayLink callbacks. vo_libmpv's render loop is blocked on the reported swap which we call from the cvdisplaylink callback. However CAOpenGLLayer effectively already synchronizes things against a vsync for us, so I wonder whether the two might interact badly, causing unnecessary delays. Do they both trigger at the same time?

* I'm curious whether this approach fixes the dropped frame issue mentioned in https://github.com/iina/iina/discussions/4652.

Overall I think best way to see is to experimentally test it. But I do feel that mpv (and libmpv) are not really designed for a "pull" based approach (since mpv wants to be in control of the timings, as opposed to a true pull-based model where you simply tell mpv what timestamp you want a frame for), and this moves further in that direction.

iina/ViewLayer.swift
```diff
@@ -95,8 +90,8 @@ class ViewLayer: CAOpenGLLayer {
 
   override func draw(inCGLContext ctx: CGLContextObj, pixelFormat pf: CGLPixelFormatObj, forLayerTime t: CFTimeInterval, displayTime ts: UnsafePointer<CVTimeStamp>?) {
     let mpv = videoView.player.mpv!
-    needsMPVRender = false
 
+    CGLLockContext(ctx)
```
krackers, line 94: with this approach the cgllockcontext should no longer be needed, since all drawing is done on the main thread

svobs:
This is no good. I found that this solution creates unacceptable timing problems.

The video seems to play fine when nothing else is happening, but when the flow is interrupted, things sort of...snag.

When I try to drag the window around, there's a delay of almost a second from mouseDown and when the window actually responds. This was unexpected. But after some thought, maybe what's happening is that now that the render calculations are happening on the main thread, it's got very little bandwidth left for other things, and the window manager prioritizes drawing over user input events such as clicks.

And there seem to be various situations where the video falls behind in frames and then speeds up for a while to try to catch up. At least I'm assuming. It's been hard to spot and pin down exactly when this occurs. I think it may happen when the window is fully occluded for a while (maybe MacOS stops rendering the window during this time?) and then is brought back to the front.

Also, I found that removing the `CVDisplayLink` at this point produces functionally identical results & symptoms (to my eyes anyway). I pushed another commit with it commented out, for reference.

I have more to add but I'd like to go back to #4668's comments for that, since I don't think this PR will go any further.

iina/ViewLayer.swift
low-batt: The calls to `CGLLockContext` and `CGLUnlockContext` **should be** removed from the `draw` method, but not for the reason cited.

OpenGL context must be locked even if the main thread is used exclusively, because the application does not control this. Quartz performs some operations asynchronously that access the context using a separate thread. We noticed this in https://github.com/mpv-player/mpv/issues/11681. See [this comment](https://github.com/mpv-player/mpv/issues/11681#issuecomment-1586568804).

**However** in the case of the `draw` method  the locking should be removed because:

```text
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
  * frame #0: 0x00000001fb4334f8 OpenGL`CGLLockContext
    frame #1: 0x000000019de92998 QuartzCore`CAOpenGLLayerDraw(CAOpenGLLayer*, double, CVTimeStamp const*, unsigned int) + 548
    frame #2: 0x000000019de925a4 QuartzCore`-[CAOpenGLLayer _display] + 520
    frame #3: 0x00000001044b7c5c IINA`closure #1 in ViewLayer.display(isUninited=false, self=0x0000600002769400) at ViewLayer.swift:160:13
```

That is `CAOpenGLLayer` locking the context before calling `ViewLayer.draw` as discussed in the documentation for [CAOpenGLLayer.draw](https://developer.apple.com/documentation/quartzcore/caopengllayer/1522316-draw). Wait a minute. The documentation does not say a thing about this. Nothing. Nada. Zip. Zilch. Not one iota. Could it be somebody got nailed by this causing a deadlock and thus previously has had to debug `CAOpenGLLayer` and try and figure out what it was doing…

As `OpenGL` uses a reentrant lock the locking in `draw` works, but adds needless overhead and should be removed.

low-batt:
I don't think `mpv` is happy with this change:
```text
[  16.277][v][vo/libmpv] mpv_render_context_render() not being called or stuck.
[  16.501][v][vo/libmpv] mpv_render_context_render() not being called or stuck.
[  16.718][v][vo/libmpv] mpv_render_context_render() not being called or stuck.
[  16.941][v][vo/libmpv] mpv_render_context_render() not being called or stuck.
[  17.203][v][vo/libmpv] mpv_render_context_render() not being called or stuck.
[  18.074][v][vo/libmpv] mpv_render_context_render() not being called or stuck.
[  18.327][v][vo/libmpv] mpv_render_context_render() not being called or stuck.
[  18.551][v][vo/libmpv] mpv_render_context_render() not being called or stuck.
```

And when I play a bunch of videos at the same time and then quit IINA it triggers a crash:
```text
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libmpv.2.dylib                	       0x107961dfc mpv_render_context_render + 132
1   IINA                          	       0x1052df274 closure #1 in closure #1 in ViewLayer.draw(inCGLContext:pixelFormat:forLayerTime:displayTime:) + 1020 (ViewLayer.swift:118)
2   IINA                          	       0x1052e13ac partial apply for closure #1 in closure #1 in ViewLayer.draw(inCGLContext:pixelFormat:forLayerTime:displayTime:) + 32
3   libswiftCore.dylib            	       0x1a5582ff4 withUnsafeMutablePointer<A, B>(to:_:) + 28
4   IINA                          	       0x1052dee24 closure #1 in ViewLayer.draw(inCGLContext:pixelFormat:forLayerTime:displayTime:) + 1448 (ViewLayer.swift:112)
5   IINA                          	       0x1052e0f84 partial apply for closure #1 in ViewLayer.draw(inCGLContext:pixelFormat:forLayerTime:displayTime:) + 36
```

This particular area of IINA is hazardous to modify because history has shown that thread related problems may not show up during development.

iina/ViewLayer.swift
krackers: Nice finding, I'll add this to my mental list of "undocumented things apple forgot to tell you". This might also explain why in https://github.com/mpv-player/mpv/issues/11681 the issue is only seen during shutdown; almost all other calls first go through the `[CAOpenGLLayer display]`, but during shutdown the deinit is called without the lock on current context.

Edit: That being said, apple's (somewhat outdated but much more comprehensive) docs in https://developer.apple.com/library/archive/documentation/GraphicsImaging/Conceptual/OpenGL-MacProgGuide/opengl_threading/opengl_threading.html seem to imply that a `CGLLockContext` is only needed if you start rendering on things other than the main thread, and none of the sample code I've seen uses CGLLockContext for purely singled threaded drawing. But it's possible that modern macOS changed things (those docs were likely written before `CoreAnimation` existed, and now a lock is needed even if you're on the main thread as well (or at the very least, needed if you use opengl via coreanimation).

krackers:
>I think it may happen when the window is fully occluded for a while 

Yeah likely due to app nap introduced in 10.9, osx has behavior to lower priority of the main thread if the window is fully occluded and reduce timer precision.

>mpv_render_context_render() not being called or stuck.

Yeah this is part of what I worried about. The model that libmpv has is that when it has a frame ready for you to render, it expects render to be called basically asap, and then it takes care of internally waiting as needed before unblocking present. With this approach, if you play a low fps video (say 12fps) it may take 1/12 seconds between when libmpv has a frame ready and when render is actually called, which probably messes with the a/v compensation.

iina/ViewLayer.swift
low-batt: Yes, I agree, the locking in `CAOpenGLLayer` may explain why the `mpv` crash only occurs during shutdown.

The old archived Apple documentation is in may ways significantly better than the current documentation. But as it is way out of date so some details may have changed. And there are definitely problems even in the old documentation. In [this](https://developer.apple.com/library/archive/samplecode/CALayerEssentials/Listings/ExampleCAOpenGLLayer_m.html#//apple_ref/doc/uid/DTS40008029-ExampleCAOpenGLLayer_m-DontLinkElementID_8) `CALayerEssentials` example the comment before the method `drawInCGLContext` says:
> 1. Draw your OpenGL content! (the current context has already been set)

And the method starts with:
> // Set the current context to the one given to us.
> CGLSetCurrentContext(glContext);

This shows one of the reasons why the software industry keeps trying to come up with "easier" ways to support threading. When it comes to threading poor documentation of "black box" implementations can result in severe problems that are hard to reproduce.

svobs:
Closing this PR. It was not the best approach, and PR #5013 does the job.

Issue #4057, opened 2022-11-09
The Find Subtitle Result Window was too Narrow to see the whole name.


owenzhao:
**What you want IINA to do:**

The result window of find subtitle should be much much more wider.

**What IINA does currently:**

Too narrow to see the full name, we need to see the full name to get the subtitle that is mostly match my video.

**Why you think this should be added:**

**Examples of other projects that have something similar:**

Iiii-I-I-I:
To complete the **Why you think this should be added:** field, here are some situations where I've needed to read the full filename:

- to see the year of release, because there are multiple remakes with the same title
- to see the format of the movie (streaming, Blu-ray, restoration, etc.), because the timings may be off
- to see the type of subtitle (forced, CC, etc.), because I'm not hearing impaired and do not need SDH subtitles

This type of information is usually listed at the end, where it gets cut off by IINA's search window. Making sure I'm downloading the correct file is even more important for OpenSubtitles now that they've added a daily limit.

Iiii-I-I-I:
Some possible solutions:

* show the full text when hovering over the title, the same way title text appears when hovering over an `<a>` tag on the web. Could be less work than messing with the UI?
* add more information to the metadata row (I don't know if the OpenSubtitles API gives you that kind of detail). Here is a mockup, with the current on the left and proposed on the right:

![iina](https://github.com/iina/iina/assets/647835/c5ae3581-8f7b-4336-8be0-03244816fd7b)

svobs:
> show the full text when hovering over the title, the same way title text appears when hovering over an <a> tag on the web. Could be less work than messing with the UI?

This is trivial (a checkbox in the UI). I may post a PR for it.

However, you can try another workaround, which exists because the table width is set by the width of its title. Try increasing the OSD text size in IINA's Settings:
<img width="718" alt="SCR-20240216-unyk" src="https://github.com/iina/iina/assets/2213815/e161fc99-971f-4a28-a4a9-1e14abf492a6">

Iiii-I-I-I:
Thanks all for addressing this so quickly! I look forward to the title text + new plug-ins in a future update.

low-batt:
This was changed for IINA 1.3.5.

Issue #3530, opened 2021-10-31
 Can't get the subtitles from shooter.cn


yupik123:
- macOS 
- IINA <!-- 1.1.2 and 1.2.0 -
<img width="1316" alt="iShot2021-11-01 00 15 11" src="https://user-images.githubusercontent.com/93484018/139592713-ec866fb2-e9df-44bc-b4a3-149c677b45a7.png">
->

Stephen-ps:
Same problem

PR #4678, opened 2023-11-05
Close download from Open Subtitles using REST API, #3920

### LINKED ISSUE #3920: Download from Open Subtitles using their REST API
**What you want IINA to do:**
Download from Open Subtitles using their new [REST API](https://opensubtitles.stoplight.io/docs/opensubtitles-api/e3750fd63a100-getting-started).

**What IINA does currently:**
Downloads from Open Subtitles using their original [XMLRPC API](https://trac.opensubtitles.org/projects/opensubtitles/wiki/XMLRPC).

Because of this post by Open Subtitles: [OpenSubtitles.org API deprecation and end of life](https://forum.opensubtitles.org/viewtopic.php?f=11&t=17930&sid=d6897753acd72f04f9cf733afe736831) which says:
> After 16 years of service, the time has come to turn off our original API service, to focus our time and energy on the new REST API.

And:
> 2023-12-31 end of life, API endpoint will be turned off

IINA has no choice but to switch to the new REST API before that deadlline.

N/A

low-batt:
This commit will:
- Add a new `OpenSubClient` class
- Change the `OpenSub` class to use the new client
- Change the `OpenSub` default language from `eng` to `en`
- Change name of provider from `opensubtitles.org` to `opensubtitles.com`
- Add a new `loggedIn` property to `OnlineSubtitle`
- Add a new `logout` method to `OnlineSubtitle`
- Add a new `iinaLogoutCompleted` notification to `AppData`
- Change `AppDelegate` to log out of the online subtitle provider during termination

This is a temporary stopgap solution that updates the builtin legacy code in order to meet the end of 2023 deadline imposed by Open Subtitles for switching to the new REST API. The plan is still to replace all built-in code for downloading from online subtitle providers with plug-in implementations.

- [x] This implements/fixes issue #3920.

---

**Description:**

### DIFF
--- a/iina.xcodeproj/project.pbxproj
+++ b/iina.xcodeproj/project.pbxproj
@@ -20,6 +20,7 @@
 		515B5E5A2A579903001FCD49 /* iina-plugin in Copy Executables */ = {isa = PBXBuildFile; fileRef = E38558872A484A2D0083772D /* iina-plugin */; settings = {ATTRIBUTES = (CodeSignOnCopy, ); }; };
 		5196819A29EC963F00B05D55 /* CoreDisplay.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 5196819929EC963F00B05D55 /* CoreDisplay.framework */; };
 		519872FF26879B9B00F84BCC /* AccessibilityPreferences.swift in Sources */ = {isa = PBXBuildFile; fileRef = 519872FE26879B9B00F84BCC /* AccessibilityPreferences.swift */; };
+		51AC1CAB2A9FBF3700DF7079 /* OpenSubClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = 51AC1CAA2A9FBF3700DF7079 /* OpenSubClient.swift */; };
 		51C1BA3A291CA76700C1208A /* InfoDictionary.swift in Sources */ = {isa = PBXBuildFile; fileRef = 51C1BA39291CA76700C1208A /* InfoDictionary.swift */; };
 		51CACB9529D500290034CEE5 /* VideoPIPViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 51CACB9429D500290034CEE5 /* VideoPIPViewController.swift */; };
 		51DE55C92A6646710050AD06 /* Sysctl.swift in Sources */ = {isa = PBXBuildFile; fileRef = 51DE55C82A6646710050AD06 /* Sysctl.swift */; };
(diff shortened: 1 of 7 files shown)


low-batt:
## Preferred Implementation

The plan is to replace all built-in code for downloading from online subtitle providers with plug-in implementations. Thus the preferred way to address the need to switch from the Open Subtitles XLMRPC API to their new REST API is to do so as a part of the refactoring of the code into a plug-in. 

## Potential Problem

The potential problem with this plan is due to the following from [OpenSubtitles.org API deprecation and end of life](https://forum.opensubtitles.org/viewtopic.php?f=11&t=17930&sid=d6897753acd72f04f9cf733afe736831):
> 2023-12-31 end of life, API endpoint will be turned off

Assuming:
- IINA users must not loose access to Open Subtitles
- IINA users must not be required to use a Beta or Nightly build to continue to have access to Open Subtitles
- The plug-in system will be enabled in IINA 1.4.0

Then this would require the IINA 1.4.0  release to be out before the end of the year. Possibly doable, but awfully aggressive based on the time taken for previous releases. There are quite a few tasks listed in the 1.4.0 goals. This does not seem like a release that should be rushed.

## Temporary STOPGAP

The commit in this PR **does not** contain the desired implementation. That requires a JavaScript expert. This PR is the _backup plan_ and contains a _temporary stopgap_ solution that updates the builtin legacy code. If we can come up with a way to implement and release a plug-in version before the end of year deadline then there will be no need to review this PR and it can be directly tossed into the trash.

This PR provides an option for an IINA 1.3.4 release that addresses the 1.3.3 regressions and ensures continued download support from Open Subtitles.

## Considerations for Plug-in Implementation

Another reason to not rush the refactoring of the built-in code into plug-ins is that existing issues that should be addressed when the code is rewritten. I suspect some of these issues will be easier to address with plug-ins.

The ones I can remember:
- Terrible error reporting. Users think IINA is broken due to always reporting `Network Error`
- In settings you can switch to mpv's OSD instead of IINA's. Users who do that probably think the online subtitles feature is totally broken as "nothing happens". I remember a user mentioning this
- Issue #4057 subtitles window too narrow to see full subtitle names
- Issue #3530 trouble with shooter.cn

With Open Subtitles it is especially important to clearly report the status of a user's download quota.

The Open Subtitles REST API provides lots of options that can be specified when searching. It would be good to fully support these features. Search results now support paging. There are lots of details returned about each subtitle file that users might want to see.

As this PR is only a temporary stopgap it does not address any of these problems.

## Draft

This PR is still a draft is because IINA needs to obtain an Open Subtitles [API key](https://opensubtitles.stoplight.io/docs/opensubtitles-api/e3750fd63a100-getting-started#api-key). The code has been checked in without an API key. Look for "FIXME" in `OpenSubClient`. If the `OpenSubClient` class is updated with a key then this code will work in its current state. We should discuss this on Telegram.

## Release Notes

Users with an [opensubtitles.org](https://www.opensubtitles.org/) account will be unable to download subtitles if they have not imported their accounts into [opensubtitles.com](https://www.opensubtitles.com/). Not sure if Open Subtitles users are aware of this change. This definitely needs to be called out in the release notes. Maybe something like the following:

Open Subtitles [opensubtitles.org](https://www.opensubtitles.org/) is being rebuilt as [opensubtitles.com](https://www.opensubtitles.com/). API access for downloading subtitles from [opensubtitles.org](https://www.opensubtitles.org/) will be shutdown at the end of 2023. This version of IINA has been changed to use [opensubtitles.com](https://www.opensubtitles.com/) instead of [opensubtitles.org](https://www.opensubtitles.org/). Users with  [opensubtitles.org](https://www.opensubtitles.org/) accounts **must** [import](https://www.opensubtitles.com/users/import) their accounts into [opensubtitles.com](https://www.opensubtitles.com/). See issue #3920 for details.

## Logging

The lack of logging in this area of IINA has hindered diagnosing problems downloading subtitles. Adding to this problem is that the responses from Open Subtitles are not consistent. Open Subtitles will return HTML error pages, apparently due to overload. By the time a developer tries to reproduce the problem everying is working fine.

These changes include extensive logging as can be seen in this example:
[iina.log](https://github.com/iina/iina/files/13260430/iina.log)

browser-bug:
It doesn't seem as precise as the older version. There are many video files which are not finding anything through [opensubtitles.com](). Is this a limitation to the new REST Api or did you change some parameters during the call? I cannot find the exact implementation but are we still searching by moviehash and/or by name as (I guess) we did before?

low-batt:
Lets continue this discussion in the new issue #4749.

The particular problem #4749 identifies has to do with results for TV 
episodes not being narrowed to the specific episode. Possibly that is 
the problem behind the lack of precision being reported?

The new REST API is quite different from the old API. It still uses hash 
codes and the new code is passing the hash code to Open Subtitles.

In investigating #4749 I found an Open Subtitles utility API that 
attempts to extract information from the filename. Apparently the new 
API expects clients to call this and add the extracted information to 
the search query. I am currently finishing code that does this and it 
looks like that will address the problem with TV episodes.

The report of not finding anything is concerning. To investigate that I 
need a specific example that shows this.

Communication with Open Subtitles is now fully logged. With a log file I 
can tell you exactly what happened.

The 1.3.4 release was rushed to meet the Open Subtitles deadline. Quite 
a few tasks were left undone. We are planning issuing another release 
soon that addresses those issues along with the above mentioned change 
to the Open Subtitles support.

Lets continue this discussion in #4749. If you can provide an example 
where no results are returned I will investigate that ASAP.

PR #4687, opened 2023-11-11
Fix Swift 5.9 warning

svobs:

---

**Description:**

See comments below

### DIFF
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -348,6 +348,21 @@ extension NSMutableAttributedString {
 }
 
 extension Data {
+  init<T> (bytesOf thing: T) where T: FixedWidthInteger {
+    var copyOfThing = thing
+    self.init(bytes: &copyOfThing, count: MemoryLayout<T>.size)
+  }
+
+  init(bytesOf num: Double) {
+    var numCopy = num
+    self.init(bytes: &numCopy, count: MemoryLayout<Double>.size)
+  }
+
+  init(bytesOf ts: timespec) {
+    var mutablePointer = ts
+    self.init(bytes: &mutablePointer, count: MemoryLayout<timespec>.size)
+  }
+
   var md5: String { Insecure.MD5.hash(data: self).map { String(format: "%02x", $0) }.joined() }
 
   var chksum64: UInt64 {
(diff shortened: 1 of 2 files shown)


saagarjha (changes requested):
Don't merge this yet. The changes here should be as follows:

1. Constrain to `FixedWidthInteger`.
2. Add an overload for `Double`.
3. Figure out what we need to do to serialize `timespec` correctly and in a backwards-compatible manner.

When `BitwiseCopyable` is available, we should migrate to that.

svobs:
@saagarjha alright, put a little bit more effort into it... 

The documentation for `MemoryLayout.size` is terrible, but via debugging it was clear that `MemoryLayout.size(ofValue: x)` was returning incorrect values (it was always 8), but `MemoryLayout<T>.size` gave correct values.

Not sure if I got your whole point in (3). Looks like `timespec` is just 2 Int-type values, and although I couldn't find the def for `__darwin_time_t`, the result of `MemoryLayout<timespec>.size` is 16 which seems to vouch for that idea.

Didn't dive in to the written files with a hex editor; just did spot testing. Thumbnail generation & Date Last Opened updates seemed to work fine before this, and seem to work fine after.

saagarjha:
What value were you trying this on? The two should be consistent in their results. The problem with serializing `timespec` is that it can include padding if you write it out directly, which something that you generally don't want to do.

svobs:
> What value were you trying this on? The two should be consistent in their results.

Well...I guess I was wrong... I could have sworn I saw the debugger showing `8` for both `version` (`UInt8`) and `timespec`. But just tried again and I see that they are consistent. So then everything seems to be equivalent to what it was before.

Did not learn until now that Apple's ARM is little-endian. And found [here](https://apple.stackexchange.com/questions/252542/is-any-version-of-os-x-macos-vulnerable-to-the-year-2038-problem) that `__darwin_time_t` is expected to be 64-bit also for all 64-bit processors.

svobs:
Pinging @saagarjha: this is still marked as "changes requested" but should be ready to go. Please let me know of any concerns which are blocking this.

saagarjha:
Well my point was we should serialize timespec's members and not just try to splat its contents directly.

svobs:
I dug into this some more and I don't see why breaking apart the `timespec` struct is necessary or even wise. I don't see any other version of this struct. We only support 64-bit MacOS. Both x86 & arm64 are little-endian. But I broke apart its fields into an [Int] in my new commit, so see if that is satisfactory.

The only case where this field is being used is to (comments) `set "date last opened" attribute`. I was able to confirm via `Mdls -name kMDItemLastUsedDate {file}` that the attribute is set correctly.

PR #4715, opened 2023-12-14
Fix phantom "B" tick in slider when AB-Loop is used, #4705

### LINKED ISSUE #4705: Phantom "B" tick appearing in play slider when AB-Loop is used
**System and IINA version:**

- macOS 14.1.1
- IINA 1.3.3 dev build aa5a4158cef44e972c8515c2a723db14bc97bd83

**Expected behavior:**

A and B ticks in playback position slider should match timestamps of `ab-loop-a` and `ab-loop-b` properties , respectively, if and only if each is set.

**Actual behavior:**

B tick appears when `ab-loop-b` is not set, seems to be in an arbitrary spot, and may not even be inside the slider.

SCR-20231202-rmfr
SCR-20231202-rnbc

**Steps to reproduce:**
1. Open a video file in IINA. Set a key binding with action `ab-loop`.
2. In the player window, seek to >50% of the file's duration, and then pause.
3. Press the key for the `ab-loop` binding. See if the B tick appears at some distance to the right.
4. Seek to another spot in the video and activate `ab-loop` again. See if a third tick appears.
5. Try again with "floating" OSC but seek almost to the end. B tick will be outside the OSC.

- [NA] MPV does not have this problem.

Doesn't seem to happen at the beginning of the file, or is at least not as bad. But happens most of the time.

low-batt:
This commit will:
- Change `PlaySlider.draw` to not call the loop knob draw methods when drawing under macOS 10.14+
- Change `PlaySliderLoopKnob.draw` to not use the knob position within the slider as the x-coordinate when drawing under macOS 10.14+

These changes remove a workaround that was required due to `NSSlider` not drawing its subviews. As of macOS Sonoma `NSSlider` has changed and subviews are being drawn. This is good and means the workaround is no longer needed and must not be applied as it is causing the incorrect drawing of the knobs.

- [x] This implements/fixes issue #4705.

---

**Description:**

### DIFF
--- a/iina/PlaySlider.swift
+++ b/iina/PlaySlider.swift
@@ -57,14 +57,29 @@ final class PlaySlider: NSSlider {
 
   // MARK:- Drawing
 
+  /// Draw the slider.
+  ///
+  /// The [NSSlider](https://developer.apple.com/documentation/appkit/nsslider) method is being overridden
+  /// for two reasons.
+  ///
+  /// With the onscreen controller hidden and a movie playing spindumps showed time being spent drawing the slider even though it
+  /// was not visible. Apparently `NSSlider.draw` is not calling
+  /// [hiddenOrHasHiddenAncestor](https://developer.apple.com/documentation/appkit/nsview/1483473-hiddenorhashiddenancestor)
+  /// to see if drawing can be avoided.  This was noticed under macOS Monterey.  Unknown if Apple addressed this in later macOS
+  /// releases.
+  ///
+  /// The loop knobs are added as subviews to the slider. That should have resulted in the `PlaySliderLoopKnob.draw` method
+  /// being called when the slider was being drawn. Prior to macOS Sonoma that did not occur. The assumption is that the
+  /// [NSSlider](https://developer.apple.com/documentation/appkit/nsslider) `draw` method was not calling
+  /// `super.draw` and that has now been corrected. As a workaround on earlier versions of macOS the loop knob `draw` method
+  /// is called directly.
   override func draw(_ dirtyRect: NSRect) {
-    // With the onscreen controller hidden and a movie playing spindumps showed time being spent
-    // drawing the slider even though it was not visible. Apparently NSSlider is missing the
-    // following check.
     guard !isHiddenOrHasHiddenAncestor else { return }
     super.draw(dirtyRect)
-    abLoopA.draw(dirtyRect)
-    abLoopB.draw(dirtyRect)
+    if #unavailable(macOS 14) {
+      abLoopA.draw(dirtyRect)
+      abLoopB.draw(dirtyRect)
+    }
   }
 
   override func viewDidUnhide() {
--- a/iina/PlaySlider.swift
+++ b/iina/PlaySlider.swift
@@ -76,10 +76,11 @@ final class PlaySlider: NSSlider {
   override func draw(_ dirtyRect: NSRect) {
     guard !isHiddenOrHasHiddenAncestor else { return }
     super.draw(dirtyRect)
-    if #unavailable(macOS 14) {
-      abLoopA.draw(dirtyRect)
-      abLoopB.draw(dirtyRect)
-    }
+    abLoopA.needsDisplay = true
+    abLoopB.needsDisplay = true
+    guard #unavailable(macOS 14) else { return }
+    abLoopA.draw(dirtyRect)
+    abLoopB.draw(dirtyRect)
   }
 
   override func viewDidUnhide() {
--- a/iina/PlaySliderLoopKnob.swift
+++ b/iina/PlaySliderLoopKnob.swift
@@ -132,13 +132,24 @@ final class PlaySliderLoopKnob: NSView {
     }
   }
 
+  /// Draw the knob.
+  ///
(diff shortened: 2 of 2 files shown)


svobs:
I've been using this code for a while and stumbled onto a remaining issue. If I set A and/or B positions, all looks good, but then when I resize the window they do not scale with the slider.

For example, I started with a large window when I set A & B. Then I reduced the window size until one of the knobs fell out of the slider's bounds entirely:

SCR-20231222-tkbd

low-batt:
Good catch! I found it now sometimes draws the knobs and sometimes not. Possibly `NSSlider` is still not quite right. I am now setting `needsDisplay` on the knobs. That causes the knobs to be redraw when the slider is resized.

svobs:
> I found it now sometimes draws the knobs and sometimes not.

Something's not right here. I debugged a bit and found that `draw()` is not getting called for the knobs when the slider is getting resized. Could be that the layout process sees that the knobs haven't changed position or size and thus don't need to be redrawn. This is a valid assumption to make, since it's a little bit unusual to be using `draw()` to change the position of the knobs. Ideally the knob's `draw` method would only worry about drawing the knob inside the knob's bounds (i.e. with no offset), and then a layout constraint would be used to define where the knob should be placed relative to the slider bar, I think.

Ran out of time before I could dig deeper. What I still don't understand is how Cocoa decides when to "erase" an object which has already been drawn. I suspect that might be the key. Say that if, instead of waiting till the next `draw`, you were to loop over the code inside `PlaySliderLoopKnob.draw()` several times with different offsets, I'd expect you'd draw a bunch of knobs. But you can draw them anywhere in the window. Are they all supposed to get cleared before the next call to `draw`? The existence of the `dirtyRect` param seems to imply the possibility of partial redraw, which could mess up the assumptions the code is making.

low-batt:
In Ventura and earlier `PlaySliderLoopKnob.draw` is never called. To workaround that `PlaySlider.draw` was calling  `PlaySliderLoopKnob.draw`. That required `PlaySliderLoopKnob.draw` to use an X offset within the slider when drawing.

In Sonoma AppKit started calling `PlaySliderLoopKnob.draw`. That resulted in `PlaySliderLoopKnob.draw` being called twice, once by AppKit and once by `PlaySlider.draw`. When AppKit calls draw the X value is relative to the knobs frame. So using the X offset relative to the slider was drawing the knob outside of the slider. Removing the call to  `PlaySliderLoopKnob.draw` from `PlaySlider.draw` and using a zero offset seemed to work.

But for some reason AppKit does not call `PlaySliderLoopKnob.draw` some of the time. This is what was happening during resize. For me it sometimes corrected the drawing of the knob when I stopped resizing it. Sometimes not. So now  `PlaySlider.draw` is setting `needsDisplay` to make sure the knobs are drawn.

I think this is working?

On the 1.3.4 front, key event related issues 3710, 4538, 4562 and 4582 are on the list of issues we'd like to fix for 1.3.4. I'm not clear on the state of PRs and which fixes what. PR 4567 is on the list, but I'm not on which issues that fixes.

Issue 4548 is also on the list and that one's PR is in draft.

The primary goal for 1.3.4 is addressing issue 3920. That makes it desirable to get the release out before 2024. There is talk of a backup plan of a 1.3.5 release if we can't get all the planned fixes into 1.3.4 in time. But we are trying to avoid that.

svobs:
> I think this is working?

When I tested it I thought I found cases where it didn't animate along with the window during the resize, although the knobs clicked into the correct position at the end. But my attempts to retest have been hobbled by the fact that I'm traveling internationally this week and I just discovered that I need a good Internet connection to build the project. Every time I change branches, XCode fails with a bunch of errors about "Missing package product[s]", as though it's deleting the dependencies and re-downloading them with every new checkout. Will try this again when I can get some time with stable/fast Internet.

> On the 1.3.4 front, key event related issues 3710, 4538, 4562 and 4582 are on the list of issues we'd like to fix for 1.3.4. I'm not clear on the state of PRs and which fixes what. PR 4567 is on the list, but I'm not on which issues that fixes.
> 
> Issue 4548 is also on the list and that one's PR is in draft.

- 3710 (duplicates: 3918, 3321, 3110, 3440, 3534)
  - Fixed by PR 3732. Would seriously love to have this merged, as this issue has been number 1 on my wish list and was the reason I started contributing to the project.
- 4538
  - Already fixed + merged by PR 4569.
- 4562 (dup 4692)
  - Fixed by PR 4567, but didn't get consensus on approval last I checked. But would be good to do as this is a high visibility regression.
- 4582
  - Sounds like the same underlying problem as 4562, so may also be fixed by PR 4567, but I haven't dug into it too deeply.
- 4548
  - There is no "proper" fix for this yet because it's non-trivial and I never got around to putting in the time for it. I recommend just reverting PR 4439 as a short-term fix, because more users seem to be impacted by the regression than are helped by the original fix.

low-batt:
Thanks for the status. I have updated the issues accordingly.

Issue 3710 is now critical because the latest mpv and FFmpeg updates introduced regressions with seeking requiring `seek exact` as a workaround. So I'm pushing to get that one fixed.

I confirmed issue 4582 is also fixed by PR 4567.

I too encounter Xcode deciding to reset its package caches causing it to download all the packages again from the net. I'm not exactly sure what triggers it.

uiryuu (approved):
Tested this PR and it seems to work. LGTM

PR #4718, opened 2023-12-16
Fix cannot open display link, #4520

### LINKED ISSUE #4520: Cannot open display link.  The application will now close
**System and IINA version:**

- macOS  Ventura 13.4.1 (c)

- IINA  1.3.2 

**Expected behavior:**

App stays open in the background
**Actual behavior:**
Blank window with message "Cannot open display link. The application will now close

  Cannot find any stack trace reports

<details>

```
```
</details>

<br>

<details>

```
```
</details>

<br>

**Steps to reproduce:**

Happens after the app is running in the background.  Very often.

low-batt:
This commit will change the method `PlayerCore.pause` to not call `MPVControler.setFlag` if the `mpv` core is idle or stopping.

- [x] This implements/fixes issue #4520.

---

**Description:**

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -647,7 +647,18 @@ class PlayerCore: NSObject {
     info.isPaused ? resume() : pause()
   }
 
+  /// Pause playback.
+  ///
+  /// - Important: Setting the `pause` property will cause `mpv` to emit a `MPV_EVENT_PROPERTY_CHANGE` event. The
+  ///     event will still be emitted even if the `mpv` core is idle. If the setting `Pause when machine goes to sleep` is
+  ///     enabled then `PlayerWindowController` will call this method in response to a
+  ///     `NSWorkspace.willSleepNotification`. That happens even if the window is closed and the player is idle. In
+  ///     response the event handler in `MPVController` will call `VideoView.displayIdle`. The suspicion is that calling this
+  ///     method results in a call to `CVDisplayLinkCreateWithActiveCGDisplays` which fails because the display is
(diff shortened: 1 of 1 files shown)


iina/PlayerCore.swift
```diff
@@ -647,7 +647,18 @@ class PlayerCore: NSObject {
     info.isPaused ? resume() : pause()
   }
 
+  /// Pause playback.
+  ///
+  /// - Important: Setting the `pause` property will cause `mpv` to emit a `MPV_EVENT_PROPERTY_CHANGE` event. The
```
uiryuu, line 652: Should we also block some of other `setFlag`s (if not all) when the mpv core is idle or stopping?

low-batt: I did not do any sort of cleanup or changes for potential bugs with 1.3.4 in mind. Since 1.3.3 added regressions I have been intentionally keeping fixes to the minimum required change. The problem at hand was not reproducible and had to be diagnosed through code reading and experiments. This was another reason to minimize this fix. If the user reports this fixes the problem then that validates the guess as to what is going wrong.

Ok to merge as is?

One of the goals I listed for 1.4.0 was to fix the thread related problems in `MPVController` and refactor that class as well as `PlayerCore`. I thought it was best to wait for a Beta release to due to extensive refactoring. One of the planned changes is to switch to a state enumeration to eliminate these separate flags. But first the thread related problems need to be resolved. My thoughts on that are that `MPVController` should not be doing so much work using its own queue. That queue/thread should be focused on reading the events from `mpv` and then in most cases dispatching the work to the main thread. As it is now you can see cases where it is half done on  `MPVController` thread and then dispatched to main. This is causing data races.

PR #4739, opened 2024-01-01
Fix key bindings for seek exact seeks by keyframes, #3710

### LINKED ISSUE #3710: Keybindings for `seek exact` are silently overridden to `seek keyframes`

low-batt:
This commit will:
- Change MenuController.sameKeyAction to detect the exact flag and pass it on as extra data
- Change MainMenuActionHandler.menuStep to pass the extra data to PlayerCore.seek if present

The main author of these changes is Matt Svoboda. I merely addressed some review comments.

- [x] This implements/fixes issue #3710.

---

**Description:**

### DIFF
--- a/iina/MainMenuActions.swift
+++ b/iina/MainMenuActions.swift
@@ -90,11 +90,15 @@ extension MainMenuActionHandler {
   }
 
   @objc func menuStep(_ sender: NSMenuItem) {
-    let seconds = Double(abs((sender.representedObject as? Int) ?? 5))
-    if sender.tag == 0 { // -> 5s
-      player.seek(relativeSecond: seconds, option: .relative)
-    } else if sender.tag == 1 { // <- 5s
-      player.seek(relativeSecond: -seconds, option: .relative)
+    if let args = sender.representedObject as? (Double, Preference.SeekOption) {
+      player.seek(relativeSecond: args.0, option: args.1)
+    } else {
+      let seconds = Double(abs((sender.representedObject as? Int) ?? 5))
+      if sender.tag == 0 { // -> 5s
+        player.seek(relativeSecond: seconds, option: Preference.SeekOption.defaultValue)
+      } else if sender.tag == 1 { // <- 5s
+        player.seek(relativeSecond: -seconds, option: Preference.SeekOption.defaultValue)
+      }
     }
   }
 
--- a/iina/MenuController.swift
+++ b/iina/MenuController.swift
@@ -8,36 +8,65 @@
 
 import Cocoa
 
-fileprivate func sameKeyAction(_ lhs: [String], _ rhs: [String], _ normalizeLastNum: Bool, _ numRange: ClosedRange<Double>?) -> (Bool, Double?) {
+fileprivate func sameKeyAction(_ lhs: [String], _ rhs: [String], _ normalizeLastNum: Bool, _ numRange: ClosedRange<Double>?) -> (Bool, Double?, Any?) {
   var lhs = lhs
-  if lhs.first == "seek" && (lhs.last == "exact" || lhs.last == "keyframe") {
-    lhs = [String](lhs.dropLast())
+  var extraData: Any? = nil
+  if lhs.first == "seek", rhs.first == "seek", lhs.count > 2, let last = lhs.last {
+    // This is a seek command that includes flags. Adjust the command before checking for a match.
+    if lhs.count == 4 {
+      // The original mpv seek command required that the keyframes and exact flags be passed as a
+      // 3rd parameter. This is considered deprecated but still supported by mpv. Convert this to
+      // the current command format by combining the flags using a "+" separator.
+      lhs[2] = "\(lhs[2])+\(lhs[3])"
+      lhs = [String](lhs.dropLast())
+    }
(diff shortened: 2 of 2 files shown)


low-batt:
@uiryuu I tried to address your review feedback. Please have a look.
@svobs I know you are busy, but if you get a chance look please have a look.

In my testing these key bindings matched:
- `seek 20`
- `seek -20 exact`
- `seek 20 relative+exact`
- `seek 20 exact+relative`
- `seek 20 relative exact`

These did not match (as expected):
- `seek 20 keyframes`
- `seek 20 relative+keyframes`
- `seek 20 relative keyframes`
- `seek 20 absolute`

The `keyframe` flag can't be supported without dealing with the current code confusion over `SeekOption.relative`. That value is labeled in the UI as `Keyframe seek` in the `Seek type` setting under `Control`. That makes sense. But internally it is named `relative`. The code where it is used in `PlayerCore.seek` looks like this:
```swift
    case .relative:
      mpv.command(.seek, args: ["\(relativeSecond)", "relative"], checkError: false)
```

Trouble is, if you really want to do a keyframe seek you **must** pass the `keyframes` flag. Although the seek command will default to `keyframes` that default can be changed. From the `mpv` manual entry for [hr-seek](https://mpv.io/manual/stable/#options-hr-seek):
> **--hr-seek=<no|absolute|yes|default>**
>    Select when to use precise seeks that are not limited to keyframes. Such seeks require decoding video from the previous keyframe up to the target position and so can take some time depending on decoding performance. For some video formats, precise seeks are disabled. This option selects the default choice to use for seeks; it is possible to explicitly override that default in the definition of key bindings and in input commands.
> - **no**:	Never use precise seeks.
>  -  **absolute**:	Use precise seeks if the seek is to an absolute position in the file, such as a chapter seek, but not for relative seeks like the default behavior of arrow keys.
> - **default**:	Like absolute, but enable hr-seeks in audio-only cases. The exact behavior is implementation specific and may change with new releases (default).
>- **yes**:	Use precise seeks whenever possible.
    always:	Same as yes (for compatibility).

So if you have set `hr-seek=yes` then you **must** pass the `keyframes` flag in a `seek` command if you want to seek by keyframe.

As the issue at hand is about getting `exact` to work I don't think we need to address this in the current PR.

uiryuu (approved):
I also noticed the issue you mentioned when I was reviewing the original PR, and that indeed need to be addressed in another one. The current changes LGTM.

PR #4755, opened 2024-01-07
Fix discrepancies with new Open Subtitles API Results, #4749

### LINKED ISSUE #4749: Discrepancies with new Open Subtitles API Results When Using IINA
I've encountered a significant issue when utilizing the new API through IINA. The results returned by the new API deviate substantially from those provided by the old API. As an illustrative example, when I input 'The.Sopranos.S06E14.Stage.5.720p.WEB-DL.h.264.DD5.1-NTb.mkv', the API returns the likes of 'Sopranos 1x01', which is clearly inaccurate. This discrepancy raises questions about the source of the issue. It's unclear whether the problem lies within the IINA integration, the new API itself, or potentially changes in the Open Subtitles libraries. Further investigation or insights into this matter would be greatly appreciated.

**System and IINA version:**

- macOS 14.2.1 (23C71)
- IINA 1.3.4 Build 140

low-batt:
This commit will:
- Add a guessit method to OpenSubClient class
- Add a GuessitResponse struct to hold the results of a call to the Open Subtitles guessit API
- Change the subtitles method to accept a GuessitResponse and pass information such as season and episode number as query parameters
- Add a guessit method to OpenSub class
- Change the searchForSubtitles method  to accept a GuessitResponse and pass it on when calling OpenSubClient.subtitles
- Change the OpenSub.Fetcher.fetch method to call guessit and pass the results to searchForSubtitles

With these changes it is more likely that the results of searching for a particular episode of a TV series will be narrowed to subtitles for that episode.

- [x] This implements/fixes issue #4749.

---

**Description:**

### DIFF
--- a/iina/OpenSubClient.swift
+++ b/iina/OpenSubClient.swift
@@ -267,6 +267,30 @@ class OpenSubClient {
     }
   }
 
+  /// [Guessit](https://opensubtitles.stoplight.io/docs/opensubtitles-api/7783e082edcf7-guessit)
+  /// method.
+  ///
+  /// Extracts as much information as possible from a video filename.
+  /// - Parameter filename: Video filename to extract information from.
+  /// - Returns: A `GuessitResponse` containing the information that could be extracted from the filename.
+  func guessit(filename: String) -> Promise<GuessitResponse> {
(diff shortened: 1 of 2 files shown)


uiryuu (approved):
LGTM

Issue #2139, opened 2018-12-23
How to make Repeat mode enabled by default?


SarveZir:
Hi Guys

Every time I close the app **Repeat mode** becomes disabled so I have to enable it manually. Possible to change it to **enabled** by default?

bw-space:
Can you please elaborate your issue?

SarveZir:
MPV : **--loop** to loop the playlist, **--loop-file** to loop playlist entries.
inna : ? 

(No need to say I want an inna's permanent option to work like that; Just like mpv config options)

PaulSlocum:
Really needs a way to do this.  I would want a preference in settings that specifies the default loop mode.

tertiumsquid:
I'd like this as well, as Iina has become my go-to music player.  If possible, maybe have two settings - one for music playback and one for video, as I'd rather my videos didn't loop by default.

thinker-g:
I'd like to have this feature also.
How about just to make IINA to "remember" the user's choice? Just like it remembers the timestamp of last play of each video.

abplcode:
i need this.

Acciente717:
I would also like to have this feature.

JiashengWu:
> MPV : **--loop** to loop the playlist, **--loop-file** to loop playlist entries.
> inna : ?
> 
> (No need to say I want an inna's permanent option to work like that; Just like mpv config options)

Per my understanding, I suppose in INNA setting loop/loop-file to 1, 2, ..., inf would do the same?
It's in Preference -> Advanced -> Additional mpv options

thinker-g:
Dears, I figured out how to turn on the playlist loop by default.

Just to set an additional mpv option in  Preference -> Advanced -> Additional mpv options, with following values:
name: **loop-playlist**
value: **inf**

PS: For other available values of this option, you may refer to the [official mpv-player document](https://mpv.io/manual/master/#options-loop-playlist).

PaulSlocum:
Adding "loop-playlist"/"inf" to mpv options did not cause videos to loop on startup for me.  I'm on IINA 1.2.0 on MacOS.

One of the reasons this feature is needed is because when opening short MP4 loops that are only a couple/few seconds, IINA often finishes playing and jumps to the next video before you have any time to active the loop function.

tomtomkk:
> Adding "loop-playlist"/"inf" to mpv options did not cause videos to loop on startup for me. I'm on IINA 1.2.0 on MacOS.
> 
[...]

PR #4761, opened 2024-01-08
Add an option to repeat file/playlist when opening a media

uiryuu:
- [x] This implements/fixes issue #2139.

---

**Description:**
Added an new option:

image

Wording suggestions are welcomed

### DIFF
--- a/iina/Base.lproj/PrefGeneralViewController.xib
+++ b/iina/Base.lproj/PrefGeneralViewController.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="21701" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="22154" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
     <dependencies>
         <deployment identifier="macosx"/>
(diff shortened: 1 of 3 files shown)


low-batt:
[This comment](https://github.com/iina/iina/issues/2139#issuecomment-497121250) on  issue #2139 says:
> If possible, maybe have two settings - one for music playback and one for video, as I'd rather my videos didn't loop by default.

We don't have a concept of separate settings for music mode, but we already have precedent for a `Only in "music mode"` setting as can be seen in the screenshot above. Maybe this new setting needs such a subordinate setting as well.

Doesn't the new text need to be added under `en.lproj` as well?

iina/Base.lproj/PrefGeneralViewController.xib
```diff
@@ -525,22 +524,57 @@
                         <binding destination="IBP-Mz-Maa" name="value" keyPath="values.playlistShowMetadataInMusicMode" id="nxW-hX-9co"/>
                     </connections>
                 </button>
+                <button verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="AbF-Dw-gXa">
+                    <rect key="frame" x="118" y="7" width="260" height="18"/>
+                    <buttonCell key="cell" type="check" title="Enable repeat automatically upon start" bezelStyle="regularSquare" imagePosition="left" state="on" inset="2" id="j4o-RP-pfA">
```
low-batt, line 529: Missing a colon after "start". I looked over the settings and that is a consistent convention, except for one other setting, `Check for updates` is also missing a colon at the end of the label.

svobs:
> Wording suggestions are welcomed

How about `Enable repeat for new windows:` ? Because using "start" isn't clear whether "start" means "start of playback" or "application start".

Assuming this is for new windows. Is it also set for existing windows when opening a media in an existing window? If so it would have to be something like `Enable repeat when media is opened manually`, following the existing terminology in Settings > UI > Window.

Issue #4688, opened 2023-11-11
Open Recent menu cleared with each new build


low-batt:
**System and IINA version:**

- macOS 14.1.1
- IINA `develop`

**Expected behavior:**
The list of recent files in the `Open Recent` menu item under the `File` menu is only cleared when the user clicks on `Clear Menu`.

**Actual behavior:**
Running an IINA executable built with Xcode 15.0.1 under macOS 14.1.1 clears the list of recent files. After that the `Open Recent` menu item works as expected. However if a source is modified and IINA is rebuilt, then the list will be cleared again when the new executable is run.

**Steps to reproduce:**
Under macOS Sonoma 14.1.1 and using Xcode 15.0.1 follow these instructions:
- Start the released version of IINA
- Play a video
- Open the `File` menu and move the cursor over the `Open Recent` menu item
- Notice the video is listed in the `Open Recent` menu item
- Quit the released version of IINA
- Run IINA from Xcode
- Open the `File` menu and move the cursor over the `Open Recent` menu item
- Notice the list is _empty_
- Play a video
- Open the `File` menu and move the cursor over the `Open Recent` menu item
- Notice the video is listed in the `Open Recent` menu item
- Quit the develop version of IINA
- Modify a source
- Run IINA from Xcode
- Open the `File` menu and move the cursor over the `Open Recent` menu item
- Notice the list is _empty again_

- [x] MPV does not have this problem.

It does not support `Open Recent`.

**How often does this happen?**
Every time a new executable is run.

low-batt:
Searching for others reporting this problem I found the Apple developer forum post: [Document controller's recent files list cleared between versions](https://developer.apple.com/forums/thread/703694). That post reports this occurring after building on Monterey. Unfortunately the post provides no clue to the problem.

I confirmed that [recentDocumentURLs](https://developer.apple.com/documentation/appkit/nsdocumentcontroller/1514976-recentdocumenturls) is indeed returning an empty list.

low-batt:
I ran a test where I built the develop branch of IINA using Xcode 15.0.1 under macOS Ventura 13.6.1 and brought that executable to my Mac running macOS Sonoma 14.1.1. Under Ventura that executable does not clear the list of previously opened files shown in the `Open Recent` menu item. Running that same executable under Sonoma clears the list and  `Open Recent` only shows a dimmed `Clear Menu`.

Xcode 15 links the macOS 14 SDK which uses a newer version of AppKit as shown by `otool`.

[...]

PR #4792, opened 2024-01-25
Include fix for Open Recent menu in release builds, #4688

low-batt:
This commit will:
- Add `clearRecentDocuments` and `noteNewRecentDocumentURL` methods to `AppDelegate`
- Change `PrefGeneralViewController` and `PrefUtilsViewController` to call `AppDelegate.clearRecentDocuments` instead of the `NSDocumentController` method
- Change `AppDelegate.openFile` and `PlayerCore.fileLoaded` to call `AppDelegate.noteNewRecentDocumentURL` instead of the `NSDocumentController` method
- Change `saveRecentDocuments` to not include a check for the `recordRecentFiles` preference
- Remove the compiler directives that only included this code in debug builds

These changes enhance the workaround for macOS Sonoma clearing the open recent menu to cover the case where the user clears the list and also removes conditionalization that restricted this workaround to debug builds.

- [x] This implements/fixes issue #4688.

---

**Description:**

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -1607,7 +1607,7 @@ class PlayerCore: NSObject {
         HistoryController.shared.add(url, duration: duration.second)
       }
       if Preference.bool(for: .recordRecentFiles) && Preference.bool(for: .trackAllFilesInRecentOpenMenu) {
-        DispatchQueue.main.sync { (NSApp.delegate as? AppDelegate)?.noteNewRecentDocumentURL(url) }
+        DispatchQueue.main.sync { (NSApp.delegate as! AppDelegate).noteNewRecentDocumentURL(url) }
       }
 
     }
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -1607,10 +1607,7 @@ class PlayerCore: NSObject {
         HistoryController.shared.add(url, duration: duration.second)
       }
       if Preference.bool(for: .recordRecentFiles) && Preference.bool(for: .trackAllFilesInRecentOpenMenu) {
-        NSDocumentController.shared.noteNewRecentDocumentURL(url)
-#if DEBUG
-        DispatchQueue.main.async { (NSApp.delegate as? AppDelegate)?.saveRecentDocuments() }
-#endif
+        DispatchQueue.main.sync { (NSApp.delegate as? AppDelegate)?.noteNewRecentDocumentURL(url) }
       }
 
     }
--- a/iina/PrefGeneralViewController.swift
+++ b/iina/PrefGeneralViewController.swift
@@ -58,7 +58,7 @@ class PrefGeneralViewController: PreferenceViewController, PreferenceWindowEmbed
 
   @IBAction func rememberRecentChanged(_ sender: NSButton) {
     if sender.state == .off {
-      (NSApp.delegate as? AppDelegate)?.clearRecentDocuments(self)
+      (NSApp.delegate as! AppDelegate).clearRecentDocuments(self)
     }
   }
 }
--- a/iina/PrefGeneralViewController.swift
+++ b/iina/PrefGeneralViewController.swift
@@ -58,10 +58,7 @@ class PrefGeneralViewController: PreferenceViewController, PreferenceWindowEmbed
 
   @IBAction func rememberRecentChanged(_ sender: NSButton) {
     if sender.state == .off {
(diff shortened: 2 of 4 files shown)


low-batt:
The original plan was to only use the workaround added in PR  #4730 in debug builds and to sign nightly builds to avoid users losing their list of recent documents if they run a nightly build to see if it corrects a problem. This was not ideal as developers would still lose entries as they transitioned between the released version of IINA and their own builds. Now that we have decided not to sign nightly builds, this PR updates the changes so they are appropriate for inclusion in release builds. This also greatly improves the situation for developers as the open recent menu will appear to work like it does in earlier versions of macOS.

iina/PlayerCore.swift
```diff
@@ -1607,10 +1607,7 @@ class PlayerCore: NSObject {
         HistoryController.shared.add(url, duration: duration.second)
       }
       if Preference.bool(for: .recordRecentFiles) && Preference.bool(for: .trackAllFilesInRecentOpenMenu) {
-        NSDocumentController.shared.noteNewRecentDocumentURL(url)
-#if DEBUG
-        DispatchQueue.main.async { (NSApp.delegate as? AppDelegate)?.saveRecentDocuments() }
-#endif
+        DispatchQueue.main.sync { (NSApp.delegate as? AppDelegate)?.noteNewRecentDocumentURL(url) }
```
saagarjha: I know this isn't your code but just force unwrap this. If the application delegate is not `AppDelegate` there is no point in continuing execution

```suggestion
        DispatchQueue.main.sync { (NSApp.delegate as! AppDelegate).noteNewRecentDocumentURL(url) }
```

iina/PrefGeneralViewController.swift
```diff
@@ -58,10 +58,7 @@ class PrefGeneralViewController: PreferenceViewController, PreferenceWindowEmbed
 
   @IBAction func rememberRecentChanged(_ sender: NSButton) {
     if sender.state == .off {
-      NSDocumentController.shared.clearRecentDocuments(self)
-#if DEBUG
-      (NSApp.delegate as? AppDelegate)?.saveRecentDocuments()
-#endif
+      (NSApp.delegate as? AppDelegate)?.clearRecentDocuments(self)
```
saagarjha: Same here

```suggestion
      (NSApp.delegate as! AppDelegate).clearRecentDocuments(self)
```

iina/PrefUtilsViewController.swift
```diff
@@ -129,10 +129,7 @@ class PrefUtilsViewController: PreferenceViewController, PreferenceWindowEmbedda
     Utility.quickAskPanel("clear_history", sheetWindow: view.window) { respond in
       guard respond == .alertFirstButtonReturn else { return }
       try? FileManager.default.removeItem(atPath: Utility.playbackHistoryURL.path)
-      NSDocumentController.shared.clearRecentDocuments(self)
-#if DEBUG
-      (NSApp.delegate as? AppDelegate)?.saveRecentDocuments()
-#endif
+      (NSApp.delegate as? AppDelegate)?.clearRecentDocuments(self)
```
saagarjha:
```suggestion
      (NSApp.delegate as! AppDelegate).clearRecentDocuments(self)
```

saagarjha (approved):

iina/PlayerCore.swift
low-batt: Will fix.

Please do point out anything you spot that needs fixing during a review. I am always appreciative of your review comments and happy to fix anything you notice.

At the time I coded this I searched the code base and found this same pattern appears in many places. More instances use the conditional operator, but some force the conversion. I was thinking about adding an extension to `NSApp` to eliminate the duplication but decided such cleanup should wait for 1.4.0.

I am biased towards forcing such conversions. The conditional operator turns what would have been a crash seen and fixed by a developer into odd runtime behavior that can easily escape notice and end up being released to users. Troubles me that the industry views use of force operators to always be a _code smell_. So much so there was this Swift Pitch: [Moving toward deprecating force unwrap from Swift?](https://forums.swift.org/t/moving-toward-deprecating-force-unwrap-from-swift/43455) The core team decided deprecation was out of the question, responding in [this comment](https://forums.swift.org/t/moving-toward-deprecating-force-unwrap-from-swift/43455/81). But that is an example of how strongly many are against use of the force operators.

For a case like the one at hand, if I was unable to use a force operator I would use a `guard` statement that triggerd a fatal internal error if the downcast failed.

PR #4811, opened 2024-02-09
Fix typo in plugin 'network-request' message

svobs:

---

**Description:**

Fix typo: "requsts" should be "requests"

### DIFF
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -350,7 +350,7 @@
 // Plugin permissions
 
 "permissions.network-request.name" = "Network Request";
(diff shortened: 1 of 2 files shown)


low-batt (changes requested):
Nice catch!

The `Localizable.strings` in `en.lproj` needs to be corrected as well. Both the `Base.lproj` and `en.lproj` files are managed directly. The other files must be modified through Crowdin.

This same typo is in the UK English, but that will have to be corrected through Crowdin.

svobs:
> The Localizable.strings in en.lproj needs to be corrected as well. Both the Base.lproj and en.lproj files are managed directly. 

I amended my commit to include this.

low-batt (approved):
Changes look good. Glad you noticed this typo!

Issue #4763, opened 2024-01-08
Add subtitles FPS label on found subs list


browser-bug:
**What you want IINA to do:**
When displaying the subtitles found (using Opensubtitles.com, etc.) show also the FPS of the subs if available:
![image](https://github.com/iina/iina/assets/44030022/06e4b22d-7ec6-4192-a2fb-30df67c1a662)

**What IINA does currently:**
It shows the subs list without this information. In particular it currently only shows:
- Language
- Download number
- Stars
- Upload date

**Why you think this should be added:**
This would benefit the choice of the sub since it depends on the video source FPS. It is quite common that multiple subtitles are found but I cannot choose the correct one and therefore find myself going to the site for a double check (which makes the in-player download feature almost useless).

low-batt:
@browser-bug Thanks for entering this issue.

As the spacing is rather tight I rounded the fps up. Do you think that is sufficient or is more precision required?
![subtitle-fps](https://github.com/iina/iina/assets/86170219/e7aca08b-145a-4aa6-937d-587246245635)

browser-bug:
@low-batt 
Thanks to you for addressing the issue. I guess a round up will do aswell, since the most common values are always 24, 25, 30 or 60 to my knowledge. But still, being more precise doesn't hurt. If the spacing is too tight and there are no other options I guess rounding will do just fine. Maybe an icon on the left?

low-batt:
It is looking like our next release will be focused on fixing regressions with a feature release to follow. To get this in now for the next release I wanted to minimize the changes required. That is why I wanted to avoid trying to expand the size of this UI. I do agree that more precision is desirable.

The current code for downloading subtitles is considered legacy code. The plan is to replace the existing code with a plugin implementation. Hopefully that work will include enhancements to the user interface.

On using an icon, I did look around for an icon. I found a lot of them. They were associated with online gaming. I didn't find anything I thought people would recognize as frames per second. Are you aware of something?

For this change I would not have used an icon as that would require a lot of changes as the UI is not structured for that. What looks like icons in the UI is actually characters:

⬇︎
DOWNWARDS BLACK ARROW
Unicode: U+2B07 U+FE0E, UTF-8: E2 AC 87 EF B8 8E

★
BLACK STAR
Unicode: U+2605, UTF-8: E2 98 85

I will check with the other developers to see if they think it is ok to squeeze in a couple more characters.

low-batt:
[...]

PR #4825, opened 2024-02-17
Fixes to Subtitle Chooser layout

### LINKED ISSUE #4057: The Find Subtitle Result Window was too Narrow to see the whole name.

svobs:
- [x] This implements/fixes issue #4057.

---

**Description:**

- Allow expansion tooltips (as suggested in [this comment](https://github.com/iina/iina/issues/4057#issuecomment-1949629135))
- Set a preferred width of 480 to that it doesn't collapse to size of `osdLabel`. Uses a priority of 499 so that it will still shrink if there is not enough space.
- Fixes vertical text alignment for each table item's left & right labels
- Fix layout warning

### DIFF
--- a/iina/Base.lproj/SubChooseViewController.xib
+++ b/iina/Base.lproj/SubChooseViewController.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="16097" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="22505" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="16097"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="22505"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <objects>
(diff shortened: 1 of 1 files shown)


low-batt:
I'm pretty sure @lhc70000 is currently preparing changes in this area as well. The online subtitle providers are being transitioned from the built-in implementations to plug-ins. I believe that involves enhancing the user interface as well.

svobs:
> I'm pretty sure @lhc70000 is currently preparing changes in this area as well. The online subtitle providers are being transitioned from the built-in implementations to plug-ins. I believe that involves enhancing the user interface as well.

Ah, that is good to know. I got carried away and already had more extensive changes prepared. I will hold off on those.

low-batt (approved):
Pulled PR, built IINA and tested searching for subtitles online. Made the window tiny and confirmed tool tips worked. Looks good to me.

low-batt:
Will have to see what @lhc70000 thinks, but I'm thinking these changes can go into the bug fix release with the improvements @lhc70000 is working on going into the feature release.

This change possibly calls into question my decision, namely [this comment](https://github.com/iina/iina/issues/4763#issuecomment-1888076871):
> As the spacing is rather tight I rounded the fps up.

When I was working on that issue I was expecting we would not be altering the OSD for the bug fix release. If we proceed with this change, should we switch the fps to be a floating point number with two fractional digits?

svobs:
> If we proceed with this change, should we switch the fps to be a floating point number with two fractional digits?

I haven't paid much attention to subtitles stuff until now. I'm not sure how helpful or accurate the "fps" field is. But seems like the 2 decimals should be important given its use by traditional standards like 23.98, 29.98, etc.

I guess I'll try posting a new PR with all the fixes I made (which includes that) and see what people think.

low-batt:
I will keep an eye out for the new PR.

The fps field is helpful because if the fps of the video used when authoring the subtitles does not match the fps of the video you are trying to obtain subtitles for then the time codes within the subtitles may not precisely match the video which can cause the subtitles to drift out of synchronization with the video. The accuracy of the subtitle fps I believe depends upon the author correctly reporting that value, so it could be suspect.

uiryuu:
Thanks!

PR #4819, opened 2024-02-14
Fix data races, #3829, #3830, #3936, #3937

low-batt:
This commit will:
- Change `MPVController` to perform most event processing using the main thread
- Remove dispatching to main queue from `PlayerCore` methods now being called by `MPVController` using the main thread
- Add the `Atomic` property wrapper to the `playlistTotalLength` and `playlistTotalLengthIsReady` `PlaylistViewController` properties
- Add the `Atomic` property wrapper to the following `PlayerInfo` properties:
  - aid
  - secondSid
  - sid
   - subTracks
  - thumbnails
  - thumbnailsProgress
  - thumbnailsReady
  - vid
- Change references to `subTracks` to lock the array
- Change references to `thumbnails` to lock the array
- Add the `Atomic` property wrapper to the `hooks` `MPVController` property
- Change references to `hooks` to lock the array
- Change `PlayerCore.playbackRestarted` to re-enable the OSD by dispatching to the main thread using `async` with a delay instead of using a timer
- Add the method `logPropertyValueError` to `MPVController` to report when property values can not be converted to the expected type
- Replace `MPVController` method `reEnableOSDAfterFileLoading` with inline code

These changes will correct data races between the  main thread and the thread for the `MPVController` `com.colliderli.iina.controller` queue by changing `MPVController` to dispatch work to the main queue. This will simplify the code as in some cases work was being done on both the `controller` queue and the main queue as some processing required calling AppKit methods that must be run on the main thread.

These changes will correct the data races between the main thread and PlayerCore backgroundQueue used for loading subtitles by adding the `Atomic` property wrapper to properties shared between these threads and using the `withLock` method. Similar changes correct the data races between the main thread and PlayerCore thumbnailQueue.

These changes will also fix data races with the `playlistTotalLengthIsReady` and `playlistTotalLength` properties.

- [x] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)
- [x] This implements/fixes issue #3829, #3830, #3936, #3937.

---

**Description:**

PR #4830, opened 2024-02-23
Fix ctrl-s (screenshot window) does not work, #4822

### LINKED ISSUE #4822: Ctrl-s (screenshot window) does not work
**System and IINA version:**

- macOS 13.6.3
- IINA 1.3.4

**Expected behavior:**
Using the `mpv Default` key bindings and pressing the ⌃s key takes a screenshot of IINA's window.

**Actual behavior:**
Nothing happens when the ⌃s key is pressed. No screenshot is generated.

Checking the `mpv` log shows the [screenshot window](https://mpv.io/manual/stable/#command-interface-screenshot-%3Cflags%3E) command is failing:

```text
[  34.592][d][cplayer] Run command: screenshot, flags=73, args=[flags="window", legacy="unused"]
[  34.630][e][cplayer] Taking screenshot failed.
```

**Steps to reproduce:**
- Configure IINA key bindings to use `mpv Default`
- Start playing a video
- Press ⌃s

Every time.

low-batt:
This commit will change the `MPVController.mpvInitRendering` method to enable advanced control when creating the mpv renderer context. Advanced control must be enabled for this variation of the screenshot command to work.

- [x] This implements/fixes issue #4822.

---

**Description:**

### DIFF
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -478,6 +478,12 @@ not applying FFmpeg 9599 workaround
     chkErr(setString(MPVOption.Video.gpuHwdecInterop, "auto"))
   }
 
+  /// Initialize the `mpv` renderer.
+  ///
+  /// This method creates and initializes the `mpv` renderer and sets the callback that `mpv` calls when a new video frame is available.
+  ///
+  /// - Note: Advanced control must be enabled for the screenshot command to work when the window flag is used. See issue
+  ///         [#4822](https://github.com/iina/iina/issues/4822) for details.
   func mpvInitRendering() {
     guard let mpv = mpv else {
       fatalError("mpvInitRendering() should be called after mpv handle being initialized!")
@@ -486,14 +492,16 @@ not applying FFmpeg 9599 workaround
     var openGLInitParams = mpv_opengl_init_params(get_proc_address: mpvGetOpenGLFunc,
                                                   get_proc_address_ctx: nil)
     withUnsafeMutablePointer(to: &openGLInitParams) { openGLInitParams in
-      // var advanced: CInt = 1
-      var params = [
-        mpv_render_param(type: MPV_RENDER_PARAM_API_TYPE, data: apiType),
-        mpv_render_param(type: MPV_RENDER_PARAM_OPENGL_INIT_PARAMS, data: openGLInitParams),
-        // mpv_render_param(type: MPV_RENDER_PARAM_ADVANCED_CONTROL, data: &advanced),
-        mpv_render_param()
-      ]
-      chkErr(mpv_render_context_create(&mpvRenderContext, mpv, &params))
+      var advanced: CInt = 1
+      withUnsafeMutablePointer(to: &advanced) { advanced in
+        var params = [
+          mpv_render_param(type: MPV_RENDER_PARAM_API_TYPE, data: apiType),
+          mpv_render_param(type: MPV_RENDER_PARAM_OPENGL_INIT_PARAMS, data: openGLInitParams),
+          mpv_render_param(type: MPV_RENDER_PARAM_ADVANCED_CONTROL, data: advanced),
+          mpv_render_param()
+        ]
+        chkErr(mpv_render_context_create(&mpvRenderContext, mpv, &params))
+      }
       openGLContext = CGLGetCurrentContext()
       mpv_render_context_set_update_callback(mpvRenderContext!, mpvUpdateCallback, mutableRawPointerOf(obj: player.mainWindow.videoView.videoLayer))
     }


low-batt:

lhc70000:
I remember that it was disabled on purpose. We need to take a further look.

low-batt:
YES, this change needs a close look. This was disabled as a fix for issue #2149. In digging through old `mpv` commits I thought I saw some `mpv` fixes that might have fixed the hang. I did confirm those fixes were made after this workaround was added. There is a lot of discussion in the associated issue.

We should definitely merge PR #4819 first (as long as that one passes review). This PR corrects many data races detected by TSAN by moving work off of the `MPVController` queue and on to the main thread. As a lot of that is small I thought it better to access the data only from the main thread rather than adding locks to coordinate access.

krackers:
Did we fix the threading issues that exist in iina when advanced rendering is enabled? That would seem to be a preqreq for this @low-batt

low-batt:
@krackers Please remind me what IINA is getting wrong with threading.

krackers:
@low-batt it's discussed in https://github.com/iina/iina/issues/4822#issuecomment-1996046507 -- basically enabling advanced mode will cause libmpv to deadlock in certain cases because iina does not follow the rule that 1) `mpv_render_update` is always called shortly after an update is requested and 2) that the render thread never makes use of certain properties like `vo-passes`. (2) might never have been an issue, not too sure.

With advanced mode now enabled, we are back to the state that caused https://github.com/iina/iina/issues/2149 , namely that it is possible for iina to stop servicing the libmpv thread (e.g. if the window is offscreen).

low-batt:
Time to track this down. I will see if I can trigger the hang.

low-batt:
@krackers Big picture… The commits currently being merged are for the long delayed feature release. The release plan includes beta releases. In addition to features we are merging changes considered to be a bit risky for a maintenance release, namely changes involving threads. Just FYI, a PR currently being reviewed addresses many TSAN detect data race conditions. One of the changes in that PR is to change the thread in `MPVController` that reads mpv events to queue more of the processing of those events to the main thread. Currently IINA will make calls to get mpv properties while processing events on the event thread. This change avoids the need for locks as much of the data being modified is normally only accessed by the main thread.

I did some testing with IINA being moved "off screen" while playing by switching spaces. I was not able to trigger a hang. I decided to desk check the code regarding the theory that this might have to do with `canDraw` not being called. After comparing IINA to mpv I think we can rule out `canDraw`.

mpv `GLLayer`:
```swift
    override func canDraw(inCGLContext ctx: CGLContextObj,
                          pixelFormat pf: CGLPixelFormatObj,
                          forLayerTime t: CFTimeInterval,
                          displayTime ts: UnsafePointer<CVTimeStamp>?) -> Bool {
        if inLiveResize == false {
            isAsynchronous = false
        }
        return cocoaCB.backendState == .initialized &&
               (forceDraw || libmpv.isRenderUpdateFrame())
    }
```

IINA `ViewLayer`:
```swift
  override func canDraw(inCGLContext ctx: CGLContextObj, pixelFormat pf: CGLPixelFormatObj, forLayerTime t: CFTimeInterval, displayTime ts: UnsafePointer<CVTimeStamp>?) -> Bool {
    if forceRender { return true }
    return videoView.player.mpv.shouldRenderUpdateFrame()
  }
```

So neither will call mpv if drawing is being forced. As support for `isAsynchronous` was just added to IINA we need to ask why IINA does not need to ensure it is `false` as mpv does. @uiryuu Thoughts?

mpv `LibmpvHelper`:
```swift
    func isRenderUpdateFrame() -> Bool {
        uninitLock.lock()
        if mpvRenderContext == nil {
            uninitLock.unlock()
            return false
        }
        let flags: UInt64 = mpv_render_context_update(mpvRenderContext)
        uninitLock.unlock()
        return flags & UInt64(MPV_RENDER_UPDATE_FRAME.rawValue) > 0
    }
```

IINA MPVController:
```swift
  func shouldRenderUpdateFrame() -> Bool {
    guard let mpvRenderContext = mpvRenderContext else { return false }
    let flags: UInt64 = mpv_render_context_update(mpvRenderContext)
    return flags & UInt64(MPV_RENDER_UPDATE_FRAME.rawValue) > 0
  }
```

So the only difference here is the use of locks for shutdown protection. I will need to look into that, but I don't think it has anything to do with the issue at hand.

I'm more suspicious of the other concern raised in #4822, not calling `mpv_render_context_update` because the layer is "blocked":

IINA MPVController:
```swift
fileprivate func mpvUpdateCallback(_ ctx: UnsafeMutableRawPointer?) {
  let layer = bridge(ptr: ctx!) as ViewLayer
  guard !layer.blocked else { return }

  layer.mpvGLQueue.async {
    layer.draw()
  }
}
```
IINA `ViewLayer`:
```Swift
  func suspend() {
    blocked = true
    mpvGLQueue.suspend()
  }
```

This sure seems like it violates mpv's requirements that `mpv_render_context_update` must be called.

I am going to look into this next.

krackers:
>that this might have to do with canDraw not being called

Isn't the issue that the system (i.e. CAOpenGLLayer internals) skips the entire `canDraw + draw` calls when the window is offscreen? Upstream mpv handles this case by setting a flag in `draw` and having a fallback: https://github.com/mpv-player/mpv/blob/master/video/out/mac/gl_layer.swift#L231-L232

There's actually another subtler issue with doing the mpv render queue processing inside `canDraw` (or draw) because it means that in async mode we do not process the queue immediately but wait for the system to call us, even when we don't have any such constraint because the queue processing work uses an offscreen FBO and does not need to display anything. In practice it does not matter too much because osx does not support opengl 4.4./azdo/direct rendering anyway, but if it did this would bottleneck decoding speed to the presentation speed which can lead to framedropping.

>isAsynchronous was just added to IINA we need to ask why IINA does not need to ensure it is false as mpv does

The way mpv does it is a bit weird and shouldn't be strictly necessary, I think the place it's done was selected arbitrarily. uiryuu's PR set isAsynchronous to false in a different place which is fine.

uiryuu:
> why IINA does not need to ensure it is false as mpv does.

The code you showed is how mpv revert the changes to `isAsynchronous` when the live resizing session ended. mpv also programmatically enable it when live resizing starts:

https://github.com/mpv-player/mpv/blob/cd1b63f628a99d16183961056bb8177511a85888/video/out/mac/gl_layer.swift#L93-L100

low-batt:
@krackers I will look closer into what mpv is doing. When you say "when the window is offscreen" what exactly do you mean? When I try and move a window off the screen macOS moves it back.

krackers:
Window is occluded.

low-batt:
Pretty busy today, but I was able to do some testing. Instead of triggering a hang, I triggered a crash in `shouldRenderUpdateFrame` the method shown above that differed from mpv in that mpv has a lock for coordinating shutdown. Seems IINA may need that as well. I will be looking into that.

I will try some more testing that occludes the IINA window.

low-batt:
Success. I managed to get IINA to hang. Too tired to think. Tuesday busy, but will investigate more Wednesday.
[Sample.txt](https://github.com/user-attachments/files/15964555/Sample.txt)

svobs:
> Window is occluded.

@krackers we talked about this, and looks like this case is handled in IINA pretty similarly to how it's handled in mpv, by checking for whether `needsMPVRender` was modified after the call to `display()`, and if not, doing a skip render. The skip renders should keep things synced in the occluded state.

There is one difference I see though, which is that IINA is missing a check to `shouldRenderUpdateFrame()` prior to the skip render:

```Swift
            if libmpv.isRenderUpdateFrame() { // <-- Not in IINA
                libmpv.drawRender(NSSize.zero, bufferDepth, cglContext, skip: true)
            }
```

I can't remember whether this check is important? I think not really, right? Although it's also trivial to add to the IINA code.

As for the crash during quit in `shouldRenderUpdateFrame`, that looks like a trivial fix. I made a comment in the issue @low-batt filed for that.

So everything looks accounted for except for the hang.

low-batt:
@svobs  Not important. _Critical_. This is the source of the hang. It is `isRenderUpdateFrame()` in mpv and `shouldRenderUpdateFrame()` in IINA that calls `mpv_render_context_update`. The requirement that `mpv_render_context_update` be called is rule 1 in @krackers comment [above](https://github.com/iina/iina/pull/4830#issuecomment-2184316786). What was confusing is that IINA has the equivalent fallback @krackers pointed to in [this comment](https://github.com/iina/iina/pull/4830#issuecomment-2185332334), it is just that it is slightly incorrect as you pointed out. I will post of fix for this today.

PR #4833, opened 2024-02-24
Fixes & improvements to Subtitle Chooser

svobs:
- [x] This implements/fixes issue #4057.

---

**Description:**
Like #4825 but makes additional changes:
- The message `osd.sub_found`, previously set to `%d subtitles found. Downloading…`, was not exactly appropriate as the heading for the list of subtitles to download. It looks like this was done because it was used both for the subtitle list & for the actual "downloading" message. So I changed the existing `osd.sub_found` message to remove the ` Downloading…` part, added a new `osd.sub_downloading` message whose English version is `Downloading %d subtitles`, and changed the "downloading" message to use this key instead.
- Improve text size of sub-heading and table items slightly.
- Change table style & border so it looks like [this mock-up](https://github.com/iina/iina/issues/4057#issuecomment-1949629135) from @Iiii-I-I-I. Also implemented some other changes to displayed attributes based on said mock-up:
- Remove time from upload date (seems rare that subtitles are uploaded more than once a day).
- Remove rating (seems to be 0.0 most of the time?) - am open to bringing this back.
- Add release year in parentheses.
- Add up to 2 fraction digits to `fps` attribute's displayed value.

### DIFF
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -88,7 +88,8 @@
 "osd.find_online_sub" = "Finding online subtitles…";
 "osd.find_online_sub.source" = "from";
 "osd.sub_not_found" = "No subtitles found";
-"osd.sub_found" = "%d subtitles found. Downloading…";
+"osd.sub_found" = "%d subtitles found.";
+"osd.sub_downloading" = "Downloading %d subtitles";
 "osd.sub_downloaded" = "Subtitle downloaded";
 "osd.sub_saved" = "Subtitle saved";
 "osd.network_error" = "Network error";
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -281,6 +281,15 @@ extension BinaryInteger {
   }
 }
 
+// Formats a number to max 2 digits after the decimal, rounded, but will omit trailing zeroes, and no commas or other formatting for large numbers
+fileprivate let fmtDecimalMaxFractionDigits2: NumberFormatter = {
+  let fmt = NumberFormatter()
+  fmt.numberStyle = .decimal
+  fmt.usesGroupingSeparator = false
+  fmt.maximumFractionDigits = 2
+  return fmt
+}()
+
 extension FloatingPoint {
   func clamped(to range: Range<Self>) -> Self {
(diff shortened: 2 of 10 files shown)


svobs:
Squashed + rebased to fix merge errors + improved commit msg.

iina/Base.lproj/Localizable.strings
```diff
@@ -83,7 +83,8 @@
 "osd.find_online_sub" = "Finding online subtitles…";
 "osd.find_online_sub.source" = "from";
 "osd.sub_not_found" = "No subtitles found";
-"osd.sub_found" = "%d subtitles found. Downloading…";
+"osd.sub_found" = "%d subtitles found.";
+"osd.sub_downloading" = "Downloading %d subtitles";
```
lhc70000, line 92: Please also add this to en.lproj.

iina/Extensions.swift
```diff
@@ -290,6 +299,12 @@ extension FloatingPoint {
       return self
     }
   }
+
+  /// Formats as String, rounding the number to 2 digits after the decimal
+  var stringMaxFrac2: String {
```
lhc70000: I suggest using a name like `stringWithMaxFractionDigit2` for better readability.

iina/OSDMessage.swift
```diff
@@ -69,6 +69,7 @@ enum OSDMessage {
 
   case startFindingSub(String)  // sub source
   case foundSub(Int)
+  case downloadingSub(Int, String)  // download count, ssub source
```
lhc70000: sub

lhc70000 (changes requested):

iina/Extensions.swift
svobs: Ah yeah. For utility functions like this which might be commonly used, I worry about them getting too long. But, just semantics. Done

iina/OSDMessage.swift
svobs: Thanks

svobs:
Rebased + pushed an update with all the suggested changes and more detailed commit message.

lhc70000:
Merged via 1f87c39.

low-batt:

PR #4853, opened 2024-03-19
Expand IINA screenshot functionality to screenshot flags

svobs:
- [x] This implements/fixes issue #4582 (or the second part of it as [mentioned here](https://github.com/iina/iina/issues/4582#issuecomment-1926200685)).

---

**Description:**

### Current functionality

IINA's `Playback` > `Take a Screenshot` menu item (I'll call it "IINA's screenshot") utilizes mpv's `screenshot` command (i.e. "mpv's screenshot") as its backbone, but differs in the following ways:
1. Displays a screenshot thumbnail using the OSD (if configured in prefs).
2. Uses the value of `screenshotIncludeSubtitle` pref (bool) to decide between `subtitles` or `video` flags to send to `screenshot`.
3. Can be configured to store to clipboard instead of file, or in addition to file, or...disabled entirely (is this good...?).

If a key binding is mapped to `screenshot` with no flags, then it is matched to the menu item, and so it will call IINA's screenshot. However, if any flags are supplied (e.g., `screenshot subtitles`), it will not match and will end up calling mpv's screenshot.

### PR functional changes

- Expands functionality of `PlayerCore.screenshot()` to take an optional key binding argument, and adds a call to it from `PlayerWindowController.handleKeyBinding` so it can handle `screenshot` actions.
- Adds logic so that if flags are specified by the key binding, they will be used (if possible - see caveat 1 below). If no flags are specified, then the value for pref key `screenshotIncludeSubtitle` will be used to determine the flags.

### Caveats

1. Does not add IINA screenshot support for `each-frame` flag. If this flag is present, mpv's screenshot will be used instead (sorry - maybe some future PR).
2. It *does* add handling for `screenshot window`, although the feature itself will continue to error out until the fix for #4822 is merged.
3. IINA's screenshot deviates slightly from the spec for mpv's screenshot, and by its nature this PR will make this unavoidable for more screenshot cases. As noted in the code comments:
    i. If the stored values for `Preference.Key.screenshotSaveToFile` and `Preference.Key.screenshotCopyToClipboard` are set to false, all screenshot commands will be ignored. It may not be obvious to the end user why their screenshot seems to be getting ignored. Although no one seems to have filed any issues about it.
    ii. When no flags are given with `screenshot`, the mpv spec says that default is `subtitles` flag. But IINA will use either `subtitles` or `video`, depending on the value for `Preference.Key.screenshotIncludeSubtitle`.

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -770,12 +770,6 @@ class PlayerCore: NSObject {
     }
   }
 
-  func isScreenshotEnabled() -> Bool {
-    let saveToFile = Preference.bool(for: .screenshotSaveToFile)
-    let saveToClipboard = Preference.bool(for: .screenshotCopyToClipboard)
-    return saveToFile || saveToClipboard
-  }
-
   /// Takes a screenshot, attempting to augment mpv's `screenshot` command with additional functionality & control, for example
   /// the ability to save to clipboard instead of or in addition to file, and displaying the screenshot's thumbnail via the OSD.
   /// Returns `true` if a command was sent to mpv; `false` if no command was sent.
@@ -797,10 +791,19 @@ class PlayerCore: NSObject {
   /// `Preference.Key.screenshotIncludeSubtitle` to decide between `subtitles` or `video`.
   @discardableResult
   func screenshot(fromKeyBinding keyBinding: KeyMapping? = nil) -> Bool {
-    guard isScreenshotEnabled() else {
-      Logger.log("Ignoring screenshot request\(keyBinding == nil ? "" : " from key binding") because all forms of screenshots are disabled in prefs")
+    let saveToFile = Preference.bool(for: .screenshotSaveToFile)
+    let saveToClipboard = Preference.bool(for: .screenshotCopyToClipboard)
+    guard saveToFile || saveToClipboard else {
+      Logger.log("Ignoring screenshot request: all forms of screenshots are disabled in prefs")
       return false
     }
+
+    guard let vid = info.vid, vid > 0 else {
+      Logger.log("Ignoring screenshot request: no video stream is being played")
+      return false
+    }
+
+    Logger.log("Screenshot requested by user\(keyBinding == nil ? "" : " (rawAction: \(keyBinding!.rawAction))")")
 
     var commandFlags: [String] = []
 
@@ -808,7 +811,7 @@ class PlayerCore: NSObject {
       var canUseIINAScreenshot = true
 
       guard let commandName = keyBinding.action.first, commandName == MPVCommand.screenshot.rawValue else {
-        Logger.log("Cannot take screenshot: first token in key binding action is unexpected: \(keyBinding.action)")
+        Logger.log("Cannot take screenshot: unexpected first token in key binding action: \(keyBinding.rawAction)", level: .error)
         return false
       }
       if keyBinding.action.count > 1 {
@@ -836,8 +839,6 @@ class PlayerCore: NSObject {
       }
     }
 
-    guard let vid = info.vid, vid > 0 else { return false }  // TODO: why this is needed?
-
     if commandFlags.isEmpty {
       let includeSubtitles = Preference.bool(for: .screenshotIncludeSubtitle)
       commandFlags.append(includeSubtitles ? "subtitles" : "video")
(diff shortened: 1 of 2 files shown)


low-batt:
I'm busy on Tuesdays, so I only had a chance to read through the description and peek at the code. Seems like the right approach. I will take a closer look soon.

iina/PlayerCore.swift
```diff
@@ -770,15 +770,81 @@ class PlayerCore: NSObject {
...
+  /// used. If `keyBinding` is not present or its command has no flags, the value for `Preference.Key.screenshotIncludeSubtitle` will
+  /// be used to determine the flags:
+  /// - If `true`, the command `screenshot subtitles` will be sent to mpv.
+  /// - If `false`, the command `screenshot video` will be sent to mpv.
+  ///
+  /// Note: IINA overrides mpv's behavior in some ways:
+  /// 1. As noted above, if the stored values for `Preference.Key.screenshotSaveToFile` and `Preference.Key.screenshotCopyToClipboard` are
+  /// set to false, all screenshot commands will be ignored.
+  /// 2. When no flags are given with `screenshot`: instead of defaulting to `subtitles` as mpv does, IINA will use the value for
+  /// `Preference.Key.screenshotIncludeSubtitle` to decide between `subtitles` or `video`.
```
low-batt, line 791: The documentation comment is very helpful!

If you highlight the function name and then under the Xcode `Editor` menu open the `Structure` sub-menu and click on `Add Documentation` then Xcode will add a template with the appropriate documentation tags. In the case of `screenshot` method it will add:
```swift
  /// <#Description#>
  /// - Parameter keyBinding: <#keyBinding description#>
  /// - Returns: <#description#>
```

iina/PlayerCore.swift
```diff
@@ -770,15 +770,81 @@ class PlayerCore: NSObject {
...
+  @discardableResult
+  func screenshot(fromKeyBinding keyBinding: KeyMapping? = nil) -> Bool {
+    guard isScreenshotEnabled() else {
+      Logger.log("Ignoring screenshot request\(keyBinding == nil ? "" : " from key binding") because all forms of screenshots are disabled in prefs")
```
low-batt: This log message is not entirely correct. If you disable the screenshots settings and then press `s` key the log message will not indicate the request was triggered "from key binding".

iina/PlayerCore.swift
```diff
@@ -770,15 +770,81 @@ class PlayerCore: NSObject {
...
+        }
+      }
+
+      if !canUseIINAScreenshot {
+        let returnValue = mpv.command(rawString: keyBinding.rawAction)
+        return returnValue == 0
+      }
+    }
+
+    guard let vid = info.vid, vid > 0 else { return false }  // TODO: why this is needed?
```
low-batt: Always good to emit a log message when ignoring input from the user like was done at the start of the method indicating screenshots are disabled. If that had been done in the past for this statement not only would it be possible to tell what happened from the log file, it also would have made obvious what this code is for. I suggest changing this code to something like this:
```swift
    let fromKey = keyBinding == nil ? "" : " from key binding"
    guard let vid = info.vid, vid > 0 else {
      Logger.log("Ignoring screenshot request\(fromKey) because no video stream is being played")
      return false
    }
```
And moving it right to the start of the method.

I believe the thinking was that it is better to ignore the request than interrupt the user with the alert that says "Cannot take a screenshot".

To trigger this play a `mp3` and press the `s` key.

low-batt (changes requested):
That guard statement that is checking `vid` needs to be moved to the start of the method. Also it is confusing that the `s` key is not treated as a key binding.

Otherwise, it worked fine in my tests.

svobs:
> I believe the thinking was that it is better to ignore the request than interrupt the user with the alert that says "Cannot take a screenshot".

That makes sense.

I pushed another commit with the suggestions, although I removed the talk about key bindings from the error messages (weren't really useful) but added a couple extra lines of logging to facilitate future troubleshooting.

low-batt (approved):
Pulled latest commits, built and tested.
Changes look good to me.

uiryuu (approved):
Tested and looks good. 

I think we should prevent `Preference.Key.screenshotSaveToFile` and `Preference.Key.screenshotCopyToClipboard` from both being `false`. I will consider making a PR to make these 2 options radio buttons so that only one of them can be selected. And also combining these 2 preference keys into 1 so that the user cannot set them both to false in via commands either.

PR #4860, opened 2024-03-30
fix: avoid double percent-encode for url scheme

### LINKED ISSUE #4861: Incorrect percent encoding under Sonoma
**System and IINA version:**

- macOS 14.4
- IINA 1.3.4

**Expected behavior:**
IINA correctly percent encodes URLs.

**Actual behavior:**
When running under macOS Sonoma IINA will perform percent encoding _twice_ resulting in malformed URLs. @xfoxfu provided two examples:

Opening this URL in Firefox:
```text
iina://weblink?url=https://example.com/%5Bfoo%20bar.mkv
```

IINA changes `%5Bfoo%20bar.mkv` to be `%5Bfoo%2520bar.mkv`:
```text
13:54:42.969 [iina][d] Parsing URL iina://weblink?url=https://example.com/%5Bfoo%20bar.mkv
13:54:42.969 [player0][d] Open URL: https://example.com/%5Bfoo%2520bar.mkv
13:54:42.969 [player0][d] Opening https://example.com/%5Bfoo%2520bar.mkv in main window
```

Opening this URL in Firefox:
```text
iina://weblink?url=https://example.com/%25foo%20bar.mkv
```

IINA changes `%25foo%20bar.mkv` to be `%25foo%2520bar.mkv`:
```text
13:51:00.896 [iina][d] Parsing URL iina://weblink?url=https://example.com/%25foo%20bar.mkv
13:51:00.897 [player0][d] Open URL: https://example.com/%25foo%2520bar.mkv
13:51:00.897 [player0][d] Opening https://example.com/%25foo%2520bar.mkv in main window
```

This is a change from the behavior under macOS Ventura.

**Steps to reproduce:**
- Enable logging in IINA
- Open one of the URLs given above in Firefox
- Look for the message shown above in IINA's log file

This is an issue with IINA code. I have not checked if `mpv` also has this problem.

Every time.

xfoxfu:

---

**Description:**

Remove manual percent-encoding for URL opened in URL scheme. This will make URLs with special characters no longer be encoded twice and thus openable.

Due to a recent change in Apple `URL.init`, URL will be encoded automatically, see: <https://developer.apple.com/documentation/foundation/url/3126806-init>

Therefore, if the URL is percent-encoded beforehand, it will be encoded a second time if `URL` is constructed.

```
URL Scheme:
iina://weblink?url=https://******/d/Anime/no%20Subarashii%20Sekai%20ni%20Bakuen%20wo!.mkv
Log:
16:01:57.099 [player0][d] Open URL: https://******/d/Anime/Kono%2520Subarashii%2520Sekai%2520ni%2520Bakuen%2520wo!.mkv
URL-decoded:
16:01:57.099 [player0][d] Open URL: https://******/d/Anime/Kono%20Subarashii%20Sekai%20ni%20Bakuen%20wo!.mkv

Intended URL:
16:01:19.708 [player0][d] Open URL: https://******/d/Anime/Kono Subarashii Sekai ni Bakuen wo!.mkv
```

This PR removes the first manual percent-encoding, which makes URL correct.

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -360,8 +360,26 @@ class PlayerCore: NSObject {
     if str.first == "/" {
       openURL(URL(fileURLWithPath: str))
     } else {
-      guard let url = URL(string: str) else {
-        Logger.log("Cannot parse url for \(str)", level: .error, subsystem: subsystem)
+      // For apps built with Xcode 15 or later the behavior of the URL initializer has changed when
+      // running under macOS Sonoma or later. The behavior now matches URLComponents and will
+      // automatically percent encode characters. Must not apply percent encoding to the string
+      // passed to the URL initializer if the new new behavior is active.
+      var performPercentEncoding = true
+#if compiler(>=5.9)
+      if #available(macOS 14, *) {
+        performPercentEncoding = false
+      }
+#endif
+      var pstr = str
+      if performPercentEncoding {
+        guard let encoded = str.addingPercentEncoding(withAllowedCharacters: .urlAllowed) else {
+          print("Cannot add percent encoding for \(str)")
+          return
+        }
+        pstr = encoded
+      }
+      guard let url = URL(string: pstr) else {
+        Logger.log("Cannot parse url for \(pstr)", level: .error, subsystem: subsystem)
         return
       }
       openURL(url)
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -360,8 +360,8 @@ class PlayerCore: NSObject {
     if str.first == "/" {
       openURL(URL(fileURLWithPath: str))
     } else {
-      guard let pstr = str.addingPercentEncoding(withAllowedCharacters: .urlAllowed), let url = URL(string: pstr) else {
-        Logger.log("Cannot add percent encoding for \(str)", level: .error, subsystem: subsystem)
+      guard let url = URL(string: str) else {
+        Logger.log("Cannot parse url for \(str)", level: .error, subsystem: subsystem)
         return
       }
       openURL(url)


low-batt:
How do I reproduce the problem this PR fixes?

Running under macOS 14.4 I opened the following URL in Firefox:
```text
iina://weblink?url=https://example.com/d/Anime/Kono%20Subarashii%20Sekai%20ni%20Bakuen%20wo!.mkv
```

And I'm not seeing double percent encoding in IINA's log:
```text
14:26:02.770 [iina][d] URL event: iina://weblink?url=https://example.com/d/Anime/Kono%20Subarashii%20Sekai%20ni%20Bakuen%20wo!.mkv
14:26:02.770 [iina][d] Parsing URL iina://weblink?url=https://example.com/d/Anime/Kono%20Subarashii%20Sekai%20ni%20Bakuen%20wo!.mkv
14:26:02.771 [player0][d] Open URL: https://example.com/d/Anime/Kono%20Subarashii%20Sekai%20ni%20Bakuen%20wo!.mkv
14:26:02.771 [player0][d] Opening https://example.com/d/Anime/Kono%20Subarashii%20Sekai%20ni%20Bakuen%20wo!.mkv in main window
…
14:26:06.447 [mpv][e] [stream] error: Failed to open https://example.com/d/Anime/Kono%20Subarashii%20Sekai%20ni%20Bakuen%20wo!.mkv.
```

xfoxfu:
Interesting, it seems only having whitespaces in the URL is not sufficient to trigger the bug. Please try this URL:

<iina://weblink?url=https://example/%25foo%20bar.mkv>

```console
03:47:28.786 [iina][d] URL event: iina://weblink?url=https://example/%25foo%20bar.mkv
03:47:28.786 [iina][d] Parsing URL iina://weblink?url=https://example/%25foo%20bar.mkv
03:47:28.786 [player0][d] Open URL: https://example/%25foo%2520bar.mkv
03:47:28.786 [player0][d] Opening https://example/%25foo%2520bar.mkv in main window
```

And this one: <iina://weblink?url=https://example/%5Bfoo%20bar.mkv>

```console
03:49:51.240 [iina][d] URL event: iina://weblink?url=https://example/%5Bfoo%20bar.mkv
03:49:51.240 [iina][d] Parsing URL iina://weblink?url=https://example/%5Bfoo%20bar.mkv
03:49:51.240 [player0][d] Open URL: https://example/%5Bfoo%2520bar.mkv
03:49:51.240 [player0][d] Opening https://example/%5Bfoo%2520bar.mkv in main window
```

low-batt:
Thanks for the update. Reproduced for me now.

Very busy today. I will be taking a close look into this problem.

iina/PlayerCore.swift
```diff
@@ -360,8 +360,8 @@ class PlayerCore: NSObject {
     if str.first == "/" {
       openURL(URL(fileURLWithPath: str))
     } else {
-      guard let pstr = str.addingPercentEncoding(withAllowedCharacters: .urlAllowed), let url = URL(string: pstr) else {
-        Logger.log("Cannot add percent encoding for \(str)", level: .error, subsystem: subsystem)
+      guard let url = URL(string: str) else {
+        Logger.log("Cannot parse url for \(str)", level: .error, subsystem: subsystem)
```
low-batt:
```suggestion
      // For apps built with Xcode 15 or later the behavior of the URL initializer has changed when
      // running under macOS Sonoma or later. The behavior now matches URLComponents and will
      // automatically percent encode characters. Must not apply percent encoding to the string
      // passed to the URL initializer if the new new behavior is active.
      var performPercentEncoding = true
#if compiler(>=5.9)
      if #available(macOS 14, *) {
        performPercentEncoding = false
      }
#endif
      var pstr = str
      if performPercentEncoding {
        guard let encoded = str.addingPercentEncoding(withAllowedCharacters: .urlAllowed) else {
          print("Cannot add percent encoding for \(str)")
          return
        }
        pstr = encoded
      }
      guard let url = URL(string: pstr) else {
        Logger.log("Cannot parse url for \(pstr)", level: .error, subsystem: subsystem)
```

low-batt (changes requested):
Excellent work tracking this problem to Apple's change. I really appreciate your investigation and proposed fix.

The call to `addingPercentEncoding` can't be fully removed because the new behavior is only present when running under macOS Sonoma and only if built with Xcode 15. Look for the suggested change.

I've created issue #4861 for this problem. See that issue for a full discussion. There are other places in the code that probably need to be updated. We'd want to be able to reproduce a problem before making changes.

xfoxfu:
Thanks for your suggestions, and I have tested on macOS Ventura with XCode 15.2 that the problem has been resolved, but I do not have access to older macOS or XCode.

uiryuu (approved):
Tested this PR on two machines:
1. MacOS Sonoma w/ Xcode 15.2
2. MacOS Monterey w/ Xcode 13.4.1

IINA used for testing was compiled locally on the corresponding machine. The `#if compiler(>=5.9)` check works as intended. And confirmed the proposed fix doesn't break when IINA is built by older version Xcode.

However, during testing, I found that this URL: `iina://weblink?url=https://example/%25foo%20bar.mkv` cannot be parsed on the older machine. Because we allowed `.urlAllowed`, which is customized in here:

https://github.com/iina/iina/blob/3f02bf0b37e9f13be8dcebc570dbf6b2b0cdad62/iina/Extensions.swift#L451C1-L462C2

This is where we manually added "%" into the CharSet, so that the `pstr` becomes `https://example/%foo%20bar.mkv` after `addingPercentEncoding` and thus invalid for further parsing. Not sure why we manually added "%" before. 

This is *not* caused by this PR, so this one is safe to merge.

xfoxfu:
Actually, `iina://weblink?url=https://example/%25foo%20bar.mkv` is not leading to a valid URL by RFC 3986. The `url` query parameter will be `https://example/%foo bar.mkv` where the `%` is not properly encoded to `%25`.

I think allowing `%` in URL has practical use since the URL could be `https://example.com/%2520.mkv`, where the filename is `%20.mkv` (not ` .mkv`), and in this case, the `%` before `2520` should not be encoded to preserve `%25` as a whole.

A better way of handling these "endless" corner cases easily could be utilizing the new `URL` behavior, while backporting a similar feature to older platforms which does not support `URL(string: encodingInvalidCharacters:)`.

low-batt (approved):
I pulled the latest commits and tested under Ventura and Sonoma.
Changes look good to me.

low-batt:
I've created issue #4862 for the problem with `%`. We should continue the discussion in that issue.

@xfoxfu A final big THANK YOU for investigating, reporting and fixing this problem. Very timely has we are moving closer to a release that fixes regressions. Would have been bad to miss this one.

PR #4871, opened 2024-04-07
Fix playback history window doesn't respect reduce motion, #4870

### LINKED ISSUE #4870: Playback history window doesn't respect reduce motion
**System and IINA version:**

- macOS 13.6.3
- IINA 1.3.4

**Expected behavior:**
IINA reduces animations if the user has enabled the macOS [Reduce motion](https://support.apple.com/guide/mac-help/reduce-screen-motion-mchlc03f57a1/mac) setting. From Apple's Human Interface Guidelines section [Motion](https://developer.apple.com/design/human-interface-guidelines/accessibility#Motion):
> **Play tightened animations when Reduce Motion is on.** People can turn on Reduce Motion if they tend to get distracted or experience dizziness or nausea when viewing animations that include effects such as zooming, scaling, spinning, or peripheral motion. In response to this setting, you need to turn off or reduce animations that are known to cause problems (to learn more, see [Responsive Design for Motion](https://webkit.org/blog/7551/responsive-design-for-motion/)).

**Actual behavior:**
Disclosure triangles in the `Playback History` window use a sliding animation when showing and hiding information.

**Steps to reproduce:**
- Start IINA
- Click on  `Playback History` under the `File` menu
- The `Playback History` window appears
- Click on the triangle next to one of the dates 
- The nested list of videos played on that date collapses into the date with a sliding animation

Every time.

low-batt:
This commit will change IINA to adhere to the Apple [Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/) with respects to the macOS [Reduced motion](https://developer.apple.com/design/human-interface-guidelines/accessibility#Motion) setting. With these changes IINA will reduce, not eliminate, animations when the macOS [Reduced motion](https://support.apple.com/guide/mac-help/stop-or-reduce-onscreen-motion-mchlc03f57a1/mac)  setting is enabled. When this setting is enabled IINA will fade in the sidebar instead of sliding it into view. All other changes to animation behavior, such as how IINA transitions in and out of full screen mode, are provide by AppKit.

For users who are hyper sensitive to motion this commit adds a `Disable animations` setting to the `UI` settings tab in a new `Accessibility` section at the bottom of the panel. Enabling this setting will eliminate many, but not all, animated transition effects.

- [x] This implements/fixes issue #4870.

---

**Description:**

### DIFF
--- a/iina.xcodeproj/project.pbxproj
+++ b/iina.xcodeproj/project.pbxproj
@@ -16,6 +16,7 @@
 		49542982216C34950058F680 /* ToolbarItemIcon.pdf in Resources */ = {isa = PBXBuildFile; fileRef = 49542981216C34950058F680 /* ToolbarItemIcon.pdf */; };
 		49542986216C34950058F680 /* OpenInIINA.appex in Embed App Extensions */ = {isa = PBXBuildFile; fileRef = 49542973216C34950058F680 /* OpenInIINA.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
 		496B19921E2968530035AF10 /* PIP.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 496B19911E2968530035AF10 /* PIP.framework */; };
+		512B8FDD2BC2376E00AF41BF /* OutlineView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 512B8FDC2BC2376E00AF41BF /* OutlineView.swift */; };
 		513A4FFA29B53F8100A8EA7D /* Atomic.swift in Sources */ = {isa = PBXBuildFile; fileRef = 513A4FF929B53F8100A8EA7D /* Atomic.swift */; };
 		515B5E5A2A579903001FCD49 /* iina-plugin in Copy Executables */ = {isa = PBXBuildFile; fileRef = E38558872A484A2D0083772D /* iina-plugin */; settings = {ATTRIBUTES = (CodeSignOnCopy, ); }; };
 		5196819A29EC963F00B05D55 /* CoreDisplay.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 5196819929EC963F00B05D55 /* CoreDisplay.framework */; };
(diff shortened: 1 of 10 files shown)


saagarjha (changes requested):
Reduce motion doesn't mean "turn off all animations", it means "avoid problematic transitions that are likely to cause people to feel disoriented". This animation provides important context to a user interaction and seems like a poor choice to disable unless it is causing issues for people.

low-batt:
Apple has recognized that sliding is causing issues for such people. The Human Interface Guidelines in the [Motion](https://developer.apple.com/design/human-interface-guidelines/accessibility#Motion) section provides this guidance:
> - Replace a slide with a fade to avoid motion

I couldn't find an easy way to make [NSOutlineView](https://developer.apple.com/documentation/appkit/nsoutlineview) first fade the item before making it just disappear. Is there a way to do that?

I could break the connection with the macOS reduce motion setting and create a `Use animations` checkbox in IINA's general settings in the behavior section. Should I implement that?

saagarjha:
IMO if you think this would be a problem, we should file a feedback and get Apple to implement it in NSOutlineView rather than trying to roll it ourselves.

low-batt:
I am going to move this to a draft until I have time to add an IINA setting for controlling animations.

It used to be that people sensitive to motion were using the `defaults` command to set macOS properties that controlled AppKit animations. I think it was starting with OS X Mountain Lion that some of these properties stopped working. More properties stopped working with the releases that followed. Since that time people have been giving Apple feedback on this. I'm expecting the feedback goes to the group that handles accessibility. Relative to the other kinds of impairments this group must address I would expect reducing motion sickness to be a low priority.

But recently I have noticed changes. My guess is that two things have brought some attention to this problem. The first was the need to support the CSS `prefers-reduced-motion` feature in Safari and Apple's web pages. The second is that with visionOS Apple has to worry about animations contributing to virtual reality sickness. Possibility this is why the accessibility portion of the HIG that covers motion was updated and enlarged. There is also now a separate motion section. In that section under [Best Practices](https://developer.apple.com/design/human-interface-guidelines/motion#Best-practices) it says:
> **Make motion optional.** There are several reasons why people might not experience movement in your app, so it’s essential to avoid using it as the only way to communicate important information. For example, people can turn on the Reduce Motion accessibility setting to minimize or eliminate animations. For guidance, see [Motion](https://developer.apple.com/design/human-interface-guidelines/accessibility#Motion).

I don't remember Apple ever mentioning _eliminate animations_ in the HIG before. I'm hoping Apple is starting to recognize the problems sliding and zooming trigger in some people.

It definitely would be good to get Apple to enhance `NSOutlineView`, but that won't provide a solution for older versions of macOS.

low-batt:
With `Reduced motion` now only causing IINA to replace the slide animation with a fade when showing/hiding the sidebar, I'm sure Apple would approve of IINA's behavior when this setting is enabled.

Reviewers should take a close look at the `showSideBar` and `hideSideBar` methods in `MainWindowController`. I had a hard time getting the animations to switch between sliding and fading. Is there a cleaner way to code this?

low-batt:
This is what the new setting looks like:
animation-setting

low-batt:
After thinking about the code some more I made one more fix.

svobs:
> Reviewers should take a close look at the showSideBar and hideSideBar methods in MainWindowController. I had a hard time getting the animations to switch between sliding and fading. Is there a cleaner way to code this?

Trying out the built product, the fade effect seems to be working as desired... But I see you're only manipulating `sideBarView.alphaValue` to do an explicit fade in `hideSideBar` and not `showSideBar`. Is there a reason you didn't use `alphaValue` for both? Looks like the fade is still happening for the latter case via an implicit animation which results from `sideBarView.isHidden` being toggled inside an animation group. I've found this mechanism to be less robust in general than when using `alphaValue`, though in this case it's probably safe enough because the layout changes being made are relatively simple.

low-batt:
I originally attempted to just use `alphaValue`. I had `showSideBar` set `alphaValue` to zero and then animate it to `1` to fade-in the sidebar. That did not work. The sidebar was fully opaque from the start. To get that to work I had to set `alphaValue` to zero outside of `showSideBar`. That complicated the code a lot. So I switched to animating `isHidden` to `false` to fade-in the sidebar. That worked. But animating `isHidden` to `true` did not fade-out the sidebar. It just immediately disappeared. I found others discussing this. They had to use  `alphaValue` to get the fade-out to work. I tried that and it worked. It makes no sense to me why `isHidden` works for fade-in but not for fade-out. I was hoping somebody would know a trick to get this to work in a cleaner way.

uiryuu:
I really don't think this is necessary. For one, I support @saagarjha's opinion. We really shouldn't mess `NSOutlineView` up by ourselves. With `reduce motion` enabled in the system settings, I don't observe one app (including Apple's apps, like Music, Xcode) removes the motion in the `NSOutlineView`. Besides, introducing another "enable animations" in the IINA settings, which only a very small number of users will use, along side the "reduce motion" already exists in the system accessibility setting, will cause further confusion to users.

low-batt:
Offline @uiryuu gave me some suggestions on how to make clear what this setting does to avoid confusing users.

The latest commit changes the section at the end of the `UI` tab to be `Accessibility`. The setting toggle is now `Disable animations`. There is a help button that links to the HIG [Motion](https://developer.apple.com/design/human-interface-guidelines/motion) section. The toggle has the following text under it:

When the macOS Reduce Motion accessibility setting is enabled IINA will reduce user interface animations that are known to cause problems for those with vestibular disorders. Should that be insufficient this setting can be used to reduce motion further by eliminating user interface animations.

It now looks like this:
accessibility

PR #4885, opened 2024-04-19
Make all volume icons indicate level

low-batt:
This commit will:
- Add a `volumeIcon` method to `PlayerWindowController` that returns the appropriate icon based on volume level
- Add an `updateVolume` method to `MainWindowController` that sets the icon on the mute button
- Change `MiniPlayerController.updateVolume` to set the icon on the mute button as well as the volume button

The mini window selects the image to use for the volume button from multiple images based on the volume level. This commit extends that functionality to the mute button in the main window and the mute button in the mini window. This makes the behavior of volume icons consistent.

---

**Description:**

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -2641,6 +2641,13 @@ class MainWindowController: PlayerWindowController {
     }
   }
 
+  override func updateVolume() {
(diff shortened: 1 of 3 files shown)


uiryuu (approved):
Tested and LGTM!

Issue #4468, opened 2023-06-17
Localization for "en.lproj" has typos which override "Base.lproj"


svobs:
- macOS 13.4
- IINA develop branch 39801567ac3c836e655f37041bf2af5f277cad69

**Expected behavior:**

English localizations should match the values in "Base" localization exactly.

**Actual behavior:**

Entries in `iina/en.lproj/Localizable.strings` are missing or incorrect relative to `iina/Base.lproj/Localizable.strings`, but because the "en" translation is present, it will always be used instead of the "Base", resulting in incorrect US English text.

Some examples:

| Correct text  | Actual visible text |
| ------------- | ------------- |
| "Add overlays on videos"  | "Add overlay on videos"  |
| "Error occurred during the installation:"  | "Error occured during the installation:"  |
| "Cannot connect"  | "osd.cannot_connect"  |
| "Timed Out"  | "osd.timed_out"  |

**Steps to reproduce:**
 
As @low-batt [pointed out](https://github.com/iina/iina/pull/4467#pullrequestreview-1484206035):
>The somewhat confusing IINA rules on localization require both the base and english localization files be updated. The rest are updated through [Crowdin](https://translate.iina.io/). So `Localizable.strings` under `en.lproj` also needs to be updated as a part of this PR.
> 
> I've been confused as to why we have to duplicate strings like this. Seems wrong. This PR got me to look into that. From [Enabling Base Internationalization](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/InternationalizingYourUserInterface/InternationalizingYourUserInterface.html#//apple_ref/doc/uid/10000171i-CH3-SW4):
>> Xcode modifies your project folder according to the selections you make in this dialog. Xcode creates a Base.lproj folder in your project folder and adds to it the resource files you select. Xcode creates a language folder for the development language but only adds resources that need translation to the folder. For example, if you select English as the development language, Xcode inserts the resource file in the Base.lproj project folder but not the en.lproj folder because the resource is already in English.
> 
> Checking `info.plist` confirms the development language is set to English:
```xml
<key>CFBundleDevelopmentRegion</key>
<string>en</string>
```
> From the Apple documentation it sounds like `en.lproj` is not needed.

I was able to confirm this by (1) selecting `Localizable.strings` in XCode's Navigator, and (2) deselecting `English`. When the confirmation dialog comes up, check the box to also delete the file on disk.
[...]

PR #4893, opened 2024-04-26
Fix subtitles override level not localized, #4892

### LINKED ISSUE #4892: Subtitles override level not localized
**System and IINA version:**

- macOS 13.6.3
- IINA 1.3.4

**Expected behavior:**
Where appropriate labels presented to the user are localized.

**Actual behavior:**
In IINA's settings on the `Subtitle` tab in the `ASS Subtitles` section the `Override level` setting is controlled by a slider and has a label that displays the slider value as a string. This value is always shown in English. This setting is tied to the `mpv` [sub-ass-override](https://mpv.io/manual/stable/#options-sub-ass-override) option:
> **--sub-ass-override=<yes|no|force|scale|strip>**
>    Control whether user style overrides should be applied. Note that all of these overrides try to be somewhat smart about figuring out whether or not a subtitle is considered a "sign".
> - no:	Render subtitles as specified by the subtitle scripts, without overrides.
> - yes:	Apply all the `--sub-ass-*` style override options. Changing the default for any of these options can lead to incorrect subtitle rendering (default).
> - force:	Like yes, but also force all `--sub-*` options. Can break rendering easily.
> - scale:	Like yes, but also apply `--sub-scale`.
> - strip:	Radically strip all ASS tags and styles from the subtitle. This is equivalent to the old `--no-ass` / `--no-sub-ass` options.
>
> This also controls some bitmap subtitle overrides, as well as HTML tags in formats like SRT, despite the name of the option.

IINA is missing support for the `scale` value. That will be a subject for a separate issue.

**Steps to reproduce:**
- Open the  and click on `System Settings…`
- Click on the `General` tab
- Under `General` click on `Language & Region`
- Scroll down to the `Applications` section
- Add IINA and select a language other than English
- Start IINA
- Click on `Settings…` (in the languge being used) under the `IINA` menu
- The settings panel appears
- On the left side of the panel click on `Subtitle`
- In the `ASS Subtitles` section look for the slider
- Move the slider and notice the value displayed is in English

Every time.

low-batt:
This commit will:
- Add key-value pairs for the yes, force and strip override levels to the base and en `Localizable.strings` files
- Change the `transformedValue` method in the `ASSOverrideLevelTransformer` class to return a localized string instead of the enum value string

This localizes the value displayed for the Override level setting found in the ASS Subtitles section on the Subtitle tab in IINA's settings.

- [x] This implements/fixes issue #4892.

---

**Description:**

### DIFF
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -116,6 +116,10 @@
 
 "preference.enable_adv_settings" = "Enable advanced settings";
 
+"preference.sub_override_level.force" = "force";
+"preference.sub_override_level.strip" = "strip";
+"preference.sub_override_level.yes" = "yes";
+
 "menu.volume" = "Volume: %d";
 "menu.volume_muted" = "Volume: %d (Muted)";
 "menu.audio_delay" = "Audio Delay: %.2fs";
--- a/iina/PrefSubViewController.swift
+++ b/iina/PrefSubViewController.swift
@@ -183,8 +183,16 @@ class PrefSubViewController: PreferenceViewController, PreferenceWindowEmbeddabl
   }
 
   override func transformedValue(_ value: Any?) -> Any? {
-    guard let num = value as? NSNumber else { return nil }
-    return Preference.SubOverrideLevel(rawValue: num.intValue)?.string
+    guard let num = value as? NSNumber,
+          let level = Preference.SubOverrideLevel(rawValue: num.intValue) else { return nil }
+    switch level {
+    case .yes:
+      return NSLocalizedString("preference.sub_override_level.yes", value: "yes", comment: "yes")
+    case .force:
+      return NSLocalizedString("preference.sub_override_level.force", value: "force", comment: "force")
+    case .strip:
+      return NSLocalizedString("preference.sub_override_level.strip", value: "strip", comment: "strip")
+    }
   }
 
 }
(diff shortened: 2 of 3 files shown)


low-batt:
One issue I did not address is that the text field holding the label is being sent to translators even though that value will never be used. Run this command to see the translations:
```text
grep -inr --include \*.strings 3H8-Ei-cTm *
```

I did not find a way in Xcode to exclude this text field from localization. Did I miss something?

Because Crowdin is not able to read an XIB file as discussed in #4468 we have to use the English strings file. Would deleting the entry for the object ID `3H8-Ei-cTm` from the English translations remove this translation from Crowdin?

Searching the net I found people using various methods, for example a reply to the stack overflow post [Marking some XIB/Storyboard strings as not localizable](https://stackoverflow.com/questions/27407282/marking-some-xib-storyboard-strings-as-not-localizable) suggests using as special phrase in the `Localizer Hint` XIB field and then using [XMLStarlet](https://en.wikipedia.org/wiki/XMLStarlet) to detect and delete elements from the XLIFF.

We might want to have some sort of `Localizer Hint` convention even when relying upon not putting the object ID into the English strings file to make it clear that a particular object should not be localized.

low-batt:
@uiryuu Don't forget about my question as to how we should deal with objects that should not be sent to translators.

uiryuu:
@low-batt Sorry, I didn't read the thread.

> I did not find a way in Xcode to exclude this text field from localization. Did I miss something?

I don't think there's a way to do so. I believe I have also did some research on this without success.

> Would deleting the entry for the object ID 3H8-Ei-cTm from the English translations remove this translation from Crowdin?

I'll test this, and try to find a way to do this semiautomatically with the Localizer Hint.

PR #4896, opened 2024-04-27
Fix audio delay units uses two lines in Hebrew, #4895

### LINKED ISSUE #4895: Audio delay units uses two lines in Hebrew
**System and IINA version:**

- macOS 13.6.3
- IINA 1.3.4 + PR #4922

**Expected behavior:**
The `Audio delay` setting on the `AUDIO` tab of the quick settings panel is displayed as intended in languages other than English.

**Actual behavior:**
The `Audio delay` setting contains a text field after the slider that can be used to enter a value for the delay. After the text field is another text field giving the units of the value. In English this field shows `s`.  Currently IINA does not support any right to left languages. However PR #4922 enables Hebrew. If you build IINA with that pull request and set Hebrew as the language then the units (`שנ'`) do not fit:
delay

This problem was reported by @ShlomoCode in issue #4776. As that issue reports quite a few RTL related problems I have created this issue to focus just on this particular problem.

**Steps to reproduce:**
- Pull PR #4922
- Click on the `Options` tab
- In the pull down select `Hebrew`
- Start playing a video
- Press ⇧⌘A to show the audio panel
- Notice the units shown for `Audio delay` is wrapped into multiple lines  

Every time.

low-batt:
Fix audio delay units uses two lines in Hebrew, #4895

This commit will:
- Change the `Line Break` attribute of the text field holding the units from `Word Wrap` to `Clip`
- Add a width constraint to the text field displaying slider setting as a numeric value
- Change the width constraint of the similar text field on the `SUBTITLES` tab to match the new constraint added to the `AUDIO` tab

This prevents the audio delay units field from wrapping into two lines and places a limit on the minimum size of the text field displaying the slider value such that the number it holds is not truncated when the user reduces the window size so much that the quick settings panel is compressed. 

- [x] This implements/fixes issue #4895.

---

**Description:**

### DIFF
--- a/iina/Base.lproj/QuickSettingViewController.xib
+++ b/iina/Base.lproj/QuickSettingViewController.xib
@@ -1233,6 +1233,9 @@
                                                                         </textField>
                                                                         <textField focusRingType="none" horizontalHuggingPriority="251" verticalHuggingPriority="750" allowsCharacterPickerTouchBarItem="YES" translatesAutoresizingMaskIntoConstraints="NO" id="Qiz-tS-aWT" userLabel="CustomAudioDelay Text Field">
                                                                             <rect key="frame" x="248" y="12" width="61" height="21"/>
+                                                                            <constraints>
+                                                                                <constraint firstAttribute="width" relation="greaterThanOrEqual" constant="28" id="6ch-P7-rIi"/>
+                                                                            </constraints>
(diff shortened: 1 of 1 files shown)


low-batt:
Moved this to be a draft due to the problem being discussed in the issue.

PR #4903, opened 2024-04-28
Fix screenshot cache contains many large files, #4899

### LINKED ISSUE #4899: Screenshot cache contains many large files
**System and IINA version:**

- macOS 14.4.1
- IINA 1.3.4

**Expected behavior:**
Disk caches are bounded in size.

**Actual behavior:**
@hugepower reports in [this comment](https://github.com/iina/iina/issues/4747#issuecomment-2081105353):
> I am using PNG format for my screenshots, and each one occupies approximately 2MB of space. Surprisingly, in the `~/Library/Caches/com.colliderli.iina/screenshot_cache` directory, IINA has stored nearly 10GB of past screenshot images. I am puzzled by this as I don't understand the need to retain so many screenshots, which seem to be taking up a significant amount of unnecessary storage space.

**Steps to reproduce:**
- Start IINA
- The settings panel appears
- Scroll down to the `Screenshots` section
- Uncheck the `Save to` setting
- Uncheck the `Show previews after taking screenshots`
- Start playing a video
- Take a screenshot
- Notice a file still remains in the `~/Library/Caches/com.colliderli.iina/screenshot_cache` directory

Every time with the settings shown above.

low-batt:
This commit will change the screenshotCallback method in PlayerCore to remove the cached screenshot file when not displaying a preview.

- [x] This implements/fixes issue #4899.

---

**Description:**

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -808,6 +808,9 @@ class PlayerCore: NSObject {
     guard let lastScreenshotURL = Utility.getLatestScreenshot(from: imageFolder) else { return }
     guard let image = NSImage(contentsOf: lastScreenshotURL) else {
       self.sendOSD(.screenshot)
+      if !saveToFile {
+        try? FileManager.default.removeItem(at: lastScreenshotURL)
+      }
       return
     }
     if saveToClipboard {
@@ -816,6 +819,9 @@ class PlayerCore: NSObject {
     }
     guard Preference.bool(for: .screenshotShowPreview) else {
       self.sendOSD(.screenshot)
+      if !saveToFile {
+        try? FileManager.default.removeItem(at: lastScreenshotURL)
+      }
       return
     }
 


svobs:
Could use a `defer` after initing `lastScreenshotURL`, instead of duplicating the code 3 times, like so:
```Swift
    defer {
      if !saveToFile {
        try? FileManager.default.removeItem(at: lastScreenshotURL)
      }
    }
```

low-batt:
I did not do that because I was worried that there was a reason the current code retains the file until all operations have completed. Since currently that happens asynchronously, deleting the file when exiting the method would create a race condition if there is a reason it must be retained until after it has been sent to the OSD.

How threading is handled between `PlayerCore` and `MPVController` is being refactored for the feature release. That will eliminate the "split processing" where some work occurs on the thread `MPVController` uses to read events and the main thread. This will simplify some of this code.

uiryuu:
What about clearing all the cached images when quitting (or starting) IINA? This can not only avoid synchronicity issue but also clear up previous (possibly large) cache after we ship the new version.

low-batt:
Synchronicity is only an issue with the suggestion to use `defer`.

The best practice is always to cleanup the temp file right after you finish with it. For example, I leave IINA running for weeks at a time on my non-development machine. It would be wrong to allow temp files to build up during that time. So the above code is needed regardless of whether IINA cleans up during startup/shutdown.

As noted the benefit of cleaning up during startup/shutdown is that it would clean up the files left from this defect. It would also clean up if IINA is killed with a temp file outstanding. But I'm too terrified of deleting all files in a directory to code that.

uiryuu (approved):
Oh, I was thinking that `self.sendOSD(.screenshot)` would also use the image file somehow. The changes are okay to me. Maybe we can state in our wiki that users can clear up the cache folder manually if they have used a older version of IINA.

PR #4907, opened 2024-05-03
Fix RTL language causes history indicator to flip, #4906

### LINKED ISSUE #4906: RTL language causes history progress indicator to reverse
**System and IINA version:**

- macOS 13.6.3
- IINA 1.3.4 + PR #4922

**Expected behavior:**
IINA adheres to Apple's [Internationalization and Localization Guide](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/SupportingRight-To-LeftLanguages/SupportingRight-To-LeftLanguages.html) which specifies that timeline indicators should not flip in a right-to-left language.

**Actual behavior:**
Currently IINA does not support any right to left languages. However PR #4922 enables Hebrew. If you build IINA with that pull request and set Hebrew as the language then the progress indicator in the history window changes direction as seen here:
history-window

This problem was reported by @ShlomoCode in issue #4776. As that issue reports quite a few RTL related problems I have created this issue to focus just on this problem.

**Steps to reproduce:**
- Click on the `Options` tab
- In the pull down select `Right-to-Left Pseudolanguage` near the end of the list
- Click on  `Playback History` under the `File` menu
- The `Playback History` window appears
- Notice in the direction of the progress indicator is flipped.

Or you can build with PR #4922 and set your language to Hebrew.

Every time.

low-batt:
This commit will add an `awakeFromNib` method to the `HistoryProgressCellView` class that sets the progress indicator property `userInterfaceLayoutDirection` to be `leftToRight`. This prevents the indicator from reversing direction when a RTL language is configured.

- [x] This implements/fixes issue #4906.

---

**Description:**

### DIFF
--- a/iina/HistoryWindowController.swift
+++ b/iina/HistoryWindowController.swift
@@ -349,4 +349,10 @@ class HistoryProgressCellView: NSTableCellView {
 
   @IBOutlet var indicator: NSProgressIndicator!
 
+  /// Prepares the receiver for service after it has been loaded from an Interface Builder archive, or nib file.
+  /// - Important: As per Apple's [Internationalization and Localization Guide](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/SupportingRight-To-LeftLanguages/SupportingRight-To-LeftLanguages.html)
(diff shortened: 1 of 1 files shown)


ShlomoCode:
I confirm that this PR fixes the issue for me

PR #4926, opened 2024-05-11
Log running commands and setting options and properties

low-batt:
This commit will:
- Change `PlayerCore` to pass the value of `playerCoreCounter` when creating the `MPVController` object
- Change `MPVController` to use a logger subsystem name that includes the counter to identify the player the instance is associated with
- Change `MPVController` to log commands sent to mpv
- Change `MPVController` to log setting of mpv options and properties

---

**Description:**


uiryuu:
Very useful feature!

PR #4942, opened 2024-05-24
Remove unsupported codecs from hwdec-codecs

low-batt:
This commit will:
- Add a new class `HardwareDecodeCapabilities`
- Add a call to `HardwareDecodeCapabilities.checkCapabilities` in `AppDelegate.applicationWillFinishLaunching`
- Add a new method `userOptionsContains` to `MPVController` that checks if a user has configured a `mpv` option in IINA's advanced settings
- Add a new property `mpvCodecToCodecTypes` to  `MPVController` that maps a `mpv` codec name to core media video codec types
- Add a new `adjustCodecWhiteList` method to `MPVController` that removes codecs from the `mpv` `hwdec-codecs` option if the Mac does not support hardware decoding for that codec
- Change the `applyHardwareAccelerationWorkaround` method to use the value of the `hwdec-codecs` option instead of a hardcoded value

These changes cause mpv to not attempt to use hardware decoding for a codec that is known to not be supported. This eliminates the overhead of setting up for hardware decoding only to have it fail. This is not comprehensive. This only covers the recent codecs whose support for hardware decoding varies among Macs. This merely reduces the dependence upon the FFmpeg fallback to software decoding feature in some cases.

---

**Description:**

### DIFF
--- a/iina/HardwareDecodeCapabilities.swift
+++ b/iina/HardwareDecodeCapabilities.swift
@@ -1,0 +1,103 @@
+//
+//  HardwareDecodeCapabilities.swift
+//  iina
+//
+//  Created by low-batt on 5/23/24.
+//  Copyright © 2024 lhc. All rights reserved.
+//
+
+import Foundation
+import VideoToolbox
+
+/// Cache containing information about the hardware decoding capabilities of this Mac.
+///
+/// It is desirable to cache this information because the [Video Toolbox](https://developer.apple.com/documentation/videotoolbox)
+/// method  [VTIsHardwareDecodeSupported](https://developer.apple.com/documentation/videotoolbox/vtishardwaredecodesupported(_:))
+/// is long running and should not be called on the main thread. Instead another thread is used and the information is cached for when
+/// code that is running on the main thread needs it.
+class HardwareDecodeCapabilities {
+  /// The `HardwareDecodeCapabilities` singleton object.
+  static let shared = HardwareDecodeCapabilities()
+
+  /// The hardware decoding capabilities of this Mac will be checked for the codecs in this list.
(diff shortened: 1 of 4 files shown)


low-batt:

iina/HardwareDecodeCapabilities.swift
```diff
@@ -0,0 +1,103 @@
...
+  /// - Important: This method calls
+  ///     [VTIsHardwareDecodeSupported](https://developer.apple.com/documentation/videotoolbox/vtishardwaredecodesupported(_:)),
+  ///     which if called on the main thread will cause Xcode to report: "This method should not be called on the main thread as it may
+  ///     lead to UI unresponsiveness". Use a different thread to call this method. Do not call `isHardwareDecodeSupported`
+  ///     from the main thread.
+  private func isHardwareDecodeSupported(_ codecType: CMVideoCodecType) -> Bool {
+    if #available(macOS 11.0, *) {
+      VTRegisterSupplementalVideoDecoderIfAvailable(codecType)
+    }
+    if #available(macOS 10.13, *) {
```
uiryuu: Given that IINA now will only run on macOS 10.15+, we can safely remove this version check and the code after this, since they will never be executed.

uiryuu (changes requested):

low-batt:
@lhc70000 Another PR that is affected by the commit that was amended after it was merged into `develop`. Please fix.

iina/HardwareDecodeCapabilities.swift
low-batt: Fixed

uiryuu (approved):

PR #4953, opened 2024-05-26
Fix text clipping when textfields have multiple lines in inspector

uiryuu:

---

**Description:**
A regression in https://github.com/iina/iina/pull/4531.

Before (current develop):
image

After:
image


low-batt (approved):
I reproduced the problem. Then I pulled the PR, rebuilt IINA and tested. Confirmed fixed.

Changes look good to me.

Issue #1098, opened 2017-10-12
Video Filter Presets window is too small to fit some words


euf:
**System and IINA version:**

- macOS 10.13
- IINA 0.0.13

**Expected behavior:**

Filter Presets window is wide enough to fit all parameter names.

 

**Actual behavior:**

When setting up Video Filters, the window is too narrow for Russian (and maybe other languages) words. 

![2017-10-12 15 25 09](https://user-images.githubusercontent.com/11754264/31496040-812c00fc-af62-11e7-8665-5a0d11ae5e8d.png)

In some cases it is crucial for understanding.
For example, two last options «Custom (MPV)» and «Custom (ffmpeg)» can’t be distinguished now.

euf:
As a temporary measure, I will submit a translation update, which will use shorter (but less accurate) equivalents for these terms. This would help for Russian, not sure about other languages with long words (German?)

low-batt:
This was improved in IINA 1.4.0-beta1.

PR #4955, opened 2024-05-27
Improve UI of Filters windows (fixes #4894)

### LINKED ISSUE #4894: Filter names truncated when adding filter
**System and IINA version:**

- macOS 13.6.3
- IINA 1.3.4 + PR #4922

**Expected behavior:**
Filter names are not truncated.

**Actual behavior:**
PR #4922 enables Hebrew. If you build IINA with that pull request and set Hebrew as the language then the filter name for a custom FFmpeg filter is truncated in the sheet for adding a filter:
he-filter

The table of filters should be automatically resized to adapt to the size of the filter names in other languages.

**NOTE** that this problem can also be reproduced using the `develop` branch and English. In Xcode open `FilterWindowController.xib`. Built and run IINA. The `Custom (FFmpeg)` filter name is now truncated:
en-filter

Digging through all the changes to the XIB Xcode helpfully adds even if you only open the file in the editor and don't make any changes I see:
```xml
- <tableColumn width="477" minWidth="40" maxWidth="1000" id="TbO-kN-XQm">
+ <tableColumn width="468" minWidth="40" maxWidth="1000" id="TbO-kN-XQm">
```

Xcode is messing with the width of table columns.

There are warning in the build log regarding layout problems:
```text
/* com.apple.ibtool.document.notices */
/Users/low-batt/Documents/builds/iina-official/iina/iina/Base.lproj/FilterWindowController.xib:MAd-sY-Hvg: note: NSTextField's height does not match its intrinsic content size, use "Size to Fit Content" to correct [22]
/Users/low-batt/Documents/builds/iina-official/iina/iina/Base.lproj/FilterWindowController.xib:cLl-pZ-KNW: note: View is clipping its content [13]
/Users/low-batt/Documents/builds/iina-official/iina/iina/Base.lproj/FilterWindowController.xib:ZJo-ef-lfp: note: View is clipping its content [13]
/Users/low-batt/Documents/builds/iina-official/iina/iina/Base.lproj/FilterWindowController.xib:56K-2H-lTw: note: View is clipping its content [13]
```

This problem was reported by @ShlomoCode in issue #4776. As that issue reports quite a few RTL related problems I have created this issue to focus just on this particular problem.

**Steps to reproduce:**
- Pull PR #4922
- Click on the `Options` tab
- In the pull down select `Hebrew`
- Click on  `Video Filters…` under the `Video` menu
- The `Video Filters` window appears
- Click on the `+` button
- The sheet for adding a filter appears
- Notice the last filter in the list is truncated

Every time.

svobs:
  - #4894 
  - #1098

---

**Description:**

Makes the following fixes/improvements to the `Video Filters` & `Audio Filters` windows:
• Increase size of text & icons
• Add save of window sizes & positions between launches
• Add double-click in Saved Filters table to edit saved filter

Also makes the following changes to the "Add Filter" sheet:
• Ensure Filter Presets table always has exactly 1 row selected, so that the sheet's Add button can no longer be clicked when nothing is selected
• Save size of sheet between launches
• Fix multiple layout issues which prevented the sheet from resizing correctly
• Set width of Filter Presets table at first load to fit its widest localized text (issues #1098, #4894).

### DIFF
--- a/iina/AppData.swift
+++ b/iina/AppData.swift
@@ -106,6 +106,10 @@ struct Constants {
   struct Time {
     static let infinite = VideoTime(999, 0, 0)
   }
+  struct WindowAutosaveName {
+    static let videoFilters = "VideoFilters"
+    static let audioFilters = "AudioFilters"
+  }
   struct FilterName {
     static let crop = "iina_crop"
     static let flip = "iina_flip"
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -66,14 +66,12 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
   lazy var logWindow: LogWindowController = LogWindowController()
 
   lazy var vfWindow: FilterWindowController = {
-    let w = FilterWindowController()
-    w.filterType = MPVProperty.vf
+    let w = FilterWindowController(filterType: MPVProperty.vf, autosaveName: Constants.WindowAutosaveName.videoFilters)
     return w
   }()
 
   lazy var afWindow: FilterWindowController = {
-    let w = FilterWindowController()
-    w.filterType = MPVProperty.af
+    let w = FilterWindowController(filterType: MPVProperty.af, autosaveName: Constants.WindowAutosaveName.audioFilters)
     return w
   }()
 
--- a/iina/Base.lproj/FilterWindowController.xib
+++ b/iina/Base.lproj/FilterWindowController.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="14313.13.2" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="32700.99.1234" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="14313.13.2"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="22690"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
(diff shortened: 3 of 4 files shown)


svobs:
Made a couple last-minute tweaks. Funny how I don't notice these things until they just pop out.
SCR-20240527-kquu

low-batt:
Changes look good.

There is a minor problem. I think you have fixed problems like this in the past in other tables. After the two named columns and the 3<sup>rd</sup> unnamed column that holds `save` buttons there is an additional 4<sup>th</sup> unnamed column:

extra-col

This is an existing problem, so it doesn't have to be fixed for this PR. This might be hard to fix as the saved filters table uses 4 columns. Not sure if there is code that requires both tables to have the same number of columns.

svobs:
> There is a minor problem. I think you have fixed problems like this in the past in other tables. After the two named columns and the 3rd unnamed column that holds save buttons there is an additional 4th unnamed column

Yes, this is a design limitation of AppKit's `NSTableView`. I wasn't ever able to "fix" it completely, but if you try resizing the window to make the width smaller than the sum of the columns, it should "snap" into place and the last column should hug the right side when you expand it again.

I think the Saved Filter table has a better approach. It works around the problem by using only a single column, and then doing all the layout of the views inside it. If you contrast with the Active Filters table above it, at first glance there's no reason the `Save` button should have its own column. If the `#` column were to be combined with the `Filter String` column, the only downside would be the loss of its header. So it would make sense to remove the table headers too. But by changing the column from a simple text field, we'd lose the built-in "double-click to edit inline" feature and would need to code it manually. So it quickly turns into a lot of work. Still hoping that some kind of SwiftUI table component will appear which can do everything and solve all these problems.

svobs:
One last change -

I was able to set the width of the Filter Presets table based on its content! So it will look nice with any locale. I changed the table style to "Full Width" and set the width of its clip view when the table loads. So you have the following results.

English:
SCR-20240527-mguo-2
Hebrew:
SCR-20240527-mhax-2

low-batt (approved):
Pulled PR, built IINA and tested. Looks good to me.

uiryuu:
Very little thing, the text height seems not enough to fit the filter string.

image

Other changes look good

svobs:
> Very little thing, the text height seems not enough to fit the filter string.

Strange that it wasn't cut off on my screen. But I could see in the View Debugger that the height of that row was a little bit too small. I think I fixed it. Please confirm.

I also just noticed that the `+` and `-` buttons in the window were not moving to the right side in Hebrew. That was a headache to fix because I had to migrate a bunch of things from using auto-resize masks to using constraints. But all should be good now - please check and let me know.

uiryuu (approved):
Truncated text has been fixed. LGTM

uiryuu:
Thanks!

PR #4956, opened 2024-05-28
Refactor logs and use more .warning and .verbose leveled logs

uiryuu:
**Description:**
- Added a log of mpv compile-time configuration
- Refactored `AutoFileMatcher`, `ThumbnailCache`, and `PlayerCore` which added a dedicated `log` function for their subsystems
- Uses more .warning logs
- Changed mpv property/option related logs into .verbose

### DIFF
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -577,7 +577,7 @@ not applying FFmpeg 9599 workaround
   // Send arbitrary mpv command.
   func command(_ command: MPVCommand, args: [String?] = [], checkError: Bool = true, returnValueCallback: ((Int32) -> Void)? = nil) {
     guard mpv != nil else { return }
-    log("Run command: \(command.rawValue) \(args.compactMap{$0}.joined(separator: " "))")
+    log("Run command: \(command.rawValue) \(args.compactMap{$0}.joined(separator: " "))", level: .verbose)
     var cargs = makeCArgs(command, args).map { $0.flatMap { UnsafePointer<CChar>(strdup($0)) } }
     defer {
       for ptr in cargs {
@@ -595,13 +595,13 @@ not applying FFmpeg 9599 workaround
   }
 
   func command(rawString: String) -> Int32 {
-    log("Run command: \(rawString)")
+    log("Run command: \(rawString)", level: .verbose)
     return mpv_command_string(mpv, rawString)
   }
 
   func asyncCommand(_ command: MPVCommand, args: [String?] = [], checkError: Bool = true, replyUserdata: UInt64) {
     guard mpv != nil else { return }
-    log("Asynchronously run command: \(command.rawValue) \(args.compactMap{$0}.joined(separator: " "))")
+    log("Asynchronously run command: \(command.rawValue) \(args.compactMap{$0}.joined(separator: " "))", level: .verbose)
     var cargs = makeCArgs(command, args).map { $0.flatMap { UnsafePointer<CChar>(strdup($0)) } }
     defer {
       for ptr in cargs {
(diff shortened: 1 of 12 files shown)


iina/MPVController.swift
```diff
@@ -577,7 +577,7 @@ not applying FFmpeg 9599 workaround
   // Send arbitrary mpv command.
   func command(_ command: MPVCommand, args: [String?] = [], checkError: Bool = true, returnValueCallback: ((Int32) -> Void)? = nil) {
     guard mpv != nil else { return }
-    log("Run command: \(command.rawValue) \(args.compactMap{$0}.joined(separator: " "))")
+    log("Run command: \(command.rawValue) \(args.compactMap{$0}.joined(separator: " "))", level: .verbose)
```
low-batt: I plan to take a close look at this PR. I'm busy on Tuesdays, so maybe not until Wed. In the meantime, one comment on this part of the change…

I added this logging after debugging a problem with a user. It would have saved time if commands were logged to the IINA log. Although they are in the `mpv` log the `mpv` log messages are not shown in the log viewer. We want this to be easy for users to see.

I think the interactions with `mpv` are more important to log than many of the other log messages IINA emits. It is especially important for us to see any options the user has directly set in advanced settings.

Is there a case where IINA emits these messages at a high rate? Otherwise I'd prefer these messages were emitted at the default level.

uiryuu: I changed mpv commands and options related logs to .verbose indeed because sometimes they are too many. For instance, during init phase, there're a lot of option settings; when the user drags the progress bar to seek, there will also generate a bunch of `Run command: seek ...` logs. 

> I think the interactions with mpv are more important to log than many of the other log messages IINA emits.

This is also true. So if you insist on keeping them the default level, I can revert them back.

> Although they are in the mpv log the mpv log messages are not shown in the log viewer.

What we can do maybe in the future is to display mpv logs in the log window as well. We can disable that by default and let user to enable that in the advanced settings.

low-batt: The point about options at startup is good in that a majority of those are constant and do not change in a user's environment. However a few of them are not and are important to log.

I think this means we have to do what I was thinking of doing, but didn't. We have to add a `level: Level = .debug` parameter to the methods that are used by higher level code that generate logs at a high rate and then change that higher level code to pass `.verbose` for the operations that produce too much logging.

If you want I can address this in a follow up PR.

For me the `mpv` log level seems to be set to verbose and has this same issue of tons of repeated log messages. Do you see that as well?

By the way, I did not put logging into the unused async property commands (`setFlagAsync`, etc) to avoid a merge conflict with the PR that is deleting them. If they are staying they need logging. I would delete these methods. We don't want to use asynchronous methods unless there is a significant improvement in performance. We have had problems that reproduced for users and not for developers due to asynchronous thread behavior. So we don't want to introduce additional asynchronous behavior until we have to.

uiryuu: > If you want I can address this in a follow up PR.

Please do that 🙇 I'll revert back the changes about property/option related logs.

> For me the mpv log level seems to be set to verbose and has this same issue of tons of repeated log messages. Do you see that as well?

I didn't get you. Isn't `mpv` log level set to 'warn' in IINA currently?

For the async functions, we can remove them.

low-batt: I will follow up with a log PR once we get this one merged. If the `mpv` log looks good for you then maybe it is some configuration on my machine. I will try and track that down.

iina/VideoView.swift
```diff
@@ -444,12 +444,12 @@ extension VideoView {
     guard let mpv = player.mpv else { return false }
 
     guard let primaries = mpv.getString(MPVProperty.videoParamsPrimaries), let gamma = mpv.getString(MPVProperty.videoParamsGamma) else {
-      Logger.log("HDR primaries and gamma not available", level: .debug, subsystem: hdrSubsystem)
+      Logger.log("HDR primaries and gamma not available", level: .warning, subsystem: hdrSubsystem)
```
low-batt: May need to rethink using warning for this one. If I remember correctly I think it is normal for the primaries and gamma to not initially be available.

iina/VideoView.swift
```diff
@@ -473,23 +473,23 @@ extension VideoView {
...
     default:
-      Logger.log("Unknown HDR color space information gamma=\(gamma) primaries=\(primaries)", level: .debug, subsystem: hdrSubsystem)
+      Logger.log("Unknown HDR color space information gamma=\(gamma) primaries=\(primaries)", level: .warning, subsystem: hdrSubsystem)
       return false
     }
 
     guard (window?.screen?.maximumPotentialExtendedDynamicRangeColorComponentValue ?? 1.0) > 1.0 else {
-      Logger.log("HDR video was found but the display does not support EDR mode", level: .debug, subsystem: hdrSubsystem)
+      Logger.log("HDR video was found but the display does not support EDR mode", level: .warning, subsystem: hdrSubsystem)
       return false
```
low-batt, line 482: This is a very normal occurrence. I would not log a warning for this.

low-batt (approved):
Changes look good. Might want to change the couple of logs I noted to not be warnings. We can always do that later after getting experience with it.

Lets get this merged so I can follow up with a PR to reduce the excess logging.

Might want to wait to do more refactoring until we get more of the outstanding PRs merged to avoid introducing merge conflicts.

Issue #4946, opened 2024-05-25
The right time label shows xx:xx- instead of -xx:xx when in `.remaining` mode in Hebrew


uiryuu:
**System and IINA version:**

- macOS 14.5
- IINA Latest `enable-he` branch

**Expected behavior:**
-xx:xx, as in other languages. Need to check with @ShlomoCode
![image](https://github.com/iina/iina/assets/20237141/c55a8489-17ea-4f3d-bf70-8184ed106cb7)

**Actual behavior:**
![image](https://github.com/iina/iina/assets/20237141/fe1886e5-9170-4ed8-a9f0-fedd8a4f2e54)

**Steps to reproduce:**
- Open IINA in Hebrew
- Start a video, and click on the right time label to make it to show the remaining time

- [x] MPV does not have this problem.

**How often does this happen?**
Always

ShlomoCode:
> -xx:xx, as in other languages. Need to check with @ShlomoCode

Agreed

low-batt:
@uiryuu The touch bar is still showing xx:xx-. That view needs to be fixed as well.

low-batt:
This was fixed in IINA 1.4.0-beta1.

Issue #4950, opened 2024-05-26
RTL: Incorrect arrow icon direction in browser extensions reference


ShlomoCode:
**System and IINA version:**

- macOS 14.4.1
- IINA 1.3.4 + PR #4922

**Expected behavior:**

The icon will always point to the text, even in RTL languages such as Hebrew and Arabic.

**Actual behavior:**

Currently IINA does not support any right-to-left languages. However PR #4922 allows Hebrew. If you build IINA with this pull request and set Hebrew as the language, you will notice that the arrows in the browser extension buttons (under the Utils tab in the settings) do not point in the direction of the text (the direction is hard-coded to the right)

![CleanShot 2024-05-23 at 06 21 53@2x](https://github.com/iina/iina/assets/78599753/b11f4265-3b74-4f23-9d9d-2a099762ea09)
![CleanShot 2024-05-23 at 07 03 48@2x](https://github.com/iina/iina/assets/78599753/02c0ba59-3921-430b-a558-7557736ad3b6)

**Steps to reproduce:**

- Open the IINA project in Xcode
- Under the `Product` menu open the Scheme menu and select `Edit Scheme…`
- Click on the `Options` tab
- Scroll down to the `App Language` setting
- In the pull down select `Right-to-Left Pseudolanguage` near the end of the list
- Start IINA running under Xcode
- Open Settings and go to Utils section
- Notice that the arrows are not facing the text

- [x] MPV does not have this problem.

mpv does not support localization.

**How often does this happen?**

Every time.

low-batt:
@ShlomoCode Does it look the same when the system language is Hebrew?

ShlomoCode:
> @ShlomoCode Does it look the same when the system language is Hebrew?

Yes (even after restart)
![CleanShot 2024-05-26 at 06 35 59@2x](https://github.com/iina/iina/assets/78599753/6cdf263f-e805-4418-aa8b-3bd22229b117)

low-batt:
Thanks for checking. I see the problem. I was confused by the terminology. Working on a fix.

ShlomoCode:
![CleanShot 2024-05-26 at 16 03 49@2x](https://github.com/iina/iina/assets/78599753/8fd46d84-fabe-4437-a9f3-f87ce3d5a92b)
The arrows should stay on the right side of the button, just point towards the text and not out. See also https://github.com/iina/iina/pull/4937

uiryuu:
[...]

Issue #4878, opened 2024-04-13
RTL language causes OSC to malfunction 


low-batt:
**System and IINA version:**

- macOS 13.6.3
- IINA 1.3.4 + PR #4922

**Expected behavior:**
The on screen controller works properly with right to left languages such as Arabic and Hebrew.

**Actual behavior:**
Currently IINA does not support any right to left languages. However PR #4922 enables Hebrew. If you build IINA with that pull request and set Hebrew as the language then the slider in the OSC malfunctions. When you click on the slider the knob does not move to where the cursor is. The A-B Loop knobs can not be dragged. If you hover over the slider the thumbnail preview shown is not for the correct position in the video. Some images on the OSC point in the wrong direction.

These problems were reported by @ShlomoCode in issue #4776. As that issue reports quite a few RTL related problems I have created this issue that focuses on the OSC problem due to the malfunctioning behavior of the slider.

**Steps to reproduce:**
- Open the IINA project in Xcode
- Under the `Product` menu open the `Scheme` menu and select `Edit Scheme…`  
- Click on the `Options` tab
- Scroll down to the `App Language` setting
- In the pull down select `Right-to-Left Pseudolanguage` near the end of the list
- Start IINA running under Xcode
- Start playing a video
- Click on various places in the OSC slider
- Notice the knob moves the the wrong place

Or you can build with PR #4922 and set your language to Hebrew.

- [x] MPV does not have this problem.

mpv does not support localization.

**How often does this happen?**
Every time.

low-batt:
I will be posting a PR soon that corrects the problems.

low-batt:
@ShlomoCode Do you spot anything wrong in this screenshot?:

![he-osc](https://github.com/iina/iina/assets/86170219/22a52a35-c06b-4a2e-a841-37b9afa6119c)

I did some more testing and found dragging the floating OSC around is broken. I will add a fix for that once I figure out the proper fix.

ShlomoCode:
@low-batt It looks fine! :clap:
If you want me to pull and check the PR locally please let me know

low-batt:
I suggest waiting for me to fix the problem with moving the OSC in floating mode. I may be able to get that done tomorrow.

My understanding is that the IINA project's position is that localization is a priority and must be done right. So please do point out anything you notice that is not correct even if it is minor.

I am expecting this will be an iterative process. First priority is to fix anything that is actually malfunctioning due to RTL.

low-batt:
[...]

PR #4958, opened 2024-05-29
More RTL related fixes on OSC

uiryuu:
- [x] This implements/fixes issue #4946 and #4878 .

---

**Description:**
This PR
- Revert back 1 previous commit in favor of a more correct fix
- Fix DurationDisplayTextField flipped when having minus sign in RTL #4946
- Make the entire layout of the OSC reverse (as per https://github.com/iina/iina/issues/4878#issuecomment-2067835237)
- The directions of arrows of the media control are fixed (remaining issue in #4878)
- Fix the browser extension layout and the flipped the image #4950

Floating OSC:
image

Top/bottom OSC:
image

Please try out @ShlomoCode, and if there're still OSC related issues or I have messed up anything, please report here so that I can quickly fix them.

### DIFF
--- a/iina/Assets.xcassets/Icons/backward.imageset/Contents.json
+++ b/iina/Assets.xcassets/Icons/backward.imageset/Contents.json
@@ -1,13 +1,14 @@
 {
   "images" : [
     {
+      "filename" : "back.pdf",
       "idiom" : "universal",
-      "filename" : "back.pdf"
+      "language-direction" : "left-to-right"
     }
   ],
   "info" : {
-    "version" : 1,
-    "author" : "xcode"
+    "author" : "xcode",
+    "version" : 1
   },
   "properties" : {
     "template-rendering-intent" : "template"
--- a/iina/Assets.xcassets/Icons/forward.imageset/Contents.json
+++ b/iina/Assets.xcassets/Icons/forward.imageset/Contents.json
@@ -2,7 +2,8 @@
   "images" : [
     {
       "filename" : "forward.pdf",
-      "idiom" : "universal"
+      "idiom" : "universal",
+      "language-direction" : "left-to-right"
     }
   ],
   "info" : {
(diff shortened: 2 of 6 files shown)


ShlomoCode:
The OSC looks really good! It seems to fix all major issues. Please give me some time to look if there is anything missed.
In the meantime I noticed that the PIP button points in the correct direction (to left) but the popup actually opens on the right side:
CleanShot 2024-05-29 at 05 50 21

low-batt:
@uiryuu The reason why I did not change this initially is because for whatever reason Quicktime does not change the location of the volume as seen here with language set to Hebrew:
quicktime

The button in the PiP window also does not flip with Quicktime:
pip

I don't think IINA can do anything about that as we are just using what the system provides.

ShlomoCode:
> The button in the PiP window also does not flip with Quicktime:

It doesn't look like something intentional, note that the arrows are pointing in random directions... it seems Apple just didn't pay attention to that aspect

https://github.com/iina/iina/assets/78599753/5aa6f73d-fd12-4e80-902a-b020d35f7519

ShlomoCode:
> I don't think IINA can do anything about that as we are just using what the system provides.

The icon **inside** the PIP doesn't matter that much, what matters is the location where it opens, can't you control that? 
Apple's behavior (Safari/QT) is funny - they open the PIP in the **upper corner** of the screen (see my video)...
by the way, I think that's why that it is impossible to learn from them about changing the position of the buttons

uiryuu:
I pushed another commit to fix the browser extension layout and the flipped the image.
telegram-cloud-photo-size-5-6248923411770358876-x

ShlomoCode:
Now I understand that **MacOS opens PIP in the last place where PIP was open** - that is, if you activated PIP in Safari and manually dragged it to a certain corner, then closed it and activated PIP in IINA, it will open in the place where you placed the previous PIP.
Wow, it's getting too complicated to know what the default is...

If so, and if IINA cannot control the MacOS bug of the direction of the arrow inside the PIP not pointing towards the window, the current situation (https://github.com/iina/iina/commit/d6a05e32ba3624975f16fc63bf699545fccd528a) seems fine to me. Sorry for the confusion I caused

ShlomoCode:
> I pushed another commit to fix the browser extension layout and the flipped the image. 

notice that the icon has changed to fill style (maybe you changed it on purpose)

uiryuu:
I did change it intentionally. We used system template previously, however it's very hard to flip for RTL language, so I just used the image from our own assets. Hopefully that wouldn't cause any problem...

low-batt:
IINA is using `PIPViewController`, which is private and not documented, so not easy to control.

low-batt (approved):
Changes look good to me.

lhc70000:
I just checked that Xcode can correctly mirror the newly added `forward` and `backward` icons.

@uiryuu Please update the last commit to set the icons as "Let to right, mirrors". No new outlets or code should be needed (force push if possible).

Issue #3724, opened 2022-05-06
Monitor FSEvents changes to auto-reload subtitles if they are modified.


ShikiSuen:
**What you want IINA to do:**

Monitor FSEvents changes to auto-reload subtitles if they are modified.

**What IINA does currently:**

You have to manually drag the modified subtitle file to the IINA playback window. Dragging onto IINA icon on the dock **doesn't work**.

**Why you think this should be added:**

Helping subtitle makers to debug their subtitles. Some famous AAS subtitle tools like Aegisub are not updated for years and have accuracy problems in playing videos inside themselves, hence the necessary of using a separated video player to debug the subtitles. In this case, IINA is almost the best except having lack of the feature requested in this ticket.

**Examples of other projects that have something similar:**

vChewing input method, automatically reloads user phrases when they gets changed:
https://gitee.com/vchewing/vChewing-macOS
(See [FSEventStreamHelper.swift](https://gitee.com/vchewing/vChewing-macOS/blob/main/Source/Modules/FileHandlers/FSEventStreamHelper.swift) and [AppDelegate.swift](https://gitee.com/vchewing/vChewing-macOS/blob/main/Source/Modules/AppDelegate.swift) to know how they gets used.)
(Crediting Zonble for the copyright of the FSEventStreamHelper.swift in MIT license.)

svobs:
This looks amazing...

cssmagic:
Very important to me. Looking forward to its implementation.

ShikiSuen:
@svobs I apologize that I switched to another method. Please see my updated main post.

PR #4960, opened 2024-05-29
Double click in the input config table to rename a config (+ bug fixes)

uiryuu:

---

**Description:**
This PR:
- Enable the user to change the config name by double clicking in the input config table
- Refactors the configuration creation/duplication to eliminate duplicated code
- Change some of the wordings to make things clearer
- Stabilize the default input config files by using `KeyValuePairs`

Besides, this PR also fixes the usage of the `Utility.quickAskPanel`:

https://github.com/iina/iina/blob/94176bb558334e6abf1299fb02f3f921a649a860/iina/PrefKeyBindingViewController.swift#L177-L188

Previously it used the return value of the `quickAskPanel`, however, if the `quickAskPanel` is provided with the sheet window, it will always return `false` immediately without waiting for user input in the `NSAlert` panel. So 1) the "delete file" part of code could never be called, and 2) it always shows the existing file in Finder after the `NSAlert` is initiated (instead of after the user clicking the cancel button).

This PR changes the logic to show the existing file when clicking on OK, and do nothing when clicking on cancel. And also change the wording from "This should not happen..." to "A config file with the same filename already exists..." 

I also remove the file deletion code because maybe its too dangerous to remove user files with "OK" or an enter hit in a seemingly unrelated alert window (the user want to create a config, but end up with deleting their file), let alone that part of code was never been executed before due to the erroneous usage.

### DIFF
--- a/iina/PrefKeyBindingViewController.swift
+++ b/iina/PrefKeyBindingViewController.swift
@@ -267,6 +267,7 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
   /// If the target config file cannot be found, or the file cannot be parsed correctly, it will fallback to the default config.
   /// - Parameter configName: the target config name
   private func loadConfigFile(_ configName: String?) {
+    guard configName != Preference.string(for: .currentInputConfigName) else { return }
     isLoading = true
     
     func fallback() {
--- a/iina/PrefKeyBindingViewController.swift
+++ b/iina/PrefKeyBindingViewController.swift
@@ -61,6 +61,8 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
 
   var currentConfName: String!
   var currentConfFilePath: String!
+  
+  var isLoading = false
 
   var shouldEnableEdit: Bool = true
 
@@ -93,27 +95,10 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
       useMediaKeysButton.title = NSLocalizedString("preference.system_media_control", comment: "Use system media control")
     }
 
-    // Fallback default input config
-    var currentConf = "IINA Default"
-    if let confFromUd = Preference.string(for: .currentInputConfigName) {
-      if getFilePath(forConfig: confFromUd, showAlert: false) != nil {
-        currentConf = confFromUd
-      }
-    }
-    // load
-    confTableSelectRow(withTitle: currentConf)
-    currentConfName = currentConf
-    guard let path = getFilePath(forConfig: currentConf) else { return }
-    currentConfFilePath = path
-    loadConfigFile()
+    // Load the config file saved in user default
+    loadConfigFile(Preference.string(for: .currentInputConfigName))
     
     NotificationCenter.default.addObserver(forName: .iinaKeyBindingChanged, object: nil, queue: .main, using: saveToConfFile)
-  }
-
-  private func confTableSelectRow(withTitle title: String) {
-    if let index = configNames.firstIndex(of: title) {
-      confTableView.selectRowIndexes(IndexSet(integer: index), byExtendingSelection: false)
-    }
   }
 
   // MARK: - IBActions
@@ -163,7 +148,6 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
   // Check whether or not a new config file with provided filename should be created.
   // Returns the path of the new config if could be created; nil otherwise
   private func checkNewConfigFile(with filename: String) -> String? {
-    
     // Check if the name is empty
     guard !filename.isEmpty else {
       Utility.showAlert("config.empty_name", sheetWindow: self.view.window)
@@ -188,14 +172,6 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
     }
     return filePath
   }
-  
-  private func enableNewConfigFile(_ filename: String, _ filePath: String) {
-    currentConfName = filename
-    currentConfFilePath = filePath
-    confTableView.reloadData()
-    confTableSelectRow(withTitle: filename)
-    loadConfigFile()
-  }
 
   @IBAction func newConfFileAction(_ sender: AnyObject) {
     Utility.quickPromptPanel("config.new", sheetWindow: view.window) { newName in
@@ -205,10 +181,9 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
         Utility.showAlert("config.cannot_create", sheetWindow: self.view.window)
         return
       }
-      self.enableNewConfigFile(newName, newFilePath)
+      self.loadConfigFile(newName)
     }
   }
-
 
   @IBAction func duplicateConfFileAction(_ sender: AnyObject) {
     Utility.quickPromptPanel("config.duplicate", sheetWindow: view.window) { newName in
@@ -220,7 +195,7 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
         Utility.showAlert("config.cannot_create", arguments: [error.localizedDescription], sheetWindow: self.view.window)
         return
       }
-      self.enableNewConfigFile(newName, newFilePath)
+      self.loadConfigFile(newName)
     }
     
   }
@@ -229,13 +204,14 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
     guard shouldEnableEdit else { return }
     Utility.quickPromptPanel("config.rename", sheetWindow: view.window) { newName in
       guard let newFilePath = self.checkNewConfigFile(with: newName) else { return }
+
       do {
         try fm.moveItem(atPath: self.currentConfFilePath!, toPath: newFilePath)
       } catch let error {
         Utility.showAlert("config.cannot_create", arguments: [error.localizedDescription], sheetWindow: self.view.window)
         return
       }
-      self.enableNewConfigFile(newName, newFilePath)
+      self.loadConfigFile(newName)
     }
   }
 
@@ -250,12 +226,8 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
     } catch {
       Utility.showAlert("error_deleting_file", sheetWindow: view.window)
     }
-    // load
-    confTableView.reloadData()
-    currentConfName = configNames[0]
-    currentConfFilePath = getFilePath(forConfig: currentConfName)
-    confTableSelectRow(withTitle: currentConfName)
-    loadConfigFile()
+    // Fallback to default
+    loadConfigFile(configNames[0])
   }
 
   @IBAction func importConfigBtnAction(_ sender: Any) {
@@ -270,7 +242,7 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
         Utility.showAlert("config.cannot_create", arguments: [error.localizedDescription], sheetWindow: self.view.window)
         return
       }
-      self.enableNewConfigFile(newName, newFilePath)
+      self.loadConfigFile(newName)
     }
   }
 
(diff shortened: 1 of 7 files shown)


svobs:
When I saw the commit comment "don't maintain userConfigNames manually" I got excited. It would be better if `Preference.Key.inputConfigs` was no longer used at all, and instead the list of user configs was made by checking the `input_conf` directory for files ending in `.conf.

uiryuu:
In the latest commit, I removed `Preference.Key.inputConfigs` and made `userConfigs` a computed variable, so that even if the user adds a file to the folder manually, the table will show the file when refreshed. This commit needs more testing.

iina/PrefKeyBindingViewController.swift
```diff
@@ -27,18 +30,39 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
...
+
+  static var userConfigs: [String: String] {
+    do {
+      let files = try fm.contentsOfDirectory(at: Utility.userInputConfDirURL, includingPropertiesForKeys: nil)
+      let configFiles = files.filter { $0.pathExtension == "conf" }
+      return Dictionary(uniqueKeysWithValues: configFiles.map { ($0.deletingPathExtension().lastPathComponent, $0.path) })
+    } catch {
+      Logger.fatal("Cannot get user config file!")
+    }
+  }
```
lhc70000, line 58: This raises serious performance concerns (although may not be perceivable on a modern machine):

It means that `configNames` is also computed. Then, for example, `func tableView(_, objectValueFor tableColumn, row)` will access the file system for each row in the table.

iina/PrefKeyBindingViewController.swift
```diff
@@ -155,101 +144,74 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
     mappingController.remove(sender)
     NotificationCenter.default.post(Notification(name: .iinaKeyBindingChanged))
   }
+  
+  // Check whether or not a new config file with provided filename should be created.
+  // Returns the path of the new config if could be created; nil otherwise
+  private func checkNewConfigFile(with filename: String) -> String? {
```
lhc70000: It's returning a value, so better called `newConfigFilePath(forName filename:)` or something.

Also, better to use Swift doc comments.

iina/PrefKeyBindingViewController.swift
```diff
@@ -155,101 +144,74 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
...
+    // Check if there exists a config file with the same filename
+    let filePath = Utility.userInputConfDirURL.appendingPathComponent(filename + ".conf").path
+    if fm.fileExists(atPath: filePath) {
+      Utility.quickAskPanel("config.file_existing", sheetWindow: self.view.window) { result in
+        if result == .alertFirstButtonReturn {
+          NSWorkspace.shared.activateFileViewerSelecting([URL(fileURLWithPath: filePath)])
+        }
+      }
+      return nil
+    }
```
lhc70000: > I also remove the file deletion code because maybe its too dangerous to remove user files with "OK" or an enter hit in a seemingly unrelated alert window (the user want to create a config, but end up with deleting their file), let alone that part of code was never been executed before due to the erroneous usage.

At least we should tell the user how to delete this file (where it is) in the dialog.

iina/PrefKeyBindingViewController.swift
```diff
@@ -341,52 +285,58 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
...
-  private func loadConfigFile() {
-    if let mapping = KeyMapping.parseInputConf(at: currentConfFilePath) {
-      mappingController.content = nil
-      mappingController.add(contentsOf: mapping)
-      mappingController.setSelectionIndexes(IndexSet())
-    } else {
-      // on error
+  // This function firstly reloads the table data, select the config file row, then load the config file
+  // If the target config file cannot be found, or the file cannot be parsed correctly, it will fallback to IINA Default
+  private func loadConfigFile(_ configName: String?) {
```
lhc70000: Please add more comments in this function, and use Swift doc comment

iina/PrefKeyBindingViewController.swift
```diff
@@ -341,52 +285,58 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
...
+    func fallback() {
+      isLoading = false
       Utility.showAlert("keybinding_config.error", arguments: [currentConfName], sheetWindow: view.window)
-      let title = "IINA Default"
-      currentConfName = title
-      currentConfFilePath = getFilePath(forConfig: title)!
-      confTableSelectRow(withTitle: title)
-      loadConfigFile()
+      loadConfigFile(configNames[0])
       return
```
lhc70000: Don't need this

iina/PrefKeyBindingViewController.swift
```diff
@@ -71,47 +95,12 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
...
-    if !gotCurrentConf {
-      currentConf = "IINA Default"
-    }
-    // load
-    confTableSelectRow(withTitle: currentConf)
-    currentConfName = currentConf
-    guard let path = getFilePath(forConfig: currentConf) else { return }
-    currentConfFilePath = path
+    // Load the config file saved in user default
+    loadConfigFile(Preference.string(for: .currentInputConfigName))
```
lhc70000: Previously there was no `loadConfigFile()` here. So now it calls `setKeybindingsForPlayerCore()` when the view is loaded. Not a big issue, but we need to be careful when the behavior changes.

iina/PrefKeyBindingViewController.swift
```diff
@@ -155,101 +144,74 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
...
-          return
-        }
-      }
-      // - copy file
+      self.loadConfigFile(newName)
+    }
+    
+  }
+  
+  @IBAction func configFileListDoubleAction(_ sender: NSTableView) {
```
lhc70000, line 182: TBH I think when double clicking a table row people should expect the cell become editable, rather than a pop-up dialog. It's fine though.

iina/PrefKeyBindingViewController.swift
```diff
@@ -264,17 +226,8 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
...
-    if let index = userConfigNames.firstIndex(of: currentConfName) {
-      userConfigNames.remove(at: index)
-    }
-    confTableView.reloadData()
-    currentConfName = userConfigNames[0]
-    currentConfFilePath = getFilePath(forConfig: currentConfName)
-    confTableSelectRow(withTitle: currentConfName)
-    loadConfigFile()
+    // Fallback to default
+    loadConfigFile(configNames[0])
```
lhc70000: I'm not confident using `[0]` here. What if someone changed the behavior of `configNames` later? it's better to do a `filter`, and if you don't like hard-coded config names, you can make it a constant.

iina/PrefKeyBindingViewController.swift
```diff
@@ -434,11 +384,11 @@ extension PrefKeyBindingViewController: NSTableViewDelegate, NSTableViewDataSour
...
 
   func tableViewSelectionDidChange(_ notification: Notification) {
+    guard !isLoading else { return }
     if let tableView = notification.object as? NSTableView, tableView == confTableView {
-      guard let title = userConfigNames[at: confTableView.selectedRow] else { return }
-      currentConfName = title
-      currentConfFilePath = getFilePath(forConfig: title)!
-      loadConfigFile()
+      let title = configNames[at: confTableView.selectedRow]
+      guard title != currentConfName else { return }
```
lhc70000:
```swift
guard let title = configNames[at: confTableView.selectedRow], title != currentConfName else { return }
```

iina/PrefKeyBindingViewController.swift
```diff
@@ -434,11 +384,11 @@ extension PrefKeyBindingViewController: NSTableViewDelegate, NSTableViewDataSour
   }
 
   func tableViewSelectionDidChange(_ notification: Notification) {
+    guard !isLoading else { return }
```
lhc70000: Seems that `isLoading`'s only usage is to filter out didChange notifications when calling `loadConfigFile`? If so, it's better to add a parameter like `loadConfigFile(_, reloadTable:)`

lhc70000 (changes requested):

iina/PrefKeyBindingViewController.swift
uiryuu: Fixed

iina/PrefKeyBindingViewController.swift
uiryuu: The default action (OK button or enter) will bring up the Finder and show the conflicting config file

iina/PrefKeyBindingViewController.swift
uiryuu: Added a guard to prevent loading the same config

iina/PrefKeyBindingViewController.swift
uiryuu: Initially I wanted to implement this way, however, I noticed that if the renaming fails, it would be messy to bring up another dialog window to give the user info on why the action failed. So I just used the alert window as other actions (new, duplicate) in this window.

iina/PrefKeyBindingViewController.swift
uiryuu: `isLoading` is preventing infinite loops, since we call `loadConfigFile` in `tableViewSelectionDidChange`, and we also call `reloadData` in `loadConfigFile`, which will cause another `tableViewSelectionDidChange`, thus causing an infinite loop.

svobs:
> In the latest commit, I removed Preference.Key.inputConfigs and made userConfigs a computed variable, so that even if the user adds a file to the folder manually, the table will show the file when refreshed. This commit needs more testing.

Very much appreciate the update! And yes @lhc70000 is right about the performance issue when checking the file system each time. IMHO it's still a good improvement to do a lazy property for now.

Later (or hopefully sooner) a good way to do a live update of the list would be to make a `DispatchSourceFileSystemObject` for the directory so we can get notified when its contents change, and do a reload in response to those. The author of #3724 posted some helpful links showing how to set this up. But it does take more than a little code to set up, so probably better to leave for later.

iina/PrefKeyBindingViewController.swift
```diff
@@ -341,66 +263,103 @@ class PrefKeyBindingViewController: NSViewController, PreferenceWindowEmbeddable
...
+    guard let configName = configName else { fallback(); return }
+    
+    cachedConfigNames = configNames
+    confTableView.reloadData()
+    if let index = cachedConfigNames.firstIndex(of: configName) {
+      confTableView.selectRowIndexes(IndexSet(integer: index), byExtendingSelection: false)
+    }
+    currentConfName = configName
+    currentConfFilePath = getFilePath(forConfig: configName)!
+
```
svobs, line 289: If you move around the last few lines like this, you can get rid of the `isLoading` variable completely:

```Swift

    guard let configName = configName else { fallback(); return }
    
    cachedConfigNames = configNames
    currentConfName = configName
    currentConfFilePath = getFilePath(forConfig: configName)!

    confTableView.reloadData()
    if let index = cachedConfigNames.firstIndex(of: configName) {
      confTableView.selectRowIndexes(IndexSet(integer: index), byExtendingSelection: false)
    }

```

uiryuu: Thanks for your suggestion. I too, indeed want to get rid of `isLoading`, however, I tried, it caused more edge cases which will make the code messier. It's not worth it to explain how it would cause problems which requires more code to deal with, but trust me here :)

svobs: Would be good to at least change the name of the variable to something more specific (e.g., "isLoadingConf" and maybe copy your comment above about infinite loops into a code comment to make it easier to figure out its purpose.

low-batt (changes requested):
Pulled PR, built IINA, started IINA, opened settings. clicked on `Key Bindings` and the panel did not display any bindings for `IINA Default`:
bindings

If I click on another configuration the bindings are shown.

low-batt (approved):
I confirmed the latest commit fixes the issue with the bindings not being drawn when the panel first comes up.

I would think it would be better to follow the conventions of `Finder` where you press return and the name becomes editable. `Finder` also has to deal with the case where the file can't be renamed, like when you make the directory read only. `Finder` puts up an alert when that happens.

uiryuu:
> I would think it would be better to follow the conventions of Finder where you press return and the name becomes editable. Finder also has to deal with the case where the file can't be renamed, like when you make the directory read only. Finder puts up an alert when that happens.

We can make this happen in another PR

lhc70000 (approved):

low-batt:
This was added in IINA 1.4.0.

PR #4963, opened 2024-05-29
Reduce MVPController logging

low-batt:
This commit will:
- Add a log level parameter to MVPController methods for setting properties and executing commands
- Change code that emits a lot of log message to use verbose for the log level

---

**Description:**


uiryuu:
What about making `.verbose` the default and only manually change to `.debug` if we think it's important?

low-batt:
We could do that. I intentionally did it this way to match the default used by the log method.

uiryuu (approved):
> I intentionally did it this way to match the default used by the log method.

I think it's fair. We can merge this first and adjust later based on our needs later.

PR #4964, opened 2024-05-30
Close enable screensaver while playing audio files, #4954

### LINKED ISSUE #4954: Enable Screensaver While Playing Audio Files in IINA Media Player
**Summary**:  
I would like to request an enhancement for the IINA media player on Mac to enable the screensaver while playing audio files. This feature would allow users to conserve energy and prevent screen burn-in when only audio playback is needed.

**Description**:  
Currently, IINA disables the screensaver for both video and audio playback. While this is beneficial for video files to prevent interruptions, it is not necessary for audio files.

low-batt:
This commit will:
- Add new settings `preventDisplaySleepForAudio` and `preventDisplaySleepForVideo`
- Change the `PlayerCore` `checkCurrentMediaIsAudio` method into an `isAudio` `PlaybackInfo` property
- Add a `Require display to stay on while actively playing video` checkbox to the settings window `General` tab
- Add a similar checkbox for playing audio
- Change the `PlayerCore.checkStatusForSleep` method to support the new settings
- Add logging to `SleepPreventer`

- [x] This implements/fixes issue #4954.

---

**Description:**

### DIFF
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -184,7 +184,6 @@
 
 "alert.fatal_error" = "Fatal error: %@ \nThe application will exit now.";
 "alert.mpv_error" = "Internal error %@ (%@) when setting option %@.";
-"alert.sleep" = "Cannot prevent display sleep! (0x%0X8)";
 
 "alert.unsupported_audio" = "Unsupported external audio file.";
 "alert.unsupported_sub" = "Unsupported external subtitle.";
--- a/iina/Base.lproj/PrefGeneralViewController.xib
+++ b/iina/Base.lproj/PrefGeneralViewController.xib
@@ -24,19 +24,19 @@
             <point key="canvasLocation" x="-1607" y="64"/>
         </customView>
         <customView id="6zR-96-iKB" userLabel="Prefs &gt; General &gt; Behavior">
-            <rect key="frame" x="0.0" y="0.0" width="480" height="501"/>
+            <rect key="frame" x="0.0" y="0.0" width="480" height="557"/>
             <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMinY="YES"/>
             <subviews>
-                <textField identifier="SectionTitleBehavior" horizontalHuggingPriority="251" verticalHuggingPriority="750" allowsCharacterPickerTouchBarItem="YES" translatesAutoresizingMaskIntoConstraints="NO" id="38y-2I-MGs">
-                    <rect key="frame" x="-2" y="477" width="66" height="16"/>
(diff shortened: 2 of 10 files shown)


uiryuu:
I don't think there's a need for allowing to go in screensaver for audio and not for video. Then the options could be simplified into:
- [ ] Prevent screensaver
    - [ ] Not apply when playing audio file or in music mode

This way we can also handle music mode correctly, since maybe someone will play a video file but actually only want the audio, so they enter music mode. Of cause you can even differentiate audio file and music mode and split them into 2 options.

low-batt:
I will make that change. I struggled with what to name this setting. I based it on how Apple describes the setting we are using, [kIOPMAssertionTypeNoDisplaySleep](https://developer.apple.com/documentation/iokit/kiopmassertiontypenodisplaysleep):
> The idle display will not sleep when enabled, and consequently the system will not idle sleep.

I do think the wording needs to make it clear that this only kicks in when playback is active.

low-batt:
I think I have to make more changes to properly implement this. Is the functionality we normally want for audio/music mode "allow screen saver, but prevent system sleep"? If so, maybe instead of [kIOPMAssertionTypeNoDisplaySleep](https://developer.apple.com/documentation/iokit/kiopmassertiontypenodisplaysleep) we should switch to using [beginActivity](https://developer.apple.com/documentation/foundation/processinfo/1415995-beginactivity) and the flags [idleDisplaySleepDisabled](idleDisplaySleepDisabled) and [idleSystemSleepDisabled](https://developer.apple.com/documentation/foundation/processinfo/activityoptions/1409849-idlesystemsleepdisabled). When not suppressing the screen saver for audio we would still not allow the system to sleep. The `beginActivity` method does not report errors, so we would be loosing the current error reporting we do. Not sure what that method would do if macOS was broken as was encountered in the past.

Another question is whether we should be calling `beginActivity` with [userInitiated](https://developer.apple.com/documentation/foundation/processinfo/activityoptions/1412233-userinitiated) when playback is active.

Thoughts?

uiryuu:
I think using `ProcessInfo` methods to control the system behavior is clearly the better way than what we are currently doing. We can definitely try using that in the upcoming beta versions.

> Another question is whether we should be calling beginActivity with [userInitiated](https://developer.apple.com/documentation/foundation/processinfo/activityoptions/1412233-userinitiated) when playback is active.

I don't have a strong opinion on this, but maybe I slightly prefer `.background`. Need to hear other devs' opinions.

low-batt:
I will prepare an update to this that includes switching to `ProcessInfo`. We definitely do not want to use `.background`. Playing audio is a real time operation. Failing to post a buffer in time can result stuff like pops and clicks.

Definitely run this past the other developers, especially @saagarjha.

low-batt:
This is what the new settings look like:
prevent-2

uiryuu:
Merged locally by eff7512e2c52065bab02a0aacda0260b5a4010c0 to avoid conflicts.

PR #4967, opened 2024-06-01
Fix Touch Bar time remaining label in RTL, #4946

low-batt:
This commit will change the touchBar method in TouchBarSupport to set baseWritingDirection to leftToRight on the DurationDisplayTextField that shows time remaining.

This is a follow-up to PR #4958 which changed the baseWritingDirection in the XIB for the OSC. That fix is correct, but does not fix the DurationDisplayTextField created at runtime by  TouchBarSupport. This commit corrects the TouchBarSupport use case.

- [x] This implements/fixes issue #4946.

---

**Description:**


uiryuu (approved):
Pulled and tested.

PR #4970, opened 2024-06-01
Fix RTL causes date in about window to be malformed, #4969

### LINKED ISSUE #4969: The compilation date in the debug build in RTL languages seems not rendered correctly
**System and IINA version:**

- macOS 14.5
- IINA 1.3.5

**Expected behavior:**
Correctly rendered

**Actual behavior:**
image

**Steps to reproduce:**
- Use a debug (or nightly) build in Hebrew
- Click "About IINA" in the menu
- Observe the date field

Always

low-batt:
This commit will:
- Change the `Text Direction` to be `Left To Right` for the `Date` text field in `AboutWindowController.xib`
- Change the `Mirror` attribute to be `Never` for the `Date` text field in `AboutWindowController.xib`
- Change the type of the `buildDate` property in `InfoDictionary` to be `Date?` instead of `String?`
- Change `windowDidLoad` in `AboutWindowController` to format `buildDate` as a localized string
- Change `logBuildDetails` in `AppDelegate` to format `buildDate` using the `en_US` locale

These changes correct the localized build date displayed in the about window. They also change the build date displayed in the log file to not use the user's locale.

- [x] This implements/fixes issue #4969.

---

**Description:**

### DIFF
--- a/iina/AboutWindowController.swift
+++ b/iina/AboutWindowController.swift
@@ -71,11 +71,16 @@ class AboutWindowController: NSWindowController {
     mpvVersionLabel.stringValue = PlayerCore.active.mpv.mpvVersion
     ffmpegVersionLabel.stringValue = "FFmpeg \(String(cString: av_version_info()))"
 
+    // Use a localized date for the build date.
+    let toString = DateFormatter()
+    toString.dateStyle = .medium
+    toString.timeStyle = .medium
+
     switch InfoDictionary.shared.buildType {
     case .nightly:
       if let buildDate = InfoDictionary.shared.buildDate,
          let buildSHA = InfoDictionary.shared.shortCommitSHA {
-        buildDateLabel.stringValue = buildDate
+        buildDateLabel.stringValue = toString.string(from: buildDate)
         buildDateLabel.isHidden = false
         buildBranchButton.title = "NIGHTLY " + buildSHA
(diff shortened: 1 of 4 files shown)


ShlomoCode:
Additional issues in the About window in the "Contributors" tab (the arrows should be reversed like in https://github.com/iina/iina/pull/4958, and the line "powered by Crowdin" should be localizable):
CleanShot 2024-06-02 at 04 19 12@2x

This should also be localized:
CleanShot 2024-06-02 at 04 43 42@2x

uiryuu:
> This should also be localized:
CleanShot 2024-06-02 at 04 43 42@2x

It's not worthy to maintain the localization for this. The only valid string to localize is "Website", but to maintain this file needs a lot of effort.

ShlomoCode:
Why not translate for example "The modern video player for macOS"?

uiryuu:
It's the slogan of the app, which is fairly important, but we cannot verify and guarantee that every localization is good. Plus translating it is very subjective, people might disagree on it.

low-batt:
This is what the about window looks like with the latest commit:
about

uiryuu:
A last confirmation from @ShlomoCode before we can merge this...

ShlomoCode:
yes it's normal now
I don't really like the time format when it's PM, I think a time in 24 hour format is more readable, but I understand it's Apple's design?

PR #4971, opened 2024-06-02
Fix incorrect text alignment in the battery icon in RTL, #4957

### LINKED ISSUE #4957: Incorrect text alignment in the battery icon of the OSD in RTL
**System and IINA version:**

- macOS 14.4.1
- IINA 1.3.4 + PR #4922 + PR #4889 or #develop (https://github.com/iina/iina/commit/13d16eb223d50575d931ef68e9561b3730f7163b)

The text inside the battery icon (OSD) should be aligned to the right in RTL 

CleanShot 2024-05-29 at 03 07 51@2x

**Expected behavior:**

The text is right aligned

**Actual behavior:**

The text is aligned to the left

**Alternatives I considered:**

Always align to the center

**Steps to reproduce:**

- Under the `Product` menu open the Scheme menu and select `Edit Scheme…`
- Click on the `Options` tab
- In the pull down select `Right-to-Left Pseudolanguage` near the end of the list
- Enable OSD and "Display time and battery info when in full screen" in setings > UI
- Open some video and go to full screen

Every time.

low-batt:
This commit will change the `updateAdditionalInfo` method in `MainWindowController` to adjust the centering constraint for the label inside the battery icon based on the capacity value to be displayed.

- [x] This implements/fixes issue #4957.

---

**Description:**


low-batt:
@ShlomoCode in [this comment](https://github.com/iina/iina/issues/4957#issuecomment-2145396871) shows that the alignment of the capacity within the battery icon could also use some adjustment in LTR. I did not address that as I was unsure as to whether the way I corrected this for RTL was good or whether there was a simple way to deal with this that I missed. Let me know if I should address the LTR alignment as well.

ShlomoCode:
Now in LTR it is aligned to the left and not the center. You wrote "The intent is that the text is centered", so it seems that this should also be taken care of, no?

low-batt:
Yes. LTR needs fixing.

The root cause of the problem is that the space reserved for the capacity in the battery icon is not centered. The icon is larger on the side that represents the plus sign. The icon could be corrected, either by increasing the width of the line represending the minus side of the battery, or by adding white space such that the space for the capacity is centered. That way a simple center X constraint works correctly.

Instead, for the RTL fix I added code that adjusts the constraint at runtime. This works, but might be considered hacky by the senior developers that know more than me about UI programming. Before I put time into doing something similar for LTR I want to know this approach is acceptable.

@lhc70000 Thoughts?

lhc70000:
Updated the icons and removed the constraint. If it looks good to you then we can merge.

ShlomoCode:
Looks good now

PR #4975, opened 2024-06-04
Fix if a NAS share cannot be reached, iina fails to start, #4973

### LINKED ISSUE #4973: If a NAS share cannot be reached, iina fails to start
**System and IINA version:**

- IINA: 1.3.5 Build 141

**Expected behavior:**

Double clicking a local file should just show the movie.

**Actual behavior:**

Double clicking a file opens Iina which then tries to check checks previous links/files (for no reason at all). As my NAS wasn't online I got an never ending popups of ```There was a problem connecting to the server “MyNasServerName”. Check the server name or IP address, and then try again. If you continue to have problems, contact your system administrator.```.

Starting Iina was impossible until I reconnected to the Nas.

**Steps to reproduce:**

I would guess:
* Open a movie on the nas
* Close Iina
* Disconnect the nas
* Open a local file

Every time; force quitting didn't help.

low-batt:
This commit will:
- Add a new preference `enableRecentDocumentsWorkaround`
- Change the `AppDelegate` methods `restoreRecentDocuments` and `saveRecentDocuments` to not do anything if `enableRecentDocumentsWorkaround` has not been set to `true`
- Change `restoreRecentDocuments` to use the options `withoutMounting` and `withoutUI` when creating a URL from bookmark data
- Add code to `restoreRecentDocuments` to create a URL from a string if creating using bookmark data fails

This has to do with the workaround added for issue #4688 where starting with Sonoma macOS erases the list of recent documents when an unsigned IINA executable is run. This affects developers as well as normal users that test nightly builds. These changes correct the problem with IINA causing macOS to prompt the user to mount volumes when the list of recent documents contains files on volumes that are not mounted.

However it is not possible to correct the long delays when the volume is mounted, but unreachable, such as a NAS that macOS does not know yet has been shutdown. For this reason this commit adds a preference that must be set using terminal and the defaults command to enable the workaround.

- [x] This implements/fixes issue #4973.

---

**Description:**

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -1000,6 +1000,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
   /// be restored when `sharedfilelistd` clears its list.
   ///
   /// If the following is true:
+  /// - Workaround has been enabled by setting `enableRecentDocumentsWorkaround`
   /// - Running under macOS Sonoma and above
   /// - Recording of recent files is enabled
   /// - The list in  [NSDocumentController.shared.recentDocumentURLs](https://developer.apple.com/documentation/appkit/nsdocumentcontroller/1514976-recentdocumenturls) is empty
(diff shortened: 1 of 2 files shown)


low-batt:
This fixes the problem with this workaround causing macOS to prompt the user to mount a dismounted volume. _However_ I could not find a way to stop the workaround from causing macOS to prompt the user about a volume it thinks is mounted, but can't be accessed.

As this workaround is for users who test unsigned IINA nightly builds (and anyone who builds IINA themselves), it must not disrupt users who only use signed releases. I don't see a way to accomplish that. So I have added a new preference `enableRecentDocumentsWorkaround` that is set to `false` by default. People needing this workaround will have to execute this command in [Terminal](https://support.apple.com/guide/terminal/welcome/mac):
```text
defaults write com.colliderli.iina enableRecentDocumentsWorkaround true
```

Unfortunate as it is very annoying to have your list of recent documents cleared. But that is a macOS issue. I'm still troubled that I have not been able to find a discussion on the net of this new macOS Sonoma behavior. I would think developers would have noticed this.

lhc70000:
Merged via 9978a1d.

PR #4992, opened 2024-06-13
Always display Find Online Subtitle OSD messages

### LINKED ISSUE #4422: External Subtitles > Search Online... doesn't do anything
**System and IINA version:**

- macOS 10.13.6
- IINA 1.3.1 Build 133

**Expected behavior:**

Clicking on `Search Online...` does something, presumably what is seen on https://github.com/iina/iina/issues/4142

**Actual behavior:**

Nothing happens, not with any of the three provided sources for subtitle search under `Preferences > Subtitle > Online Subtitle`.

This has always happened to me, for years I thought the option was broken and didn't work for anyone at all. Finding the issue above made me realize otherwise. My iina config is too laborious to start fresh and see if it changes anything, but I've got the feeling it has to do with running an old version of macOS and the way the window is called.

low-batt:
This commit will:
- Add an isAlwaysShown property to OSDMessage that is true for OSD messages emitted by the Find Online Subtitles feature
- Change PlayerCore.sendOSD to send messages whose isAlwaysShown property is true even when the enableOSD setting is false
- Change MainWindowController.displayOSD to display messages whose isAlwaysShown property is true even when the IINA OSD is disabled due to the user enabling mpv's OSD

These changes correct a problem where IINA's Find Online Subtitles feature appears to do nothing when the OSD is disabled. This feature uses the OSD to interact with the user. Failing to display the messages breaks the feature.

---

**Description:**

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -2158,7 +2158,7 @@ class PlayerCore: NSObject {
 
   func sendOSD(_ osd: OSDMessage, autoHide: Bool = true, forcedTimeout: Float? = nil, accessoryView: NSView? = nil, context: Any? = nil, external: Bool = false) {
     // querying `mainWindow.isWindowLoaded` will initialize mainWindow unexpectly
-    guard mainWindow.loaded, Preference.bool(for: .enableOSD) || osd.isAlwaysShown else { return }
+    guard mainWindow.loaded, Preference.bool(for: .enableOSD) || osd.alwaysEnabled else { return }
     if info.disableOSDForFileLoading && !external {
       guard case .fileStart = osd else {
         return
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -2158,7 +2158,7 @@ class PlayerCore: NSObject {
 
   func sendOSD(_ osd: OSDMessage, autoHide: Bool = true, forcedTimeout: Float? = nil, accessoryView: NSView? = nil, context: Any? = nil, external: Bool = false) {
     // querying `mainWindow.isWindowLoaded` will initialize mainWindow unexpectly
-    guard mainWindow.loaded && Preference.bool(for: .enableOSD) else { return }
+    guard mainWindow.loaded, Preference.bool(for: .enableOSD) || osd.isAlwaysShown else { return }
     if info.disableOSDForFileLoading && !external {
       guard case .fileStart = osd else {
         return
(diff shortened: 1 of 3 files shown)


low-batt:
This does not address `customWithDetail` which is used by plugins. I'm guessing there may be a similar problem with messages emitted by plugins.

Thoughts?

lhc70000:
`isAlwaysShown` looks confusing to me. Initially I thought it would make the OSD persistently displayed on the screen and not hiding anymore. Maybe it can be something more straightforward, like `alwaysEnabled` or `ignoreDisableOSD`?

low-batt:
Makes sense. I was too close to the code to notice that it could be interpreted that way. I renamed it to `alwaysEnabled`.

We still need to figure out how to properly solve the similar issue for plugins.

svobs:
> Thoughts?

In PR #4991 you wrote this:

> A solution that exposes all (excluding the mandatory ones) OSD messages really should be data driven, so adding a new enum to OSDMessage automatically appears in the UI. Very doable, but a lot more work. This also means the enum can't have "generic" messages that actually represent multiple different kinds of messages like the current "custom" entries do.

This sounds like you have some idea for what you would do for a more comprehensive fix? To throw in my suggestion, I think it would be less code to maintain in the long-term if, instead of adding booleans to `OSDMessage`, just add a `canShow(msg)` func prior to `displayOSD` which will discard the message if it can't be shown. That way, more complex logic can be used, e.g., a switch which checks for all the messages to deny when `PK.enableOSD` is false, in addition to checks for all the individually disabled messages, and the logic will all be in one place. Any added enums could simply default to `true` unless code is added there. Enums for plugins could attach a property containing their own enablement state or specify a block of code to execute to determine it.

Of course if this is just a stopgap fix, then hard to argue with the winding path of progress :)

iina/PlayerCore.swift
```diff
@@ -2158,7 +2158,7 @@ class PlayerCore: NSObject {
 
   func sendOSD(_ osd: OSDMessage, autoHide: Bool = true, forcedTimeout: Float? = nil, accessoryView: NSView? = nil, context: Any? = nil, external: Bool = false) {
     // querying `mainWindow.isWindowLoaded` will initialize mainWindow unexpectly
-    guard mainWindow.loaded && Preference.bool(for: .enableOSD) else { return }
+    guard mainWindow.loaded, Preference.bool(for: .enableOSD) || osd.alwaysEnabled else { return }
```
lhc70000, line 2161: Why using `,` here instead of `&&`?

low-batt: I tried to find what Swift considers the best practice in this case and only found people debating that. The advice in the Medium post [Chaining logical expressions with commas in Swift](https://medium.com/@austinlouden/chaining-logical-expressions-with-commas-in-swift-c3755ee784e9) about  expressions using commas being easier to read than complex expressions reflects what I was thinking. Using `&&` would require adding `()` around the `||` expression. I was also thinking that the comma was more like swift has you have to use it when using `let`.

I'm good with whatever style the group wants to use. Let me know if you want me to switch this to `&&`.

lhc70000: I was just asking and also found it a little bit confusing. I think it's alright to use `,` here to avoid the parentheses.

low-batt:

PR #4996, opened 2024-06-15
Allow closing sub download OSD using escape

Cykelero:

---

Currently, it's impossible to dismiss the “online subtitle download” OSD using the keyboard. This PR allows using escape to do just that.

- Now, when escape is pressed, if a persistent OSD is being shown (non-interruptible, and without a scheduled hide), `MainWindowController` gives the OSD view the opportunity to handle the event.
- The `keyBinding.normalizedMpvKey == "ESC"` condition might be safely removed, allowing OSD VCs to opt into handling any event; but I'm not confident enough that it won't disrupt anything.  
Specifically, the forwarding is made necessary by [the `MainWindow.keyDown()` override](https://github.com/iina/iina/blob/7e4ed44666496133299c2c24b8a4b01fba926028/iina/MainWindow.swift#L18), which I don't fully understand.
- It's necessary to check `hideOSDTimer` in addition to `isShowingPersistentOSD`, as this boolean is sometimes true even for auto-dismissed OSDs (i.e. the “screenshot taken” OSD)

### DIFF
--- a/iina/Base.lproj/SubChooseViewController.xib
+++ b/iina/Base.lproj/SubChooseViewController.xib
@@ -134,6 +134,9 @@
                     <buttonCell key="cell" type="square" title="CANCEL" bezelStyle="shadowlessSquare" alignment="center" inset="2" id="pPz-om-Xit">
                         <behavior key="behavior" pushIn="YES" lightByBackground="YES" lightByGray="YES"/>
                         <font key="font" metaFont="systemBold"/>
+                        <string key="keyEquivalent" base64-UTF8="YES">
+Gw
+</string>
                     </buttonCell>
                     <connections>
                         <action selector="cancelBtnAction:" target="-2" id="trN-ei-1wo"/>
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -833,19 +833,22 @@ class MainWindowController: PlayerWindowController {
 
   // MARK: - Mouse / Trackpad events
 
-  @discardableResult
-  override func handleKeyBinding(_ keyBinding: KeyMapping) -> Bool {
-    if
-      keyBinding.normalizedMpvKey == "ESC",
-      isShowingPersistentOSD,
-      hideOSDTimer?.isValid != true,
-      let currentEvent = NSApp.currentEvent
-    {
-      if osdStackView.performKeyEquivalent(with: currentEvent) {
-        return true
+  override func keyDown(with event: NSEvent) {
+    if isShowingPersistentOSD {
+      let keyCode = KeyCodeHelper.mpvKeyCode(from: event)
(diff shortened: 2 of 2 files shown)


svobs:
Looks like a good idea to me, but it needs some changes to the implementation. I would keep the changes you made to `SubChooseViewController.xib`, but replace the changes to `MainWindowController.swift` with this:

```Swift
  override func keyDown(with event: NSEvent) {
    if isShowingPersistentOSD {
      let keyCode = KeyCodeHelper.mpvKeyCode(from: event)
      let normalizedKeyCode = KeyCodeHelper.normalizeMpv(keyCode)
      if normalizedKeyCode == "ESC", osdStackView.performKeyEquivalent(with: event) {
        Logger.log("ESC key was handled by OSD", level: .verbose)
        return
      }
    }

    super.keyDown(with: event)
  }
```

Rationale:
1. You need to add code in `keyDown` instead of `handleKeyBinding`, because that will only get called if there is a key binding for `ESC` in the `Settings` > `Key Bindings` table. (And that binding would have to be for a different action because there is no support for any kind of "dismiss OSD" action there.)
2. Check for `isShowingPersistentOSD` first because it's the fastest check and will require less CPU when it is false (which is most of the time).
3. No need to check `hideOSDTimer` now. The function `osdStackView.performKeyEquivalent` will return `true` only if it recognized the event and accepted it. Other OSDs such as "screenshot taken" do not recognize `ESC` and will return false, so those will fall through and call `super.keyDown` which will allow the key to be processed normally.

Cykelero:
Thank you for the feedback! These are good points—I've updated the PR with exactly the code you suggested.

lhc70000 (commented):
Looks good to me, but I think `keyUp` is a better place for key shortcuts. `keyDown` can be called repeatedly if the key is long-pressed, while `keyUp` is only fired once and represents the action better. Could you change it?

uiryuu:
> Looks good to me, but I think keyUp is a better place for key shortcuts. keyDown can be called repeatedly if the key is long-pressed, while keyUp is only fired once and represents the action better. Could you change it?

I don't think so. Changing it to `keyUp` the window will be close when the user releases the ESC key, however this is not how the system handle ESC when it's been used to close aux window.  Try opening Music.app, and open the settings of it. The settings window can be closed via ESC, and it is immediately closed when you press ESC key, instead of waiting until you release the key. Besides, using `keyDown` shouldn't cause any performance issue in this context.

Cykelero:
I'd agree with staying on `keyDown` for this.
In general it's the way macOS handles key equivalents, but here specifically, switching to `keyUp` actually makes the functionality glitchier: since the key down event is let through, it causes the window to exit fullscreen (or perform any other action that's bound to the key) in addition to then closing the OSD.

lhc70000:
Sounds reasonable, you are right. @Cykelero thanks for your contribution!

low-batt:

Issue #3976, opened 2022-10-10
"Reveal in Finder" should be renamed to "Show in Finder"


svobs:
**System and IINA version:**

- macOS 12.6
- IINA 1.3.0 Build 132

Not sure when it happened, but the standard English language label for this functionality appears to be "Show in Finder", not "Reveal in Finder". IINA should be updated to match.

Examples:

1. Open Chrome. Go to "chrome://downloads/"
2. Open a project in XCode. Right-click on a file in the Navigator.
3. Right-click on any application in the Dock > Options > Show in Finder

**Expected behavior:**
Preferences > Key Bindings has a button labeled "Show config file in Finder".
File > Playback History -> (right-click on a row in Player History window) > context menu has an item labelled "Show in Finder"
Preferences > Plugins has a button labeled "Show in Finder"

**Actual behavior:**
Preferences > Key Bindings has a button labeled "Reveal config file in Finder"
File > Playback History -> (right-click on a row in Player History window) > context menu has an item labelled "Reveal in Finder"
Preferences > Plugins has a button labeled "Reveal in Finder"

low-batt:
Other actions in IINA that bring up a `Finder` window:
-  `Go to Screenshot Folder` in the `Video` menu
- `Open log directory` in `Advanced` preferences

uiryuu:
I'll try to make a fix. It might be complicated than just renaming the texts since I also want to rename the corresponding key (`pl_menu.reveal_in_finder`), and I'm not sure how Crowdin would handle such key change.

uiryuu:
`"alert.config.file_existing.message" = "This should not happen. Choose OK to overwrite, Cancel to reveal the file in Finder.";`
Another "reveal" spotted.

uiryuu:
<img width="405" alt="image" src="https://user-images.githubusercontent.com/20237141/195552216-b988c737-4fb1-4555-8463-bbccfc944adf.png">
This is in the OSD after a screenshot is taken. What do you guys think about this one? (I'm not a native English speaker)

svobs:
> This is in the OSD after a screenshot is taken. What do you guys think about this one? (I'm not a native English speaker)

My vote is to change to "Show" as well. It looks like another case of "Show in Finder" functionality, but due to the limited space and use of block character font it makes sense to shorten it to one word - to avoid ambiguity about the bounds of the button/link.

svobs:
[...]

PR #5005, opened 2024-06-21
Add “Show Current File in Finder” menu item (#5001)

### LINKED ISSUE #5001: add Show in Finder menu command
**What you want IINA to do:**

- Conveniently show the currently playing media's file in the Finder via menu command.
 
- Even more conveniently do so with a keyboard shortcut. 

Of the two, I think the keyboard shortcut is more important, so I'd be happy with that without the menu item. I vote for a default keyboard shortcut of `⌘R`.

**What IINA does currently:**

The only way that I can find to do this is by showing the playlist, right-click the item, and choose `Show in Finder`. It is a pain, and  it is not keyboard-shortcut-able.

One can right-click on the filename in the titlebar, then click on the containing folder in the dropdown that appears, but this is also inconvenient and not keyboard-shortcut-able.

It's a common need. 

IINA has a `File:Delete Current File` command, so it seems odd that the more general ability to do whatever you want with the file via the Finder is missing.

Many apps have such a menu command, though named differently:

`Show in Finder`: XCode, macOS Dock, BBEdit
`Reveal in Finder`: Sublime, Transmission, VLC
`Open in Finder`:  Jetbrains IDEs
`Open File Location`: Typora

ShlomoCode:
- [x] This implements/fixes issue #5001.

---

**Description:**
Adds a menu bar item (under `File`) to reveal the current file in Finder.
~~For some reason I was unable to set a Cmd+R shortcut, even though I set it through the interface builder, the shortcut doesn't work, so the PR does not currently include a shortcut.~~
Built and tested with XCode 15.1, on MacOS 14.5.

### DIFF
--- a/iina/Base.lproj/KeyBinding.strings
+++ b/iina/Base.lproj/KeyBinding.strings
@@ -52,6 +52,7 @@
 "iina.toggle-flip" = "Toggle flip";
 "iina.toggle-mirror" = "Toggle mirror";
 "iina.save-playlist" = "Save current playlist";
+"iina.show-current-file-in-finder" = "Show current file in Finder";
 "iina.delete-current-file" = "Delete current file";
 "iina.delete-current-file-hard" = "Delete current file directly without moving to trash";
 "iina.find-online-subs" = "Find online subtitles";
--- a/iina/Base.lproj/MainMenu.xib
+++ b/iina/Base.lproj/MainMenu.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="32700.99.1234" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="22505" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="22690"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="22505"/>
     </dependencies>
     <objects>
         <customObject id="-2" userLabel="File's Owner" customClass="NSApplication">
@@ -130,6 +130,9 @@
                                 </connections>
                             </menuItem>
                             <menuItem isSeparatorItem="YES" id="N0Z-h7-JSg"/>
+                            <menuItem title="Show Current File in Finder" id="che-dy-QUU">
+                                <modifierMask key="keyEquivalentModifierMask"/>
+                            </menuItem>
                             <menuItem title="Delete Current File" id="npH-8B-XA4">
                                 <modifierMask key="keyEquivalentModifierMask"/>
                             </menuItem>
@@ -781,6 +784,7 @@ CA
                 <outlet property="resetAudioDelay" destination="H3d-jR-YCr" id="Qx9-n3-F90"/>
                 <outlet property="resetSubDelay" destination="SLr-6g-Mgh" id="8oW-hm-TNq"/>
                 <outlet property="resetTextSize" destination="P1a-ot-P31" id="fJk-GP-GBM"/>
+                <outlet property="showCurrentFileInFinder" destination="che-dy-QUU" id="p1H-Xj-dbg"/>
                 <outlet property="rotationMenu" destination="5xN-ZB-VLI" id="YGx-Ae-oMp"/>
                 <outlet property="saveDownloadedSub" destination="9zk-co-roT" id="lGr-mb-oHw"/>
                 <outlet property="savePlaylist" destination="Mjr-lL-oEt" id="ftc-bn-4Qi"/>
--- a/iina/MainMenuActions.swift
+++ b/iina/MainMenuActions.swift
@@ -45,6 +45,12 @@ class MainMenuActionHandler: NSResponder {
     }
   }
 
+  @objc func menuShowCurrentFileInFinder(_ sender: NSMenuItem) {
+    guard let url = player.info.currentURL else { return }
+    guard !player.info.isNetworkResource else { return }
+    NSWorkspace.shared.activateFileViewerSelecting([url])
+  }
+
   @objc func menuDeleteCurrentFile(_ sender: NSMenuItem) {
     guard let url = player.info.currentURL else { return }
     do {
--- a/iina/MainMenuActions.swift
+++ b/iina/MainMenuActions.swift
@@ -9,7 +9,7 @@
 import Cocoa
 
 
-class MainMenuActionHandler: NSResponder {
+class MainMenuActionHandler: NSResponder, NSMenuItemValidation {
 
   unowned var player: PlayerCore
 
@@ -466,6 +466,16 @@ extension MainMenuActionHandler {
     }
   }
 
+  func validateMenuItem(_ menuItem: NSMenuItem) -> Bool {
+    switch menuItem.action {
+    case #selector(menuShowCurrentFileInFinder(_:)):
+      return player.info.currentURL != nil && !player.info.isNetworkResource
+    default:
+      break
+    }
+    return menuItem.isEnabled
+  }
+  
   // MARK: - Plugin
 
   @objc func reloadAllPlugins(_ sender: NSMenuItem) {
--- a/iina/MainMenuActions.swift
+++ b/iina/MainMenuActions.swift
@@ -46,8 +46,7 @@ class MainMenuActionHandler: NSResponder, NSMenuItemValidation {
   }
 
   @objc func menuShowCurrentFileInFinder(_ sender: NSMenuItem) {
-    guard let url = player.info.currentURL else { return }
-    guard !player.info.isNetworkResource else { return }
+    guard let url = player.info.currentURL, !player.info.isNetworkResource else { return }
     NSWorkspace.shared.activateFileViewerSelecting([url])
   }
 
--- a/iina/IINACommand.swift
+++ b/iina/IINACommand.swift
@@ -30,6 +30,7 @@ enum IINACommand: String {
   case fitToScreen = "fit-to-screen"
 
   case saveCurrentPlaylist = "save-playlist"
+  case showCurrentFileInFinder = "show-current-file-in-finder"
   case deleteCurrentFile = "delete-current-file"
   case deleteCurrentFileHard = "delete-current-file-hard"
 
--- a/iina/KeyBindingDataLoader.swift
+++ b/iina/KeyBindingDataLoader.swift
@@ -57,6 +57,7 @@ class KeyBindingDataLoader {
     KBI("open-file", type: .iinaCmd),
     KBI("open-url", type: .iinaCmd),
     KBI("save-playlist", type: .iinaCmd),
+    KBI("show-current-file-in-finder", type: .iinaCmd),
     KBI("delete-current-file", type: .iinaCmd),
     KBI("delete-current-file-hard", type: .iinaCmd),
     KBI.separator(),
--- a/iina/MenuController.swift
+++ b/iina/MenuController.swift
@@ -86,6 +86,7 @@ class MenuController: NSObject, NSMenuDelegate {
   @IBOutlet weak var openURL: NSMenuItem!
   @IBOutlet weak var openURLAlternative: NSMenuItem!
   @IBOutlet weak var savePlaylist: NSMenuItem!
+  @IBOutlet weak var showCurrentFileInFinder: NSMenuItem!
   @IBOutlet weak var deleteCurrentFile: NSMenuItem!
   @IBOutlet weak var newWindow: NSMenuItem!
   @IBOutlet weak var newWindowSeparator: NSMenuItem!
@@ -221,6 +222,7 @@ class MenuController: NSObject, NSMenuDelegate {
     stringForOpenURLAlternative = openURLAlternative.title
 
     savePlaylist.action = #selector(MainMenuActionHandler.menuSavePlaylist(_:))
+    showCurrentFileInFinder.action = #selector(MainMenuActionHandler.menuShowCurrentFileInFinder(_:))
     deleteCurrentFile.action = #selector(MainMenuActionHandler.menuDeleteCurrentFile(_:))
 
     if Preference.bool(for: .enableCmdN) {
@@ -826,6 +828,7 @@ class MenuController: NSObject, NSMenuDelegate {
 
   func updateKeyEquivalentsFrom(_ keyBindings: [KeyMapping]) {
     var settings: [(NSMenuItem, Bool, [String], Bool, ClosedRange<Double>?, String?)] = [
+      (showCurrentFileInFinder, true, ["show-current-file-in-finder"], false, nil, nil),
       (deleteCurrentFile, true, ["delete-current-file"], false, nil, nil),
       (savePlaylist, true, ["save-playlist"], false, nil, nil),
       (quickSettingsVideo, true, ["video-panel"], false, nil, nil),
--- a/iina/PlayerWindowController.swift
+++ b/iina/PlayerWindowController.swift
@@ -634,6 +634,8 @@ class PlayerWindowController: NSWindowController, NSWindowDelegate {
       menuActionHandler.menuToggleMirror(.dummy)
     case .saveCurrentPlaylist:
       menuActionHandler.menuSavePlaylist(.dummy)
+    case .showCurrentFileInFinder:
+      menuActionHandler.menuShowCurrentFileInFinder(.dummy)
     case .deleteCurrentFile:
       menuActionHandler.menuDeleteCurrentFile(.dummy)
     case .findOnlineSubs:
--- a/iina/config/iina-default-input.conf
+++ b/iina/config/iina-default-input.conf
@@ -48,6 +48,7 @@ Meta+2 set window-scale 2
 #@iina Meta+= bigger-window
 
 #@iina Ctrl+Meta+p toggle-pip
+#@iina Shift+Meta+R show-current-file-in-finder
 Ctrl+Meta+f cycle fullscreen
 Ctrl+Meta+t cycle ontop
 
--- a/iina/config/input.conf
+++ b/iina/config/input.conf
@@ -31,6 +31,7 @@
 #@iina Shift+Meta+p playlist-panel
 #@iina Shift+Meta+c chapter-panel
 #@iina Ctrl+Meta+p toggle-pip
+#@iina Shift+Meta+R show-current-file-in-finder
 
 MBTN_LEFT     ignore              # don't do anything
 MBTN_LEFT_DBL cycle fullscreen    # toggle fullscreen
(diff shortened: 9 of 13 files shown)


iina/Base.lproj/KeyBinding.strings
```diff
@@ -52,6 +52,7 @@
 "iina.toggle-flip" = "Toggle flip";
 "iina.toggle-mirror" = "Toggle mirror";
 "iina.save-playlist" = "Save current playlist";
+"iina.reveal-current-file" = "Reveal current file in Finder";
```
low-batt: I will try and find time for a proper review tomorrow. As per issue #3976 this must be show, not reveal. I'd prefer just `Show in Finder` as that is the common label for such a menu item. I think it is ok that it differs from the label for deleting. See what the other reviewers say.

low-batt:
@svobs Should be able to explain how to add the shortcut for the menu item.

iina/Base.lproj/KeyBinding.strings
low-batt: The way our integration with Crowdin works whenever you change a file in Base.lproj the corresponding file in en.lproj needs to be updated as well. No other language files should be changed. They will be updated through Crowdin.

iina/MainMenuActions.swift
```diff
@@ -44,7 +44,12 @@ class MainMenuActionHandler: NSResponder {
       }
     }
   }
-
```
uiryuu, line 47: Remove trailing spaces

uiryuu:
I think adding the keybinding in "IINA Default" input config and it can be automatically matched and show up in the menu.

iina/Base.lproj/MainMenu.xib
```diff
@@ -130,6 +130,9 @@
                                 </connections>
                             </menuItem>
                             <menuItem isSeparatorItem="YES" id="N0Z-h7-JSg"/>
+                            <menuItem title="Reveal Current File in Finder" id="che-dy-QUU">
```
low-batt: Similar to KeyBinding.strings, the MainMenu.strings in en.lproj must be updated with the new text.

iina/Base.lproj/KeyBinding.strings
uiryuu: I'd like "Show current file in Finder". "Show in Finder" is kinda weird to me since there's no subject.

uiryuu (changes requested):
Apart from what @low-batt has indicated, I have nothing to say. We should consider adding this into the default configuration, and we can add after merging this PR. The problem is should we add this to all default configs (mpv, VLC, Movist)?

iina/Base.lproj/KeyBinding.strings
ShlomoCode: I don't see a significant difference in popularity
https://github.com/search?q=%22Reveal+in+Finder%22&type=code 10.2k
https://github.com/search?q=%22Show+in+Finder%22&type=code 10.6k
In VSCode it's "Reveal", in XCode it's "Show". Change to Show?

iina/Base.lproj/MainMenu.xib
ShlomoCode: Am I supposed to do it manually or is there some sync command in XCode?

iina/Base.lproj/KeyBinding.strings
uiryuu: We had a discussion in #3976 before and the conclusion was to use "show" instead of "reveal" all across IINA. So please change to show to make it consistent with other parts of IINA.

iina/MainMenuActions.swift
ShlomoCode: done

iina/Base.lproj/KeyBinding.strings
low-batt: Go with @uiryuu "Show current file in Finder" although I prefer "Show in Finder" as that is the typical Apple menu item, like in Xcode. The capitalization needs to be "Show Current File in Finder" as per Apple HIG.

The way I update the en.lproj strings file for a XIB is to use [ibtool](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/MaintaingYourOwnStringsFiles/MaintaingYourOwnStringsFiles.html#//apple_ref/doc/uid/10000171i-CH19-SW6) like so:
```text
ibtool iina/Base.lproj/MainMenu.xib --generate-strings-file temp.strings
```

Then I run [meld](https://github.com/yousseb/meld):
```text
meld temp.strings iina/en.lproj/MainMenu.strings
```

And merge in only the string I added.

iina/Base.lproj/MainMenu.xib
low-batt: See my latest comment on how I use `ibtool` to do it.

iina/Base.lproj/KeyBinding.strings
ShlomoCode: BTW note that the screenshot OSD use "REVEAL" word

iina/MainMenuActions.swift
```diff
@@ -45,6 +45,11 @@ class MainMenuActionHandler: NSResponder {
     }
   }
 
+  @objc func menuShowCurrentFileInFinder(_ sender: NSMenuItem) {
+    guard let url = player.info.currentURL else { return }
+    NSWorkspace.shared.activateFileViewerSelecting([url])
```
low-batt, line 50: Missing check for `isNetworkResource`. Not sure how `activateFileViewerSelecting` will react if given a URL to a video on a web site. Have a look at the code in `PlaylistViewController` that calls `activateFileViewerSelecting`.

low-batt:
We do need to get the menu shortcut working. The one suggested is ⌘R. I checked [Mac keyboard shortcuts](https://support.apple.com/en-us/102650) and this is how Apple uses that key:
> Command-R: (1) When an alias is selected in the Finder: show the original file for the selected alias. (2) In some apps, such as Calendar or Safari, refresh or reload the page. (3) In Software Update, check for software updates again.

I don't think it is unreasonable to use ⌘R, but I'm not that familiar with IINA's key binding code. Need to hear from @svobs and @lhc70000.

iina/Base.lproj/KeyBinding.strings
ShlomoCode: done

iina/MainMenuActions.swift
ShlomoCode: The approach taken in the `PlaylistViewController` is to simply ignore such items, and this is indeed necessary to avoid an error when using the keyboard shortcut. I applied it. However, in our case it would have been better to disable the button in advance (and display it as gray). Is it possible?

ShlomoCode: I noticed that the similar item, "Delete current file", throws an error when trying to delete a network file:
341644475-8f15b23e-6b94-4d5d-bb33-abfef2af616f
We will have to apply the check there as well.

uiryuu (approved):
I'm okay with the changes. I'll try to abstract the logic of `@IBAction func contextMenuShowInFinder(_ sender: NSMenuItem)` after merging this PR. Not sure whether or not that's worthwhile

svobs:
> @svobs Should be able to explain how to add the shortcut for the menu item.

Add the 2 lines below to the case statement of `handleIINACommand`, in `PlayerWindowController.swift`:

```Swift
internal func handleIINACommand(_ cmd: IINACommand) {
    switch cmd {
    // […]
    case .showCurrentFileInFinder:
      menuActionHandler.menuShowCurrentFileInFinder(.dummy)
    // […]
```

svobs:
Having connectivity problems here...strange... After adding the code I posted above, everything will be in place for the end user to add a row like this one to your Key Bindings configuration:
Key column: `Meta+r`
Action column: `#@iina show-current-file-in-finder`

iina/MainMenuActions.swift
ShlomoCode: > In our case it would have been better to disable the button in advance (and display it as gray). Is it possible?

@uiryuu?

ShlomoCode:
I confirm that it is now possible to set a keyboard shortcut in the settings and it works.
CleanShot 2024-06-21 at 14 18 08@2x

One thing I noticed that when I use two screens MacOS opens the Finder on the built-in screen and does not focus on it, meaning the focus remains on the IINA window. When I only use the built-in screen it doesn't happen. But I'm guessing this is a macOS bug and not something we should be dealing with.

iina/MainMenuActions.swift
ShlomoCode: > In our case it would have been better to disable the button in advance (and display it as gray). Is it possible?

I implemented this using `validateMenuItem`, like in `PlaylistViewController` and `HistoryWindowController`.

We will have to handle `menuDeleteCurrentFile` in the same way (add an internal `guard` to prevent an error when using the keyboard shortcut, and add it to `validateMenuItem` to disable the menu item). But I didn't know if it was appropriate to include it in the current PR, let me know whether to add it as well.

svobs:
> One thing I noticed that when I use two screens MacOS opens the Finder on the built-in screen and does not focus on it, meaning the focus remains on the IINA window. When I only use the built-in screen it doesn't happen. 

Try changing this setting in your Finder dock icon:
SCR-20240621-kqdk

Don't ask me how this is supposed to work though. In fact, if anyone knows what the `None` option does or what it means when multiple options are checked, please let me know 😅

svobs:
> One thing I noticed that when I use two screens MacOS opens the Finder on the built-in screen and does not focus on it, meaning the focus remains on the IINA window. When I only use the built-in screen it doesn't happen

Try changing this option in your Finder's Dock icon:
SCR-20240621-kvod

Don't ask me how this is supposed to work though. In fact, if anyone can explain what the `None` option is supposed to do, or what it means when multiple options are checked, please let me know. 😅

svobs:
> One thing I noticed that when I use two screens MacOS opens the Finder on the built-in screen and does not focus on it, meaning the focus remains on the IINA window. When I only use the built-in screen it doesn't happen.

I believe it has something to do with the `Assign To` setting in the Dock. Try setting this to different values for Finder and for IINA, respectively.
SCR-20240621-kvod

Some combinations appear to focus on the new Finder window and some do not. For me, using `Desktop on Display 1` (main display) for Finder and `All Desktops` for IINA correctly changed the focus. So I suspect when some combinations don't give focus to the Finder window, that is by design for some reason.

I haven't been able to find documentation for this feature. From playing around, it looks like the `All Desktops` option will open windows for the app in the same screen that the mouse is currently in at the time of open. But I can't figure out what the `None` option is supposed to do. And I have no idea why sometimes multiple options are checked. If anyone has any info, please share :)

iina/MainMenuActions.swift
low-batt: Lets treat the problem with deleting as a separate issue. I will enter one for it.

low-batt: I entered #5011 for the problem with `Delete Current File` being enabled while streaming.

low-batt:
The shortcut key still needs to be added to `iina-default-input.conf`, yes? I've not looked to see if that key is available in the other key bindings.

iina/MainMenuActions.swift
```diff
@@ -45,6 +45,12 @@ class MainMenuActionHandler: NSResponder {
     }
   }
 
+  @objc func menuShowCurrentFileInFinder(_ sender: NSMenuItem) {
+    guard let url = player.info.currentURL else { return }
+    guard !player.info.isNetworkResource else { return }
```
low-batt: Could combine these.
```suggestion
    guard let url = player.info.currentURL, !player.info.isNetworkResource else { return }
```

svobs:
> Add the 2 lines below to the case statement of handleIINACommand, in PlayerWindowController.swift

I was in a little bit of a hurry when I wrote that, but I found time to re-evaluate and have determined that adding a case to `handleIINACommand` is no longer needed. There is a newer mechanism introduced by the fix to #4567 which makes it mostly obsolete. Although it shouldn't hurt anything to leave it there, it can be removed.

I filed a PR, #5012, which has more info if curious.

iina/MainMenuActions.swift
ShlomoCode: I prefer the explicit form, but I don't know the Swift conventions. I followed your suggestion.

ShlomoCode:
I'm not sure that Cmd+R fits here, because it's usually used to refresh (in browsers, XCode, etc.)
Might we want it for a streaming refresh? Thoughts?

low-batt:
I too am concerned about the choice of key.

From [Spotlight keyboard shortcuts on Mac](https://support.apple.com/guide/mac-help/spotlight-keyboard-shortcuts-mh26783/mac):
> See a file in an app or the Finder
> Command-R or Command-Double-click

So we have an Apple example of ⌘R being used to bring up `Finder`.

I need to defer to @svobs and @lhc70000 on the choice of key.

ShlomoCode:
> I believe it has something to do with the `Assign To` setting in the Dock. Try setting this to different values for Finder and for IINA, respectively.

I use an external screen, not space (macOS virtual screen). It seems to only work for spaces, it doesn't show up for me for the external screen, only when I create two spaces on the same physical screen.
CleanShot 2024-06-23 at 02 32 52@2x
macOS 14.5

saagarjha:
I do want to note that this is feature already pretty much available because we present a document proxy icon, which makes it easy to locate the file and open a Finder window for it. I’m not opposed to also adding a “Reveal in Finder” menu item but it definitely should not be squatting on shortcut prime real estate like ⌘R by default.

svobs:
> I do want to note that this is feature already pretty much available because we present a document proxy icon, which makes it easy to locate the file and open a Finder window for it.

The document icon is not available in legacy full screen or in music mode. It is lost when either the `titled` window style is removed or the window has `titleVisibility = .hidden`. I have bent over backwards trying to figure out how to add it back in manually as one can do with the other title buttons but can't seem to make it work, and can't find documentation anywhere.

svobs:
Also agree that ⌘R is probably not the best default. I would associate it more with Reload or Refresh or some other command starting with R, particularly since the action is no longer called "Reveal in Finder". But I wouldn't use anything with S or F either as those have much stronger existing associations. Maybe just add another modifier key? ⇧⌘R? Looking at the existing IINA defaults, there doesn't seem to be a strong pattern to follow for which modifier(s) to add.

low-batt:
Apple guidance on custom keyboard shortcuts can be found [here](https://developer.apple.com/design/human-interface-guidelines/keyboards#Custom-keyboard-shortcuts). Don't be confused by the entries in the `Recommended usage`  column that are in the wrong row. It indicates the ⌘ key is preferred:
> Prefer the Command key as the main modifier key in a custom keyboard shortcut.

Focusing on the keys that trigger IINA actions we can see there isn't a consistent pattern but many use ⇧⌘:
keys

So ⇧⌘ and one of the keys not already used seems right.

uiryuu:
Not meant to divert the discussion too much, but there's an alternative is to not set any default keybindings for that and let the user to set whatever they want.

low-batt:
I thought about that, but configuring key bindings is a bit advanced. I would not be surprised if many users looked at the new menu item, saw there was no keyboard shortcut and gave up. Therefore I think it is important we provide a default key binding. I'm thinking ⌃⌘R due to the use of ⌘R by Spotlight.

lhc70000:
I suggest ⇧⌘R.

We use ⇧⌘ when possible and I prefer it over ⌃⌘ because it's more comfortable when pressing. The reason we have other modifiers for some shortcuts are

- `⌘3, ⌘+, ⌘- for adjusting window size`: these are major functions and should have corresponding mpv commands, but we have custom commands because they need extra handling
- `⌃⌘P for PIP`: ⇧⌘P is occupied (show playlist)
- `⌥⌘M for music mode`: TBH I couldn't remember 😢

low-batt:
I agree, ⇧⌘R.

low-batt:
We also use the ⇧⌘ convention for the mpv defaults as well, but it looks like we switched to ⌥⌘ for Movist and VLC. So ⌥⌘R for Moviest and VLC?

ShlomoCode:
> We use ⇧⌘ when possible and I prefer it over ⌃⌘ because it's more comfortable when pressing

I think it's the opposite :)

ShlomoCode:
But Apple [says](https://developer.apple.com/design/human-interface-guidelines/keyboards#Custom-keyboard-shortcuts) "Prefer the Shift key as a secondary modifier that complements a related shortcut" and "Use the Option modifier sparingly for less-common commands or power features" 🤷

ShlomoCode:
Note that Apple [says](https://developer.apple.com/design/human-interface-guidelines/keyboards#Custom-keyboard-shortcuts):
- Prefer the Shift key as a secondary modifier that complements a related shortcut
- Use the Option modifier sparingly for less-common commands or power features
- Avoid using the Control key as a modifier. The system uses Control in many systemwide features and shortcuts, like moving focus or capturing screenshots

low-batt:
⇧⌘R meets those requirements, yes? We can use that for the IINA and mpv defaults. We will have to hear from others as to why the Moviest and VLC did not use that and instead used option, ⌥, instead of shift, ⇧. To match up with the existing conventions in those key bindings we should use ⌥⌘R in them. See what the other developers say.

svobs:
> * Prefer the Shift key as a secondary modifier that complements a related shortcut
> * Use the Option modifier sparingly for less-common commands or power features

That seems almost backwards, no? Some menu items can have "alternate" versions which offer slightly different functionality. These can display an alternate title by holding down the Option key when clicking on the menu. Their key equivalent also incorporates Option.

Example in IINA: `Open In New Window…` with alternate, `Open…`

ShlomoCode:
I added ⇧⌘R to `iina-default-input.conf`. Waiting for a decision about VLC and Movist files.

low-batt:
Please also add ⇧⌘R to `input.conf` which is the mpv bindings.

Filtering the VLC key bindings shows for some reason ⌥⌘ was used instead of the ⇧⌘ convention:
VLC

I'd go ahead and change the Moviest and VLC key bindings to use ⌥⌘R to match up with the existing convention in those key bindings.

ShlomoCode:
I have added

low-batt:
I verified the key bindings. I would remove the ⇧ from the Moviest and VLC key bindings and just use ⌥⌘R to match up with many of the other key bindings that bind to IINA functions:
VLC-key

ShlomoCode:
> I would remove the ⇧ from the Moviest and VLC key bindings and just use ⌥⌘R to match up with many of the other key bindings that bind to IINA functions

Isn't that what I did? I'm confused
>`#@iina Alt+Meta+R show-current-file-in-finder`

ShlomoCode:
i got it. The `R` should be a lowercase letter, otherwise `Shift` is added to the shortcut. pushes a fix...

low-batt:
Great! I will see if I can get the other developers to get this one approved and merged.

low-batt:
@ShlomoCode Thanks for addressing #5001

cmbail:
It doesn't appear in my file menu and doesn't work when added as a key binding. 1.3.5 Build 141|

Untitled

low-batt:
@cmbail This change has been merged into the development version of IINA. It is not in the current release of IINA. 

Issue #5001 was closed automatically by GitHub when this pull request was merged. The downside of that is that it makes it look to users like the issue has been addressed and should be working in the latest release of IINA. I have been trying to re-open issues and leave them open until the change is in the released version of IINA, but it is exhausting to keep up with GitHub's automatic actions. GitHub users have been asking for a way to turn off this auto close feature, but last I looked GitHub has not provided a way to do that.

It is a philosophical different. GitHub views issues as tasks for developers to do. Once a change has been merged then the developer has nothing more to do and GitHub closes the issue. The IINA project views issues as the public defect and feature request database. With that point of view you would not close issues until the features are available to users. Currently the GitHub view dominates.

Sorry about the confusion.

cmbail:
Thanks for clearing that up.

PR #5006, opened 2024-06-21
Fix HistoryController uses deprecated methods, #5004

### LINKED ISSUE #5004: HistoryController uses deprecated methods
**System and IINA version:**

- macOS 13.6.3
- IINA `develop` at 2ed5d72568eac27aba57685686b48b8c88dd1834

**Expected behavior:**
As much as possible building IINA does not generate compiler warnings.

**Actual behavior:**
Building with Xcode 15.2 generates a number of new warnings since the `MACOSX_DEPLOYMENT_TARGET` setting was raised to 10.15. This issue is only about the following two warnings:
```text
/Users/low-batt/Documents/builds/iina-official/iina/iina/HistoryController.swift:27:34 'unarchiveObject(withFile:)' was deprecated in macOS 10.14: Use +unarchivedObjectOfClass:fromData:error: instead

/Users/low-batt/Documents/builds/iina-official/iina/iina/HistoryController.swift:31:34 'archiveRootObject(_:toFile:)' was deprecated in macOS 10.14: Use +archivedDataWithRootObject:requiringSecureCoding:error: and -writeToURL:options:error: instead
```

The compiler is flagging use of the deprecated methods [NSKeyedUnarchiver.unarchiveObject(withFile:)](https://developer.apple.com/documentation/foundation/nskeyedunarchiver/1417153-unarchiveobject) and [NSKeyedArchiver.archiveRootObject(_:toFile:)](https://developer.apple.com/documentation/foundation/nskeyedarchiver/1410621-archiverootobject). Apple has deprecated these methods in order to encourage conformance to [NSSecureCoding](https://developer.apple.com/documentation/foundation/nssecurecoding) which protects against object substitution attacks.

Correcting these warnings requires changing `PlaybackHistory` to implement [NSSecureCoding](https://developer.apple.com/documentation/foundation/nssecurecoding) and then changing the methods in `HistoryController` that read and save history. I believe this will require the history database to be migrated.

**Steps to reproduce:**
Perform a full clean build of the `develop` branch.

Every time.

low-batt:
This commit will:
- Change PlaybackHistory to conform to NSSecureCoding
- Change HistoryController to use methods that support secure coding instead of the deprecated methods

- [x] This implements/fixes issue #5004.

---

**Description:**

### DIFF
--- a/iina/HistoryController.swift
+++ b/iina/HistoryController.swift
@@ -24,15 +24,26 @@ class HistoryController: NSObject {
   }
 
   private func read() {
-    history = (NSKeyedUnarchiver.unarchiveObject(withFile: plistURL.path) as? [PlaybackHistory]) ?? []
+    // Avoid logging a scary error if the file does not exist.
+    guard FileManager.default.fileExists(atPath: plistURL.path) else { return }
+    do {
+      let data = try Data(contentsOf: plistURL)
+      let object = try NSKeyedUnarchiver.unarchivedObject(ofClasses: [NSArray.self, PlaybackHistory.self],
+                                                          from: data)
+      history = object as? [PlaybackHistory] ?? []
+    } catch {
+      Logger.log("Failed to read playback history file \(plistURL.path): \(error)", level: .error)
+    }
   }
 
   func save() {
-    let result = NSKeyedArchiver.archiveRootObject(history, toFile: plistURL.path)
-    if !result {
-      Logger.log("Cannot save playback history!", level: .error)
+    do {
+      let data = try NSKeyedArchiver.archivedData(withRootObject: history, requiringSecureCoding: true)
+      try data.write(to: plistURL)
+      NotificationCenter.default.post(Notification(name: .iinaHistoryUpdated))
+    } catch {
+      Logger.log("Failed to save playback history to file \(plistURL.path): \(error)", level: .error)
     }
-    NotificationCenter.default.post(Notification(name: .iinaHistoryUpdated))
   }
 
   func add(_ url: URL, duration: Double) {
--- a/iina/PlaybackHistory.swift
+++ b/iina/PlaybackHistory.swift
@@ -15,7 +15,13 @@ fileprivate let KeyPlayed = "IINAPHPlayed"
 fileprivate let KeyAddedDate = "IINAPHDate"
 fileprivate let KeyDuration = "IINAPHDuration"
 
-class PlaybackHistory: NSObject, NSCoding {
+/// An entry in the playback history file.
+/// - Important: This class conforms to [NSSecureCoding](https://developer.apple.com/documentation/foundation/nssecurecoding).
+///     When making changes be certain the requirements for secure coding are not violated by the changes.
+class PlaybackHistory: NSObject, NSCoding, NSSecureCoding {
+
+  /// Indicate this class supports secure coding.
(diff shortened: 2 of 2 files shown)


iina/PlaybackHistory.swift
```diff
@@ -15,7 +15,13 @@ fileprivate let KeyPlayed = "IINAPHPlayed"
 fileprivate let KeyAddedDate = "IINAPHDate"
 fileprivate let KeyDuration = "IINAPHDuration"
 
-class PlaybackHistory: NSObject, NSCoding {
+/// An entry in the playback history file.
+/// - Important: This class conforms to [NSSecureCoding](https://developer.apple.com/documentation/foundation/nssecurecoding).
+///     When making changes be certain the requirements for secure coding are not violated by the changes.
+class PlaybackHistory: NSObject, NSCoding, NSSecureCoding {
```
uiryuu: `NSSecureCoding` already inherits from `NSCoding`, do we need to confirm these two at the same time?

low-batt: Sorry I missed that. Fixed.

iina/HistoryController.swift
```diff
@@ -24,15 +24,26 @@ class HistoryController: NSObject {
...
 
   private func read() {
-    history = (NSKeyedUnarchiver.unarchiveObject(withFile: plistURL.path) as? [PlaybackHistory]) ?? []
+    // Avoid logging a scary error if the file does not exist.
+    guard FileManager.default.fileExists(atPath: plistURL.path) else { return }
+    do {
+      let data = try Data(contentsOf: plistURL)
+      let object = try NSKeyedUnarchiver.unarchivedObject(ofClasses: [NSArray.self, PlaybackHistory.self],
+                                                          from: data)
+      history = object as? [PlaybackHistory] ?? []
```
uiryuu, line 33: I think we should throw an error there if `as?` fails so that we can log the type casting failed, tho I don't think that will happen very often...

uiryuu (approved):
You can decide whether or not to throw an error, I'm okay with both.

svobs:
Pulled this PR and ran it, added an item to the play history so that it was saved in the secure-encoded format, then quit & opened IINA 1.3.5 & IINA 1.3.4. Looks like older versions can still read it seamlessly. Nice.

iina/HistoryController.swift
low-batt, line 33: I coded it that way with error logging and then simplified it and put it back to the way the code originally was. I prefer simplifying further to the code below. The other error handling needs to be there as you can never assume an external file hasn't been messed with. But the call to `unarchivedObject` will throw an error if the file contains any classes other than an array or `PlaybackHistory` objects. A developer would have to change `save` to save something like nested arrays, or just a single  `PlaybackHistory` object to cause `unarchivedObject` to be successful and the cast to fail. I don't see the need to cover that case, but I'm happy to add in the check and logging if you prefer an more cautious approach. Thoughts?
```suggestion
      history = try NSKeyedUnarchiver.unarchivedObject(ofClasses: [NSArray.self, PlaybackHistory.self],
                                                       from: data) as! [PlaybackHistory]
```

lhc70000: I think we don't need extra handling here.

uiryuu:
Thanks!

PR #5012, opened 2024-06-22
Key bindings for IINA commands: clean up & remove redundant code

svobs:

---

**Description:**

Now that the fix for #4567 has been merged, it's clear that a lot of code for handling key bindings for IINA commands is now redundant because most of them are now being mapped to hidden menu items. This became apparent while examining PR #5005, which is adding a new command.

To be specific, of all the commands enumerated in `IINACommand.swift`, I found that nearly all of them are already being mapped to menu items in `MenuController.updateKeyEquivalentsFrom()`, so the checks for these can be simply deleted from `PlayerWindowController.handleIINACommand()` and its overrides; they will never be used. I handled the remaining 5 commands as follows:

- `flip`, `mirror`: I migrated handling for these to `updateKeyEquivalentsFrom`. It was easy because all the objects were already created and easily referenced there.
- `openFile` and `openURL`: I tried migrating these to `updateKeyEquivalentsFrom`, but for some reason, the `Option` key was being added to their equivalents. Did not dig into why this was happening; just decided to leave them in `handleIINACommand` instead.
- `deleteCurrentFileHard`: this is a less-used command, and it didn't have an `NSMenuItem` object in `MenuController`, so it would have required more code to migrate. I did move it up from `MainWindowController` to `PlayerWindowController` so that it can also be recognized in music mode. I'm assuming that was an oversight when that feature was created.

Note that this should be considered lower-priority and not merged too close to a release.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -2987,44 +2987,6 @@ class MainWindowController: PlayerWindowController {
 
   // MARK: - Utility
 
-  internal override func handleIINACommand(_ cmd: IINACommand) {
-    super.handleIINACommand(cmd)
-    switch cmd {
-    case .togglePIP:
-      if #available(macOS 10.12, *) {
-        menuTogglePIP(.dummy)
-      }
-    case .videoPanel:
-      menuShowVideoQuickSettings(.dummy)
-    case .audioPanel:
-      menuShowAudioQuickSettings(.dummy)
-    case .subPanel:
-      menuShowSubQuickSettings(.dummy)
-    case .playlistPanel:
-      menuShowPlaylistPanel(.dummy)
-    case .chapterPanel:
-      menuShowChaptersPanel(.dummy)
-    case .toggleMusicMode:
-      menuSwitchToMiniPlayer(.dummy)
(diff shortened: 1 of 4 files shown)


svobs:
Rebased and resolved conflicts.

uiryuu:
I think it's good to merge, given that our next release will be a beta release and the changes look logical. Or do you want to deal with this short after we release the first beta (i.e. to include this in beta 2)?

svobs:
> I think it's good to merge, given that our next release will be a beta release and the changes look logical.

Sounds great to me. I put the disclaimer in there because I don't have much info about the project roadmap.

uiryuu:
Found a problem when testing this out, but not introduced by this PR. I have another app which registered cmd+shift+p as a global shortcut, in IINA this is used to toggle the playlist. When I press cmd+shift+p while IINA is focused, the other app handles this shortcut. I checked the menu, the menu item for this action is enabled. Is this the intended behavior of macOS?

lhc70000:
@uiryuu Yes, this can happen if the event is consumed (by a event tap) before reaching IINA.

PR #5017, opened 2024-06-23
Fix CC_MD5 deprecated warning

low-batt:
This commit will:
- Remove the `NSData` extension that was using the deprecated Common Crypto method
- Change the `md5` property in the `Data` extension to generate the MD5 hash using `CryptoKit` instead of calling the `NSData` method

These changes correct a compiler warning that CC_MD5 is deprecated because it is cryptographically broken and should not be used in security contexts. This warning is not generated for the `CryptoKit` method. With `CryptoKit` you must access the method using the `CryptoKit` `Insecure` enum, thus there is no need to warn the programmer that MD5 is insecure.

---

**Description:**

### DIFF
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -7,6 +7,7 @@
 //
 
 import Cocoa
+import CryptoKit
 
 extension NSSlider {
   /** Returns the position of knob center by point */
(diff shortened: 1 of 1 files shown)


saagarjha (commented):
This seems like an excellent time to convert this to being an extension on Data and returning a String

low-batt:
We already have such a `Data` extention. It calls this `NSData` `md5` function. I just tried commenting out the `NSData` extension and did not get a build error. Should I remove the `NSData` extension?

saagarjha:
Yes that seems appropriate

low-batt:
Updated as discussed.

PR #5018, opened 2024-06-23
Fix icon(forFileType:) deprecated warning

low-batt:
The commit will change the call to the deprecated method icon(forFileType:) in History WindowController.outlineView to call icon(forFile:) instead.

---

**Description:**

### DIFF
--- a/iina/HistoryWindowController.swift
+++ b/iina/HistoryWindowController.swift
@@ -215,7 +215,7 @@ class HistoryWindowController: NSWindowController, NSOutlineViewDelegate, NSOutl
         let fileExists = !entry.url.isFileURL || FileManager.default.fileExists(atPath: entry.url.path)
         filenameView.textField?.stringValue = entry.url.isFileURL ? entry.name : entry.url.absoluteString
         filenameView.textField?.textColor = fileExists ? .controlTextColor : .disabledControlTextColor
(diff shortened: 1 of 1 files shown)


low-batt:
Worried about this one. This was the recommended solution by someone on the net, but I'm worried it is not just looking at the path, but is accessing the file. Can't tell from the Apple documentation.

ShlomoCode:
> Worried about this one. This was the recommended solution by someone on the net, but I'm worried it is not just looking at the path, but is accessing the file. Can't tell from the Apple documentation.

Apple [says](https://developer.apple.com/documentation/appkit/nsworkspace/1527879-icon#:~:text=Use%20icon(for%3A)%20instead.):
>Use [icon(for:)](https://developer.apple.com/documentation/appkit/nsworkspace/3553230-icon) instead"

`icon(for:)` accepts a [`UTType`](https://developer.apple.com/documentation/uniformtypeidentifiers/uttype-swift.struct) as a parameter, which has an [`init(filenameExtension:)`](https://developer.apple.com/documentation/uniformtypeidentifiers/uttype-swift.struct/init(filenameextension:conformingto:)) init option. Is it not suitable?

low-batt:
I coded it this way based on recommendations from the net, but was worried that developer might have gotten it wrong. I'll convert this to a draft for now. Tied up with other stuff.

low-batt:
Closing as this has been addressed by PR #5216

PR #5021, opened 2024-06-24
Fix the window is shifted when pinching to resize

### LINKED ISSUE #4681: Pinch to Adjust video size shift the window position

uiryuu:
- [x] This implements/fixes issue #4681.

---

**Description:**
This commit:
- Caches the window frame when started pinching to resize
- Used the cached window frame instead of the current window frame to avoid accumulated rounding errors

Before:
https://github.com/iina/iina/assets/20237141/3bfc10d4-5fbc-419d-80e1-9d41ca6ea623

After:
https://github.com/iina/iina/assets/20237141/eff9cb20-38a0-43d3-a70b-63cdd1d10c6f

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -148,6 +148,7 @@ class MainWindowController: PlayerWindowController {
   var isPausedPriorToInteractiveMode: Bool = false
 
   var lastMagnification: CGFloat = 0.0
+  var frameWhenStartedPinching = NSRect()
 
   /** Views that will show/hide when cursor moving in/out the window. */
   var fadeableViews: [NSView] = []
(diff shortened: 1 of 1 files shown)


low-batt (approved):
Reproduced problem with 1.3.5. Pulled PR, built, tested, confirmed fixed. Looks good to me.

PR #5029, opened 2024-06-27
Fix crash while quitting in mpv_render_context_update, #5023

### LINKED ISSUE #5023: Crash while quitting in mpv_render_context_update
**System and IINA version:**

- macOS 13.6.3
- IINA `develop` at db436141ffeb6fdb5e76a13a595687ad36b82d32

**Expected behavior:**
IINA does not crash.

**Actual behavior:**
While quitting IINA crashed in `MPVController.shouldRenderUpdateFrame` in the call to `mpv_render_context_update`:
```swift
  func shouldRenderUpdateFrame() -> Bool {
    guard let mpvRenderContext = mpvRenderContext else { return false }
    let flags: UInt64 = mpv_render_context_update(mpvRenderContext)
    return flags & UInt64(MPV_RENDER_UPDATE_FRAME.rawValue) > 0
  }
```

<details>

```
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = hit program assert
    frame #0: 0x00000001a8298744 libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x000000010588ab84 libsystem_pthread.dylib`pthread_kill + 288
    frame #2: 0x00000001a81ddae8 libsystem_c.dylib`abort + 180
    frame #3: 0x00000001a81dce44 libsystem_c.dylib`__assert_rtn + 272
    frame #4: 0x0000000105015c44 libmpv.2.dylib`mp_dispatch_queue_process.cold.4 + 40
    frame #5: 0x0000000104f62438 libmpv.2.dylib`mp_dispatch_queue_process + 516
    frame #6: 0x0000000104fe616c libmpv.2.dylib`mpv_render_context_update + 28
  * frame #7: 0x00000001013c63c4 IINA`MPVController.shouldRenderUpdateFrame(self=0x0000600000c39950) at MPVController.swift:603:25
    frame #8: 0x00000001012686f4 IINA`ViewLayer.canDraw(ctx=0x122244a00, pf=0x00006000010bc300, t=342764.37683787511, ts=0x16ef4b760, self=0x0000600001dbcbe0) at ViewLayer.swift:93:33
    frame #9: 0x000000010126876c IINA`@objc ViewLayer.canDraw(inCGLContext:pixelFormat:forLayerTime:displayTime:) at <compiler-generated>:0
    frame #10: 0x00000001af943754 QuartzCore`CAOpenGLLayerDraw(CAOpenGLLayer*, double, CVTimeStamp const*, unsigned int) + 608
    frame #11: 0x00000001af946908 QuartzCore`CAOpenGLLayer_timer_callback(__CFRunLoopTimer*, void*) + 820
    frame #12: 0x00000001a83c8338 CoreFoundation`__CFRUNLOOP_IS_CALLING_OUT_TO_A_TIMER_CALLBACK_FUNCTION__ + 32
    frame #13: 0x00000001a83c7fe0 CoreFoundation`__CFRunLoopDoTimer + 940
    frame #14: 0x00000001a83c7b38 CoreFoundation`__CFRunLoopDoTimers + 356
    frame #15: 0x00000001a83ad2b8 CoreFoundation`__CFRunLoopRun + 1852
    frame #16: 0x00000001a83ac428 CoreFoundation`CFRunLoopRunSpecific + 612
    frame #17: 0x00000001b1c00df0 HIToolbox`RunCurrentEventLoopInMode + 292
    frame #18: 0x00000001b1c00a80 HIToolbox`ReceiveNextEventCommon + 220
    frame #19: 0x00000001b1c00984 HIToolbox`_BlockUntilNextEventMatchingListInModeWithFilter + 76
    frame #20: 0x00000001ab5d4908 AppKit`_DPSNextEvent + 636
    frame #21: 0x00000001ab5d3aa4 AppKit`-[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 716
    frame #22: 0x00000001ab7f517c AppKit`-[NSApplication _shouldTerminate] + 700
    frame #23: 0x00000001ab7e7e68 AppKit`-[NSApplication terminate:] + 612
    frame #24: 0x00000001ab77c458 AppKit`-[NSApplication(NSResponder) sendAction:to:from:] + 440
    frame #25: 0x00000001ab848424 AppKit`-[NSMenuItem _corePerformAction] + 336
    frame #26: 0x00000001ab8481a0 AppKit`-[NSCarbonMenuImpl performActionWithHighlightingForItemAtIndex:] + 104
    frame #27: 0x00000001ab883b2c AppKit`-[NSMenu performActionForItemAtIndex:] + 200
    frame #28: 0x00000001ab883a4c AppKit`-[NSMenu _internalPerformActionForItemAtIndex:] + 76
    frame #29: 0x00000001ab8838a0 AppKit`-[NSCarbonMenuImpl _carbonCommandProcessEvent:handlerCallRef:] + 108
    frame #30: 0x00000001ab8307b4 AppKit`NSSLMMenuEventHandler + 640
    frame #31: 0x00000001b1bd93a4 HIToolbox`DispatchEventToHandlers(EventTargetRec*, OpaqueEventRef*, HandlerCallRec*) + 1092
    frame #32: 0x00000001b1bd8828 HIToolbox`SendEventToEventTargetInternal(OpaqueEventRef*, OpaqueEventTargetRef*, HandlerCallRec*) + 356
    frame #33: 0x00000001b1beea40 HIToolbox`SendEventToEventTarget + 40
    frame #34: 0x00000001b1c4c8e4 HIToolbox`SendHICommandEvent(unsigned int, HICommand const*, unsigned int, unsigned int, unsigned char, void const*, OpaqueEventTargetRef*, OpaqueEventTargetRef*, OpaqueEventRef**) + 416
    frame #35: 0x00000001b1c70564 HIToolbox`SendMenuCommandWithContextAndModifiers + 56
    frame #36: 0x00000001b1c704f4 HIToolbox`SendMenuItemSelectedEvent + 352
    frame #37: 0x00000001b1c70320 HIToolbox`FinishMenuSelection(SelectionData*, MenuResult*, MenuResult*) + 100
    frame #38: 0x00000001b1c70ce0 HIToolbox`MenuSelectCore(MenuData*, Point, double, unsigned int, OpaqueMenuRef**, unsigned short*) + 560
    frame #39: 0x00000001b1c70a00 HIToolbox`_HandleMenuSelection2 + 416
    frame #40: 0x00000001ab7268a4 AppKit`_NSHandleCarbonMenuEvent + 256
    frame #41: 0x00000001ab7266c8 AppKit`_DPSEventHandledByCarbon + 60
    frame #42: 0x00000001ab5d4090 AppKit`-[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 2232
    frame #43: 0x00000001ab5c7f08 AppKit`-[NSApplication run] + 464
    frame #44: 0x00000001ab59f358 AppKit`NSApplicationMain + 880
    frame #45: 0x00000001010f9918 IINA`main at AppDelegate.swift:24:7
    frame #46: 0x00000001a7f77f28 dyld`start + 2236
(lldb)
```
</details>

<br>

**Steps to reproduce:**
This crash can be hard to reproduce. I was not able to reproduce this outside of Xcode. The first steps are about fully enabling animations as that seemed to make the crash more likely to occur.

- Disable the macOS `Reduce motion` setting if it is enabled
- Start IINA running in Xcode
- The settings panel appears
- Disable the setting `Use legacy full screen` if it is enabled
- Click on the `UI` tab
- Scroll to the bottom of the window
- Make certain the `Disable animations` setting is not checked
- Start playing a video
- Enter full screen mode
- Press ⌃→ to switch to a new [space](https://support.apple.com/guide/mac-help/work-in-multiple-spaces-mh14112/mac)
- Start playing another video
- Enter full screen mode
- Press ⌃→ to switch to a new [space](https://support.apple.com/guide/mac-help/work-in-multiple-spaces-mh14112/mac)
- Repeat this two more times so four videos are playing in separate spaces
- Click on  `Quit IINA` under the `IINA` menu
- IINA will crash sometimes

Sometimes.

low-batt:
This commit will:
- Change where `isUninited` is locked in `VideoLayer`
- Add a `displayLock` property to `VideoLayer`
- Use `displayLock` to single thread the display method
- Move the skip render code from the draw method to the display method

This ensures the render context has not been freed when `shouldRenderUpdateFrame` is called. This also changes the location of the skip render code to match how mpv handles skipping.

- [x] This implements/fixes issue #5023.

---

**Description:**

### DIFF
--- a/iina/ViewLayer.swift
+++ b/iina/ViewLayer.swift
@@ -64,7 +64,7 @@ class ViewLayer: CAOpenGLLayer {
                         forLayerTime t: CFTimeInterval, displayTime ts: UnsafePointer<CVTimeStamp>?) -> Bool {
     videoView.$isUninited.withLock() { isUninited in
       guard !isUninited else { return false }
-      guard !forceRender else { return true }
+      if forceRender { return true }
       return videoView.player.mpv.shouldRenderUpdateFrame()
     }
   }
--- a/iina/ViewLayer.swift
+++ b/iina/ViewLayer.swift
@@ -17,7 +17,10 @@ class ViewLayer: CAOpenGLLayer {
   let mpvGLQueue = DispatchQueue(label: "com.colliderli.iina.mpvgl", qos: .userInteractive)
   @Atomic var blocked = false
 
-  private let displayLock = NSRecursiveLock()
+  private let cglContext: CGLContextObj
+  private let cglPixelFormat: CGLPixelFormatObj
+
+  private let displayLock = Lock()
 
   private var fbo: GLint = 1
 
@@ -25,6 +28,8 @@ class ViewLayer: CAOpenGLLayer {
   private var forceRender = false
 
   override init() {
+    cglPixelFormat = ViewLayer.createPixelFormat()
+    cglContext = ViewLayer.createContext(cglPixelFormat)
     super.init()
 
     isOpaque = true
@@ -45,50 +50,11 @@ class ViewLayer: CAOpenGLLayer {
     fatalError("init(coder:) has not been implemented")
   }
 
-  override func copyCGLPixelFormat(forDisplayMask mask: UInt32) -> CGLPixelFormatObj {
-    var attributeList: [CGLPixelFormatAttribute] = [
-      kCGLPFADoubleBuffer,
-      kCGLPFAAllowOfflineRenderers,
-      kCGLPFAColorFloat,
-      kCGLPFAColorSize, CGLPixelFormatAttribute(64),
-      kCGLPFAOpenGLProfile, CGLPixelFormatAttribute(kCGLOGLPVersion_3_2_Core.rawValue),
-      kCGLPFAAccelerated,
-    ]
+  override func copyCGLPixelFormat(forDisplayMask mask: UInt32) -> CGLPixelFormatObj { cglPixelFormat }
 
-    if (!Preference.bool(for: .forceDedicatedGPU)) {
-      attributeList.append(kCGLPFASupportsAutomaticGraphicsSwitching)
-    }
+  override func copyCGLContext(forPixelFormat pf: CGLPixelFormatObj) -> CGLContextObj { cglContext }
 
-    var pix: CGLPixelFormatObj?
-    var npix: GLint = 0
-
-    for index in (0..<attributeList.count).reversed() {
-      let attributes = Array(
-        attributeList[0...index] + [_CGLPixelFormatAttribute(rawValue: 0)]
-      )
-      CGLChoosePixelFormat(attributes, &pix, &npix)
-      if let pix = pix {
-        Logger.log("Created OpenGL pixel format with \(attributes)", level: .debug)
-        return pix
-      }
-    }
-
-    Logger.fatal("Cannot create OpenGL pixel format!")
-  }
-
-  override func copyCGLContext(forPixelFormat pf: CGLPixelFormatObj) -> CGLContextObj {
-    let ctx = super.copyCGLContext(forPixelFormat: pf)
-
-    var i: GLint = 1
-    CGLSetParameter(ctx, kCGLCPSwapInterval, &i)
-
-    CGLEnable(ctx, kCGLCEMPEngine)
-
-    CGLSetCurrentContext(ctx)
-    return ctx
-  }
-
-  // MARK: Draw
+  // MARK: - Draw
 
   override func canDraw(inCGLContext ctx: CGLContextObj, pixelFormat pf: CGLPixelFormatObj,
                         forLayerTime t: CFTimeInterval, displayTime ts: UnsafePointer<CVTimeStamp>?) -> Bool {
@@ -130,7 +96,7 @@ class ViewLayer: CAOpenGLLayer {
               mpv_render_param(type: MPV_RENDER_PARAM_FLIP_Y, data: .init(flip)),
               mpv_render_param()
             ]
-            mpv_render_context_render(context, &params);
+            mpv_render_context_render(context, &params)
             ignoreGLError()
           }
         } else {
@@ -168,42 +134,92 @@ class ViewLayer: CAOpenGLLayer {
   }
 
   override func display() {
-    displayLock.lock()
-    defer { displayLock.unlock() }
+    displayLock.withLock {
+      super.display()
+      CATransaction.flush()
+      videoView.$isUninited.withLock() { isUninited in
+        guard !isUninited else { return }
 
-    super.display()
-    CATransaction.flush()
-
-    videoView.$isUninited.withLock() { isUninited in
-      guard !isUninited else { return }
-
-      guard !forceRender else {
-        forceRender = false
-        return
-      }
-      guard needsMPVRender else { return }
-
-      // Neither canDraw nor draw(inCGLContext:) were called by AppKit, needs a skip render.
-      // This can happen when IINA is playing in another space, as might occur when just playing
-      // audio. See issue #5025.
-      videoView.player.mpv.lockAndSetOpenGLContext()
-      defer { videoView.player.mpv.unlockOpenGLContext() }
-      if let renderContext = videoView.player.mpv.mpvRenderContext,
-         videoView.player.mpv.shouldRenderUpdateFrame() {
-        var skip: CInt = 1
-        withUnsafeMutablePointer(to: &skip) { skip in
-          var params: [mpv_render_param] = [
-            mpv_render_param(type: MPV_RENDER_PARAM_SKIP_RENDERING, data: .init(skip)),
-            mpv_render_param()
-          ]
-          mpv_render_context_render(renderContext, &params);
+        guard !forceRender else {
+          forceRender = false
+          return
         }
+        guard needsMPVRender else { return }
+
+        // Neither canDraw nor draw(inCGLContext:) were called by AppKit, needs a skip render.
+        // This can happen when IINA is playing in another space, as might occur when just playing
+        // audio. See issue #5025.
+        videoView.player.mpv.lockAndSetOpenGLContext()
+        defer { videoView.player.mpv.unlockOpenGLContext() }
+        if let renderContext = videoView.player.mpv.mpvRenderContext,
+           videoView.player.mpv.shouldRenderUpdateFrame() {
+          var skip: CInt = 1
+          withUnsafeMutablePointer(to: &skip) { skip in
+            var params: [mpv_render_param] = [
+              mpv_render_param(type: MPV_RENDER_PARAM_SKIP_RENDERING, data: .init(skip)),
+              mpv_render_param()
+            ]
+            mpv_render_context_render(renderContext, &params)
+          }
+        }
+        needsMPVRender = false
       }
-      needsMPVRender = false
     }
   }
 
-  // MARK: Utils
+  // MARK: - Core OpenGL Context and Pixel Format
+
+  private static func createContext(_ pixelFormat: CGLPixelFormatObj) -> CGLContextObj {
+    var ctx: CGLContextObj?
+    CGLCreateContext(pixelFormat, nil, &ctx)
+
+    guard let ctx = ctx else {
+      Logger.fatal("Cannot create OpenGL context")
+    }
+
+    // Sync to vertical retrace.
+    var i: GLint = 1
+    CGLSetParameter(ctx, kCGLCPSwapInterval, &i)
+
+    // Enable multi-threaded GL engine.
+    CGLEnable(ctx, kCGLCEMPEngine)
+
+    CGLSetCurrentContext(ctx)
+    return ctx
+  }
+
+  private static func createPixelFormat() -> CGLPixelFormatObj {
+    var attributeList: [CGLPixelFormatAttribute] = [
+      kCGLPFADoubleBuffer,
+      kCGLPFAAllowOfflineRenderers,
+      kCGLPFAColorFloat,
+      kCGLPFAColorSize, CGLPixelFormatAttribute(64),
+      kCGLPFAOpenGLProfile, CGLPixelFormatAttribute(kCGLOGLPVersion_3_2_Core.rawValue),
+      kCGLPFAAccelerated,
+    ]
+
+    if (!Preference.bool(for: .forceDedicatedGPU)) {
+      attributeList.append(kCGLPFASupportsAutomaticGraphicsSwitching)
+    }
+
+    var pix: CGLPixelFormatObj?
+    var npix: GLint = 0
+
+    for index in (0..<attributeList.count).reversed() {
+      let attributes = Array(
+        attributeList[0...index] + [_CGLPixelFormatAttribute(rawValue: 0)]
+      )
+      CGLChoosePixelFormat(attributes, &pix, &npix)
+      if let pix = pix {
+        Logger.log("Created OpenGL pixel format with \(attributes)", level: .debug)
+        return pix
+      }
+    }
+
+    Logger.fatal("Cannot create OpenGL pixel format!")
+  }
+
+  // MARK: - Utils
 
   /** Check OpenGL error (for debug only). */
   func gle() {
--- a/iina/ViewLayer.swift
+++ b/iina/ViewLayer.swift
@@ -17,6 +17,8 @@ class ViewLayer: CAOpenGLLayer {
   let mpvGLQueue = DispatchQueue(label: "com.colliderli.iina.mpvgl", qos: .userInteractive)
   @Atomic var blocked = false
 
+  private let displayLock = NSRecursiveLock()
+
   private var fbo: GLint = 1
 
   private var needsMPVRender = false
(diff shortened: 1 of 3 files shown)


low-batt:
I am troubled that I can not explain why trying to match the way mpv handles the layer locking resulted in a deadlock when used in IINA. The deadlock occurred then I used a `NSLock` for `displayLock`, which is what mpv does. Seems the call to `CATransaction.flush` in the `display` method ends up calling the `display` method again. So I was forced to use a `NSRecursiveLock`. Anyone have any guesses as to how mpv is able to just use a `NSLock`? @svobs, @krackers, thoughts?
<details>
<summary><b>Deadlock:</b></summary>

```text
Call graph:
    8260 Thread_836017   DispatchQueue_1: com.apple.main-thread  (serial)
    + 8260 start  (in dyld) + 2236  [0x195babf28]
    +   8260 main  (in IINA) + 12  [0x102d4daec]  AppDelegate.swift:24
    +     8260 NSApplicationMain  (in AppKit) + 880  [0x1991d3358]
    +       8260 -[NSApplication run]  (in AppKit) + 500  [0x1991fbf2c]
    +         8260 -[NSApplication _handleEvent:]  (in AppKit) + 60  [0x199584bd4]
    +           8260 -[NSApplication(NSEvent) sendEvent:]  (in AppKit) + 1556  [0x199334abc]
    +             8260 -[NSWindow(NSEventRouting) sendEvent:]  (in AppKit) + 284  [0x199335778]
    +               8260 -[NSWindow(NSEventRouting) _reallySendEvent:isDelayedEvent:]  (in AppKit) + 364  [0x199335ab8]
    +                 8260 -[NSWindow(NSEventRouting) _handleMouseDownEvent:isDelayedEvent:]  (in AppKit) + 3476  [0x1993aae7c]
    +                   8260 @objc InitialWindowViewActionButton.mouseDown(with:)  (in IINA) + 52  [0x102b1e0cc]  <compiler-generated>:0
    +                     8260 InitialWindowViewActionButton.mouseDown(with:)  (in IINA) + 2412  [0x102b1e01c]  InitialWindowController.swift:470
    +                       8260 PlayerCore.openURL(_:shouldAutoLoad:)  (in IINA) + 164  [0x102f7f598]  PlayerCore.swift:355
    +                         8260 PlayerCore.openURLs(_:shouldAutoLoad:)  (in IINA) + 1500  [0x102f7efe8]  PlayerCore.swift:339
    +                           8260 PlayerCore.open(_:shouldAutoLoad:)  (in IINA) + 1408  [0x102f7e988]  PlayerCore.swift:296
    +                             8260 PlayerCore.openMainWindow(path:url:isNetwork:)  (in IINA) + 1380  [0x102f808b4]  PlayerCore.swift:401
    +                               8260 -[NSWindowController window]  (in AppKit) + 84  [0x199372098]
    +                                 8260 -[NSWindowController _windowDidLoad]  (in AppKit) + 620  [0x199375ca4]
    +                                   8260 @objc MainWindowController.windowDidLoad()  (in IINA) + 36  [0x102c76790]  <compiler-generated>:0
    +                                     8260 MainWindowController.windowDidLoad()  (in IINA) + 4956  [0x102c74564]  MainWindowController.swift:573
    +                                       8260 PlayerCore.initVideo()  (in IINA) + 320  [0x102f84ad4]  PlayerCore.swift:508
    +                                         8260 @objc ViewLayer.display()  (in IINA) + 36  [0x102ebfb18]  <compiler-generated>:0
    +                                           8260 ViewLayer.display()  (in IINA) + 204  [0x102ebe9a8]  ViewLayer.swift:174
    +                                             8260 CA::Transaction::commit()  (in QuartzCore) + 648  [0x19d4cc2c0]
    +                                               8260 CA::Context::commit_transaction(CA::Transaction*, double, double*)  (in QuartzCore) + 456  [0x19d65d4d4]
    +                                                 8260 CA::Layer::display_if_needed(CA::Transaction*)  (in QuartzCore) + 760  [0x19d4e9ae8]
    +                                                   8260 @objc ViewLayer.display()  (in IINA) + 36  [0x102ebfb18]  <compiler-generated>:0
    +                                                     8260 ViewLayer.display()  (in IINA) + 116  [0x102ebe950]  ViewLayer.swift:170
    +                                                       8260 _pthread_mutex_firstfit_lock_slow  (in libsystem_pthread.dylib) + 248  [0x1074f3c98]
    +                                                         8260 _pthread_mutex_firstfit_lock_wait  (in libsystem_pthread.dylib) + 96  [0x1074f3d6c]
    +                                                           8260 __psynch_mutexwait  (in libsystem_kernel.dylib) + 8  [0x195ec6be8]
```
</details>

svobs:
In the current HEAD of `develop`, I see this comment above `CATransaction.flush()`:
https://github.com/iina/iina/blob/172e091776e9f02815aa5402c1366825c8965f5c/iina/ViewLayer.swift#L195-L197

But in this patch I see this:
https://github.com/iina/iina/blob/b4031123f23ec8eda357af2a6179192fcace4e9f/iina/ViewLayer.swift#L170-L175

If the original comment is correct, then `display` is being called while `displayLock` is still held. Perhaps `CATransaction.flush` is getting compiled into `CA::Transaction::commit`.

I question whether introducing another lock which overlaps the `isUninited` lock is the best approach. Seems like it would make for much cleaner & easier to understand code to refactor so that rendering only ever done via `mpvGLQueue`. I believe @krackers was the first to recommend this.

low-batt:
Excellent question. This:
```swift
// Must not call flush while holding isUninited's lock as that method may call display and our 
// locks do not support recursion. 
````
Has to do with `isUninited` using an implementation that does not support a thread locking the same lock a second time. Doing so will deadlock. This is not a problem for `displayLock` because it is using `NSRecursiveLock` which permits a thread to lock a lock more than once.

This particular locking pattern was chosen to match up with how mpv does it. _Except_ mpv is using a plain `NSLock` for `displayLock`. It really bothers me that I can not explain why `display` is not being called recursively in mpv allowing use of `NSLock`.

At one point I tried to change rendering to always use the `mpvGLQueue`. Users reported problems that I was unable to reproduce. Had to back out those changes.

**Comment by svobs** [comment]
Oh sorry, I misread your question.

I think it might be a special case. Based on my testing the deadlock only seems to occur when `display` is first called in `PlayerCore`:
```Swift
  func initVideo() {
    // init mpv render context.
    // The video layer must be displayed once to get the OpenGL context initialized.
    mainWindow.videoView.videoLayer.display()  // <-- here
```

I tried this code (https://github.com/iina/iina/commit/0ad796bd9f11cc83ae56a7a99dab302570a7f2d1) as a crude test and it allowed a regular `NSLock` to be used, which seems to support this idea.

low-batt:
There is a better way to create the OpenGL context. I will experiment with that and see if I can switch this to `NSLock`.

low-batt:
I confirmed your findings, but I'm still trying to figure out why the mpv code does not need to do this.

low-batt:
Marked this as a draft while I investigate further.

low-batt:
With regards to the question of why mpv does not need to use a recursive lock and yet IINA deadlocks unless a recursive lock is used…

To answer that we have to guess why `CATransaction.flush()` sometimes calls `ViewLayer.display`:
```text
 8260 PlayerCore.initVideo()  (in IINA) + 320  [0x102f84ad4]  PlayerCore.swift:508
  8260 @objc ViewLayer.display()  (in IINA) + 36  [0x102ebfb18]  <compiler-generated>:0
    8260 ViewLayer.display()  (in IINA) + 204  [0x102ebe9a8]  ViewLayer.swift:174
      8260 CA::Transaction::commit()  (in QuartzCore) + 648  [0x19d4cc2c0]
        8260 CA::Context::commit_transaction(CA::Transaction*, double, double*)  (in QuartzCore) + 456  [0x19d65d4d4]
          8260 CA::Layer::display_if_needed(CA::Transaction*)  (in QuartzCore) + 760  [0x19d4e9ae8]
            8260 @objc ViewLayer.display()  (in IINA) + 36  [0x102ebfb18]  <compiler-generated>:0
              8260 ViewLayer.display()  (in IINA) + 116  [0x102ebe950]  ViewLayer.swift:170
```

The stack trace indicates `CA::Layer::display_if_needed` decided `display` needed to be called. Debugging shows [needsDisplay](https://developer.apple.com/documentation/appkit/nsview/1483360-needsdisplay) is `false` before calling [flush](https://developer.apple.com/documentation/quartzcore/catransaction/1448270-flush), however [needsLayout](https://developer.apple.com/documentation/appkit/nsview/1526912-needslayout) is `true`. It _appears_ that `display_if_needed` is calling `display` because layout is needed.

The code identified by @svobs shows that `PlayerCore.initVideo` is calling `display` only to initialize the OpenGL context:
```swift
    // The video layer must be displayed once to get the OpenGL context initialized.
    mainWindow.videoView.videoLayer.display()
    mpv.mpvInitRendering()
```

To fix this the latest commit forms the OpenGL context (and pixel format) when the `ViewLayer` object is constructed. With that change the call to `display` in `initVideo` can be removed.

However fixing that uncovered a second call that may occur before laytout. This one is in `MainWindowController.windowWillOpen`:
```swift
    videoView.videoLayer.draw(forced: true)
```
This call was added in commit f921f4a47dc8da437e414c8c77daa08287b39287 with the comment:
> Draw black screen before playing a video

I discussed this with @uiryuu and he indicated this call can be removed.

With these changes I was able to use IINA's `Lock` class instead of a recursive lock.

low-batt:
I did some more testing and managed to find another way to trigger a deadlock. Investigating…

low-batt:
The new deadlock I hit was with switching out of music mode. Given the behavior of  `CATransaction.flush` I think it is risky to not use a recursive lock. So the latest commit switches this back to `NSRecursiveLock`.

iina/ViewLayer.swift
```diff
@@ -43,92 +54,62 @@ class ViewLayer: CAOpenGLLayer {
...
-      CGLChoosePixelFormat(attributes, &pix, &npix)
-      if let pix = pix {
-        Logger.log("Created OpenGL pixel format with \(attributes)", level: .debug)
-        return pix
-      }
+  override func canDraw(inCGLContext ctx: CGLContextObj, pixelFormat pf: CGLPixelFormatObj,
+                        forLayerTime t: CFTimeInterval, displayTime ts: UnsafePointer<CVTimeStamp>?) -> Bool {
+    videoView.$isUninited.withLock() { isUninited in
+      guard !isUninited else { return false }
+      guard !forceRender else { return true }
```
uiryuu:
```suggestion
      if forceRender { return true }
```

iina/ViewLayer.swift
```diff
@@ -17,12 +17,23 @@ class ViewLayer: CAOpenGLLayer {
...
+  /// Lock to single thread calls to `display`.
+  ///
+  /// A recursive lock is needed because the call to `CATransaction.flush()` in `display` calls `display_if_needed`
+  /// which will then call `display` if layout is needed. See the discussion in PR #5029.
+  private let displayLock = NSRecursiveLock()
+
   private var fbo: GLint = 1
 
   private var needsMPVRender = false
   private var forceRender = false
```
uiryuu, line 32: I think we should combine these two variables into a single variable to improve readibility, something like
```swift
private var needsRender = [.no, .yes. force]
```

iina/ViewLayer.swift
```diff
@@ -154,10 +135,18 @@ class ViewLayer: CAOpenGLLayer {
...
     // the lock and our locks do not support recursion.
     display()
+  }
 
+  override func display() {
+    displayLock.lock()
+    defer { displayLock.unlock() }
+
+    super.display()
+    CATransaction.flush()
```
uiryuu, line 145: Why we need to flush here?

uiryuu:
> To fix this the latest commit forms the OpenGL context (and pixel format) when the ViewLayer object is constructed. With that change the call to display in initVideo can be removed.

I remember because of this, when we open a window, we must draw a black window first in order to make the GLLayer init correctly; at that point, mpv isn't ready, so we don't know the size of the video, which causes the window being stretched to the final window size. With this fix, I think we can finally display the window after the size is available, thus avoiding the arguably ugly stretch animation (which some users has complained about).

svobs:
> I did some more testing and managed to find another way to trigger a deadlock. Investigating…

Can you elaborate on this? It sounds like you are saying that doing a "forced" draw can trigger a deadlock? Or is the deadlock happening each time a `layout` is done?

It would be great to identify all the places the video layer's `display` method is being called, because they should be very limited. I recently found documentation on [layerContentsRedrawPolicy](https://developer.apple.com/documentation/appkit/nsview/1483514-layercontentsredrawpolicy) which states the following:

> For a view that has no associated layer, or that has been assigned a developer-provided layer (a layer-hosting view) using the [layer](https://developer.apple.com/documentation/appkit/nsview/1483298-layer) property, the default contents redraw policy is [NSView.LayerContentsRedrawPolicy.never](https://developer.apple.com/documentation/appkit/nsview/layercontentsredrawpolicy/never) and the [layerContentsPlacement](https://developer.apple.com/documentation/appkit/nsview/1483375-layercontentsplacement) property is set to [NSView.LayerContentsPlacement.scaleAxesIndependently](https://developer.apple.com/documentation/appkit/nsview/layercontentsplacement/scaleaxesindependently). These policies tell AppKit not to replace the layer’s content and to provide the same content placement as the [resize](https://developer.apple.com/documentation/quartzcore/calayercontentsgravity/1410811-resize) option.

I confirmed via testing that `VideoView` has redraw policy = `never`. So it would follow that its `needsDisplay` will always be `false` as well (unless changed in our code). So what are the consequences of that? Well, I stumbled onto [this dense post](https://stackoverflow.com/a/11213382/1347529) which I am still digesting. It might also be interesting for reasons outside this PR.

lhc70000:
> This call was added in commit https://github.com/iina/iina/commit/f921f4a47dc8da437e414c8c77daa08287b39287 with the comment:
> 
> Draw black screen before playing a video
> 
> I discussed this with @uiryuu and he indicated this call can be removed.

I remember that this was added to fix a rare issue: sometimes when opening a video, the last frame of the video played before will be displayed in the window before the playback starts. This happened more than once when the user complains IINA reveals some private video they played last time. We wasn't able to readily reproduce this, so we added a general fix.

IMO the most possible reason is that the newly created layer may contain random or undefined data, and we should manually clear it before use. So even this call is removed, we need to call `glClear()` somewhere after creating the layer and before displaying it.

iina/ViewLayer.swift
low-batt: Implicit transactions that are not on the main thread must be manually flushed. On the main thread they are automatically flushed at the end of the run loop. There is a question as to whether we should be creating implicit transactions off the main thread. See issue #5038.

iina/ViewLayer.swift
low-batt: We could, but that would deviate from how mpv handles this.

iina/ViewLayer.swift
low-batt: Will fix.

low-batt:
When I went back to using a recursive lock I added the drawing back in to `windowDidOpen`. It bothered me that I could not explain why it was necessary to black out the screen, so I was worried about removing it.

That it has to do with privacy and window reuse explains why IINA differs from mpv in this case as it is unlikely mpv is reusing windows. This is also an example of code whose purpose is not immediately clear and could use a comment. I will add one.

I believe this means the problem @uiryuu mentioned [above](https://github.com/iina/iina/pull/5029#issuecomment-2205228734) with IINA drawing before the size of the video is known is still present and needs to be addressed.

On this:
> newly created layer may contain random or undefined data

I think it is more straight forward than that? The cached player contains the window, view and layer. The previous layer is being reused. I set a breakpoint in `ViewLayer.init` and it was only hit once when the window was initially created. Closing the window and opening another video did not create a new layer.

That should mean that instead of drawing a black screen at the start and thus causing the issue @uiryuu mentioned we should be able to clear the layer when the window is closing. This should be addressed in a separate PR.

krackers:
@svobs 

>So what are the consequences of that?

Fwiw option 2 in that post is what iina does currently in the non live-resize case, while option 1 is what it (now) does during live-resize. The `setNeedsDisplayOnBoundsChange:YES` is interesting, I'm not sure if setting this would solve the blurriness during live-resize.

Also the line
> Internally a NSOpenGLView creates a NSOpenGLLayer when you make it layered,

is no longer true, on 10.14+ (?) a layer-backed NSOpenGLView differs from a NSOpenGLLayer. In fact even if you don't declare an NSOpenGLView as explicitly layer-backed, it still appears to composite to a layer. This is the only way to get push based semantics for opengl drawing with layer-backing.



---

>newly created layer may contain random or undefined data, and we should manually clear it before use

I've seen this happen with raw nsopenglview on older os (which makes sense since an opengl surface is basically mapped almost directly onto a chunk of vram), but I would have thought that a layer-backed view would properly handle this to avoid compositing garbage. Maybe you could set `backgroundColor` on the layer? 

Edit: Missed lowbatt's comment, if it's indeed the same layer being reused (and not a new layer) that would explain it.

low-batt:
On the question from @svobs about when the deadlock occurs…

I don't think this Apple code is open source? At least I have not been able to find it. So we have to guess what is going on.

From that stack trace we know the call to `CATransaction.flush` causes the implicit transaction to be committed. As a part of that `display_if_needed` is called and that method decides `display` should be called. This we know for certain.

As mpv is not using a recursive lock apparently how they are drawing never results in `display_if_needed` calling `display`. Seems something IINA is doing affects the behavior of `flush`.

From logging the state of `needsLayout` before the call to `CATransaction.flush` it appears there is a correlation between needing layout and `display_if_needed` calling `display`. The cases I found so far where this happened was the early calls to display when starting to play a video and switching from music mode to the main window. These are all operations mpv does not need to do.

@saagarjha is attempting to confirm this is expected `CATransaction.flush` behavior.

lhc70000:
> I think it is more straight forward than that? The cached player contains the window, view and layer. The previous layer is being reused. I set a breakpoint in ViewLayer.init and it was only hit once when the window was initially created. Closing the window and opening another video did not create a new layer.

Yes, that might also be a reason. However I vaguely remember that this could also happen when you open a fresh IINA instance. Of course it was a very old issue, and things may have changed now.

low-batt:
For a fresh IINA process to display a frame from a previously played video likely means macOS had a security defect. The OS can cache and reuse some things, such as file buffers. But a new process should not have any access to graphics memory from some other process.

Maybe someone thought IINA was being opened fresh when it was still open in the dock, just with no windows currently open?

svobs:
> The cases I found so far where this happened was the early calls to display when starting to play a video and switching from music mode to the main window.

Did some more testing to bring out the deadlock. I reproduced it via the "forced" display while exiting music mode, but not entering... Oddly, I was not able to make it deadlock when toggling PiP, which should be very similar to the music mode toggle. Possibly a race condition? Was hoping to find more of a pattern, but doubt it's crucial to keep dwelling on it if the recursive lock fixes it.

PR #5042, opened 2024-07-03
Remove the custom Switch and use NSSwitch

uiryuu:

---

**Description:**
This PR removes the `Switch.swift` file and replaced the 3 Switches used in the video tab of QuickSettingView with NSSwitch. One potential problem is that the translations are missing for the text fields, before they were set programmatically. We can either leave it as is and let the Crowdin community to translate them, or I can set them programmatically too. Thoughts?


low-batt (approved):
Pulled PR, built IINA, confirmed the operation of each of the three switches.
Changes look good to me.

Issue #4553, opened 2023-08-02
User Opt In to Hide the Progress Bar || Lock Window


cursed-github:
**What you want IINA to do:**
Ability to hide the progress bar so that the user can opt-out to see the time left similar to a movie theater, accidental or non-accidental mouse touch can ruin the scene.

**What IINA does currently:**

**Why you think this should be added:**
The user does not get distracted by the progress bar by an accidental mouse click

**Examples of other projects that have something similar:**
QuickTime player on Mac has a separate window for the progress bar, this separate window can be minimized for distraction-free video watching experience

low-batt:
As IINA can be controlled using keyboard shortcuts, the media keys and an Apple TV remote (with additional software), users should be able to disable the OSC.

This should be implemented by adding a `Enable OSC` setting in IINA's settings under the `UI` tab to match up with the existing `Enable OSD` setting.

**NOTE** that this may not entirely address the desire expressed of a "distraction-free video watching experience". To me that would include not displaying the cursor. There are already some cursor handling problems that need to be investigated. We recently had to add a workaround due to a regression in AppKit causing the cursor to inappropriately be shown. That problem has been reported to Apple. No response last I looked. With AppKit malfunctioning, IINA may need to switch to manually handling cursor visibility.

low-batt:
It is important to stay consistent with mpv and whenever possible support mpv options. With respects to cursor handling mpv provides the [cursor-autohide](https://mpv.io/manual/stable/#options-cursor-autohide) option:
> **--cursor-autohide=<number|no|always>**
>    Make mouse cursor automatically hide after given number of milliseconds (default: 1000 ms). no will disable cursor autohide. always means the cursor will stay hidden.

And the [cursor-autohide-fs-only](https://mpv.io/manual/stable/#options-cursor-autohide-fs-only) option:
> **--cursor-autohide-fs-only**
>    If this option is given, the cursor is always visible in windowed mode. In fullscreen mode, the cursor is shown or hidden according to `--cursor-autohide`.

svobs:
> QuickTime player on Mac has a separate window for the progress bar, this separate window can be minimized for distraction-free video watching experience

What, really? Curious how to activate this feature. I didn't think QuickTime Player had any customization options for its on-screen controls.

low-batt:
@svobs This is the issue I was thinking of concerning the `Enable OSC` setting.

svobs:
[...]

PR #5044, opened 2024-07-04
Add ability to disable title bar & OSC via enableTitleBarAndOSC pref key

svobs:
- [x] This implements/fixes issue #4553.

---

**Description:**
* Adds new Preference.Key: `.enableTitleBarAndOSC`, default `true`
* Minor reordering of OSC-related pref keys to make more sense, and update some of their comments which fell out of date (sorry, couldn't help myself), as well as making it clearer how some of them work together
* Modify `setupOnScreenController` and logic while entering & exiting full screen to support the new pref
* Add observer for the new key to `MainWindowController` so that `setupOnScreenController` is called as soon as it is changed

This PR only adds the feature itself and its pref entry. I'll submit another PR which adds controls for it to the Settings UI. For those who want to test this for the time being, they can enter the following in the Terminal:
```
# Disable OSC
defaults write com.colliderli.iina enableTitleBarAndOSC 0
```

```
# Enable OSC
defaults write com.colliderli.iina enableTitleBarAndOSC 1
```

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -312,14 +312,14 @@ class MainWindowController: PlayerWindowController {
   private var oscIsInitialized = false
 
   // Cached user default values
-  private lazy var enableOSC: Bool = Preference.bool(for: .enableOSC)
+  private lazy var enableTitleBarAndOSC: Bool = Preference.bool(for: .enableTitleBarAndOSC)
   private lazy var oscPosition: Preference.OSCPosition = Preference.enum(for: .oscPosition)
   private lazy var arrowBtnFunction: Preference.ArrowButtonAction = Preference.enum(for: .arrowButtonAction)
   private lazy var pinchAction: Preference.PinchAction = Preference.enum(for: .pinchAction)
   lazy var displayTimeAndBatteryInFullScreen: Bool = Preference.bool(for: .displayTimeAndBatteryInFullScreen)
 
   private let localObservedPrefKeys: [Preference.Key] = [
-    .enableOSC,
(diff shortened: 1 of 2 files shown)


low-batt:
I noticed it does not include suppressing the title. For the use case we are targeting it should also suppress the title. Or we could control the title separately. Or maybe not show it upon mouse movement unless the mouse is at the top of the screen where the title would be. Like the Mac menu bar behaves when in full screen mode.

lhc70000:
Continue from #4553...

Is this feature implemented to solve #4553? If so, then I think it's better to make this a runtime property rather than a pref key, because the user probably only wants to disable the OSC when watching the movie. It would be a hassle if they must open the settings window to enable/disable this feature. And there's little reason to remember this setting permanently.

svobs:
> Is this feature implemented to solve https://github.com/iina/iina/issues/4553?

I probably should have filled an issue which better fit this PR...
@lhc70000 I'll respond in the comments for #4553.

svobs:
As noted in #4553, I've refactored so that the setting now covers both title bar & OSC.

It's designed to co-exist with `disableUI`, so that in effect either setting will disable the "fadeable views" in the window. I noticed that `disableUI` has some quirks. For example, when `disableUI` is true, the OSD is still shown with a vertical offset which is influenced by the size of the top bar, so if the OSC is configured to be at "top", the OSD will be lower, like this:
SCR-20240725-cgap

I decided to use a constant vertical offset when `enableTitleBarAndOSC` is false which matches the title bar height instead, like this: 
SCR-20240725-cdld

PR #5054, opened 2024-07-08
Don't resize on opening media and switching modes

uiryuu:
---
**Description:**
Thanks to https://github.com/iina/iina/pull/5029, now initiating rendering doesn't require a forced draw to the main window. Previously, IINA first draw the video in a small window, then resize to the desired window size. This PR change the behavior to show the main window after the geometry is calculated, therefore no resizing animation is required, the window will show directly to the final size.

Specifically, this PR:
- Removed `windowWillOpen()` and moved contents to newly added `override showWindow(_:)`
- Removed `windowDidOpen()` and moved contents to the end of `windowDidLoad()`
  - `PlayerWindowController.windowDidOpen()` also got removed.
- Removed code which explicitly shows the main window when opening media and creating the main window
- Calls `showWindow(_:)` after the final window geometry is calculated and set as the window frame
- Show mini player after videoreconfig

`windowWillOpen()` and `windowDidOpen()` are not standard NSWindow functions, and they are easy to be confused with other NSWindow lifecycle functions, so I moved them and move the code into override functions.

The current implementation is far from clean, but in order to not refactor too much code, I only made changes to necessary parts. Also I feel the animation for changing modes (music -> normal and reverse) bit clunky and not consistent, but that's not in the scope of this PR.

**Demo:**
Before:
https://github.com/iina/iina/assets/20237141/241eca90-66b5-47e9-92c1-551316fc1e36

After:
https://github.com/iina/iina/assets/20237141/3ea7629e-38b2-47ba-ba5c-9c36c9707990

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -180,6 +180,9 @@ class MainWindowController: PlayerWindowController {
   var isShowingPersistentOSD = false
   var osdContext: Any?
 
+  /** This variable is true when the window should show but waiting for size from mpv */
+  var pendingShow = false
+
   // MARK: - Enums
 
   // Window state
@@ -650,7 +653,6 @@ class MainWindowController: PlayerWindowController {
 
     player.events.emit(.windowLoaded)
 
-    isClosing = false
     // Must workaround an AppKit defect in some versions of macOS. This defect is known to exist in
     // Catalina and Big Sur. The problem was not reproducible in early versions of Monterey. It
     // reappeared in Ventura. The status of other versions of macOS is unknown, however the
@@ -2567,7 +2569,8 @@ class MainWindowController: PlayerWindowController {
       if let screenFrame = window.screen?.frame {
         rect = rect.constrain(in: screenFrame)
       }
-      if !player.mainWindow.window!.isVisible && !player.isInMiniPlayer {
+      if pendingShow && !player.isInMiniPlayer {
+        pendingShow = false
         window.setFrame(rect, display: false, animate: false)
         showWindow(nil)
       } else if player.disableWindowAnimation || Preference.bool(for: .disableAnimations) {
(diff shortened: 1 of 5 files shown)


low-batt (approved):
Pulled PR, built IINA, tested. Looks good to me.

svobs:
Looks like some much needed refactoring. Although I for one began to enjoy the zoom animation 😂

uiryuu:
Known issues: when opening online resources, the window will still resize after the buffering is done. We need a UI to reflect the buffering process.

lhc70000:
Try this:

- Enable auto switching to music mode (enabled by default)
- Open a music file, IINA opens in music mode
- Click the back button to return to normal mode

Expected: The main window appears with the album art.
Actual: No video window shown, the welcome window is shown instead. The player hangs when trying to quit IINA.

lhc70000:
@svobs I actually feel nostalgic about the animation. Maybe I'm going to implement a more sleek opening animation and make it an option.

iina/MainWindowController.swift
```diff
@@ -2584,7 +2570,10 @@ class MainWindowController: PlayerWindowController {
       if let screenFrame = window.screen?.frame {
         rect = rect.constrain(in: screenFrame)
       }
-      if player.disableWindowAnimation || Preference.bool(for: .disableAnimations) {
+      if !player.mainWindow.window!.isVisible && !player.isInMiniPlayer {
+        window.setFrame(rect, display: false, animate: false)
+        showWindow(nil)
```
lhc70000: Why put it here in `adjustFrameByVideoSize()`? This can be called whenever the video's size is changed.

Consider the following case: I open an IINA window and hide it using Cmd+H. Then it automatically continues to the next video (or some plugin/mpv script changed the video/window size), and this line will show the window unexpectedly.

low-batt:

PR #5065, opened 2024-07-12
Minor UI improvements to Settings Network > Cache

svobs:

---

**Description:**

See [Cache](https://mpv.io/manual/stable/#options-cache) section from mpv manual.

* Change wording from `Default cache size` to `Maximum cache size` to better reflect underlying property (`demuxer-max-bytes`). Existing UI is almost misleading. This is a hard limit and the most important number. Calling it "default" doesn't make sense (unless I'm missing something?)
* Moves the max cache size to the actual bottom line to signify its importance (along with the stronger wording). The seconds to prefetch (corresponding to `cache-secs`) is subordinate to this.


uiryuu:
Thanks!

Issue #4639, opened 2023-10-05
The EQ setting cannot be saved. We need this feature. Thanks!


Gabchie:
**What you want IINA to do:**

**What IINA does currently:**

**Why you think this should be added:**

**Examples of other projects that have something similar:**

Gabchie:
The EQ setting cannot be saved. We need this feature. Thanks!

PR #5070, opened 2024-07-14
Added EQ presets and EQ profile saving functionalities

### LINKED ISSUE #525: Saving audio EQ settings
- Save audio EQ sets / set as default
- Sync UI with audio filter `@iina_aeq` it it exists

uiryuu:
- [x] This implements/fixes issue:
  - [x] #525
  - [x] https://github.com/iina/iina/issues/4639

---

**Description:**

This PR added the audio EQ settings. A bunch of EQ settings from iTunes are added as presets (thanks @aizcutei for extracting data from iTunes on windows!). Use defined EQs can be saved to UserDefaults.

Other changes:
- "31.25" and "62.5" was changed to "32" and "64" respectively, for a cleaner UI and following the way of Music.app.
- Use `equalizer` instead of `anequalizer`
- When adjusting EQ, only add one audio filter; previous we used 10 audio filters for EQ.

The new EQ settings:
image

Thanks @lhc70000 for the new design and UI adjustments!

### DIFF
--- a/iina/Equalizations.swift
+++ b/iina/Equalizations.swift
@@ -20,12 +20,11 @@ class EQProfile: Codable {
 
 class PresetEQProfile: EQProfile {
   let localizationKey: String
-  var name: String {
-    NSLocalizedString(localizationKey, comment: localizationKey)
-  }
+  let name: String
 
   init(_ name: String, _ values: [Double]) {
     self.localizationKey = "eq.preset." + name
+    self.name = NSLocalizedString(localizationKey, comment: localizationKey)
     super.init(values)
   }
   
--- a/iina/Equalizations.swift
+++ b/iina/Equalizations.swift
@@ -6,7 +6,7 @@
 //  Copyright © 2024 lhc. All rights reserved.
 //
 
-struct EQProfile: Codable {
+class EQProfile: Codable {
   var gains = [Double](repeatElement(0.0, count: 10))
 
   init(_ values: [Double]) {
@@ -22,5 +22,11 @@ let presetEQs: KeyValuePairs = ["Flat": EQProfile([0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
                                 "Acoustic": EQProfile([0, 0, 0, 0, 12, 0, 0, 0, 0, 0]),
                                 ]
 
-var userEQs = ["test": EQProfile([12, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
-              ]
+var userEQs: Dictionary<String, EQProfile> = [:] {
+  didSet {
+    let encoder = JSONEncoder()
+    if let encoded = try? encoder.encode(userEQs) {
+      UserDefaults.standard.set(encoded, forKey: Preference.Key.userEQPresets.rawValue)
+    }
+  }
+}
--- a/iina/Equalizations.swift
+++ b/iina/Equalizations.swift
@@ -18,9 +18,46 @@ class EQProfile: Codable {
   }
 }
 
-let presetEQs: KeyValuePairs = ["Flat": EQProfile([0, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
-                                "Acoustic": EQProfile([0, 0, 0, 0, 12, 0, 0, 0, 0, 0]),
-                                ]
+class PresetEQProfile: EQProfile {
+  let localizationKey: String
+  var name: String {
+    NSLocalizedString(localizationKey, comment: localizationKey)
+  }
+
+  init(_ name: String, _ values: [Double]) {
+    self.localizationKey = "eq.preset." + name
+    super.init(values)
+  }
+  
+  required init(from decoder: any Decoder) throws {
+    fatalError("init(from:) has not been implemented")
+  }
+}
+
+let presetEQs: [PresetEQProfile] = [
+  .init("flat", [0,0,0,0,0,0,0,0,0,0]),
+  .init("acoustic", [5.0, 4.9, 3.95,1.05,2.15,1.75,3.5,4.1,3.55,2.15]),
+  .init("classical", [4.75,3.75,3,2.5,-1.5,-1.5,0,2.25,3.25,3.75]),
+  .init("dance", [3.57,6.55,4.99,0,1.92,3.65,5.15,4.54,3.59,0]),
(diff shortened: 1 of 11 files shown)


iina/QuickSettingViewController.swift
```diff
@@ -214,6 +221,19 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
...
+      let decoder = JSONDecoder()
+      let dict = try? decoder.decode(Dictionary<String, EQProfile>.self, from: data)
+      userEQs = dict ?? [:]
+    }
+
+    eqPopUpButton.menu!.delegate = self
+    presetEQs.forEach { preset in
+      eqPopUpButton.menu?.addItem(withTitle: preset.name, tag: 1, obj: preset)
+    }
+    eqPopUpButton.selectItem(withTag: 1000)
```
lhc70000: Make this a constant (also for -1, -2, etc)

iina/QuickSettingViewController.swift
```diff
@@ -214,6 +221,19 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
     switchHorizontalLine2.wantsLayer = true
     switchHorizontalLine2.layer?.opacity = 0.5
 
+    if let data = UserDefaults.standard.data(forKey: Preference.Key.userEQPresets.rawValue) {
+      let decoder = JSONDecoder()
+      let dict = try? decoder.decode(Dictionary<String, EQProfile>.self, from: data)
+      userEQs = dict ?? [:]
+    }
```
lhc70000, line 234:
```swift
if let data = UserDefaults.standard.data(forKey: Preference.Key.userEQPresets.rawValue),
   let dict = try? JSONDecoder().decode(Dictionary<String, EQProfile>.self, from: data) {
  userEQs = dict
}
```

iina/QuickSettingViewController.swift
```diff
@@ -8,7 +8,7 @@
 
 import Cocoa
 
-class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTableViewDelegate, SidebarViewController {
+class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTableViewDelegate, NSMenuDelegate, SidebarViewController {
```
lhc70000: I'd prefer using an extension

iina/QuickSettingViewController.swift
```diff
@@ -825,26 +845,110 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
...
+      }
+      if nameList.contains( where: { $0 == input } ) {
+        firstButton.isEnabled = false
+      } else {
+        firstButton.isEnabled = true
+        self.inputString = input
+      }
+    }
+    return panel
+  }
```
lhc70000: We should reuse `Utility.quickPromptPanel()`. You can make it accept a validation block or something like this.

It would be even better if you can add some help text to tell the user why the input is not accepted, with some default messages like "Name already exists", which can also be reused elsewhere.

Or you can leave this to me after merging this PR.

iina/QuickSettingViewController.swift
```diff
@@ -825,26 +845,110 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
...
+        userEQs[inputString] = profile
+        menuNeedsUpdate(eqPopUpButton.menu!)
+        eqPopUpButton.selectItem(withTitle: inputString)
+      } else {
+        eqPopUpButton.selectItem(withTitle: lastUsedProfileName)
+      }
+    case -1: // delete
+      userEQs.removeValue(forKey: lastUsedProfileName)
+      menuNeedsUpdate(eqPopUpButton.menu!)
+      eqPopUpButton.selectItem(withTag: 1000)
```
lhc70000: You use "remove the current EQ" in the UI. My intuition is that it will also reset the EQ to default. However it only deletes the current preset. I suggest more precise working like "Save the current EQ as preset" "Delete the current EQ preset".

iina/QuickSettingViewController.swift
lhc70000: There's another bug: you also need to make sure the user defined name is not "Manual" (or localized equivalent).

I'm thinking about whether it's safe to trust localized names in the code. If there are duplicated preset names (whether due to translation mistake or whatever reason), code like `eqPopUpButton.selectItem(withTitle:)` will not work properly. So perhaps it's better to use some internal enum or identifiers.

iina/Equalizations.swift
```diff
@@ -0,0 +1,69 @@
...
+  init(fromCurrentSliders sliders: [NSSlider]) {
+    gains = sliders.map { $0.doubleValue }
+  }
+}
+
+class PresetEQProfile: EQProfile {
+  let localizationKey: String
+  var name: String {
+    NSLocalizedString(localizationKey, comment: localizationKey)
+  }
```
lhc70000: Use `lazy var` so it can be cached.

(Why make it computed? Why not setting the name directly in `init()`?)

iina/PlayerCore.swift
```diff
@@ -1340,19 +1340,13 @@ class PlayerCore: NSObject {
...
+      return "equalizer=f=\(freq):t=h:width=\(Double(freq) / 1.224744871):g=\(gains[index])"
     }
-    filters.forEach { _ = addAudioFilter($0) }
-    info.audioEqFilters = filters
-  }
-
-  func removeAudioEqFilter() {
-    info.audioEqFilters?.compactMap { $0 }.forEach { _ = removeAudioFilter($0) }
-    info.audioEqFilters = nil
+    let filter = MPVFilter(name: "lavfi", label: Constants.FilterName.audioEq, paramString: "[\(filterStrings.joined(separator: ","))]")
```
lhc70000:
```swift
let freqList = [32, 64, 125, 250, 500, 1000, 2000, 4000, 8000, 16000]
let paramString = freqList.enumerated().map { (index, freq) in
  "equalizer=f=\(freq):t=h:width=\(Double(freq) / 1.224744871):g=\(gains[index])"
}.joined(separator: ",")
let filter = MPVFilter(name: "lavfi", label: Constants.FilterName.audioEq, paramString: "[\(paramString)]")
```

Slightly better?

lhc70000 (changes requested):

lhc70000 (approved):

PR #5071, opened 2024-07-14
Honor “open in new window” setting when opening multiple files at once

### LINKED ISSUE #3531: Open Multiple Windows
I keep figuring this has got to be a pref, but I don't see it.

iina seems to open only one video file at a time. I'd like to see it able to open multiple windows.

Cykelero:

---

## Motivation

Right now, when opening multiple files at once (e.g. dragging them on the app's Dock icon), IINA opens a single window, with the files as playlist items—even when the “Always open media in new window” setting is on.
To me, this isn't the expected behavior; this setting essentially allows user to choose between “I prefer using a single window with a playlist” and “I prefer using one window for each file”, and the batch open behavior fails to honor that preference.

(As a personal data point: for the longest time I thought this was a bug in IINA, that it failed to open multiple files at once; not realizing a playlist was being created. It's only when I started digging into the source to fix the bug that I discovered the actual behavior.)

## Change

With this PR, the app instead opens each file in its own window, if the `alwaysOpenInNewWindow` setting is true. If the setting is false, the current behavior is kept.

## Bonus change

The PR also fixes a related bug (when the “When media is opened > Pause” setting is enabled) where opening multiple files at once sometimes allows them to play for a while before being paused

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -720,7 +720,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
 
     // open pending files
     pendingFilesForOpenFile.removeAll()
-    if PlayerCore.activeOrNew.openURLs(urls) == 0 {
+    if PlayerCore.openURLs(urls) == 0 {
       Utility.showAlert("nothing_to_open")
     }
   }
@@ -882,8 +882,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
         }
       }
       let isAlternative = (sender as? NSMenuItem)?.tag == AlternativeMenuItemTag
-      let playerCore = PlayerCore.activeOrNewForMenuAction(isAlternative: isAlternative)
-      if playerCore.openURLs(panel.urls) == 0 {
+      if PlayerCore.openURLs(panel.urls, inverseOpenInNewWindowPref: isAlternative) == 0 {
         Utility.showAlert("nothing_to_open")
       }
     }
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -86,6 +86,37 @@ class PlayerCore: NSObject {
     let useNew = Preference.bool(for: .alwaysOpenInNewWindow) != isAlternative
     return useNew ? newPlayerCore : active
   }
+  
+  /**
+   Depending on the `alwaysOpenInNewWindow` pref, opens the URLs in a single window, or multiple windows.
+
+   - Returns: `nil` if no further action is needed, like opened a BD Folder; otherwise the
+   count of playable files.
+   */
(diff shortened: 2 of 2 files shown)


low-batt:
I believe this addresses issue #3531.

uiryuu (changes requested):
When I run IINA on this branch, it always start with a black screen together with the welcome window. Can you fix this?

Cykelero:
Ack, sorry I didn't catch that bug—I clearly didn't test the changes enough.

I moved the pause call to fix the issue. I'm not sure it belongs there, but this seems to work without side-effects.
I'll keep using this build to see if anything else seems off.

uiryuu (approved):
LGTM. Just a side note, `func application(_ sender: NSApplication, openFile filename: String) -> Bool` only handles file openings from the system, i.e. opening multiple files in Finder and drag and drop to the app icon. There're 2 other ways to open files in IINA: drag and drop to the welcome window and the main window. I think in these cases opening multiple files in the same window is reasonable.

Cykelero:
I found and fixed an additional bug: when opening IINA with a file open (e.g. dragging a file onto the Dock icon), the file would correctly open, but the welcome window would open on top. That was caused by a flag no longer being set.

uiryuu:
Merged locally via 16e271402a12916bce42a0c92b4d879f477c48b5. Thanks for you work!

uiryuu:
Could you please rebase this PR to develop branch? Thanks

Cykelero:
Rebased!
(I also squashed a couple of commits that didn't need to be separate)

Cykelero:
Actually, it looks like I haven't managed to rebase the fix for the “file plays a bit when opened” bug. The rebased fix commit doesn't seem to be helping much.

I've just spent a bunch of time trying to work around the bug, to no avail—@low-batt, do you have some insight into this?
The [previous version of the fix](https://github.com/iina/iina/commit/3cc7f0193695945f62e42f964f03b7a267ad5835) worked, but I haven't been able to port that to the new `PlayerState`-based code.
(easy bug repro: close IINA, and open it by dragging a movie onto its icon. you should hear sound for a fraction of a second)

Cykelero:
After a break, I restarted this branch from scratch; some of the more complicated stuff had become irrelevant.

It now has two commits:
- The first commit honors the “open in new window” setting when opening a file (both when opening through a macOS mechanism like dragging files on the app icon; and when using FIle > Open)
- The second commit makes sure a duplicate window isn't created when opening a file that's already opened

I tested all the configurations I could think of, and things seem good!

uiryuu:
Thanks! Changes will be included in the upcoming 1.4.0-beta2.

Cykelero:
Awesome, thank you!

low-batt:
@Cykelero Thanks for all your work on this!

PR #5074, opened 2024-07-16
Combine separate flags into a player state property

low-batt:
This commit will:
- Add a `PlayerState` struct that defines the mutually exclusive states a `PlayerCore` can be in
- Add a property to `PlaybackInfo` to hold a player's state
- Remove the `fileLoading`, `isStopping`, `isStopped`, `isShuttingDown` and `isShutdown` properties from `PlayerCore`
- Change the `isIdle`, i`sPaused`, `isPlaying` and `isSeeking` `PlaybackInfo` properties to be computed properties that check the state of the player
- Add an i`sActive` computed property to `PlaybackInfo` that checks if the player is in an active state
- Move remaining `MPV_EVENT_SHUTDOWN` processing from `MPVController` to `PlayerCore` and the main thread to avoid the need for locks
- Change handling of `MPV_EVENT_END_FILE` to check the player state instead of `MPV_END_FILE_REASON_STOP`
- Add an `isActive` guard to `PlaylistViewController.reloadData`

This makes the state the player is in explicit in one property. This also corrects some errors in the app termination sequence.

---

**Description:**

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -429,9 +429,9 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
       // For debugging list players that have not terminated.
       for player in PlayerCore.playerCores {
         let label = player.label ?? "unlabeled"
-        if !player.isStopped {
+        if player.info.state == .stopping {
           Logger.log("Player \(label) failed to stop", level: .warning)
-        } else if !player.isShutdown {
+        } else if player.info.state == .shuttingDown {
           Logger.log("Player \(label) failed to shutdown", level: .warning)
         }
       }
@@ -501,7 +501,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
     // termination is initiated.
     allPlayersHaveShutdown = true
     for player in PlayerCore.playerCores {
-      if !player.isShutdown {
+      if player.info.state != .shutDown {
         allPlayersHaveShutdown = false
         break
       }
@@ -585,7 +585,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
       if !allPlayersHaveShutdown {
         // If any player has not shutdown then continue waiting.
         for player in PlayerCore.playerCores {
-          guard player.isShutdown else { return }
+          guard player.info.state == .shutDown else { return }
         }
         allPlayersHaveShutdown = true
         // All players have shutdown.
@@ -646,7 +646,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
 
     // Instruct any players that are already stopped to start shutting down.
     for player in PlayerCore.playerCores {
-      if player.isStopped && !player.isShutdown {
+      if player.info.state == .stopped || player.info.state == .idle {
         player.shutdown()
       }
     }
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -646,7 +646,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
 
     // Instruct any players that are already stopped to start shutting down.
     for player in PlayerCore.playerCores {
-      if player.info.state == .stopped || player.info.state == .idle {
+      if player.info.state == .idle {
         player.shutdown()
       }
     }
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -117,10 +117,6 @@ class MPVController: NSObject {
   var recordedSeekStartTime: CFTimeInterval = 0
   var recordedSeekTimeListener: ((Double) -> Void)?
 
-  var receivedEndFileWhileLoading: Bool = false
-
-  var fileLoaded: Bool = false
-
   @Atomic private var hooks: [UInt64: MPVHookValue] = [:]
   private var hookCounter: UInt64 = 1
 
@@ -1088,7 +1084,7 @@ class MPVController: NSObject {
       }
 
     case MPV_EVENT_FILE_LOADED:
-      DispatchQueue.main.async { self.onFileLoaded() }
+      DispatchQueue.main.async { self.player.fileLoaded() }
 
     case MPV_EVENT_SEEK:
       DispatchQueue.main.async { [self] in
@@ -1112,7 +1108,7 @@ class MPVController: NSObject {
         // When playback is paused the display link may be shutdown in order to not waste energy.
         // The display link will be restarted while seeking. If playback is paused shut it down
         // again.
-        if player.info.isPaused {
+        if player.info.state == .paused {
           player.mainWindow.videoView.displayIdle()
         }
         if needRecordSeekTime {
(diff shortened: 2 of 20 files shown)


low-batt:
@lhc7000 @uiryuu This is the refactoring to eliminate the multiple player boolean flags.

low-batt:
You will see I left some extra logging in that will need to be changed to level `.verbose` before the beta release.

iina/PlayerState.swift
```diff
@@ -0,0 +1,105 @@
...
+    case .shuttingDown: return "shuttingDown"
+    case .shutDown: return "shutDown"
+    default:
+      var result = ""
+      for state in PlayerState.allStates {
+        guard self.contains(state) else { continue }
+        if !result.isEmpty {
+          result += ","
+        }
+        result += state.description
```
uiryuu: I don't like the recursive implementation here, it's not very straightforward and prone to bugs. I think we can simply declare a dictionary of [Option, String] and use `filter` + `join` to form the description.

iina/PlaybackInfo.swift
```diff
@@ -34,12 +34,67 @@ class PlaybackInfo {
...
+  /// `True` if the player is in an active state, `false` otherwise.
+  /// - Important: When the player is not in an active state the operations on the player **must be** restricted. In particular
+  ///         player methods that access the mpv core **must not** be called when the player is in the `shuttingDown` or
+  ///         `shutDown` states. Accessing the mpv core when the player is in these states can trigger a crash.
+  var isActive: Bool { PlayerState.activeStates.contains(state) }
+
+  var isIdle: Bool { state == .idle }
+  var isPaused: Bool { state == .paused }
+  var isPlaying: Bool { state == .playing }
+  var isSeeking: Bool { state == .seeking}
```
uiryuu:
```suggestion
  var isSeeking: Bool { state == .seeking }
```

iina/PlaybackInfo.swift
```diff
@@ -34,12 +34,67 @@ class PlaybackInfo {
...
+
+  /// The state the `PlayerCore` is in.
+  /// - Note: A computed property is used to prevent inappropriate state changes. When IINA terminates players that are actively
+  ///     playing will first be stopped and then shutdown. Once a player has stopped the mpv core will go idle. This happens
+  ///     asynchronously and could occur after the quit command has been sent to mpv. Thus we must be sure the state does not
+  ///     transition from `.shuttingDown` to `.idle`.
+  var state: PlayerState {
+    get { _state }
+    set {
```
uiryuu: I suggest get rid of two state variables and merge the `set` logic here into `didSet`. `didSet` should be called directly after the value has been set and on the same thread, so there shouldn't be any thread issue. We can check if the change of state is valid, and revert back to `oldValue` if not valid.

Generally reverting back to the original valid in `didSet` is legal as per [Apple's documentation](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/properties/):

> Similarly, if you implement a didSet observer, it’s passed a constant parameter containing the old property value. You can name the parameter or use the default parameter name of oldValue. **If you assign a value to a property within its own didSet observer, the new value that you assign replaces the one that was just set.**

iina/PlayerState.swift
```diff
@@ -0,0 +1,105 @@
...
+//  iina
+//
+//  Created by low-batt on 7/13/24.
+//  Copyright © 2024 lhc. All rights reserved.
+//
+
+import Foundation
+
+/// The mutually exclusive states a `PlayerCore` can be in.
+struct PlayerState: OptionSet, CustomStringConvertible {
```
uiryuu: Why `PlayerState` is implemented as an `OptionSet`? If I didn't miss anything, all the cases are mutually exclusive and we didn't use `PlayerState.insert()`. So a plain enum should do, no?

iina/AppDelegate.swift
```diff
@@ -646,7 +646,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
 
     // Instruct any players that are already stopped to start shutting down.
     for player in PlayerCore.playerCores {
-      if player.isStopped && !player.isShutdown {
+      if player.info.state == .stopped || player.info.state == .idle {
```
uiryuu: Here `player.info.state == .idle` is used; in other places `player.isIdle` is used. We should use only one of them in all places. 

For me, I don't see a particular advantage of using the computed properties (other than it will be a little bit shorter). And why we only have `isIdle, isPaused, isPlaying, isSeeking`, what about other states? So I suggest remove all the computed properties, and only use `player.info.state == __`. This style will remind us and new programmers explicitly that we are using `PlayerInfo.State` other than some unknown logic to compute the current state.

uiryuu (changes requested):

iina/AppDelegate.swift
low-batt: Will remove the computed properties and replace with direct checking of the state.

iina/PlaybackInfo.swift
low-batt: Good catch. As per the other comment these will be removed.

iina/PlaybackInfo.swift
low-batt: Will remove the `_state` variable and just use `didSet`.

iina/PlayerState.swift
low-batt: As per offline discussion I will add comment to explain why an `OptionSet` is being used. Then we will see what other reviewers think it is worthwhile using an `OptionSet` for performance.

iina/PlayerState.swift
low-batt: Will change the implementation to use a dictionary.

iina/AppDelegate.swift
svobs: > So I suggest remove all the computed properties, and only use player.info.state == __ 

In many places we need to check against a range of states. To ensure clarity, how about adding to `PlayerState`:

```Swift
  func isAtLeast(_ minState: PlayerState) -> Bool {
    return rawValue >= minState.rawValue
  }

  func isNotYet(_ state: PlayerState) -> Bool {
    return rawValue < state.rawValue
  }
```
So in many places, you'd have e.g., `guard player.info.state.isNotYet(.stopping) else { return }`.

svobs: Responding to myself after reading more of the existing PR code...
> So in many places, you'd have e.g., `guard player.info.state.isNotYet(.stopping) else { return }`.

I found that in practice, the `[.playing, .paused, .seeking]` states are sort of off on their own in relation to the other states.

But writing `player.info.state.isAtLeast(.playing)` doesn't seem very intuitive.

But you don't have to minimize the number of states! You could add another state above `.playing`, called `.started`. That way you could write `player.info.state.isAtLeast(.started)`. But you'd never even have to set that one explicitly. Or you could use it to indicate that some kind of init is done.

low-batt: I like this idea. I will see how it works out. Working on updating this PR today.

iina/PlayerState.swift
low-batt: Switched to an `enum`, so this code was eliminated.

iina/PlaybackInfo.swift
low-batt: These computed properties have been removed.

iina/PlayerState.swift
low-batt: Switched to an `enum`.

iina/AppDelegate.swift
low-batt: I liked the idea of these general methods, but I found there were only two cases where checking for multiple states, so instead I added `active` and `loaded` computed properties in the `enum`. They are checking the `rawValue` as suggested.

svobs: > I liked the idea of these general methods, but I found there were only two cases where checking for multiple states, so instead I added active and loaded computed properties in the enum.

Sounds fine to me. But you should name them `isActive` & `isLoaded` which is the naming convention for booleans in most projects. You probably were following `mainWindow.loaded`, but that's a special case...it looks it was created as a substitute for `isLoaded` which already exists but has unwanted side effects.

low-batt: I first coded them as `isActive` and `isLoaded` because I too am used to that convention. I renamed them because I didn't see that convention consistently used. Happy to change them back. Lets first see what @uiryuu has to say about IINA conventions.

iina/PlayerCore.swift
```diff
@@ -1794,15 +1802,32 @@ class PlayerCore: NSObject {
...
     postNotification(.iinaFileLoaded)
     events.emit(.fileLoaded, data: info.currentURL?.absoluteString ?? "")
+    // mpvResume()
+    if !(info.justOpenedFile && Preference.bool(for: .pauseWhenOpen)) {
+      mpv.setFlag(MPVOption.PlaybackControl.pause, false, level: .verbose)
+    }
+    syncUI(.playlist)
+  }
+
+  func fileEnded(doToStopCommand: Bool) {
```
uiryuu: Did you mean `dueToStopCommand`?

uiryuu (commented):
The changes look much cleaner now.

iina/PlayerCore.swift
low-batt: Good catch! Fixed.

iina/AppDelegate.swift
low-batt: I had a chat with @uiryuu about boolean flag naming conventions. The "is" prefix convention is not preferred. The flags you see in IINA that convention were likely added by me. I was used to that convention from other projects.

uiryuu (approved):

iina/MPVController.swift
```diff
@@ -1109,12 +1105,10 @@ class MPVController: NSObject {
 
     case MPV_EVENT_PLAYBACK_RESTART:
       DispatchQueue.main.async { [self] in
-        player.info.isIdle = false
-        player.info.isSeeking = false
         // When playback is paused the display link may be shutdown in order to not waste energy.
```
lhc70000, line 1109: No state update after seeking? This doesn't seem to be correct.

low-batt: Good catch. The `isSeeking` flag needs to be kept separate. It can't be included in state enum. I've pushed an update.

PR #5086, opened 2024-07-23
Correct ViewLayer shadow copy initializer

low-batt:
This commit will:
- Change the `ViewLayer.init(layer:)` initializer to invoke the corresponding superclass implementation
- Add copying of additional properties to the shadow layer

This addresses concerns raised in discussion #5046.

---

**Description:**

### DIFF
--- a/iina/ViewLayer.swift
+++ b/iina/ViewLayer.swift
@@ -21,19 +21,22 @@ class ViewLayer: CAOpenGLLayer {
   private let cglPixelFormat: CGLPixelFormatObj
 
   /// Lock to single thread calls to `display`.
-  ///
-  /// A recursive lock is needed because the call to `CATransaction.flush()` in `display` calls `display_if_needed`
-  /// which will then call `display` if layout is needed. See the discussion in PR #5029.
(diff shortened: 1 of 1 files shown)


low-batt:
To cause the `ViewLayer.init(layer:)` initializer to be used I followed the method discussed by @krackers and:
- Connected my MacBook Pro in closed clamshell mode to my Apple/LG 5K monitor
- Changed the scale of the LG monitor to be `More Space`
- Started IINA playing a video
- Unplugged the LG monitor
- At that point the shadow layer was created

I was not able to get IINA to call the `display` or `draw` methods on the shadow layer. So these changes have not been tested as much as I would like.

krackers:
Changes look good to me.

My speculation is that perhaps the only reason the shadow layer is only created to have a copy of the backing IOSurface, which is why `draw` is never called. Or maybe the property animation codepaths end up being used even when the modified property isn't actually animated (thus the shadow layer is created, but not actually used). CoreAnimation has so many mysteries...

uiryuu (approved):

PR #5108, opened 2024-08-01
Fix app can quit while playback history is being saved, #5107

### LINKED ISSUE #5107: App can quit while playback history is being saved
**System and IINA version:**

- macOS 13.6.3
- IINA `develop` @ ae5d2e3d1cd6d063793efb92812800769a97132e

**Expected behavior:**
Playback history is reliable.

**Actual behavior:**
Under heavy load IINA may terminate while playback history is being saved to the history plist file. Very rarely this can corrupt the file resulting in this error the next time IINA starts:
```text
22:00:16.449 [iina][e] Failed to read playback history file /Users/low-batt/Library/Application Support/com.colliderli.iina/history.plist: Error Domain=NSCocoaErrorDomain Code=4864 "*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is empty; did you forget to send -finishEncoding to the NSKeyedArchiver?" UserInfo={NSDebugDescription=*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: data is empty; did you forget to send -finishEncoding to the NSKeyedArchiver?}
```

**Steps to reproduce:**
This is very difficult to reproduce. Run a stress test that opens many files at once that use software decoding such that the Mac is overloaded and cause IINA to quit. Repeat for a long time.

This is an IINA feature.

Very rarely.

low-batt:
This commit will:
- Change the call to `write` in the `HistoryController.save` method to use the `atomic` option
- Add a` tasksOutstanding` property to `HistoryController` to keep track of how many tasks are in the background queue
- Change the `HistoryController.history` property to be atomic
- Change the methods in `HistoryController`, `HistoryWindowController` and `JavascriptAPICore` to lock the `history` property when accessing it
- Add a `iinaHistoryTaskFinished` notification to `AppData`
- Change `HistoryController.add` to post this notification when a background task finishes
- Change `AppDelegate.applicationShouldTerminate` to establish an observer for the `iinaHistoryTaskFinished` notification and wait until all outstanding `HistoryController` tasks have finished before proceeding with termination
- Add logging to the `HistoryController`
- Remove unused `removeAll` method from `HistoryController`

- [x] This implements/fixes issue #5107.

---

**Description:**

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -457,9 +457,8 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
 
     // Check if there are any players that are not shutdown. If all players are already shutdown
     // then application termination can proceed immediately. This will happen if there is only one
-    // player and shutdown was initiated by typing "q" in the player window. That sends a quit
-    // command directly to mpv causing mpv and the player to shutdown before application
-    // termination is initiated.
+    // player and shutdown was initiated by sending a quit command directly to mpv through it's IPC
+    // interface causing mpv and the player to shutdown before application termination is initiated.
     allPlayersHaveShutdown = true
     for player in PlayerCore.playerCores {
       if player.info.state != .shutDown {
@@ -482,6 +481,11 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
     if OnlineSubtitle.loggedIn {
       canTerminateNow = false
       Logger.log("Waiting for log out of online subtitles provider to complete")
+    }
+
+    if HistoryController.shared.tasksOutstanding != 0 {
+      canTerminateNow = false
+      Logger.log("Waiting for saving of playback history to complete")
     }
 
     // If the user pressed Q and mpv initiated the termination then players will already be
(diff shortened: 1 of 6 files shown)


iina/AppDelegate.swift
```diff
@@ -573,13 +577,16 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
         // All players have shutdown.
         Logger.log("All players have shutdown")
       }
-      guard !OnlineSubtitle.loggedIn else { return }
-      // All players have shutdown. No longer logged into an online subtitles provider.
+      // If still logged in to subtitle providers or still saving playback history then continue
+      // waiting.
+      guard !OnlineSubtitle.loggedIn, HistoryController.shared.tasksOutstanding == 0 else { return }
```
lhc70000: I suggest using two separate guards.

lhc70000 (commented):
Looks good to me.

I was thinking whether an `OperationQueue` will be more suitable for this case, but all other checks in `proceedWithTermination()` use the notification based mechanism. So the current implementation is fine.

iina/AppDelegate.swift
low-batt: Fixed. I initially had it as two guards as I thought that was easier to read with separate comments. I decided to combine it into one guard right before posting as a lot of the code uses guards with multiple expressions. In this case with the separate guards it is quicker to read and understand.

low-batt:
I too considered `OperationQueue` as well as [DispatchWorkItem](https://developer.apple.com/documentation/dispatch/dispatchworkitem) but I didn't see an advantage to using them over the current event driven system.

Issue #3158, opened 2020-10-15
Use SF Symbols For Consistency 


Unichy:
**What you want IINA to do:**
Replace its own symbols with SF Symbols.

**What IINA does currently:**
INNA recently updated its symbols, but I prefer if it used the SF Symbols like the rest of MacOS. 

**Why you think this should be added:**
MacOS Big Sur is getting Sf Symbols system wide. Every Apple app and a growing list of third party apps already use SF Symbols. Using SF Symbols would make users more familiar to the interface.

**Examples of other projects that have something similar:**

lhc70000:
SF Symbols are used widely in v1.1.0 in many places, such as the sidebar tabs, playlist loop/shuffle button, and the volume indicator. In future versions, I may also expand its usage to the preferences panel.

However, for small buttons in the on-screen controller, SF Symbols look blurry and the readability is very poor especially on non-retina screens, so I had to make original pixel-perfect icons for them. Additionally, most of these icons are not even covered by SF Symbols.

low-batt:
SF Symbols are being used as of IINA 1.4.0-beta1.

PR #5125, opened 2024-08-11
Update Settings' tab icons & toolbar icons using SF symbols

uiryuu:

---

**Description:**
Now that the plugin system has been enabled by default (f78497434211254b220389bc3de9526ff2ada060), we need a new icon for that. To take advantage of this opportunity, I chose some icons from SF symbols to replace the current ones, which are really old btw. Icons are not available at runtime are simply fallback to the old ones.

before -> after
image image

I also updated the icons for toolbar items:
image

The paddings in this one needs to be adjusted later.

Feel free to propose better ones

### DIFF
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -535,6 +535,23 @@ extension NSImage {
     newImage.unlockFocus()
     return newImage
   }
+
+  /// Try to find a SF Symbol. If the symbol is found, then return the symbol with the configuration applied; otherwise,
+  /// return the fallback image.
+  ///
+  /// - Parameters:
+  ///   - name: The name of the SF Symbol
+  ///   - configuration: The symbol configuration for the SF symbol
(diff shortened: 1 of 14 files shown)


low-batt:
Two quick thoughts…

Issue #3158 has [this response](https://github.com/iina/iina/issues/3158#issuecomment-709572165) from @lhc70000 regarding use of SF Symbols for the OSC:
> …for small buttons in the on-screen controller, SF Symbols look blurry and the readability is very poor especially on non-retina screens, so I had to make original pixel-perfect icons for them

Does this problem with small SF Symbols still exist?

Instead of a function in `Utility` is it possible to extend `NSImage` with a new initializer that takes a `fallbackName` parameter?

uiryuu:
I'm actually using a 100% screen, and I don't see a readability issue here.

image

low-batt (approved):
I pulled latest commits and tested under macOS 10.15 to confirm fallback works. Looks good to me.

low-batt:

PR #5129, opened 2024-08-11
Fix some deprecated issues

siieo:
- [T] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)
- [X] This implements/fixes issue #.

---

**Description:**

In reviewing the code, I identified several deprecated APIs that need to be updated to ensure compatibility with the latest macOS versions. To address these issues, I have replaced the deprecated methods with their modern counterparts, aligning with current best practices and API recommendations.

----

Build works fine with Xcode 15.4

### DIFF
--- a/iina/LegacyMigration.swift
+++ b/iina/LegacyMigration.swift
@@ -61,10 +61,10 @@ class LegacyMigration {
 
       // Deserialize & convert legacy pref value to modern string format:
       guard let legacyData = Preference.data(for: legacyKey) else { continue }
-      guard let color = NSUnarchiver.unarchiveObject(with: legacyData) as? NSColor,
+      guard let color = try? NSKeyedUnarchiver.unarchivedObject(ofClass: NSColor.self, from: legacyData),
             let mpvColorString = color.usingColorSpace(.deviceRGB)?.mpvColorString else {
-        Logger.log("Failed to convert color value from legacy pref \(legacyKey.rawValue)", level: .error)
-        continue
+          Logger.log("Failed to convert color value from legacy pref \(legacyKey.rawValue)", level: .error)
+          continue
       }
       // Store string under modern pref key:
       Preference.set(mpvColorString, for: modernKey)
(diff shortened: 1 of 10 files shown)


low-batt:
@siieo Thank you very much for investigating the deprecated warnings and proposing fixes. This definitely needs to be addressed.

From a quick look I see two problems. The continuous integration test failed. I expect you have a Mac with an Apple Silicon chip. For changes sensitive to the version of macOS the both Apple and Intel and builds must be checked. To build a universal binary in Xcode change the build destination to be "Any Mac" as shown in the red box in the screenshot below. As you can see after I changed that setting and executed a build, errors were reported:

errors

The Intel build supports macOS 10.15+. Anything introduced in macOS 11 must use an available test. This is what is causing the CI test to fail.

I will post about the other problem in a minute.

uiryuu:
The gif part seems not being used for a long time, no? I remember there is another PR trying to delete that part completely.

iina/LegacyMigration.swift
```diff
@@ -61,10 +61,10 @@ class LegacyMigration {
 
       // Deserialize & convert legacy pref value to modern string format:
       guard let legacyData = Preference.data(for: legacyKey) else { continue }
-      guard let color = NSUnarchiver.unarchiveObject(with: legacyData) as? NSColor,
+      guard let color = try? NSKeyedUnarchiver.unarchivedObject(ofClass: NSColor.self, from: legacyData),
```
low-batt, line 64: This will not work. The archive in question is _not_ keyed, so `NSKeyedUnarchiver` will not be happy. The `LegacyMigration` class is about migrating an old archive stored in user defaults to a new format. So there is no choice here, the code must use the old deprecated `NSUnarchiver` to read the old archive during the migration. In better languages we'd be able to add a pragma to suppress the warning with a comment was to the deprecated method is required. But last I heard the proposals to add such a feature to Swift were turned down.

For the other changes I will have to defer to other IINA developers who know more about UI code than I do.

low-batt (changes requested):
The change to LegacyMigration must be removed. See the comment on that change. And of course the build failures must be addressed.

siieo:
I will close this pull and will add another one

PR #5131, opened 2024-08-12
Add an option to enable repeat for a new window, fix #2139

### LINKED ISSUE #2139: How to make Repeat mode enabled by default?

uiryuu:
- [x] This implements/fixes issue #2139.

---

**Description:**
This is a remake of #4761. The "repeat playlist" option is the default.

image

### DIFF
--- a/iina/Base.lproj/PrefGeneralViewController.xib
+++ b/iina/Base.lproj/PrefGeneralViewController.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="21701" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="23090" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
     <dependencies>
         <deployment identifier="macosx"/>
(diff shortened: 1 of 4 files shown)


iina/Base.lproj/PrefGeneralViewController.xib
```diff
@@ -566,17 +566,51 @@
                         <binding destination="IBP-Mz-Maa" name="value" keyPath="values.playlistShowMetadataInMusicMode" id="nxW-hX-9co"/>
                     </connections>
                 </button>
+                <button verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="H91-rO-GsT">
+                    <rect key="frame" x="118" y="7" width="269" height="18"/>
+                    <buttonCell key="cell" type="check" title="Repeat when media is opened manually:" bezelStyle="regularSquare" imagePosition="left" state="on" inset="2" id="jok-kc-HjX">
```
low-batt: Under the `History` section it says: `Otherwise only files that opened manually…`
To be consistent with terms `media` should be `file`.
The wording is a bit odd because `Repeat` is used twice. And `Repeat` is not used in the OSD or the menus. To be consistent `Loop` should be used. Maybe something like:
`When a file is opened manually:` `Loop playlist`

Also, the new text must be added to the English translation file.

low-batt (changes requested):
Missing English translation file. Also see comment about wording.

Otherwise worked in my testing.

iina/Base.lproj/PrefGeneralViewController.xib
uiryuu: Sorry I kept forgetting adding the strings

uiryuu:
Now:
image

low-batt (approved):
Pulled latest commit and tested. Changes look good to me.

Issue #5064, opened 2024-07-11
Mute has 2-second lag


svobs:
**System and IINA version:**

- macOS 14.5
- IINA post-1.3.5 `develop` branch, 0ababb4e2c91bbc95909b36d434d1c46755f6584

**Expected behavior:**
Toggling `mute` should take effect immediately.

**Actual behavior:**
Toggling `mute` does not take effect until 2 seconds have passed (for both mute and unmute)

**Steps to reproduce:**
1. Open a video with the latest develop version of IINA. Wait until it starts playing.
2. Use the `Audio` > `Mute` menu item or key equivalent. Wait 2 seconds for mute to take effect.
3. Use the same command to test unmute.

- [x] MPV 0.38.0 does not have this problem.
- [x] IINA 1.3.5 does not have this problem.

**How often does this happen?**
Always

low-batt:
Reproduced for me with mpv 0.38.0. Try adding `--ao=avfoundation`.

low-batt:
To workaround this follow these instructions:
- Start IINA
- Click on  `Settings…` under the `IINA` menu
- The settings panel appears
- On the left side of the panel click on `Advanced`
- Slide the `Enable advanced settings` toggle button to be on (blue)
- In the `Additional mpv options` section click on `+`
- A new entry appears in the table
- Double click on `name` and replace it with `ao`
- Double click on `value` and replace it with `coreaudio`
- **Restart IINA** to activate the new settings

With a quick search I did not find a mpv issue reporting this problem. I recently noticed another `avfoundation` mpv issue that might also be a problem for IINA. I think that had to do with audio synchronization if playback speed is changed.

I'm probably too busy to get to it tomorrow, but I will report this to mpv.

This may mean we will need to allow the user to choose between [Core Audio](https://developer.apple.com/documentation/coreaudio) and [AVFoundation](https://developer.apple.com/documentation/avfoundation).

svobs:
Thanks for your quick diagnosis. I will have to do some research to investigate the relative pros & cons of using Core Audio vs. AVFoundation...

uiryuu:
Actually I also noticed the audio delay when adjusting the EQ. The EQ seems always have a delay until it takes effect.

low-batt:
[...]

PR #5137, opened 2024-08-14
Add a setting to let user to choose coreaudio or avfoundation as the audio driver

uiryuu:
- [x] This implements/fixes issue #5064.

---

**Description:**
Added descriptive text to indicate AVFoundation is an experimental audio driver which may cause issues.

image

### DIFF
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -513,11 +513,10 @@ class MPVController: NSObject {
     chkErr(setOptionString(MPVOption.ProgramBehavior.resetOnNextFile,
             "\(MPVOption.PlaybackControl.abLoopA),\(MPVOption.PlaybackControl.abLoopB)", level: .verbose))
 
-    // As mpv support for audio using the AVFoundation framework is new we enable it before applying
-    // user's settings. This allows a user to roll back to the Core Audio framework should a problem
-    // be encountered with the new code. Set audio driver based on user settings.
-    let driver = Preference.bool(for: .audioDriverEnableAVFoundation) ? "avfoundation" : "coreaudio"
-    chkErr(setOptionString(MPVOption.Audio.ao, driver, level: .verbose))
+    setUserOption(PK.audioDriverEnableAVFoundation, type: .other, forName: MPVOption.Audio.ao,
+                  level: .verbose) { key in
+      Preference.bool(for: key) ? "avfoundation" : "coreaudio"
+    }
 
     // Set user defined conf dir.
     if Preference.bool(for: .enableAdvancedSettings),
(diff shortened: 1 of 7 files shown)


iina/MPVController.swift
```diff
@@ -517,8 +517,9 @@ class MPVController: NSObject {
 
     // As mpv support for audio using the AVFoundation framework is new we enable it before applying
     // user's settings. This allows a user to roll back to the Core Audio framework should a problem
-    // be encountered with the new code.
-    chkErr(setOptionString(MPVOption.Audio.ao, "avfoundation", level: .verbose))
+    // be encountered with the new code. Set audio driver based on user settings.
+    let driver = Preference.bool(for: .audioDriverEnableAVFoundation) ? "avfoundation" : "coreaudio"
+    chkErr(setOptionString(MPVOption.Audio.ao, driver, level: .verbose))
```
low-batt:
```suggestion
    setUserOption(PK.audioDriverEnableAVFoundation, type: .other, forName: MPVOption.Audio.ao,
                  level: .verbose) { key in
      Preference.bool(for: key) ? "avfoundation" : "coreaudio"
    }
```

iina/MPVController.swift
```diff
@@ -517,8 +517,9 @@ class MPVController: NSObject {
 
     // As mpv support for audio using the AVFoundation framework is new we enable it before applying
     // user's settings. This allows a user to roll back to the Core Audio framework should a problem
```
low-batt: This comment was about the user being able to add "ao=coreaudio" in advanced settings to rollback from AVFoundation to Core Audio. So the comment is a bit out of date now that we have an explicit setting for this.

low-batt (commented):
Changes look good. With the suggested code change the audio driver can be dynamically changed. There is a couple of seconds delay and then mpv starts the audio again using the selected driver. This might be useful to users who like Spatial Audio, but don't like the mute delay bug with AVFoundation. So they might switch to AVFoundation only when playing a video with Spatial Audio. I'm thinking make this dynamic, just because I think apps should always avoid the requirement to restart after changing a setting, if possible. If we do make it dynamic we should update the Inspector window to understand the driver can change. Thoughts?

low-batt:
Noticed one other thing. The help button is way over to the right edge whereas other help buttons are close to the setting.

uiryuu:
> The help button is way over to the right edge whereas other help buttons are close to the setting.

I intentionally did this, because I also added a "flask" icon to indicate AVFoundation is an experimental option. I moved the question mark away to separate those 2 icons.

low-batt:

PR #5152, opened 2024-08-20
Add a loading window when loading network resources

uiryuu:

---

**Description:**
After #5054, the main window will show after the video is loaded by mpv and the video size is available from mpv. Therefore when the user is loading a video from Internet through yt-dlp, nothing will show up until the video is loaded, which will take a considerate amount of time. This PR adds a loading window (which reuses the `OpenURLWindow`) when opening a URL.

Changes in this PR:
- Added an overlay view in the `OpenURLWindow` which works as a loading screen for when loading internet resources
- Added autosave name for `OpenURLWindow`
- When user closes the `OpenURLWindow` when loading a URL, stop the player core
- Fixed an issue where the "Unable to open URL" error dialog doesn't popup
  - By changing `info.state == .loading` to `info.state == .starting` when deciding whether or not mpv couldn't load the URL
  - @low-batt please confirm whether or not this is the correct fix
- Improved error handling: if mpv couldn't load the URL and `OpenURLWindow` is open with the loading screen, `OpenURLWindow` will change back to normal mode with error message, so that the user can input another URL and try again. See the video below:

https://github.com/user-attachments/assets/b2747bc3-c3cd-4d15-a341-ad4cb3ecda3b

https://github.com/user-attachments/assets/40df5b5b-d42f-4aea-b1cc-449a621c6720

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -896,8 +896,8 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
   @IBAction func openURL(_ sender: AnyObject) {
     Logger.log("Menu - Open URL")
     openURLWindow.isAlternativeAction = sender.tag == AlternativeMenuItemTag
+    openURLWindow.resetWindowState()
     openURLWindow.showWindow(nil)
-    openURLWindow.resetFields()
   }
 
   @IBAction func menuNewWindow(_ sender: Any) {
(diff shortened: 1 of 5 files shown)


svobs:
Looks nice! Definitely a good idea.

One request: could there be a "Stop" or "Cancel" button once it starts to load? In case it gets stuck on a bad connection, that would give an impatient user a way out. Would be ideal if it could just show the form again without clearing any of the fields which were filled out.

uiryuu:
> One request: could there be a "Stop" or "Cancel" button once it starts to load? In case it gets stuck on a bad connection, that would give an impatient user a way out. Would be ideal if it could just show the form again without clearing any of the fields which were filled out.

Will add this in a follow up PR.

uiryuu:
@svobs https://github.com/iina/iina/commit/d14f573c340f122429cefc5de1e150325f6027c3 added a button to stop loading.

low-batt:

PR #5175, opened 2024-09-16
Improvements to speed UI (fixes #5147, #5168)

### LINKED ISSUE #5147: Change playback speed delimiter from "," to "." (x1.5)
**What you want IINA to do:**

I want IINA to use speed markings in values such as `1.2` and `1.5` (as opposite to the current `1,2` and `1,5`)

**What IINA does currently:**

The UI uses `.` for the marking beneath the settings (see 0.25x)

2024-08-18 at 01 47 44

I believe it used to be default behaviour of IINA, until on some version I began getting the error bleep when I wrote the more internationally accepted `.` delimiter within a value. I only recently understood that it wasn't a bug, but a behaviour, and that I had to write `1,2` instead of `1.2`. It is my understanding that Americans favor the comma, but just about every app out there that allows changing the speed shows the value using a period (YouTube app, Podcasts on iOS, etc).

### LINKED ISSUE #5168: Change playback speed up to 5 decimal places
Is there any way to change the playback speed to 5 decimal places. I am trying to sync a 23.976 fps video to a 25 fps video but i can only alter the speed to 3 decimal places.

svobs:
- [x] This implements/fixes issue #5168.
- [x] Also fixes issue #5147.
- Add `speedResetBtn` which will reset speed to 1x
- Update `osd.speed` message; and `menu.speed`, `menu.speed_up`, menu.speed_down` menu item titles to support unlimited digits after decimal and also properly localize numeric values.
- Update slider labels at load using properly localized numeric values for consistency (i.e. [issue 5147](https://github.com/iina/iina/issues/5147)).
- Update `customSpeedTextField`, `speedSliderIndicator` to support up to 6 digits after decimal (will round to 6th decimal place, rounding down if 7th decimal place is a `5`, matching mpv's behavior).
- Increase width of `customSpeedTextField` to fit at least 6 total digits
- Update speed slider action to round to 2 digits after decimal.
- Fix knob clipping issue with video speed slider & audio delay slider in MacOS Ventura & earlier.

---

**Description:**
This attempts to address concerns mentioned in #5168, while striking a good balance between usability, versatility, and technical constraints.

_~~It's not clear how many decimal places mpv will support. It looks like it will allow whatever will fit in a `double`. But it's doubtful that mpv has thoroughly tested very very precise speeds.~~_
[EDIT] It has been determined that mpv supports 6 digits after the decimal and has a minimum value of 0.01, so this is now enforced when changing the speed.

This PR ensures that the OSD message will display the exact speed mpv is using. _~~It looks like text fields want a maximum number of decimal places, so the `customSpeedTextField` in the Video Settings sidebar is now 9 decimals instead of 3. This also means that the `speedSliderIndicator` label above the slider will also be rounded to 9 decimals because it uses the number from `customSpeedTextField`, but they do not limit the actual value used by mpv. So key bindings can be used to set speed with any precision.~~_

The width of `speedSlider` had to be decreased slightly to expand the width of `customSpeedTextField`, which now can comfortably fit 5 digits after the decimal.

I also modified the speed slider's action to round the selected speed to 2 digits after the decimal. This should be more usable because having more precision when dragging the slider wouldn't seem to make much sense and would only make the speed value harder to read.

This also expanded to fix number formatting for everything speed-related, so that locales which use commas instead of decimals are supported. There is more work to do for audio delay & subtitle controls, however.

Localized slider labels

Localized menu items

### DIFF
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -230,9 +230,19 @@ extension Double {
 
   func roundedTo2Decimals() -> Double {
     let scaledUp = self * 1e2
-    let scaledUpRounded = scaledUp.rounded(.toNearestOrEven)
+    let scaledUpRounded = scaledUp.rounded(.up)
     let finalVal = scaledUpRounded / 1e2
     return finalVal
+  }
+  
+  /// Formats this number as a decimal string, using default locale.
+  ///
+  /// This should be used in most places where decimal numbers need to be printed.
+  var string: String {
+    /// Output up to 15 digits after the decimal. This should be good enough for almost all cases, but can be increased in the future.
+    /// (It's not clear what the maximum allowable value for `NumberFormatter.maximumFractionDigits` actually is. Trying to use
+    /// `NSIntegerMax` seems to result in `maximumFractionDigits` being silently set to `6` instead.)
+    return fmtDecimalMaxFractionDigits15.string(from: self as NSNumber) ?? "NaN"
   }
 }
 
@@ -254,6 +264,17 @@ fileprivate let fmtDecimalMaxFractionDigits2: NumberFormatter = {
   fmt.numberStyle = .decimal
   fmt.usesGroupingSeparator = false
   fmt.maximumFractionDigits = 2
+  return fmt
+}()
+
+fileprivate let fmtDecimalMaxFractionDigits15: NumberFormatter = {
+  let fmt = NumberFormatter()
+  fmt.numberStyle = .decimal
+  fmt.usesGroupingSeparator = true
+  fmt.maximumSignificantDigits = 15
+  fmt.minimumFractionDigits = 0
+  fmt.maximumFractionDigits = 15
+  fmt.usesSignificantDigits = false
   return fmt
 }()
 
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -235,13 +235,25 @@ extension Double {
     return finalVal
   }
   
-  /// Formats this number as a decimal string, using default locale.
+  /// Formats this number as a decimal string, using the default locale.
   ///
-  /// This should be used in most places where decimal numbers need to be printed.
+  /// This should be used in most places where decimal numbers need to be printed. Do not rely on string interpolation alone
+  /// because the number will not be localized.
+  ///
+  /// For example, if the user's locale formats numbers like `1.234.567,89` (in particular, using
+  /// a comma to signify the decimal):
+  /// ```
+  /// let num: Double = 12.34
+  /// let badStr = "Value is \(num)"          // badStr will *always* be "Value is 12.34"
+  /// let goodStr = "Value is \(num.string)"  // goodStr will be "Value is 12,34"
+  /// ```
+  ///
+  /// Currently the output string is limited to 15 digits after the decimal. This should be more than
+  /// enough for any imaginable use right now, but the limit can and should be increased in the future if
+  /// needed. (It's not clear what the maximum allowed value for `NumberFormatter.maximumFractionDigits`
+  /// actually is. An attempt to set it equal to `NSIntegerMax` seemed to result in it being silently set to
+  /// `6` instead.)
   var string: String {
-    /// Output up to 15 digits after the decimal. This should be good enough for almost all cases, but can be increased in the future.
-    /// (It's not clear what the maximum allowable value for `NumberFormatter.maximumFractionDigits` actually is. Trying to use
-    /// `NSIntegerMax` seems to result in `maximumFractionDigits` being silently set to `6` instead.)
     return fmtDecimalMaxFractionDigits15.string(from: self as NSNumber) ?? "NaN"
   }
 }
@@ -271,7 +283,7 @@ fileprivate let fmtDecimalMaxFractionDigits15: NumberFormatter = {
   let fmt = NumberFormatter()
   fmt.numberStyle = .decimal
   fmt.usesGroupingSeparator = true
-  fmt.maximumSignificantDigits = 15
+  fmt.maximumSignificantDigits = 25
   fmt.minimumFractionDigits = 0
   fmt.maximumFractionDigits = 15
   fmt.usesSignificantDigits = false
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -227,6 +227,13 @@ extension Double {
       return "\(rounded)"
     }
   }
+
+  func roundedTo2Decimals() -> Double {
+    let scaledUp = self * 1e2
+    let scaledUpRounded = scaledUp.rounded(.toNearestOrEven)
+    let finalVal = scaledUpRounded / 1e2
+    return finalVal
+  }
 }
 
 extension Comparable {
--- a/iina/QuickSettingViewController.swift
+++ b/iina/QuickSettingViewController.swift
@@ -15,6 +15,22 @@ fileprivate let eqRenameMenuItemTag = -2
 fileprivate let eqSaveMenuItemTag = -3
 fileprivate let eqCustomMenuItemTag = 1000
 
+/// Formatter for `customSpeedTextField`.
+///
+/// Configure the number formatter in code instead of the XIB so it is easier to follow.
+fileprivate let speedFormatter: NumberFormatter = {
+  let fmt = NumberFormatter()
+  fmt.numberStyle = .decimal
+  fmt.usesGroupingSeparator = true
+  fmt.maximumSignificantDigits = 25  // just make very big
+  fmt.minimumFractionDigits = 0
+  fmt.maximumFractionDigits = 6  // matches mpv behavior
+  fmt.usesSignificantDigits = false
+  fmt.roundingMode = .halfDown   // matches mpv behavior
+  fmt.minimum = NSNumber(floatLiteral: AppData.mpvMinPlaybackSpeed)
+  return fmt
+}()
+
 class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTableViewDelegate, SidebarViewController {
   override var nibName: NSNib.Name {
     return NSNib.Name("QuickSettingViewController")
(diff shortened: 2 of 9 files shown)


low-batt:
Thanks for picking this one up! Will have to see what the other developers think about some of the UI details of this…

I think mpv is truncating `speed` to 6 digits. When I obtain it directly from mpv through the IPC interface I get `1.042709`. 

The `Playback` menu only shows a couple of digits. There is room for more. Should that be changed as well?

I like the reset button. Should that have a tool tip?

For some reason when I set the language to `he` the OSD messages all show `-0.00` for the speed. I checked `develop` and this didn't happen.

The thumb of the slider is slightly cut off when it is at the end of the slider. This is an existing problem. Happens with `develop` as well.

OSD

svobs:
> Thanks for picking this one up! Will have to see what the other developers think about some of the UI details of this…

Sure. I am not super happy about the text field having to scroll. But it should be uncommon for most cases... I suppose I could move the text field onto another row, which would allow the slider to fill the whole width, which would be nice. But it would look awkward aesthetically if the text field is all alone.

Also wanted to highlight that I moved the number formatting out of the localized strings for `osd.speed`, and replaced it with just a string `%@`. Seems to me like most localized strings should be formatted like this (or at least the OSD strings) so that changing the number's format doesn't require string changes for every locale (and the number formatter should be localizing the numbers anyway). Unfortunately it does mean that all the localized strings should be migrated to use `%@`, which is something to do after this PR is merged.

> I think mpv is truncating speed to 6 digits. When I obtain it directly from mpv through the IPC interface I get 1.042709.

When I enter `1.23456789` as the custom speed, I do see in the mpv log:
```
[mpv0][d] Set property: speed=1.23456789
```
And then `MPVController` is sent a `MPVOption.PlaybackControl.speed` message which does indeed contain that exact number.

But it is helpful to know that the IPC interface stops at 6 digits. Maybe that's all anyone has ever needed? Maybe `customSpeedTextField` should stop at 6 digits as well?

Another issue. I also just now discovered there are problems setting very low speeds.

It looks like `0.01`x is the lowest speed which works properly. But trying a lower speed, for example `0.001`x, fails. I had to modify IINA's code to check the return value, but it returns `-9`, which seems to be `"unsupported format for accessing property"`.

Not sure why this particular error code might show up. And there are no problems when higher values are used (for example, `0.010002` is fine). Need to dig in to this more... But at the very least it needs better handling, because right now it will silently fail and the UI will be lying about the actual speed.

> The Playback menu only shows a couple of digits. There is room for more. Should that be changed as well?

Mmm yes, forgot about that. Will post an update.

> I like the reset button. Should that have a tool tip?

I thought it would be easy enough to guess, since there are buttons like it down below for the equalizers. A tooltip could be added though. Maybe just saying "Reset speed to 1x" or something?

> For some reason when I set the language to he the OSD messages all show -0.00 for the speed. I checked develop and this didn't happen.

Sounds like it might be the way I am doing the rounding. Will post an update to fix it. Would be great to learn more about Apple's "negative 0", "NaN", "infinity", etc and how it makes those & stores them.

> The thumb of the slider is slightly cut off when it is at the end of the slider. This is an existing problem. Happens with develop as well.

Hmm. My guess is that the slider thumb can be drawn slightly outside the slider's actual bounds. Because it looks similar to an issue which can happen for rounded buttons. Ugly, ugly AppKit... But it only does this for Hebrew? I will see if I can add a workaround.

svobs:
> The Playback menu only shows a couple of digits. There is room for more. Should that be changed as well?

I added a commit which changes the `menu.speed` indicator & various `menu.speed_up` & `menu.speed_down` speed menu items so they now support unlimited decimal places (at least whatever `player.info.playSpeed` is).

> For some reason when I set the language to he the OSD messages all show -0.00 for the speed. I checked develop and this didn't happen.

I dug in to this, and it's just due to a mismatch in formatting strings, a result of the `he` locale not yet being updated to use `%@` instead of `%.2f`. For some reason, passing in a `String` is allowed and is treated as a negative `Double`. Or something. Something which can be ignored for now.

> The thumb of the slider is slightly cut off when it is at the end of the slider. This is an existing problem. Happens with develop as well.

Unable to reproduce. Tried dark & light themes; Hebrew & English; Retina display & regular display; `Show scroll bars:`: `Always` & `When scrolling`. Using MacOS Sonoma 14.6.1 (23G93). [Not sure when a build ID started showing up after double-clicking the version in About This Mac!]

low-batt:
i will defer to @uiryuu on the change to how the numbers are localized.

When I paste `1.23456789` into the text field the mpv log shows this for me:
```text
[2797.829][v][cplayer] Set property: speed=1.234568 -> 1
```

Why is the log you showed missing `-> 1`?

The OSD message shows the full value. It is truncated to just `1.2345678` in the `Inspector` window when configured to display the `speed`. Same when pulled using the IPC interface. My guess on this is that the mpv property change event is sending IINA the value IINA sent mpv and mpv is truncating that value afterwards. I'm not liking that behavior. BUT I can't explain why your log message differs. I'm thinking IINA should stop at 6 digits, but it bothers me that I'm not fully understanding what is happening.

The documentation for [--speed](https://mpv.io/manual/stable/#options-speed) says:
> **--speed=<0.01-100>**

So the problem with values less than 0.01 is expected mpv behavior.

"Reset speed to 1x" is fine for the tool tip. As for whether the tool tip is needed or not, I rather like tool tips, but I don't know how the other developers would feel about one for this button. They may think it is obvious what it does and shouldn't need one. Lets see what they think.

The problem with the slider thumb being cut off is not specific to `he`. I see it with `en` as well and when using `default`. It is just the very edge of the thumb that is cut off: I am testing under Ventura 13.6.9. I will test on my Mac running Sonoma. Maybe AppKit has fixed AppKit. Tuesday busy for me, so might not get to that until tomorrow.

clip

The hidden macOS build number has been there for a long time. When I was adding build information to IINA's about window I copied that convention where build information was not shown until you clicked on it. But following the macOS convention got a thumbs down from the reviewers.

svobs:
> Why is the log you showed missing -> 1?

Good to point this out. My error. I was looking at the IINA log and assumed it was a message output by mpv itself. In my fork I have the mpv logging turned up and I forgot that it's only set to `w` here. And the msg looks almost identical and uses the same `mpvX` subsystem as the mpv logs.

OK I am seeing the same behavior as you now. It looks like it rounds to 6 digits (half down), not truncates. IINA should try to match this behavior. But I am seeing consistent numbers in the Inspector window also. I don't see more than 6 digits after the decimal.

> So the problem with values less than 0.01 is expected mpv behavior.

Oh ha. Guess I should have read the manual. But how strange that it won't support anything slower. It seems like that they should get the behavior for free, considering the fine-grained control they have at faster speeds. Maybe it starts to generate false positive alerts from their audio/video sync detection / buffer underrun subsystems.

I'll set the minimum to 0.01 then.

> The problem with the slider thumb being cut off is not specific to he. I see it with en as well and when using default. It is just the very edge of the thumb that is cut off: I am testing under Ventura 13.6.9. I will test on my Mac running Sonoma. Maybe AppKit has fixed AppKit. Tuesday busy for me, so might not get to that until tomorrow.

Hmm. If it is a legacy problem, it will be hard for me to test. But I can take a shot at a fix.

One more thing which became apparent. I found that numbers in IINA are not being localized properly in lots of places. So if the locale uses a `,` instead of a `.` to signify a decimal, that is not being followed.

SCR-20240917-lbkq

In my latest commit I added an extension to `Double` which adds a `string` computed property which should properly localize the number.

svobs:
Regarding the slider knob clipping problem, I found it and fixed it. I was able to reproduce it by changing `Clips Bounds` to `YES` for the slider's parent view. Recall that it was always `YES` in older versions of MacOS but defaults to `NO` now, which is most likely why it is not a problem for more recent versions.

I also found & fixed another minor regression: a horizontal line in the Audio tab did not go all the way across the sidebar.

iina/QuickSettingViewController.swift
```diff
@@ -723,6 +749,10 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
     view.layout()
   }
 
+  @IBAction func resetSpeedction(_ sender: AnyObject) {
```
low-batt: Missing "A" in `resetSpeedction`

iina/Extensions.swift
```diff
@@ -227,6 +227,23 @@ extension Double {
...
+    return finalVal
+  }
+  
+  /// Formats this number as a decimal string, using default locale.
+  ///
+  /// This should be used in most places where decimal numbers need to be printed.
+  var string: String {
+    /// Output up to 15 digits after the decimal. This should be good enough for almost all cases, but can be increased in the future.
+    /// (It's not clear what the maximum allowable value for `NumberFormatter.maximumFractionDigits` actually is. Trying to use
+    /// `NSIntegerMax` seems to result in `maximumFractionDigits` being silently set to `6` instead.)
```
low-batt: I believe markdown comments can only be used on declarations. I believe comments inside code blocks should just be simple (//) comments. I tried finding Swift documentation that explains where markdown comments can be used and struck out. I don't think there is any way to get Xcode to show this comment in the quick help. It will show the comment on the declaration if you click on `string`.

For small blocks of code like this I will put the fancy comment outside of the code and in the comment on the declaration.

low-batt:
Glad you found the problem with clipping the thumb. I was just about to report that the clipping did not reproduce for me under macOS Sonoma. I tested the latest commits under Ventura and the clipping is fixed.

This is looking good. Need to check with @uiryuu as to how we can change the format specifiers in the files for other locals.

iina/QuickSettingViewController.swift
svobs: Oops. Thanks for catching that. Fixed

iina/Extensions.swift
svobs: I tend to default to the "fancy comments" because of their ability to highlight `code snippits` in them. It's also annoying to start with a "regular" double-slash comment and then have to change it. But in this case I think you are right to point out that I should have just put these details above the declaration because they could be important. Fixed in latest.

uiryuu:
Looks nice, thank you! If there are still unfinished tasks regarding this issue, please open another PR.

low-batt:

PR #5178, opened 2024-09-19
Allow drop of playlist files into Welcome window

### LINKED ISSUE #5311: Drag-and-drop support for m3u and m3u8 playlists
**What you want IINA to do:**
I would love for IINA to adopt drag-and-drop support for m3u, m3u8 and similar playlist files.

**What IINA does currently:**
This is my setup:

- IINA 1.3.5 Build 141
- mpv 0.35.0-419-gf79458476b-dirty
- FFmpeg 6.0
- macOS 15.2 (24C101)

In my case, I have m3u8 files with references to a bunch of local mp3 files. No remote sources are referenced in these.

I can open these m3u8 playlists via the "Open..." dialogue. I can open the mp3 files either via "Open..." or by dragging and dropping them into IINA (either anywhere onto the launch window or onto a specific position within the queue).

I cannot drag and drop playlist files at all, even those created with IINA. I could reproduce this missing feature with m3u files as well. Not sure about other playlist formats.

I'm suggesting this because it would lead to more consistent behaviour.

VLC Player already supports this.

svobs:

---

**Description:**

Allows playlist files such as `*.m3u` to be opened by dropping them onto the Welcome window, so it matches the drag & drop functionality of the IINA dock icon.

This cleans up a little bit of the affected code in `DraggingDetect.acceptFromPasteboard` so it more closely resembles the code in `PlayerCore.openURLs`, but the key part is the added check for `Utility.playlistFileExt.contains(url.absoluteString.lowercasedPathExtension)`.

### DIFF
--- a/iina/DraggingDetect.swift
+++ b/iina/DraggingDetect.swift
@@ -140,6 +140,7 @@ extension PlayerCore {
     if types.contains(.nsFilenames) {
       guard var paths = pb.propertyList(forType: .nsFilenames) as? [String] else { return [] }
       paths = Utility.resolvePaths(paths)
+      
       // check 3d lut files
       if paths.count == 1 && Utility.lut3dExt.contains(paths[0].lowercasedPathExtension) {
         return .copy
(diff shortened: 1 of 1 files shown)


uiryuu:
Thanks!

low-batt:

PR #5183, opened 2024-09-22
Quick settings UI improvements

ghost:

---

**Description:**
The current track selectors are bulky, yet too small for comfortable use. The new design resolves this issue, making it easier to find and select tracks, especially in long lists.

| Before | After |
|--------|--------|
| 1 before | 2 after |


uiryuu:
I'm pretty sure that we intended to make most of the options/selection of different tracks available for only one 1-click, in order to achieve the "Quick" objective as in "Quick Settings Panel". While I'm not oppose this change, need to @lhc70000's opinion here.

PR #5204, opened 2024-10-10
Add live update of color-related options in Settings > Codec, enforce valid values

svobs:

---

**Description:**
- Adds monitoring of the 4 color mapping related pref settings to `PlayerWindowController` and reloads them when changed, so the user doesn't have to quit & reopen IINA to guarantee they take effect.
- Modifies `VideoView.refreshEdrMode` so it doesn't cancel other changes if `colorspace` is already set, so the above can take effect.
- Add action to `PrefCodecViewController` to restrict `Target peak` user entry to valid values, instead of silently failing & using `400`.

The affected settings:
Affected settings

### DIFF
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -245,6 +245,8 @@
 "alert.error_saving_file" = "Error occurred when saving %@: %@";
 "alert.error_deleting_file" = "Cannot delete the file.";
 
+"alert.target_peak.bad_value" = "The value \"%@\" is invalid.\nPlease provide a value betwen 10 and 10000, or 0.";
+
 "alert.config.new.title" = "New Input Configuration";
 "alert.config.new.message" = "Please enter a name for the new configuration.";
 "alert.config.duplicate.title" = "Duplicate Configuration";
--- a/iina/PrefCodecViewController.swift
+++ b/iina/PrefCodecViewController.swift
@@ -110,8 +110,6 @@ class PrefCodecViewController: PreferenceViewController, PreferenceWindowEmbedda
     hwdecDescriptionTextField.stringValue = hwdec.localizedDescription
   }
 
-  // TODO: add "auto" option instead of 0 value
-  // TODO: add option to support "inf"
   @IBAction func toneMappingTargetPeakAction(_ sender: NSTextField) {
     let newValue = sender.integerValue
     // constrain to valid mpv values
--- a/iina/PrefCodecViewController.swift
+++ b/iina/PrefCodecViewController.swift
@@ -110,6 +110,17 @@ class PrefCodecViewController: PreferenceViewController, PreferenceWindowEmbedda
     hwdecDescriptionTextField.stringValue = hwdec.localizedDescription
   }
 
+  // TODO: add "auto" option instead of 0 value
+  // TODO: add option to support "inf"
+  @IBAction func toneMappingTargetPeakAction(_ sender: NSTextField) {
+    let newValue = sender.integerValue
+    // constrain to valid mpv values
+    let validValue = newValue == 0 ? 0 : newValue.clamped(to: 10...10000)
+    if newValue != validValue {
+      Preference.set(validValue, for: .toneMappingTargetPeak)
+    }
+  }
+
   @IBAction func toneMappingHelpAction(_ sender: Any) {
     NSWorkspace.shared.open(URL(string: AppData.toneMappingHelpLink)!)
   }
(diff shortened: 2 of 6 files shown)


uiryuu (approved):

iina/VideoView.swift
```diff
@@ -482,12 +482,11 @@ extension VideoView {
...
-    if videoLayer.colorspace?.name == name {
-      logHDR("HDR mode already enabled, skipping")
-      return true
-    }
+    logHDR("Using HDR color space instead of ICC profile")
 
-    logHDR("Will activate HDR color space instead of using ICC profile")
+    if videoLayer.colorspace?.name != name {
+      videoLayer.colorspace = CGColorSpace(name: name!)
+    }
```
low-batt: Remove if statement and setting of `videoLayer.colorspace`. No need for this statement? The `colorspace` is unconditionally set on line 491 below.

iina/PrefCodecViewController.swift
```diff
@@ -110,6 +110,17 @@ class PrefCodecViewController: PreferenceViewController, PreferenceWindowEmbedda
     hwdecDescriptionTextField.stringValue = hwdec.localizedDescription
   }
 
+  // TODO: add "auto" option instead of 0 value
```
low-batt: Might want to support `auto`, but it should _not_ be a replacement for `0`. Setting the [--target-peak](https://mpv.io/manual/stable/#options-target-peak) option to `auto` has a specific meaning, documented in the mpv manual. Zero has been grabbed by IINA for its own implementation that attempts to actually detect the peak of the display. We'd want to add something like `IINA detected`. Something to distinguish from the mpv behavior.

iina/PrefCodecViewController.swift
```diff
@@ -110,6 +110,17 @@ class PrefCodecViewController: PreferenceViewController, PreferenceWindowEmbedda
...
+  // TODO: add option to support "inf"
+  @IBAction func toneMappingTargetPeakAction(_ sender: NSTextField) {
+    let newValue = sender.integerValue
+    // constrain to valid mpv values
+    let validValue = newValue == 0 ? 0 : newValue.clamped(to: 10...10000)
```
low-batt: Merely constraining the value is not the convention used when the user enters invalid values for other settings. Trying entering `-1`. Other settings pop-up an alert box.

low-batt (changes requested):
Glad you are enhancing this setting.

I'm thinking this needs to put up an alert for invalid values similar to other settings.

iina/VideoView.swift
svobs: Oh yes. Good catch. Removed.

iina/PrefCodecViewController.swift
svobs: Good point. I hadn't put too much thought into it. So there are actually 3 special values apart from the normal number range. Might want to add additional UI for those in the future...

The TODOs have been removed :)

iina/PrefCodecViewController.swift
svobs: I kinda loathe the built-in modal popup. It's a really old paradigm, it's obtrusive, and not even very helpful. It doesn't tell the user what the valid values actually are, and if they choose the `OK` button the field is allowed to be left blank, which leaves them guessing about the current value. At least the `clamped` approach chooses the closest valid value, and doesn't even require reading.

Also...I actually don't know how to make the built-in popup work for `0` because it's outside the contiguous range of 10-10,000. As far as I can tell, the popup is provided via the text field cell's `NumberFormatter`, which checks against accepted min and max only. I don't see any places to hook into or override it to provide more complex validation.

low-batt: Sorry, I didn't mean to imply we needed to use the alert provided by AppKit. I was expecting we would use `Utility.showAlert` to show our own alert with the appropriate information about what values are allowed.

I agree, the one provided by `NumberFormatter` can't be used due to the lack of a contiguous range. And yes, I too dislike the lack of info in the alert `NumberFormatter` provides.

low-batt: @svobs Still waiting for an alert to be added.

svobs: I pushed an update which uses an alert instead. I tried to mimic Apple's alert mostly, but it has only an OK button and resets the field to the current (valid) prefs value.

low-batt: This looks good:
alert

low-batt (approved):
Looks good to me.

iina/Base.lproj/Localizable.strings
```diff
@@ -245,6 +245,8 @@
 "alert.error_saving_file" = "Error occurred when saving %@: %@";
 "alert.error_deleting_file" = "Cannot delete the file.";
 
+"alert.target_peak.bad_value" = "The value \"%@\" is invalid.\nPlease provide a value betwen 10 and 10000, or 0.";
```
saagarjha, line 248: Between. Not sure why we have a typos CI job if it can’t catch these…

uiryuu: I think that only checks for typos in code and comments

low-batt:

PR #5213, opened 2024-10-14
Fix  colormanagement broken in 1.3.0, #3929

### LINKED ISSUE #3929: Colormanagement broken in 1.3.0
Bildschirmfoto 2022-08-28 um 22 52 21

This is how I usually tell Iina to use the correct ICC-Profile for my monitor. It all worked well and Iina gave me correct colors and gamut until I updated from 1.2.0 to 1.3.0. Colors were off (no color management any more). Does anybody have the same issue?

low-batt:
This commit will:
- Add a new method `setRenderICCProfile` to `VideoLayer` that sets `MPV_RENDER_PARAM_ICC_PROFILE`
- Remove setting of the mpv option `icc-profile` from the `VideoView` method `setICCProfile`
- Change `VideoView` to use `icc-profile-auto` instead of `icc-profile` to control use of an ICC profile

Using `icc-profile-auto` frees `icc-profile` for use by users as  `icc-profile` overrides `icc-profile-auto`. IINA must set `MPV_RENDER_PARAM_ICC_PROFILE` in order for `icc-profile-auto` to work.

- [x] This implements/fixes issue #3929.

---

**Description:**

### DIFF
--- a/iina/VideoView.swift
+++ b/iina/VideoView.swift
@@ -305,7 +305,8 @@ class VideoView: NSView {
       logHDR("Not using ICC profile due to user preference")
       player.mpv.setFlag(MPVOption.GPURendererOptions.iccProfileAuto, false)
     } else if let screenColorSpace {
-      logHDR("Using ICC profile")
+      let name = screenColorSpace.localizedName ?? "unnamed"
+      logHDR("Using the ICC profile of the color space \(name)")
       player.mpv.setFlag(MPVOption.GPURendererOptions.iccProfileAuto, true)
       videoLayer.setRenderICCProfile(screenColorSpace)
     }
--- a/iina/VideoView.swift
+++ b/iina/VideoView.swift
@@ -299,39 +299,17 @@ class VideoView: NSView {
     RunLoop.current.add(displayIdleTimer!, forMode: .default)
   }
 
-  func setICCProfile(_ displayId: UInt32) {
+  private func setICCProfile() {
+    let screenColorSpace = player.mainWindow.window?.screen?.colorSpace
     if !Preference.bool(for: .loadIccProfile) {
-      logHDR("Not using ICC due to user preference")
-      player.mpv.setString(MPVOption.GPURendererOptions.iccProfile, "")
-    } else {
-      logHDR("Loading ICC profile")
-      typealias ProfileData = (uuid: CFUUID, profileUrl: URL?)
-      guard let uuid = CGDisplayCreateUUIDFromDisplayID(displayId)?.takeRetainedValue() else { return }
-
-      var argResult: ProfileData = (uuid, nil)
-      withUnsafeMutablePointer(to: &argResult) { data in
-        ColorSyncIterateDeviceProfiles({ (dict: CFDictionary?, ptr: UnsafeMutableRawPointer?) -> Bool in
-          if let info = dict as? [String: Any], let current = info["DeviceProfileIsCurrent"] as? Int {
-            let deviceID = info["DeviceID"] as! CFUUID
-            let ptr = ptr!.bindMemory(to: ProfileData.self, capacity: 1)
-            let uuid = ptr.pointee.uuid
-
-            if current == 1, deviceID == uuid {
-              let profileURL = info["DeviceProfileURL"] as! URL
-              ptr.pointee.profileUrl = profileURL
-              return false
-            }
-          }
-          return true
-        }, data)
-      }
-
-      if let iccProfilePath = argResult.profileUrl?.path, FileManager.default.fileExists(atPath: iccProfilePath) {
-        player.mpv.setString(MPVOption.GPURendererOptions.iccProfile, iccProfilePath)
-      }
+      logHDR("Not using ICC profile due to user preference")
+      player.mpv.setFlag(MPVOption.GPURendererOptions.iccProfileAuto, false)
+    } else if let screenColorSpace {
+      logHDR("Using ICC profile")
+      player.mpv.setFlag(MPVOption.GPURendererOptions.iccProfileAuto, true)
+      videoLayer.setRenderICCProfile(screenColorSpace)
     }
 
-    let screenColorSpace = player.mainWindow.window?.screen?.colorSpace
     let sdrColorSpace = screenColorSpace?.cgColorSpace ?? VideoView.SRGB
     if videoLayer.colorspace != sdrColorSpace {
       let name: String = {
@@ -435,7 +413,7 @@ extension VideoView {
     if player.info.hdrAvailable != edrAvailable {
       player.mainWindow.quickSettingView.setHdrAvailability(to: edrAvailable)
     }
-    if edrEnabled != true { setICCProfile(displayId) }
+    if edrEnabled != true { setICCProfile() }
   }
 
   func requestEdrMode() -> Bool? {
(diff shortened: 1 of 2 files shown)


low-batt:
This is a draft because we need to update the `libmpv` on our download server with the latest patches.

uiryuu (commented):
Tested with `libmpv`+latest patches, confirmed user can override ICC profile when "Load ICC Profile" checkbox is checked. (The recompiled `libmpv` hasn't been uploaded yet)

A few questions here, not blocking this PR tho:

1. `videoLayer`'s colorspace is always set to window's colorspace, even though the user provide their own ICC profile. Is this desired?
2. The `vcolorspaceField` now will only show the color profile of the screen (system setting), even though the user provide their own ICC profile. Is this correct?

low-batt:
I don't know enough about color spaces to provide an authoritative answer. What I can say is that the mpv code does not take into account the icc-profile option when setting the color space of the layer.

svobs:
>A few questions here, not blocking this PR tho:
> 
> 1. videoLayer's colorspace is always set to window's colorspace, even though the user provide their own ICC profile. Is this desired?
> 2. The vcolorspaceField now will only show the color profile of the screen (system setting), even though the user provide their own ICC profile. Is this correct?

1. Not that I can completely answer this, but I think I have some understanding... 
As I talked about in my analysis in [this post](https://github.com/iina/iina/pull/5154#issuecomment-2393249344), setting the layer's colorspace is a Mac-only extra layer of translation, albeit one that must be configured to enable [E|H]DR. Whereas `icc-profile` relates to the screen, `cocoa-cb-output-csp` relates to the layer. But mpv is already doing colorspace management, right? So you wouldn't want too much more color mapping being done between "layer" and "screen"...
Well, according to the docs for [cocoa-cb-output-csp](https://mpv.io/manual/stable/#options-cocoa-cb-output-csp): `To get correct results, this needs to be set to the color primaries/transfer characteristics of the VO target`. So when using `icc-profile-auto`with `cocoa-cb-output-csp=auto` (as this PR is doing), both will be set to the screen's colorspace, which should ensure good results.
If some power user specifies a different `icc-profile`, presumably that user needs to also figure out the relevant value for `cocoa-cb-output-csp`, as well as `target-trc` and `target-prim` as the doc also notes. If they are crazy enough to provide all of those, then they probably have enough time on their hands to do the tons of fine-tuning that will be needed...
2. Sounds like the ICC profile information is missing from the Inspector and should be added? It sounds like both `icc-profile` & `icc-profile-auto` are fairly new options (relative to the Inspector) which could have been missed.

low-batt:
Too late here for me to be thinking, so just one comment. The `cocoa-cb-output-csp` mpv option was added after the mpv 0.38.0 release and is missing from the `libmpv` IINA is using.

uiryuu:
Update: libmpv is uploaded. Convert this PR to ready for review.

svobs:
> Too late here for me to be thinking, so just one comment. The cocoa-cb-output-csp mpv option was added after the mpv 0.38.0 release and is missing from the libmpv IINA is using.

To be clear, I am not suggesting that `cocoa-cb-output-csp` should be configurable in IINA's UI. I was just pointing out that IINA is in effect implementing `cocoa-cb-output-csp=auto` when it sets `videoLayer.colorspace` equal to the screen's colorspace.

iina/ViewLayer.swift
```diff
@@ -347,6 +347,40 @@ class ViewLayer: CAOpenGLLayer {
     return ctx
   }
 
+  // MARK: - ICC Profile
```
svobs, line 350: Why was this put in `ViewLayer`? Looks like it would fit better in `MPVController` (or even just `VideoView`). You could pull out the `videoView.$isUninited.withLock` into the calling method to keep it inside `VideoView` and reduce level of nested code.

iina/VideoView.swift
```diff
@@ -299,39 +299,17 @@ class VideoView: NSView {
...
-        }, data)
-      }
-
-      if let iccProfilePath = argResult.profileUrl?.path, FileManager.default.fileExists(atPath: iccProfilePath) {
-        player.mpv.setString(MPVOption.GPURendererOptions.iccProfile, iccProfilePath)
-      }
+      logHDR("Not using ICC profile due to user preference")
+      player.mpv.setFlag(MPVOption.GPURendererOptions.iccProfileAuto, false)
+    } else if let screenColorSpace {
+      logHDR("Using ICC profile")
```
svobs: Might be good to change this message to "Configuring auto ICC profile" or something with "auto" in it

iina/ViewLayer.swift
```diff
@@ -347,6 +347,40 @@ class ViewLayer: CAOpenGLLayer {
...
+  /// option.
+  ///
+  /// This method fulfills the mpv requirement that applications using libmpv with the render API provide the ICC profile via
+  /// `MPV_RENDER_PARAM_ICC_PROFILE` in order for the `--icc-profile-auto` option to work. The ICC profile data will not
+  /// be used by mpv unless the option is enabled.
+  ///
+  /// The IINA `Load ICC profile` setting is tied to the `--icc-profile-auto` option. This allows users to override IINA using
+  /// the [--icc-profile](https://mpv.io/manual/stable/#options-icc-profile) option.
+  func setRenderICCProfile(_ profile: NSColorSpace) {
+    videoView.$isUninited.withLock() { isUninited in
```
svobs, line 366: I merged this PR into my own code yesterday and have since been getting frequent crashes at startup when using the Xcode debugger, coming from inside `mpv_render_context_render`. The last 2 I saved:
1. `Assertion failed: (tex->params.render_src), function update_uniform, file ra_gl.c, line 990.`
2. `Assertion failed: (h->canary == CANARY), function ta_dbg_check_header, file ta.c, line 291.`

I did some experiments and it looks like I fixed it by locking the OpenGL context around the `isUninited` lock here in `setRenderICCProfile`. I realize my code is different than `develop` but I do suspect the problem exists here but just may be more rare. I suspect mpv needs calls to `mpv_render_context_set_parameter` to be isolated from all other rendering-related operations.

iina/ViewLayer.swift
low-batt: Sorry I'm slow in responding. Been focused on the JPEG XL problems.

The plan is to put it in `MPVRenderer`, along with all the render related code currently in `MPVController`, similar to how mpv structures the code. But I got interrupted doing that refactoring. As we are now looking to freeze changes, test and release the first beta, this refactoring will have to wait for a future release. So this method is here temporarily.

iina/VideoView.swift
low-batt: Will fix.

iina/ViewLayer.swift
low-batt: I will make this a draft and look into this.

low-batt:
Changed to draft while I look into the crashes @svobs encountered.

iina/VideoView.swift
low-batt: I updated the log message to include the name of the color space:
```text
19:10:15.183 [hdr0][d] Using the ICC profile of the color space LG UltraFine
```
I didn't mention any thing about automatically, because IINA doesn't take advantage of that aspect of the option. IINA is manually turning `icc-profile-auto` on and off.

iina/ViewLayer.swift
low-batt: Yes, that was it. I added the code to lock the OpenGL context.

low-batt:

PR #5216, opened 2024-10-15
Fix XCode 16 warnings & improve Set as Default for MacOS 11+

svobs:

---

**Description:**
- Add `@preconcurrency` annotation to `import WebKit` to fix warnings
- Create Utility.icon() which uses `UTType`s, to quash deprecation warning about `NSWorkspace.shared.icon`
- Trivial updates to fix `@Sendable` warnings
- Minor refactor to use `NSShadow.shadowColor` instead of deprecated `NSColor.controlShadowColor` in `PlaySliderCell`
- Refactor `PrefUtilsViewController.setAsDefaultOKBtnAction` logic to use `UTType`s for MacOS 11+, and ensure that all UT identifiers for all supported suffixes are changed to IINA

This is a merge of various warnings-related fixes I made - sorry for not breaking it up.

IMO the only item which needs discussion is the last one. As I dug in deeper and did some tests. I found that:
- Modern MacOS (since when? Not sure) uses UTTypes, not file suffixes, to determine which application to open. But the system does store file suffix info with each UTType in a many-to-many relation, which means that each UTI can be associated with multiple suffixes, and each suffix can be associated with multiple UTIs.
- The present solution starts with the "import" UTIs listed in IINA's bundle, and grabs the file suffixes listed for each. It then uses the now-deprecated `UTTypeCreatePreferredIdentifierForTag` to get a single UTI from a single file suffix, and then uses the UTI to change the default app. But this seems to be...not completely correct? Its [documentation](https://developer.apple.com/documentation/coreservices/1448939-uttypecreatepreferredidentifierf) states: `If there is more than one possible UTI for the specified tag, the UTI that will be returned is undefined`. But most of the suffixes (or "tags") being passes have more than one UTI associated with them. In practice, this is more well-defined than the docs state, because it seems to send the most relevant one. But it does miss a lot.
- In this PR I created a solution for MacOS 11+ which grabs all UTIs associated with each file extension. This will include some duplicates so I store them all in a `Set`. Then I pass each of them to `LSSetDefaultRoleHandlerForContentType` as before. On my system, this solution finds and changes 118 UTIs, where formerly it found 65.
- Worth noting: it didn't work to use the "import" UTIs directly. These all start with `io.iina.`, which IINA already associates with. It seems that once a file is assigned a UTI by the OS, it keeps that UTI. So a given `.mkv` file's UTI may be `io.mpv.mkv`, `io.iina.mkv`, `org.matroska.mkv`, or so one.
- If reviewer(s) agree that this is correct solution for MacOS 11+, perhaps the pre-MacOS 11 solution should be modified as well? Looks like this would involve calling `UTTypeCreateAllIdentifiersForTag` instead of `UTTypeCreatePreferredIdentifierForTag`. Or if the previous solution is favored for some reason, I would likely modify the MacOS 11+ code to only grab the first UTI for each file extension.

### DIFF
--- a/iina/GuideWindowController.swift
+++ b/iina/GuideWindowController.swift
@@ -7,7 +7,7 @@
 //
 
 import Cocoa
-import WebKit
+@preconcurrency import WebKit
 
 fileprivate let highlightsLink = "https://iina.io/highlights"
 
--- a/iina/HistoryWindowController.swift
+++ b/iina/HistoryWindowController.swift
@@ -215,7 +215,7 @@ class HistoryWindowController: NSWindowController, NSOutlineViewDelegate, NSOutl
         let fileExists = !entry.url.isFileURL || FileManager.default.fileExists(atPath: entry.url.path)
         filenameView.textField?.stringValue = entry.url.isFileURL ? entry.name : entry.url.absoluteString
         filenameView.textField?.textColor = fileExists ? .controlTextColor : .disabledControlTextColor
-        filenameView.docImage.image = NSWorkspace.shared.icon(forFileType: entry.url.pathExtension)
+        filenameView.docImage.image = Utility.icon(for: entry.url)
       } else if identifier == .progress {
         // Progress cell
         let progressView = cell as! HistoryProgressCellView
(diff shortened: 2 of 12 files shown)


uiryuu (approved):
Changes look good to me. It seems not harmful to grab all UTIs and set IINA to default to all of them, so I agree to use `UTTypeCreateAllIdentifiersForTag` for macOS 11-.

svobs:
Pushed a new commit which changes to `UTTypeCreateAllIdentifiersForTag`. From my testing on my Mac it also gets 118 UTIs.

PR #5233, opened 2024-10-27
Improve decimal formatting in OSD & menu items

svobs:

---

Decimal-formatting

**Description:**

Adding to the fix for #5147, here is the "remaining" work that I could find in relation to decimal-to-string number formatting issues.
- Replace decimal formatting specifiers with '%@' in i18n strings & use Double.string property extension in most places to ensure proper localized decimal & grouping separator punctuation.
- This commit extends to audio & subtitles (& seek menu items) what was already changed for speed-related items.
- Removal of rounding/truncation allows for more precise number display.

In many of these cases I needed to replace `Integer` values with `Double`s to support additional precision. It was unclear to me whether some things (e.g., volume display in the menu) were truncated/rounded to whole numbers for aesthetic or other reasons. It seems to me that they should support fractional digits of numbers wherever mpv is able to store them as such.

### DIFF
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -84,7 +84,7 @@
 // OSDMessage.swift
 "osd.pause" = "Pause";
 "osd.resume" = "Resume";
-"osd.volume" = "Volume: %i";
+"osd.volume" = "Volume: %@";
 "osd.speed" = "Speed: %@x";
 "osd.aspect" = "Aspect Ratio: %@";
 "osd.crop" = "Crop: %@";
@@ -92,19 +92,19 @@
 "osd.deinterlace" = "Deinterlace: %@";
 "osd.hwdec" = "Hardware Decoding: %@";
 "osd.audio_delay.nodelay" = "Audio Delay: No Delay";
-"osd.audio_delay.later" = "Audio Delay: %.2fs Later";
-"osd.audio_delay.earlier" = "Audio Delay: %.2fs Earlier";
+"osd.audio_delay.later" = "Audio Delay: %@s Later";
+"osd.audio_delay.earlier" = "Audio Delay: %@s Earlier";
 "osd.sub_delay.nodelay" = "Subtitle Delay: No Delay";
-"osd.sub_delay.later" = "Subtitle Delay: %.2fs Later";
-"osd.sub_delay.earlier" = "Subtitle Delay: %.2fs Earlier";
-"osd.subtitle_pos" = "Subtitle Position: %.1f";
+"osd.sub_delay.later" = "Subtitle Delay: %@s Later";
+"osd.sub_delay.earlier" = "Subtitle Delay: %@s Earlier";
+"osd.subtitle_pos" = "Subtitle Position: %@";
 "osd.sub_hidden" = "Subtitles Hidden";
 "osd.sub_visible" = "Subtitles Visible";
 "osd.sub_second_delay.nodelay" = "Secondary Subtitle Delay: No Delay";
-"osd.sub_second_delay.later" = "Secondary Subtitle Delay: %.2fs Later";
-"osd.sub_second_delay.earlier" = "Secondary Subtitle Delay: %.2fs Earlier";
+"osd.sub_second_delay.later" = "Secondary Subtitle Delay: %@s Later";
+"osd.sub_second_delay.earlier" = "Secondary Subtitle Delay: %@s Earlier";
 "osd.sub_second_hidden" = "Secondary Subtitles Hidden";
-"osd.sub_second_pos" = "Secondary Subtitle Position: %.1f";
+"osd.sub_second_pos" = "Secondary Subtitle Position: %@";
 "osd.sub_second_visible" = "Secondary Subtitles Visible";
 "osd.mute" = "Mute";
 "osd.unmute" = "Unmute";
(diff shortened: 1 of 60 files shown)


low-batt:
I just encountered the Int / Double issue with subtitle position. I found `MPVController` was treating the values of the subtitle position options as integers when they are doubles. I changed that has if IINA wants to treat a mpv option that has a double value as an integer that is fine, but it should still treat the value as a Double when interfacing to mpv.

The subtitle position handling looks broken to me. Although the slider is using integers, the OSD is using a Double for position and the settings panel allows you to enter a float for the subtitle position. This should be consistent. Corrections are needed.

@uiryuu Did we want the position slider to use integers for style reasons?

low-batt:
I figured out the reason an integer is being used for subtitle position instead of Double. Previously the mpv option was an integer. So this was correct with older versions of mpv.

svobs:
Force-pushed the wrong commit (whoops). Should be properly rebased now.

svobs:
@HarukaMa something going on with the build automation? This builds fine on my local machine.

HarukaMa:
```
/Users/runner/work/iina/iina/iina/Assets.xcassets:./Icons/puzzlepiece.extension.symbolset/[universal][][][]: error: The SVG file provided for the symbol image set ’puzzlepiece.extension‘ is not suitable: Template format 6.0 is newer than the version that this software supports (5.0)
```

it seems currently the assets file needs to be compiled with xcode 16+.

HarukaMa:
I just updated system and Xcode version the CI uses. You can rebase to the latest develop branch if you need the result of CI build.

svobs:
Rebased and fixed conflicts to account for PR #5480.

uiryuu:
Can you change the format of the strings for all localization for this PR? Otherwise all the localization will lost their translations after this PR is merged.

svobs:
> Can you change the format of the strings for all localization for this PR? Otherwise all the localization will lost their translations after this PR is merged.

You mean for example, change
```
"osd.audio_delay.later" = "音频延迟：延迟 %.2f 秒";
```
to this?
```
"osd.audio_delay.later" = "音频延迟：延迟 %@ 秒";
```

I was under the impression that all the `Localizable.strings`, except for Base and `en` were downstream from Crowdin. Aren't they normally changed there after the base locale is merged into `develop`?

uiryuu:
> I was under the impression that all the Localizable.strings, except for Base and en were downstream from Crowdin. Aren't they normally changed there after the base locale is merged into develop?

You are correct. However I don't think Crowdin will be smart enough to automatically identify that the only change is the format and automatically replace the format for us. It will just delete all the already translated strings and wait the translators to translate them again.

Because it seems straightforward to change the format for all languages (we didn't make any semantics, therefore no new translation is needed), I think we should do so to prevent all the edited strings to fall back to English.

svobs:
OK... Rebased and updated *all* the localizations.

I came up with this Bash script to do the find & replace. Maybe it will be useful in the future:
```Bash
#!/bin/bash
# Place in project's base directory, then run from there.

PDIR=`pwd` # project base dir
# Space-separated replacement items. Each line consists of: {Key} {StringToFind} {ReplacementString}
REPS=(
  "osd.volume %i %@"
  "osd.audio_delay.later %.2f %@"
  "osd.audio_delay.earlier %.2f %@"
  "osd.sub_delay.later %.2f %@"
  "osd.sub_delay.earlier %.2f %@"
  "osd.subtitle_pos %.1f %@"
  "osd.sub_second_delay.later %.2f %@"
  "osd.sub_second_delay.earlier %.2f %@"
  "osd.sub_second_pos %.1f %@"
  "osd.subtitle_scale %.2f %@"
  "menu.volume %d %@"
  "menu.volume_muted %d %@"
  "menu.audio_delay %.2f %@"
  "menu.sub_delay %.2f %@"
  "menu.seek_forward %.0f %@"
  "menu.seek_backward %.0f %@"
  "menu.volume_up %.0f %@"
  "menu.volume_down %.0f %@"
  "menu.audio_delay_up %.1f %@"
  "menu.audio_delay_down %.1f %@"
  "menu.sub_delay_up %.1f %@"
  "menu.sub_delay_down %.1f %@"
)
for FILE in `ls "$PDIR"/iina/*.lproj/Localizable.strings`
do
  for REP in "${REPS[@]}"
  do
    TOKENS=($REP) # Break line into tokens by spaces
    sed -i '' 's/\("'"${TOKENS[0]}"'".*\)\('${TOKENS[1]}'\)\(.*\)/\1'${TOKENS[2]}'\3/' "$FILE"
  done
done
```

uiryuu:
@svobs Sorry for the late reply. When I tried to run IINA, it crashes at `updateMenuItem` due sometimes `valObj` is not converted to `String` when the format is `%@`.

https://github.com/iina/iina/blob/8a9bfaff5bfdc610eb0b62d2e5478ae24c55155e/iina/MenuController.swift#L954-L957

I think here we have an incomplete list for the keys requires a string. By trailing and error I got a list to prevent it from crashing:

```
        "speed_up",
        "speed_down",
        "volume_up",
        "volume_down",
        "seek_forward",
        "seek_backward",
        "audio_delay_down",
        "audio_delay_up",
        "sub_delay_down",
        "sub_delay_up"
```

After putting these keys (please if they are complete!), IINA will still crash when opening menus (for example, the audio menu).

svobs:
@uiryuu Sorry about that. I must have been lacking sleep or something, and did not QA it. Even the script above had defects (I have updated it with corrections), & I have pushed the code fixes.

All of the "menu.*" strings in `Localizable.strings` _which have parameters_ now use Strings, so I was able to remove the case statement. And more code changes were needed for `OSDMessage` to properly use the new formatting (not sure why I missed those too). Lastly, I modified `OSDMessage.volume` to use a Double instead of an Int so that it can display more precise values if needed.

uiryuu:
No problem, and thanks for your work!

low-batt:
This was fixed in IINA 1.4.0.

PR #5239, opened 2024-10-29
Finer grained audio delay slider (fixes issue #5010)

### LINKED ISSUE #5010: Fine grained audio delay slider
**What you want IINA to do:**
Provide a way to change the `Audio delay` slider on the `AUDIO` tab of the quick settings panel to be able to make fine grained (0.1s) adjustments to audio delay.

**What IINA does currently:**
The `Audio Delay` menu items and the `Audio delay` slider on the `AUDIO` tab of the quick settings panel change the delay by 0.5s at a time.

In the IINA default key bindings the key ⌥( is bound to `add audio-delay 0.1` and the key ⌥) is bound to `add audio-delay -0.1`, so these keys can be used to make fine grained changes, but it would be nice to be able to make small changes using the slider.

See discussion #5003.

VLC changes the delay by 50ms for each key press.

svobs:
- [x] This implements/fixes issue #5010.

---

**Description:**
- Changes the audio delay slider so it uses increments of 50ms instead of 500ms.
- ~~The slider's is still required to land on tick marks as it moves, so it was necessary to increase the number of tick marks per second by 10x. This turns the slider into a millipede. I changed its control size from "small" to "mini" to reduce the prominence of the taller marks, and I think it looks a little better.~~
- ~~Reduces the slider's range (was: `-5` ↔︎ `+5`) to: `-3` ↔︎ `+3`. The slider's relatively small size made it difficult to move very precisely otherwise. But I think it should be a reasonable tradeoff because it should be very uncommon for an audio track to be farther out of sync than that.~~

### DIFF
--- a/iina/QuickSettingViewController.swift
+++ b/iina/QuickSettingViewController.swift
@@ -799,15 +799,20 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
 
   @IBAction func audioDelayChangedAction(_ sender: NSSlider) {
     let eventType = NSApp.currentEvent!.type
-    if eventType == .leftMouseDown {
-      sender.allowsTickMarkValuesOnly = true
+    let sliderValue: Double
+    switch eventType {
+    case .leftMouseDown, .leftMouseDragged, .leftMouseUp:
+      // When dragging slider with the mouse, snap to the nearest 50ms (1/20 sec)
+      // Although it is possible to show tick marks at every step of 0.05 in the slider, it is visually unpleasant.
+      // So we draw less tick marks, and keep "Only stop on tick marks" disabled, and add our own logic to stop on
+      // "virtual tick marks" for these values.
+      sliderValue = (sender.doubleValue * 20.0).rounded() / 20.0
+      sender.doubleValue = sliderValue
+    default:
+      sliderValue = sender.doubleValue
     }
-    if eventType == .leftMouseUp {
-      sender.allowsTickMarkValuesOnly = false
-    }
-    let sliderValue = sender.doubleValue
     customAudioDelayTextField.doubleValue = sliderValue
-    redraw(indicator: audioDelaySliderIndicator, constraint: audioDelaySliderConstraint, slider: audioDelaySlider, value: "\(customAudioDelayTextField.stringValue)s")
+    redraw(indicator: audioDelaySliderIndicator, constraint: audioDelaySliderConstraint, slider: audioDelaySlider, value: "\(sliderValue)s")
     if let event = NSApp.currentEvent {
       if event.type == .leftMouseUp {
         player.setAudioDelay(sliderValue)


low-batt:
I'll defer to the other reviewers on this one.

With my eyes I had to take a screenshot and magnify it to be able to see that there are tick marks. The marks seem to be missing at the edges?:
slider


The `0` labels don't line up (under Ventura). That can be addressed in some other PR.

Note that the audio delay is not just about miss-timed audio tracks. It is also needed when using AirPlay. Users report needing to use 2 second delays.

svobs:
> With my eyes I had to take a screenshot and magnify it to be able to see that there are tick marks. The marks seem to be missing at the edges?:

This seems to be the intent of the designers... if you move the slider knob, you'll see there's another missing tick in the very center. It looks like it's purely aesthetic and the slider clicks along as though the missing ticks are there. I did experiment with different sizes and found that the slider looks best when (pseudocode) `slider.frame.width == tickMarkCount * Int(N) + 4`. In the current PR code, N=2. Changing the slider's width by just a couple pixels in either direction resulted in spaces getting distributed unevenly between the ticks, which looked much worse.

> Note that the audio delay is not just about miss-timed audio tracks. It is also needed when using AirPlay. Users report needing to use 2 second delays.

Ah, yes. Well, 3 sec either direction should be still be enough for anyone then. 😎

> The 0 labels don't line up (under Ventura). That can be addressed in some other PR.

I actually spent a large amount of time on this and could not find a real solution. I think it's a great example of why AppKit UI development is so infuriating. Since it only is noticeable for the 0 value, it might be good enough to add a special constraint to align the centers of the two labels, and use code to add / remove the constraint depending on if the slider knob is at the 0 value...

uiryuu:
Tested against the latest develop branch, here are some of my thoughts.

- The slider is not visually good. Apart from the seemingly missing ticks, it's also too different from other sliders in the quick settings panel. Users will potentially have a hard time understanding it's just like other sliders with finer steps.
- In my opinion, we can add a (+, -) button somewhere around the slider, with a dropdown menu to adjust the step. And to maintain the visual consistency, we can do the same thing for every sliders in the quick settings.
- Maybe we could reconsider whether we should constraint the value can be only land on the ticks when dragged by mouse? We can also adjust `altIncrementValue` to fine tune its behavior.

uiryuu:
A quick mockup:

image
image

or

image

svobs:
Did not mean to close this issue! Inadvertently did a force-push which resulted in 0 commits...

Still working on this one. Putting in Draft status for now.

svobs:
Interesting behavior by GitHub... everything should be ok now, hopefully.

Sorry for being very late getting back to this one.

Submitting for your approval now: something different, and simpler.

The new commit keeps the existing tick marks, and only makes changes to `audioDelayChangedAction`. It removes the built-in snapping behavior and uses custom logic so that when dragging the slider, it snaps to the nearest 50ms.

This keeps things looking nice while preserving more fine-grained dragging. And with my trackpad (not sure about mouse?) I found I was able to easily drag to any multiple of 50ms within the [-5, 5] range. So I didn't see the need to reduce the range to [-3, 3].

@uiryuu I definitely like the idea of having (+, -) buttons and "Step:" drop-down(s) for all the sliders. But I stumbled onto this solution and it worked better than I expected, so I am lazy to do more. I actually have been wanting to have (+, -) buttons for the video equalizers in particular.

uiryuu (approved):
Tried out the new commit, works great and also visually good. Maybe later I'll add a bunch of +,- buttons all together

Issue #3331, opened 2021-02-22
Progress bar does not reach end when i finish playing a video


qtiuto:
**System and IINA version:**

- macOS 11.2.1
- IINA 1.1.2 Build 128

When I play a mp4 to the end, the progress bar does not reach end with one frame gap. It 's obvious with pretty short video

low-batt:
### Reproducing

As qtiuto stated the issue is obvious in short clips. I created the following file to reproduce the issue:
https://user-images.githubusercontent.com/86170219/141697007-df47cb57-c430-47f6-b073-63fa1448da0b.mp4

When this file is played not only does the thumb on the progress bar stop before it reaches the end as described by qtiuto, with the onscreen controller time labels configured to show milliseconds they too indicate play stopped short of the end of the file:
![issue-3331-iina-screen-shot](https://user-images.githubusercontent.com/86170219/141696991-e89d7401-35f9-4bf0-b5da-c2bc6f7f2630.jpg)

### Analysis

The values `videoDuration` and `videoPosition` in `PlaybackInfo` are used to determine the location of the thumb on the slider. The values of these two properties originate from the mpv properties `duration` and `time-pos`. The following debug logging was added to `MPVController.handlePropertyChange` to check the state of mpv properties when playback stops:

```swift
case MPVOption.PlaybackControl.pause:
  if let paused = UnsafePointer<Bool>(OpaquePointer(property.data))?.pointee {
    Logger.log("MPVOption.PlaybackControl.pause \(paused)")
    Logger.log("mpv property duration \(getDouble(MPVProperty.duration))")
    Logger.log("mpv property eof-reached \(getFlag(MPVProperty.eofReached))")
    Logger.log("mpv property estimated-frame-count \(getDouble(MPVProperty.estimatedFrameCount))")
    Logger.log("mpv property estimated-frame-number \(getDouble(MPVProperty.estimatedFrameNumber))")
    Logger.log("mpv property file-size \(getDouble(MPVProperty.fileSize))")
    Logger.log("mpv property percent-pos \(getDouble(MPVProperty.percentPos))")
    Logger.log("mpv property playback-time \(getDouble(MPVProperty.playbackTime))")
    Logger.log("mpv property playtime-remaining \(getDouble(MPVProperty.playtimeRemaining))")
    Logger.log("mpv property stream-end \(getDouble(MPVProperty.streamEnd))")
    Logger.log("mpv property stream-pos \(getDouble(MPVProperty.streamPos))")
[...]

PR #5256, opened 2024-11-07
Fix does not update progress bar with .mkv videos, #5254

### LINKED ISSUE #5254: IINA does not update progress bar properly with .mkv videos
### IINA Version

1.3.5

### Release Channel

Nightly

### macOS Version

15.2

### Steps to Reproduce

- Play .mkv file using IINA
- Pause at some point several minutes into video playback (note the time shown at the time of pausing on the progress bar and OSD)
- Unpause (the progress bar should jump to catch up to where the actual progress of video playback should be in the video)
- The problem is consistent when the OSC is hidden during pausing

I do not have any problem with video playback via YouTube in IINA. I have not tried other forms of video containers to test if this issue is exclusively an .mkv file format issue. I have consistently replicated this issue with other .mkv videos other than the one I have listed. I also do not have this issue in the release version of 1.3.5 of IINA which makes me think that some update in the Nightly has broken this feature

### Expected Behavior

Progress bar should be accurate at all points when pausing and unpausing

### Actual Behavior

Progress bar does not accurately track the progress of video playback and requires pausing and unpausing to show proper progress

_No response_

### Logs

_No response_

### Sample Files

https://drive.google.com/file/d/14NMC3wYzpVcXVxZrtsSLUyvujHgzFT0K/view?usp=sharing 

Fairly consistently over longer periods of playback

### Required Actions

### Optional Actions

low-batt:
This commit will change `PlayerCore.refreshSyncUITimer` to call `syncUITime` if the timer was not running even if the timer is not being restarted. This is needed to update the playback time when showing the OSD or OSC when playback is paused.

This regression also reintroduced issue #3331, "Progress bar does not reach end when i finish playing a video".

- [x] This implements/fixes issue #5254.

---

**Description:**

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -2251,14 +2251,14 @@ class PlayerCore: NSObject {
       }
     }
 
-    guard useTimer else { return }
-
-    // Timer will start
-
     if !wasTimerRunning {
       // Do not wait for first redraw
       syncUITime()
     }
+
+    guard useTimer else { return }
+
+    // Timer will start
 
     syncUITimer = Timer.scheduledTimer(
       timeInterval: timeInterval,
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -2251,8 +2251,10 @@ class PlayerCore: NSObject {
       }
     }
 
+    // When fadeable views are hidden the time can get out of sync. This method will be called when
+    // the view becomes visible to sync the time. If the timer was not running the view must be
+    // updated now. Playback may be paused. If that is the case then the timer will not be started.
     if !wasTimerRunning {
-      // Do not wait for first redraw
       syncUITime()
     }
 


low-batt:
@svobs Does this fix look correct to you?

uiryuu (approved):

svobs:
> @svobs Does this fix look correct to you?

Yeah this looks correct. The call `refreshSyncUITimer()` was intended to be robust enough to be used as a substitute for `syncUITime()`. After scrutinizing the code, it was relying on an assumption that the UI would be perfectly synced when playback stopped. But as you noted above there are a few instances where that cannot be assumed. Seems like best practice to ensure all "fadeable views" are synced when `showUI` is called. And with this fix, that will happen via `showUI`'s call to `refreshSyncUITimer`. Thanks for taking care of this @low-batt.

svobs:
P.S. ~ Might be good also to add a quick comment like I stated above instead of leaving as `// Do not wait for first redraw`. Something like "A call to `refreshSyncUITimer` be expected to ensure `syncUITime` is called at least once, even if the timer is not running and does not need to be started. Thus, any use of `syncUITime()` can simply be replaced by a call to `refreshSyncUITimer()`, removing any need to decide which of the two methods to call."

low-batt:
Thanks for reviewing! I've updated the comment on that if statement.

low-batt:
@svobs I think there is still an issue with UI syncing. I noticed the play/pause button was out of sync with playback after the OSC appeared from hiding. It indicated playback was playing when it was paused.

svobs:
> @svobs I think there is still an issue with UI syncing. I noticed the play/pause button was out of sync with playback after the OSC appeared from hiding. It indicated playback was playing when it was paused.

@low-batt There are big holes in the play button's state due to it being a toggle button. See my recent updates to #4568.

It's also unclear if some versions of MacOS, or some features like App Nap, can cause the UI to drop updates under certain conditions. It's my belief that best practices should include an explicit sync of *all* a window's UI when `windowDidBecomeKey` is called, or in response to some similar mechanism which signals that the window has become active. There are too many unknowns and possibilities for state leaking.

low-batt:
@svobs Really glad you pointed out #4568. I took a quick look and agree the button should be updated in response to the mpv event.

The two issues that still need to be fixed for B1 are #5264, which is the pause/play problem and issue #5263. And I'm trying to finish up a change to `MPVController` logging that I hope to sneak into B1.

The issue associated with PR #4568, #3785 is on the Beta 2 schedule. At this point we are sort of in a freeze right before releasing Beta 1. I know that we have called a halt to merging PRs that introduce new localizations.

Could you prepare a fix for #5264 that just contains the fixes to the play/pause button and not the new functionality/setting?

svobs:
> The issue associated with PR https://github.com/iina/iina/pull/4568, https://github.com/iina/iina/issues/3785 is on the Beta 2 schedule. At this point we are sort of in a freeze right before releasing Beta 1. I know that we have called a halt to merging PRs that introduce new localizations.

Great to hear. Would be great to have more transparency with the roadmap.

> The two issues that still need to be fixed for B1 are https://github.com/iina/iina/issues/5264, which is the pause/play problem and issue https://github.com/iina/iina/issues/5263. And I'm trying to finish up a change to MPVController logging that I hope to sneak into B1.

Issue #5263 sounds worrisome. How is that going? I tried but was unable to reproduce it, although maybe I didn't download the file in the right format.

> Could you prepare a fix for https://github.com/iina/iina/issues/5264 that just contains the fixes to the play/pause button and not the new functionality/setting?

I will do so.

low-batt:
My own worry with providing visibility into scheduling is that it will be a source of heartache for users hoping to see a particular feature added. The goals for 1.4.0 has changed drastically from the original plan. The main goal of enabling the plugin system remains, but the planned merging of quite a few of the outstanding PRs has been dropped and pushed off to a future release. Because mpv added new controls for secondary subtitles, for which IINA has multiple requests, focus turned to supporting those new controls and subtitles in general. The issue that pointed out differences between IINA and mpv when it comes to color introduced a new priority. And we had a bunch of issues with regressions due to the upgrades of the dependent libraries.

Currently there is a lot on the Beta 2 schedule. That needs to be reviewed and drastically reduced as given all the improvements in Beta 1 it is desirable to get 1.4.0 finished up and released. Changes considered risky or extensive will definitely get pushed out to the next release to minimize the chance of introducing regressions in B2.

Yes, issue #5263 is very worrisome. I've been distracted by all the issues that have been pouring in. On #5263 I've been trying to code up a script that reproduces the problem. For some reason I am now having trouble reproducing that one. I need to find a good bit of time to be able to sink into that one and come up with a reliable way of reproducing it.

I will keep an eye out for a PR for #5264.

PR #5259, opened 2024-11-08
Fix setting initial window size causes crash at startup, #5250

### LINKED ISSUE #5250: Setting initial window size causes crash at startup
### IINA Version

latest

### Release Channel

Not Sure

### macOS Version

latest

### Steps to Reproduce

I set initial window size to 100% by percent and now I can't launch the app, getting error message

### Expected Behavior

it crashes after applying the setting restarting the app

### Actual Behavior

...

_No response_

### Logs

_No response_

### Sample Files

_No response_

I can't launch the app now!

### Required Actions

### Optional Actions

low-batt:
This commit will:
- Add an `enableAllMenus` method to `MenuController`
- Add a `disableMenus` parameter to `Utility.showAlert`
- Change `Utility.showAlert` to disable menus while the alert is being shown if caller set `disableMenus` to `true`
- Change calls to `Utility.showAlert` in `MPVController` methods used during startup to use the new `disableMenus` parameter

The `MPVController` will show alerts when mpv reports errors. This can happen while `PlayerCore.first` is being created. Disabling menus prevents the `MenuController` from accessing `PlayerCore.first` while it is in the process of being created which was the source of the crash reported in the issue.

- [x] This implements/fixes issue #5250.

---

**Description:**

### DIFF
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -530,7 +530,7 @@ class MPVController: NSObject {
       setOptionString("config", "yes", level: .verbose)
       let status = setOptionString(MPVOption.ProgramBehavior.configDir, userConfDir)
       if status < 0 {
-        Utility.showAlert("extra_option.config_folder", arguments: [userConfDir])
+        Utility.showAlert("extra_option.config_folder", arguments: [userConfDir], disableMenus: true)
       }
     }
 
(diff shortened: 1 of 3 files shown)


low-batt:
Reviewers, is there a better way to handle this?

This PR only addresses the crash. More fixes are needed to fully address the issue. The settings panel needs to prevent the IINA setting from being set to a value that mpv will reject. Also the alert message needs to be improved. This is a quick PR to get the crash fixed for beta 1. The remaining problems can be fixed in beta 2.

svobs:
@low-batt I'm late to comment (and maybe you won't see this) but I have found the blocking nature of `Utility.showAlert` to be a major liability, especially at startup. But in most of these cases, the worst can be avoided by just wrapping it in a `DispatchQueue.main.async` block.

low-batt:
I think dispatching would have worked in this case. In some cases I think there is a downside to dispatching. If it is due to an internal error you really want to report that right when it happens. If the app is allowed to proceed it might crash before the queue'd alert gets a chance to be displayed and you have lost a clue to the problem.

I went with disabling the menus because I saw that the `Quit` menu item was disabled when the alert was displayed. So I made the IINA menus match that behavior. I was really unsure on this fix. I wasn't able to find anyone on the net talking about this scenario and how to address it.

PR #5278, opened 2024-11-21
Refactor playButton so its image is always set from player.state (fixes #5264)

### LINKED ISSUE #5264: Pause/Play Controls for IINA Nightly do not work after updating Nightly build
### IINA Version

1.3.5

### Release Channel

Nightly

### macOS Version

15.2

### Steps to Reproduce

- Open a video in IINA
- Close IINA
- Upgrade/Change Version of IINA being used from Nightly
- Open IINA and click resume for the video you were playing previously in IINA

Was also able to make this occur when enabling additional MPV options in IINA settings, so it seems to be tied to video playback while changing something in IINA. The issue is fixed as well if you click somewhere else on the progress bar compared to where video playback originally continues from.

### Expected Behavior

IINA pause/play control playback of previously in progress video works normally even between/after updates

### Actual Behavior

IINA pause/play controls during playback of video do not work at all both with keybinds and in player controls upon resuming previously in progress video

_No response_

### Logs

_No response_

### Sample Files

https://drive.google.com/file/d/14NMC3wYzpVcXVxZrtsSLUyvujHgzFT0K/view?usp=sharing

I was able to also replicate this with .mp4 files I had on hand as well with the same results

Every time

### Required Actions

### Optional Actions

svobs:
- [x] This implements/fixes issue #5264.

---

**Description:**

- Changes `playButton` in both main & mini player windows from type "toggle" to "push-in", so they do not automatically change their icon when clicked.
- Change `updatePlayButtonState` to change the play button's image directly. It is now the only way the image will change.
- Remove call to `updatePlayButtonState` from `arrowButtonAction` to avoid possible race.

This was extracted from PR #4568 at @low-batt's request.

IMHO, `muteButton` & `titlebarOnTopButton` should also be changed to ensure consistent behavior, but those are much less prominent.

### DIFF
--- a/iina/Base.lproj/MainWindowController.xib
+++ b/iina/Base.lproj/MainWindowController.xib
@@ -561,7 +561,7 @@
                 <button translatesAutoresizingMaskIntoConstraints="NO" id="gxw-pJ-Lcg">
                     <rect key="frame" x="62" y="0.0" width="24" height="24"/>
                     <buttonCell key="cell" type="square" bezelStyle="shadowlessSquare" image="play" imagePosition="only" alignment="center" alternateImage="pause" refusesFirstResponder="YES" state="on" imageScaling="proportionallyUpOrDown" inset="2" id="mWQ-R0-GGr">
-                        <behavior key="behavior" pushIn="YES" changeContents="YES" lightByContents="YES"/>
+                        <behavior key="behavior" pushIn="YES" lightByBackground="YES" lightByGray="YES"/>
                         <font key="font" metaFont="system"/>
                     </buttonCell>
                     <constraints>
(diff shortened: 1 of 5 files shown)


low-batt (changes requested):
Thanks for creating the separate PR.

Changes look good to me. I pulled the PR, built IINA and tested. I was not able to generate a failure, but I have found this problem hard to reproduce.

Please rebase with `develop` to pickup the fix in PR #5279 so the CI build will work. The CI configuration needed to be updated because IINA has to be built with Xcode 16.1 or the build will fail.

low-batt (approved):
Changes look good to me.

Issue #5246, opened 2024-11-02
Only audio, no video streaming from YouTube


low-batt:
### IINA Version

1.3.5

### Release Channel

Release

### macOS Version

13.6.9

### Steps to Reproduce

- Start IINA
- Click on `Open URL…` under the `File` menu
- Open the YouTube video [The World in HDR in 4K (ULTRA HD)](https://www.youtube.com/watch?v=tO01J-M3g0U)
- IINA starts playing audio, but the window is black

### Expected Behavior

The video plays fine as it did the last time I tried playing it with IINA .1.3.5.

### Actual Behavior

IINA starts playing audio, but the window is black. The IINA log shows FFmpeg failed:
```text
20:13:15.787 [mpv0][e] [ffmpeg/demuxer] error: mov,mp4,m4a,3gp,3g2,mj2: error reading header
20:13:15.793 [mpv0][e] [lavf] error: avformat_open_input() failed
```

The inspector shows IINA does not have a video stream to play:
![inspector](https://github.com/user-attachments/assets/eeadbb37-0165-4e12-8c11-fcbd9ee314f3)

This was with the latest version of [yt-dlp](https://github.com/yt-dlp/yt-dlp):
```text
low-batt@gag ~$ yt-dlp --version
2024.10.22
```

### Logs

[iina.log](https://github.com/user-attachments/files/17605035/iina.log)
[mpv.log](https://github.com/user-attachments/files/17605036/mpv.log)

### Sample Files

Stream the YouTube file [The World in HDR in 4K (ULTRA HD)](https://www.youtube.com/watch?v=tO01J-M3g0U)

### How often can you reproduce the issue

Every time.

### Required Actions

- [X] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [ ] I tested with the Nightly build and confirmed the issue is not fixed.
- [X] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

low-batt:
Steaming the YouTube video [The World in HDR in 4K (ULTRA HD)](https://www.youtube.com/watch?v=tO01J-M3g0U) works fine with the current development version of IINA. It appears YouTube has changed something about its streaming format that FFmpeg 6.0 is unable to parse. The development version of IINA is using FFmpeg 7.0.2 and it does not report the failure seen with IINA 1.3.5.

I've entered this issue to document the problem.

susl:
[...]

PR #5344, opened 2025-01-18
Fix play/pause button stops responding, #5343

### LINKED ISSUE #5343: Play/pause button stops working
### IINA Version

1.3.5 + PR #5342

### Release Channel

Other

### macOS Version

macOS 15.2

### Steps to Reproduce

- Build IINA with the fix in PR #5342
- Open IINA with the 0.9.2 release of the [Online Media](https://github.com/iina/plugin-online-media) plugin installed and enabled
- Open the YouTube video [A Rare Winter Storm Is Coming...](https://www.youtube.com/watch?v=zeNUoh5kwg4)
- While the video is playing click on the `Plugin` menu
- Under the `Audio Quality` submenu click on a different audio stream
- Wait for playback to resume
- Move the mouse to show the OSC
- Click on the pause button
- Nothing happens, playback continues

### Expected Behavior

The state of playback responds to the play/pause button.

### Actual Behavior

Pressing the pause button fails to stop playback.

The log shows clicking the pause button is setting the mpv `pause` property to `false` instead of `true` and mpv is reporting IINA is failing to call `mpv_render_report_swap`:
```text
[  44.514][v][cplayer] Set property: pause=false -> 1
[  44.766][v][vo/libmpv] mpv_render_report_swap() not being called.
```

_No response_

### Logs

[iina.log](https://github.com/user-attachments/files/18462828/iina.log)

[mpv.log](https://github.com/user-attachments/files/18462829/mpv.log)

### Sample Files

Stream the YouTube video [A Rare Winter Storm Is Coming...](https://www.youtube.com/watch?v=zeNUoh5kwg4)

Most of the time.

### Required Actions

### Optional Actions

low-batt:
This commit will change `PlayerCore.fileLoaded` to check to see if playback is in progress or is paused and then set the player state accordingly and synchronize the play/pause button.

- [x] This implements/fixes issue #5343.

---

**Description:**

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -1867,7 +1867,15 @@ class PlayerCore: NSObject {
   func fileLoaded() {
     guard info.state.active else { return }
     log("File loaded")
-    info.state = .paused
+
+    // Normally playback will be paused at this point because PlayerCore sets playback to be paused
+    // before loading a file. This is required to be able to support the setting to pause when media
+    // is opened, otherwise there would be a race condition as to whether IINA can pause playback
+    // before mpv has starting playing the media. However plugins have direct access to mpv and can
+    // load files, therefore the mpv state must be checked.
+    info.state = mpv.getFlag(MPVOption.PlaybackControl.pause) ? .paused : .playing
+    syncUI(.playButton)
+
     // Get video size and set the initial window size
     let width = mpv.getInt(MPVProperty.width)
     let height = mpv.getInt(MPVProperty.height)
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -1871,10 +1871,11 @@ class PlayerCore: NSObject {
     // Normally playback will be paused at this point because PlayerCore sets playback to be paused
     // before loading a file. This is required to be able to support the setting to pause when media
     // is opened, otherwise there would be a race condition as to whether IINA can pause playback
-    // before mpv has starting playing the media. However plugins have direct access to mpv and can
-    // load files, therefore the mpv state must be checked.
-    info.state = mpv.getFlag(MPVOption.PlaybackControl.pause) ? .paused : .playing
+    // before mpv has started playing the media. However plugins have direct access to mpv and can
+    // load files. Ensure mpv is paused so IINA and mpv are in sync on the state of playback.
+    info.state = .paused
     syncUI(.playButton)
+    mpv.setFlag(MPVOption.PlaybackControl.pause, true, level: .verbose)
 
     // Get video size and set the initial window size
     let width = mpv.getInt(MPVProperty.width)
--- a/iina/VideoView.swift
+++ b/iina/VideoView.swift
@@ -227,11 +227,13 @@ class VideoView: NSView {
     checkResult(CVDisplayLinkSetOutputCallback(link, displayLinkCallback, mutableRawPointerOf(obj: self)),
                 "CVDisplayLinkSetOutputCallback")
     checkResult(CVDisplayLinkStart(link), "CVDisplayLinkStart")
+    log("Display link started", level: .verbose)
   }
 
   @objc func stopDisplayLink() {
     guard let link = link, CVDisplayLinkIsRunning(link) else { return }
     checkResult(CVDisplayLinkStop(link), "CVDisplayLinkStop")
+    log("Display link stopped", level: .verbose)
   }
 
   // This should only be called if the window has changed displays


low-batt:
I'm a little worried about plugins being able to by-pass `PlayerCore` and access mpv directly. Much of `PlayerCore` keys off mpv events so most of the time that should not be a problem. But `PlayerCore` does take some actions outside of event processing that won't be done when a plugin accesses mpv directly.

uiryuu (approved):
I came up with the same solution yesterday, but this still cannot fix the issue that the video stops playing after resuming for me; the play button can only control the audio playback. In my case, mpv stops calling the `mpvUpdateCallback()` function thus no new frames are being generated on the screen.

If this fix works for you then it's still good to be merged.

low-batt:
That means this fix is no good and should not be merged.

Busy day today, but I will try and figure out exactly how that happens and update the PR. Sounds like the display link is shutdown. That is mainly controlled by the `PlayerCore` code that handles changes to the mpv `pause` flag.

In general allowing plugins direct access to mpv seems somewhat problematic as it is `PlayerCore` that coordinates changes to IINA that are required based on changes to mpv. Although a lot of that happens in response to mpv events, some of that happens before certain commands are sent to mpv.

I'm starting to think we need to intercept the commands being sent to mpv by the plugins and send them through `PlayerCore` so that IINA state can be adjusted when needed. Hopefully the plugins are running on the main thread when they send the commands to mpv as IINA state changes are coordinated using the main thread in order to avoid the need for a lot of locking.

low-batt:
I tried hard but could not reproduce the problem. I had to take a stab in the dark. See if the updated fix works.

uiryuu:
Things become complicated when I tried to investigate this issue more...

I applied the latest commit, and the problem still exists. The logs show that the display link is started when the problem happened. Even if I commented out all code in `displayIdle()` to prevent `stopDisplayLink()` being called, the issue still happens.

During testing, I noticed a more worrying problem. When using the online media plugin to play a youtube video (I used the video you provided: [A Rare Winter Storm Is Coming...](https://www.youtube.com/watch?v=zeNUoh5kwg4)), sometimes the audio starts but the "OpenURLWindowController" doesn't disappear and the main window doesn't appear either. This happens around 50% or more of the time. I logged all the mpv events and our internal states during opening the online media and found that some times mpv doesn't report "MPV_EVENT_VIDEO_RECONFIG", on which we relied to show up the main window and close the open url window. mpv is supposed to know the video size after video reconfig happens. When audio had started but video reoconfig event is not emitted, I logged "MPVProperty.dwidth" and it shows 0. Still not sure whether its a yt-dlp bug or mpv internal bug...

So then I got back to the original issue I described, I found that after adjusting the audio quality, video reconfig event is also not being emitted. So I guess the root cause of the problems I encountered should be the same: under certain scenarios, for some reason mpv cannot the the information about the video stream, therefore no video reconfig events. Hence I still suggest merge this PR first, since the issue I encountered is most likely another completely different issue.

low-batt:
Apparently YouTube has recently strengthened its defenses against bots, possibly due to an increase in AI training bots. This can result in [403 Forbidden](https://en.wikipedia.org/wiki/HTTP_403) failures and other problems. I believe this is an ongoing problem. Behavior can differ depending upon the region you are in and whether YouTube decides your IP address represents a bot. The most recent [yt-dlp](https://github.com/yt-dlp/yt-dlp) version is working a little better for me:
```text
low-batt@urp iina (play-pause $%=)$ yt-dlp --version
2025.01.15
```

What version are you using?

Issue #5246 shows a similar failure when an older version of yt-dlp was used causing only audio to be played. Have a look at the IINA and mpv log files attached to that issue. At least in that case there is a lot of logging of the interaction with yt-dlp in the mpv log file. In the IINA log file we can see that FFmpeg was unable to open a stream:
```text
20:13:15.787 [mpv0][e] [ffmpeg/demuxer] error: mov,mp4,m4a,3gp,3g2,mj2: error reading header
20:13:15.793 [mpv0][e] [lavf] error: avformat_open_input() failed
```

What can you tell from the IINA log file? Probably the Online Media plugin needs to be enhanced to log the interaction with yt-dlp similar to how mpv logs in detail what is happening.

I agree this seems like a separate issue. I'm thinking we probably don't need to hold up the release for this, but given the current situation with yt-dlp we should improve the logging in the Online Media plugin expecting we may face trouble in this area.

I will go ahead and merge this PR in its current form.

uiryuu:
I'm using the latest yt-dlp (managed by homebrew). I didn't see any error message from FFmpeg or mpv in the logs as in #5246. I tried to lower the video resolution in the preferences to "Video not larger than 720p", and that mitigates the problem. Not only could I open the video faster (~20s to ~5s), but also I didn't experience the issues I encountered before (both no video when opening a link and video playback is not restarted after adjusting audio quality). Admittedly I do have a limited network bandwidth (100M VDSL) but with that I can watch youtube video without problems. I'll keep an eye on future yt-dlp releases and see whether or not this issue is solved.

PR #5348, opened 2025-01-20
More mappings for click and scroll actions via preferences

evnoj:
- [ x ] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)

---

**Description:**
This adds more actions that click and scroll actions can be set to in preferences. The new click actions are the setting/clearing of A-B loop points (like the "A-B loop" action that keybinds can be mapped to), and resetting the playback speed to 1x. The new scroll action is to adjust the playback speed, with configurable sensitivity.

I added this feature because I am frequently using my macbook to learn a physical skill or musical routine by watching tutorial videos, and I am often looping sections or slowing them down. I wanted to be able to access these controls via the trackpad, since it is closer to me, which matters when I don't have my hands on the keyboard, but am reaching towards the macbook from a distance away, or when I am using a magic trackpad and the keyboard is too far away to reach without moving. The fast and easy fine-grained control over playback speed afforded by scroll control is also very nice. I've been using this feature for several months, and use it very frequently.

Screenshot 2025-01-20 at 12 31 58 PM

Possible changes:

- I'm not sure if it's really necessary to have configurable sensitivity for the playback speed adjustment, It may be fine to have it fixed. I've found that multiplying the scroll delta by `0.005` is what works best for me, this is the 2nd-highest value in the sensitivity slider (with the mappings that I picked). If we don't want this configurable sensitivity, `45e8a15c` can just be dropped.

- The playback speed limits that scrolling can reach are hardcoded to a minimum of 0.05x and a maximum of 4x. I don't think it's necessary to add a preference that controls these limits, but we may want to adjust them.

### DIFF
--- a/iina/AppData.swift
+++ b/iina/AppData.swift
@@ -39,6 +39,7 @@ struct AppData {
   static let seekAmountMap = [0, 0.05, 0.1, 0.25, 0.5]
   static let seekAmountMapMouse = [0, 0.5, 1, 2, 4]
   static let volumeMap = [0, 0.25, 0.5, 0.75, 1]
+  static let playbackSpeedMap = [0, 0.001, 0.002, 0.005, 0.01]
 
   static let encodings = CharEncoding.list
 
--- a/iina/Base.lproj/PrefControlViewController.xib
+++ b/iina/Base.lproj/PrefControlViewController.xib
@@ -133,6 +133,8 @@
                                 </menuItem>
                                 <menuItem title="Pause / Resume" tag="2" id="Dm7-JG-fLd"/>
                                 <menuItem title="Toggle Picture-in-Picture" tag="4" id="rvS-xA-7bv"/>
+                                <menuItem title="Set/clear A-B loop points" tag="5" id="GHK-4G-DAj"/>
+                                <menuItem title="Reset playback speed" tag="6" id="Zbp-ok-mMJ"/>
                                 <menuItem title="None" id="wkV-Gq-B0T"/>
                             </items>
                         </menu>
@@ -163,6 +165,8 @@
                                 </menuItem>
                                 <menuItem title="Pause / Resume" tag="2" id="mOX-Fi-oTv"/>
                                 <menuItem title="Toggle Picture-in-Picture" tag="4" id="ibM-4r-SQA"/>
+                                <menuItem title="Set/clear A-B loop points" tag="5" id="65z-Gm-oZu"/>
+                                <menuItem title="Reset playback speed" tag="6" id="ig1-kk-RZu"/>
                                 <menuItem title="None" id="oOk-y6-lek"/>
                             </items>
                         </menu>
@@ -259,6 +263,8 @@
                                     <modifierMask key="keyEquivalentModifierMask"/>
                                 </menuItem>
                                 <menuItem title="Toggle Picture-in-Picture" tag="4" id="3Pm-nn-HpB"/>
+                                <menuItem title="Set/clear A-B loop points" tag="5" id="iLX-jW-Tdw" userLabel="Set/clear A-B loop points"/>
+                                <menuItem title="Reset playback speed" tag="6" id="Vv4-AU-kGj"/>
                                 <menuItem title="None" id="unj-zt-RyU"/>
                             </items>
                         </menu>
--- a/iina/Base.lproj/PrefControlViewController.xib
+++ b/iina/Base.lproj/PrefControlViewController.xib
@@ -23,7 +23,7 @@
         </customView>
         <userDefaultsController representsSharedInstance="YES" id="Yrj-d0-KaY"/>
         <customView id="Wmg-PT-BGO" userLabel="Prefs &gt; Control &gt; Mouse">
-            <rect key="frame" x="0.0" y="0.0" width="480" height="310"/>
+            <rect key="frame" x="0.0" y="0.0" width="480" height="330"/>
             <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMinY="YES"/>
             <subviews>
                 <textField horizontalHuggingPriority="251" verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="zSQ-pa-Urx">
@@ -43,7 +43,7 @@
                     </textFieldCell>
                 </textField>
                 <popUpButton verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="VlR-LI-plv">
-                    <rect key="frame" x="227" y="221" width="227" height="25"/>
+                    <rect key="frame" x="227" y="241" width="227" height="25"/>
                     <popUpButtonCell key="cell" type="push" title="Keyframe seek" bezelStyle="rounded" alignment="left" lineBreakMode="truncatingTail" state="on" borderStyle="borderAndBezel" imageScaling="proportionallyDown" inset="2" selectedItem="wBB-fx-Mao" id="BFn-Bz-29A">
                         <behavior key="behavior" lightByBackground="YES" lightByGray="YES"/>
                         <font key="font" metaFont="menu"/>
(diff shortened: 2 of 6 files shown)


low-batt:
This looks like a nice enhancement. Currently IINA is on the verge of releasing the first beta of the next release, so only some fixes for regressions are currently being merged until we get the beta out. As it turns out some of the changes planned for the second beta have to do with input controls, this kind of enhancement. So this submission is perfectly timed for B2. Today is busy for me so I only had a chance to take a very quick look. I will take a closer look when I have a chance.

One change that is needed is due to localization. IINA's integration with [Crowdin](https://translate.iina.io/) requires that whenever new translations are added to `Base.lproj` file the related file in `en.lproj` must also be updated. This is mentioned in the [CONTRIBUTING](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md#localizations) doc. For this PR that means the new strings added in the XIB need to be added to `en.lproj/PrefControlViewController.strings`. I think this updated file contains the needed changes:

[PrefControlViewController.strings.zip](https://github.com/user-attachments/files/18496851/PrefControlViewController.strings.zip)

Check and make sure you agree and if so, add a commit with that file to the PR.

low-batt:
I had a chance to try out the changes. I'm really liking these enhancements.

On this:
> The playback speed limits that scrolling can reach are hardcoded to a minimum of 0.05x and a maximum of 4x.

I agree, we don't need settings to change these limits, but we should consider using different values for the limits.

The first consideration are the limits imposed by mpv on the value of the mpv [speed](https://mpv.io/manual/stable/#options-speed) option which is what IINA is adjusting. From the [mpv manual](https://mpv.io/manual/stable/): 
> **--speed=<0.01-100>**
> Slow down or speed up playback by the factor given as parameter.
>
> If `--audio-pitch-correction` (on by default) is used, playing with a speed higher than normal automatically inserts the scaletempo2 audio filter.

Whatever we decide, IINA must stay within these limits.

IINA already has hardcoded constants related to speed that can be found here:
https://github.com/iina/iina/blob/d8ce99878aa0ed40e3ba5b33aa44912c9ee800aa/iina/AppData.swift#L23-L28

I'm thinking we probably should match up with the limits used for the playback speed slider in the video side panel. I'm a junior IINA developer. Let's see if the senior developers agree. We probably won't hear from them until the first beta is out, which hopefully will be any day now.

A side note. I found it surprising that IINA has a `mpvMinPlaybackSpeed` property but does not have a  `mpvMaxPlaybackSpeed`. I tested setting the speed and sure enough IINA is not enforcing the 100 limit. I will be entering an issue for that as that is a bug.

low-batt:
On this:
> I'm not sure if it's really necessary to have configurable sensitivity for the playback speed adjustment

I like the ability to configure the sensitivity. I definitely think we should keep that.

evnoj:
> I'm thinking we probably should match up with the limits used for the playback speed slider in the video side panel.

Personally, I'd love for the minimum to be lower than 0.25x. It's not uncommon for me to actually use a playback speed in the 0.05-0.1x range, with a very short loop, as I am closely analyzing very fast motions.

low-batt:
I would be fine with that. Lets see what the senior developers have to say. I defer to them on UI changes.

evnoj:
I added the localization updates as you created them, thank you for that guidance. I have not worked with localization in Xcode before, it seems like you may have used a tool that automatically created the entries in that `.strings` file? Given that the strings are being mapped to arbitrary object IDs (requiring the same string to have multiple entries), instead of being given a label and then using that label in the `.xib` (I'm not positive if this is actually possible).

If you're able to point me towards the tool/resource you used to do this, it would be much appreciated - but no need to take time to write out instructions just to show me.

low-batt:
I'm always happy to explain things. Here is more than you probably want to know…

I am not the expert on IINA localization, @uiryuu is. In fact I'm a little worried as to whether my knowledge is out of date. Brain not working well tonight. I remember we were trying to move away from the current way we do localization to using to the new [string catalogs](https://developer.apple.com/documentation/xcode/localizing-and-varying-text-with-a-string-catalog) feature. I vaguely remember that effort had to be postponed because we found [Crowdin](https://crowdin.com/) did not support string catalogs? @uiryuu may correct me on this.

So what IINA is doing is the old way of doing localization. Of course XIBs are also the old way of doing things. I know there has been some investigation into slowly moving to [Swift UI](https://developer.apple.com/xcode/swiftui/). But there are more pressing matters that need to be addressed first.

To produce that file I used two tools. I first used [ibtool](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/MaintaingYourOwnStringsFiles/MaintaingYourOwnStringsFiles.html#//apple_ref/doc/uid/10000171i-CH19-SW6) to generate a new strings file from the XIB. I directed the output to a temporary file like so:
```text
ibtool iina/Base.lproj/PrefControlViewController.xib --generate-strings-file tmp.strings
```

I have found that this produces extraneous differences with the existing source due to declarations not being in the same order, frequently due to developers not using `ibtool` and just adding new text at the end. Although these extra changes are not harmful I usually make the effort to not include them in a PR by using a merge tool to merge the changes into the actual strings source file and then finish up with hand editing if needed to produce a strings file that only has the expected changes.

For merging I use [Meld](https://meldmerge.org/).  Unfortunately the Meld project does not directly support macOS. I installed it using brew. _However_ I did not have it installed on my second development machine and just this week went to go install it and found the [homebrew formula](https://formulae.brew.sh/cask/meld) is **disabled**. The problem is that the 3<sup>rd</sup> party macOS meld builds come from [yousseb](https://github.com/yousseb/meld/releases) who hasn't generated a new release for a while. I had to grab it from my other machine and then struggle to get past the new Sequoia protections against apps that are not officially signed. Although I recommend and use Meld it is a bummer that it can't be installed right now using brew.

I also frequently use Meld to eliminate the many extraneous changes added by Xcode when I make changes to XIBs.

Once a PR has been merged with changes to Base and en translations then some sort of script is run to push the changes to [IINA's instance of Crowdin](https://translate.iina.io/). Periodically another script is run to pull translations from Crowdin and merge them into the strings files for other languages. @uiryuu knows the details of this.

uiryuu:
Nice addition and very clean PR. I don't find any issue in this PR, changes regarding localization looks fine to me.

Now we have more actions, the menu looks a bit disordered (mixed initial case, space between slash):

image

My suggestion:

Toggle full-screen
Toggle pause (or "Toggle pause/play"? "Toggle pause/resume"?)
Toggle picture-in-picture
Set/clear AP loop points
Reset playback speed
None

Thoughts from native speakers?

uiryuu (approved):

evnoj:
I think that the list as you proposed it works great, with any of those options for "Toggle pause/play" being perfectly acceptable. 

Personally, I like "Toggle play/pause" (most programs with this option put "play" before "pause"), which would make the list:

Toggle fullscreen
Toggle play/pause
Toggle picture-in-picture
Set/clear A-B loop points
Reset playback speed
None

low-batt:
For capitalization and naming rules we should look to Apple's Human Interface Guidelines, examples in Apple's own apps and English grammar rules. And we want the terminology IINA uses to be consistent.

The current `Toggle fullscreen` is incorrect as Apple refers to this using two words "full screen". You can see that under IINA's `Video` menu. Although Apple does use the term "toggle" when referring to certain types of controls, I'm thinking  this should match up with the naming used in the menu, that would be: `Enter/Exit full screen`. There is some debate, but Microsoft and Wordpress use the rule "capitalize the word after the slash if the word before it is capitalized", so I capitalized `Exit`. I think that is easier to read.

If we apply the same argument of matching up with the terminology used in menus then it would be `Enter/Exit picture-in-picture`.

The current `Pause / Resume` is incorrect as when `/` is used to mean "or" no spaces should be used between single words. Changing it to `Toggle play/pause` makes it deviate from the terminology used in the `Playback` menu as well as the messages you see in the on screen display. I think this should be `Pause/Resume` to stay consistent with how it is described in other areas of IINA.

Following the above capitalization rules and shortening it I'm thinking `Set/Clear A-B loop`, without the "points".

As it is just `Reset Speed` in the menus, I'm thinking we can drop "playback" and make it `Reset speed`.

So I'm thinking:
- `Enter/Exit full screen`
- `Pause/Resume`
- `Enter/Exit picture-in-picture`.
- `Set/Clear A-B loop`
- `Reset speed`

As for capitalization of the following words, in the macOS System Settings it seemed like there wasn't one convention. I found some capitalized like in menus and some not. I did not find any recommendations or useful examples in the Apple HIG. In other settings IINA is using lowercase, so this menu should too.

I had to write this in a big rush, hopefully it is somewhat coherent. Tuesday is the day of the week I'm usually busy.

low-batt:
I am not fully comfortable with `Set/Clear A-B loop` as that makes it sound like a toggle, which it isn't. Should it just be `A-B loop`? I'm unsure.

low-batt:
I'm now thinking just `A-B loop` as that is what the menu uses is fine. Basically we are tying a click to the same operation the menu performs and its behavior is implied.

evnoj:
Thanks for the input. I agree that just "A-B loop" is sufficient, I mostly deviated from that to make the intention clearer for the PR, and because it will be in a menu that I suspect more people look at than the keybindings menu, so the action may be exposed to people who are seeing it for the first time. That being said, I think that calling it just "A-B loop" is fine, and people who use it for the first time will have no problem understanding how it works.

uiryuu:
@evannjohnson Thanks for your input. Can you help us to rename the menu items in this PR. That would be so helpful!

evnoj:
I'm happy to do that, but it might take me a few days to get to it. This is what I believe we've landed on:

Enter/Exit full screen
Pause/Resume
Enter/Exit picture-in-picture
A-B loop
Reset speed

low-batt:
Yes, that looks correct.

evnoj:
Apologies this took longer than expected. I've updated those menu item names, and I've updated the English 'title' to match in the other languages. Someone who knows about localization should take a look to ensure that I had the right idea. Assuming that looks good, I think this is ready.

Let me know if there are any questions, or any more work required here.

uiryuu:
Hi @evannjohnson, can you try to revert https://github.com/iina/iina/pull/5348/commits/6955d53b9263f0076cb4a730e564b29f1ebc9706? Updating the comments for all langs are done by Crowdin. Otherwise this PR should be good to go.

evnoj:
Gotcha - I just dropped the commit that touched the other languages. [5f16b28](https://github.com/iina/iina/pull/5348/commits/5f16b2824cb45719bc540f4b4dac1592b6c9a22f) does update the English `PrefControlViewController.strings` (`iina/en.lproj/PrefControlViewController.strings`), is that correct or would that file also be handled by Crowdin?

low-batt:
Correct. Commits must change the English strings along with the base. This requirement is due to the way the integration with Crowdin works. All other localization files are handled by the scripts that interface with Crowdin.

low-batt (approved):
Pulled and tested. Changes look good to me.

@evannjohnson Thank you very much for the hard work on this enhancement.

evnoj:
Thank you for the help with localization!

low-batt:

Issue #4875, opened 2024-04-10
More granular playback speed control


riobard:
**What IINA does currently:** 

IINA playback speed control is awkward to use:

1. In Playback menu you can only speed up by 2x / down by 0.5x each time, which is way too big a jump. 

<img width="237" alt="image" src="https://github.com/iina/iina/assets/22984/7dcafb6d-1ea1-4927-8912-0434219be737">

2. In OSD Quick Settings the playback speed slide uses logarithmic markers, which is not human-friendly.

<img width="337" alt="image" src="https://github.com/iina/iina/assets/22984/78d18164-c061-4900-82c9-3a55b3438033">

**Why you think this should be added:**

Playback speed control should be more natural, human-friendly, and easy to access. Specifically,

1. Playback menu should use speed control with linear ±0.25x each time. With keyboard shortcuts it allows users to rapidly change to desired speed quickly.
2. OSD should add YouTube-like common speed options.
3. OSD Quick Settings speed slide should use human-friendly makers, ideally at linear ±0.25x to match Playback menu.

**Examples of other projects that have something similar:**

YouTube features more natural and granular playback speed options:

<img width="242" alt="image" src="https://github.com/iina/iina/assets/22984/f6223e23-a726-4b50-9fbd-8160c64a2c85">

angelozangari:
Other than the `cmd + [` or `]`  shortcut, you can have finer playback granularity with `cmd + opt + [` or `]`. It's not linear but follows a gradually increasing exponential-like progression, starting from +0.10x.

svobs:
Regarding (1) above: it looks like there is an existing feature to increase or decrease speeds by 10%. It looks like it is meant to be enabled in the menu while the Option key is held down. But for some reason it's not working.

In MainMenu.xib:
<img width="487" alt="MainMenu.xib" src="https://github.com/user-attachments/assets/7dbcb69a-906a-4945-be99-0665261ccc67" />

If I deselect the "Alternate" checkbox for each and then run IINA, they are visible:
<img width="374" alt="Image" src="https://github.com/user-attachments/assets/25b9afdd-9d88-46c7-a1c8-9f5c0dd4f75e" />

svobs:
> Regarding (1) above: it looks like there is an existing feature to increase or decrease speeds by 10%. It looks like it is meant to be enabled in the menu while the Option key is held down. But for some reason it's not working.

Oh no...it's as I suspected. There is a regression affecting alternate menu items. And it's my fault.

[...]

PR #5350, opened 2025-01-21
Fix bug: alternate menu items without key bindings cannot be shown

svobs:

---

**Description:**
Discovered while investigating #4875.

~~Looks like this was introduced by PR #4567.~~ It looks like when the "Alternate" flag is set for a menu item, the modifier key (e.g. `Option`) needs to be set in its `keyEquivalentModifierMask` for it to be activated. But the `updateMenuItem` method, when resetting the menu item's key equivalent, has been clearing all its flags. As a result, "alternate" menu items which do not have an explicit key binding will never appear.

This patch adds a check for the `isAlternate` property, and if present, sets its modifier to `option`:

1. This restores the "alternate" behavior of the menu items if they otherwise are not matched to a user key binding.

2. Did some more testing, and discovered that, if an "alternate" menu item does end up being bound to a user key binding, the behavior depends on the bindings of it & its associated "normal" item:
- If the "alternate" item has the same binding as the "normal" item but adds a modifier key, then the "alternate" item will replace the "normal" item in the menu when the user holds down that modifier key.
- If the "alternate" item & its "normal" item have different non-modifier keys in their binding, both items will be displayed in the menu.

Seems like reasonable behavior to me.

### DIFF
--- a/iina/MenuController.swift
+++ b/iina/MenuController.swift
@@ -932,7 +932,10 @@ class MenuController: NSObject, NSMenuDelegate {
         // This is needed for the case where the menu item previously matched to a key binding, but now there is no match.
         // Obviously this is a little kludgey, but it avoids having to do a big refactor and/or writing a bunch of new code.
         let (_, value, extraData) = sameKeyAction(actions, actions, normalizeLastNum, numRange)
-        updateMenuItem(menuItem, kEqv: "", kMdf: [], l10nKey: l10nKey, value: value, extraData: extraData)
+        // An "alternate" menu item appear is intended to replace a "normal" menu item in the menu if its modifier key is held down
+        // (typically Option). But this key needs to be specified in its modifier flags, or the item may never appear.
+        let modifiers: NSEvent.ModifierFlags = menuItem.isAlternate ? [.option] : []
+        updateMenuItem(menuItem, kEqv: "", kMdf: modifiers, l10nKey: l10nKey, value: value, extraData: extraData)
       }
     }
 


low-batt:
> As a result, some of these "alternate" menu items now appear all the time, and others never appear.

Which alternate menu items appear all the time? I've been trying to compare 1.3.5 to `develop` and I haven't been able to spot the problem.

svobs:
> Which alternate menu items appear all the time? I've been trying to compare 1.3.5 to develop and I haven't been able to spot the problem.

When I wrote that I was describing what I saw as anomalies from standard behavior without having built a full understanding of the problem. I was seeing pairs of menu items which had different keys bound to them, as I described in point 2 of the description. This is actually ok.

I have updated the comment in the code to be more precise.

svobs:
> This is actually ok.

I should say I think this is actually ok. Others may have the opinion that "alternate" items should never be displayed at the same time as their "non-alternate" counterpart. That would be a more complex fix though.

Examples of 2 pairs where both are visible

low-batt:
I'm still lost. How do I reproduce the behavior that is being fixed?

I duplicated the IINA bindings and changed the key binding for `Next Frame` to period and I see them both as shown in the screenshot. But that is also how IINA 1.3.5 displays it. I'm not understanding how to see the regression that is being fixed.

svobs:
The above screenshot doesn't show the regression; it's just highlighting an AppKit quirk when a menu item has `isAlternate` set to true and has a key equivalent set, but its key equivalent does not not complement the key equivalent of the item preceding it in the menu. AppKit decides to show both items.

The regression fixed by this PR causes items to be hidden when they shouldn't be.

First, the control case to show the correct behavior:

1. Open IINA & go to Settings > Key Bindings, and select "IINA Default".
2. Begin playing a video. Open the Playback menu. Confirm that the items `Step Forward 5s` and `Step Backward 5s` are visible.
3. Hold down the `Option` key. Confirm the two items change to `Next Frame` and `Previous Frame`, respectively.

Now, to reproduce the regression:

1. Open IINA & go to Settings > Key Bindings. Create a new configuration which has no key bindings. Leave it selected.
2. Begin playing a video. Open the Playback menu. Confirm that the items `Step Forward 5s` and `Step Backward 5s` are visible as expected.
3. Hold down the `Option` key. See that the items do not change to `Next Frame` and `Previous Frame` as one would expect.
4. Try holding down `Control`, or any other key. See that nothing will result in `Next Frame` and `Previous Frame` becoming visible.
With no key bindings set

svobs:
OK, actually I now dug up some older versions of IINA and found the same behavior. So I'm going to downgrade the label of this from "regression" to simply "bug".

But I'm certain this is indeed a bug, because I didn't have to go far to find other examples of "alternate" menu items which will appear with `Option` held down, despite not having key equivalents. 

Here's the `Window` menu while in Finder. Notice how "Zoom" becomes "Zoom All" when `Option` is held down:

SCR-20250122-ugry
Screenshot 2025-01-22 at 23 31 02

low-batt (approved):
Reproduced problem to understand the issue.
Pulled PR, built IINA and confirmed fixed.
Changes look good to me.

low-batt:
Thanks for explaining it to me. I was comparing 1.3.5 to develop looking for a regression and couldn't find one. With your instructions I was able to reproduce the problem.

An unrelated oddity…
With a blank key binding configuration a few menu items still have shortcut keys assigned. I don't see any pattern as to which ones do and which don't. What is that all about? An inconsistency we should address at some point?

svobs:
> An unrelated oddity…
With a blank key binding configuration a few menu items still have shortcut keys assigned. I don't see any pattern as to which ones do and which don't. What is that all about? An inconsistency we should address at some point?

Many of the menu items in `MainMenu.xib` have key equivalents assigned there. Unless those are changed elsewhere in code, they will keep those assignments. The only place the bindings are changed (not including menu items added at runtime, e.g., custom filters or plugin items) is `MenuController.updateKeyEquivalentsFrom()`, but it doesn't cover all the items.

We have talked about how it would be desirable to make all the menu item bindings configurable. I recall a conversation sometime within the last year which went in depth about it - or maybe exactly a year ago - but I suck at finding things on GitHub. But no action items came out of it. So for now things like "Jump To…" and "Log Viewer" are basically hard-coded.

If you do see a menu item which can be configured via the Key Bindings settings, but still has a key equivalent when an empty configuration is active, that is definitely a bug.

Issue #5334, opened 2025-01-09
Support `t` / `time` parameter in URL scheme


twio142:
**What you want IINA to do:**

Support opening a weblink and jumping to a timestamp with `iina://weblink?url=...&t=600` (600 seconds from start).

**What IINA does currently:**

Not this.

**Why you think this should be added:**

This can be used to create link to a specific position of a video.

Also, when opening IINA from a web page, I could also jump right to the current position and continue watching.

**Examples of other projects that have something similar:**

njzydark:
use `mpv_start`

PR #5370, opened 2025-02-09
Seek to time in url if such query parameter is present

iThinker:
- [ ] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)
- [x] This implements/fixes issue #5334.

---

**Description:**

Parse time parameters from URL query and seek to it of present.
Tested with pasting URL from youtube.
It might be nice to improve browser extension to support adding time somehow. But I dunno how it would get time from playing web video.

### DIFF
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -547,9 +547,11 @@ extension URL {
     return (try? self.resourceValues(forKeys: [.isDirectoryKey]))?.isDirectory ?? false
   }
   
+  static let queryTimeParameterNames = Set(["t", "time"])
+  
   var queryTimeSeconds: TimeInterval? {
     let queryItems = self.urlComponents?.queryItems
-    let timeString = queryItems?.first { Set(["t", "time"]).contains($0.name) }?.value
+    let timeString = queryItems?.first { URL.queryTimeParameterNames.contains($0.name) }?.value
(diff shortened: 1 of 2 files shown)


iThinker:
PTAL @low-batt  @lhc70000 @uiryuu

svobs:
Sounds like a good addition for YT-DL. But seems there is a (probably small) but nonzero probability that this may cause problems with URLs which have "t" or "time" query params in their URLs which are expected to be passed through to the server?

iThinker:
The params are not stripped, so they would be passed to the server, aren't they? Or there is some stripping logic elsewhere?

low-batt:
Finally had some time to scan through pull requests…

Isn't this feature already available using the mpv [start](https://mpv.io/manual/stable/#options-start) option?:
> **--start=\<relative time>**
>    Seek to given time position.
>
>    The general format for times is [+|-][[hh:]mm:]ss[.ms]. If the time is prefixed with -, the time is considered relative from the end of the file (as signaled by the demuxer/the file). A + is usually ignored (but see below).
>    …

With IINA 1.4.1, this link opened the YouTube video starting at one minute into the video:
```text
iina://weblink?url=https://www.youtube.com/watch?v=tO01J-M3g0U&mpv_start=1:00
```

Or am I not understanding the proposed feature?

PR #5382, opened 2025-02-14
Add DSD (.dsf, .dff) files to supported types (issue #5381)

### LINKED ISSUE #5381: MacOS to recommend opening DSD/DSF files with IINA by default
**What you want IINA to do:**
I want IINA to tell the OS to open DSF files with IINA by default, since QuickTimePlayer doesn't.

**What IINA does currently:**

I told my dad to go into the Boston folder on my network storage and play the files in it using IINA. His response to me was that IINA wasn't under the list of apps that could open it by default. It should be, from my testing, it works great for it.

I don't know if any other ones like VLC can play DSFs, I have not tested this, nor have I tested any other programs with DSF. I just tried opening it in IINA because that's what I usually use for music and it just worked.

svobs:
- [x] This implements/fixes issue #5381.

---

**Description:**

Took a crack at descriptions. See issue #5381 for more info.

### DIFF
--- a/iina/Info.plist
+++ b/iina/Info.plist
@@ -1334,6 +1334,40 @@
 				</array>
 			</dict>
 		</dict>
+    <dict>
+      <key>UTTypeConformsTo</key>
+      <array>
+        <string>public.audio</string>
+      </array>
+      <key>UTTypeDescription</key>
+      <string>Direct Stream Digital DSF File</string>
+      <key>UTTypeIdentifier</key>
+      <string>io.iina.dsf</string>
+      <key>UTTypeTagSpecification</key>
+      <dict>
+        <key>public.filename-extension</key>
+        <array>
+          <string>dsf</string>
+        </array>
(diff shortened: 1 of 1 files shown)


iina/Info.plist
```diff
@@ -1334,6 +1334,40 @@
...
 		</dict>
+    <dict>
+      <key>UTTypeConformsTo</key>
+      <array>
+        <string>public.audio</string>
+      </array>
+      <key>UTTypeDescription</key>
+      <string>Direct Stream Digital DSF File</string>
+      <key>UTTypeIdentifier</key>
+      <string>io.iina.dsf</string>
```
low-batt, line 1345: Many of the current entries are incorrect. Per Apple you must not use your own UTI for someone else's file type. We need to try and find an official UTI for this file type.

uiryuu:
Thanks. Let's try make the UITs more correct later.

PR #5480, opened 2025-04-22
Workaround for imprecise Double values for speed provided by mpv (fixes #5472)

### LINKED ISSUE #5472: Playback Speed adds unecessary decimal places (1.700000000000001x)
### IINA Version

1.4.0-beta1 Build 152

### Release Channel

Beta

### macOS Version

Sonoma 14.7.1 (23H222)

### Steps to Reproduce

- Open Video
- Use shortcut key to incrementally increase playback speed to any speed beyond 1.6x

### Expected Behavior

playback speed should increase linearly with the proper increment for each keypress (+0.1x is default)

### Actual Behavior

- Upto 1.6x is normal.
- anything above 1.6x has unecessary decimal places.
- i.e: 1.7x changes to 1.700000000000001x instead
- reducing playback speed down to 1.6x removes the decimal place

Image

_No response_

### Logs

_No response_

### Sample Files

_No response_

every time

### Required Actions

### Optional Actions

svobs:
- [x] This implements/fixes issue #5472.

---

**Description:**

 - Rename Double's `string` property to `groupedStringUpTo6Decimals`.
 - Reduce its `maximumFractionDigits` from 15 to 6, to match standard mpv precision, and to avoid issues caused by small floating point imprecision
 - Remove its use of `maximumSignificantDigits`, which the Apple docs state as unnecessary when `usesSignificantDigits==false`.

### DIFF
--- a/iina/QuickSettingViewController.swift
+++ b/iina/QuickSettingViewController.swift
@@ -228,11 +228,10 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
     switchHorizontalLine2.layer?.opacity = 0.5
 
     // Localize decimal format of numbers
-    speedSlider0_25xLabel.stringValue = "\(0.25.string)x"
-    // Unclear if these need to be localized. Better to be safe?
-    speedSlider1xLabel.stringValue = "\(1.string)x"
-    speedSlider4xLabel.stringValue = "\(4.string)x"
-    speedSlider16xLabel.stringValue = "\(16.string)x"
+    speedSlider0_25xLabel.stringValue = "\(0.25.groupedStringUpTo6Decimals)x"
+    speedSlider1xLabel.stringValue = "1x"
+    speedSlider4xLabel.stringValue = "4x"
+    speedSlider16xLabel.stringValue = "16x"
 
     customSpeedTextField.formatter = speedFormatter
 
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -344,17 +344,13 @@ extension Double {
   /// a comma to signify the decimal):
   /// ```
   /// let num: Double = 12.34
-  /// let badStr = "Value is \(num)"          // badStr will *always* be "Value is 12.34"
-  /// let goodStr = "Value is \(num.string)"  // goodStr will be "Value is 12,34"
+  /// let badStr = "Value is \(num)"                                // badStr will *always* be "Value is 12.34"
+  /// let goodStr = "Value is \(num.groupedStringUpTo6Decimals)"  // goodStr will be "Value is 12,34"
   /// ```
   ///
-  /// Currently the output string is limited to 15 digits after the decimal. This should be more than
-  /// enough for any imaginable use right now, but the limit can and should be increased in the future if
-  /// needed. (It's not clear what the maximum allowed value for `NumberFormatter.maximumFractionDigits`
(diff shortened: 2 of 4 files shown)


svobs:
Pushed an update to change some of the comments I missed, and with slightly improved names, I hope.

iina/QuickSettingViewController.swift
```diff
@@ -228,11 +228,11 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
...
     // Localize decimal format of numbers
-    speedSlider0_25xLabel.stringValue = "\(0.25.string)x"
+    speedSlider0_25xLabel.stringValue = "\(0.25.groupedStringUpTo6Decimals)x"
     // Unclear if these need to be localized. Better to be safe?
-    speedSlider1xLabel.stringValue = "\(1.string)x"
-    speedSlider4xLabel.stringValue = "\(4.string)x"
-    speedSlider16xLabel.stringValue = "\(16.string)x"
+    speedSlider1xLabel.stringValue = "\(1.groupedStringUpTo6Decimals)x"
+    speedSlider4xLabel.stringValue = "\(4.groupedStringUpTo6Decimals)x"
+    speedSlider16xLabel.stringValue = "\(16.groupedStringUpTo6Decimals)x"
```
low-batt: No need for any of these changes to QuickSettingsViewController. These last 3 are integers. We know there isn't a problem converting 0.25 to a string. For these labels I think it is fine to use `.string`. If we did want to use a special format for these labels devs reading this class would probably expect `speedFormatter` to be used:

https://github.com/iina/iina/blob/84cc736666e77fc051e5785e1215eee7bb06f87f/iina/QuickSettingViewController.swift#L18-L32

Looks like there have been similar problems in the past as this class has code to adjust the speed:

https://github.com/iina/iina/blob/84cc736666e77fc051e5785e1215eee7bb06f87f/iina/QuickSettingViewController.swift#L707-L718

That explains why using the quick settings panel clears the problem.

low-batt (changes requested):
I think the changes to `QuickSettingsViewController` should be removed. See my inline comment. Otherwise changes look good.

iina/QuickSettingViewController.swift
svobs: > These last 3 are integers.

I went ahead and changed these back.

> We know there isn't a problem converting 0.25 to a string.

Not entirely true. Recall that decimal numbers need to go through a formatter to ensure they use a comma in certain locales (a big reason for the original PR for speed formatting).

>  If we did want to use a special format for these labels devs reading this class would probably expect speedFormatter to be used

Hmm, I had forgotten about `speedFormatter`. There is definitely an intersection in functionality now between that and `fmtDecimalGroupingMaxFractionDigits6`. Perhaps the two should be merged into a single `libmpvFormatter" or something like that?

low-batt (approved):
Pulled PR, built IINA, confirmed fixed. Changes look good.

iina/QuickSettingViewController.swift
low-batt: Yes, 0.25 needs localization. I was referring to not having a problem with imprecise conversion to double. In other words `.string` works.

There are differences between `speedFormatter` and `fmtDecimalGroupingMaxFractionDigits6`. Probably they could be merged. Not critical for the task at hand which was to get the regression fixed.

Good thing 1.4.0 is going through a beta. I've definitely broken some important stuff.

PR #5483, opened 2025-04-26
Fix no longer remembers screen to launch on, #5450

### LINKED ISSUE #5450: IINA No Longer Remembers Screen to Launch On
### IINA Version

1.4.0-beta1 Build 152

### Release Channel

Release

### macOS Version

15.4 (24E248)

### Steps to Reproduce

In a multi-monitor environment, open IINA, move to secondary monitor, close IINA. Re-open IINA.

### Expected Behavior

IINA window launches on the monitor it was on when it was last closed.

### Actual Behavior

Monitor always launches on primary display. Is this a result of the recent Sequoia update, an update to IINA, or is the software not behaving as intended?

_No response_

### Logs

_No response_

### Sample Files

_No response_

100%

### Required Actions

### Optional Actions

low-batt:
This commit will:
- Move code that reads `MainWindowLastPosition` in the `windowDidLoad` method the `MainWindowController` class to a new method `determineScreenToUse`
- Change `MainWindowController.handleVideoSizeChange` to call `determineScreenToUse` and place the window on the proper screen when the media is initially opened
- Correct `handleVideoSizeChange` to support the `Use physical resolution on Retina displays` setting even when `Resize the window to fit video size` is set to `Disabled`
- Add logging to `handleVideoSizeChange` to be able to see how it determined the position and size of the window

These changes correct multiple regressions in 1.4.0-beta1 in addition to the problem reported in issue #5450. The `Initial window position` and `Initial window size` settings were being ignored when `Resize the window to fit video size` is set to `Disabled`.

The changes also correct a problem in IINA 1.3.5 where the `Use physical resolution on Retina displays` setting is ignored if `Resize the window to fit video size` is set to `Disabled`.

- [x] This implements/fixes issue #5450.

---

**Description:**

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -671,20 +671,10 @@ class MainWindowController: PlayerWindowController {
     window.title = "Window"
 
     // As there have been issues in this area, log details about the screen selection process.
-    NSScreen.log("window!.screen", window.screen)
+    NSScreen.log("window.screen", window.screen)
     NSScreen.log("NSScreen.main", NSScreen.main)
     NSScreen.screens.enumerated().forEach { screen in
       NSScreen.log("NSScreen.screens[\(screen.offset)]" , screen.element)
-    }
-
-    var screen = window.selectDefaultScreen()
-
-    if let rectString = UserDefaults.standard.value(forKey: "MainWindowLastPosition") as? String {
-      let rect = NSRectFromString(rectString)
-      if let lastScreen = NSScreen.screens.first(where: { NSPointInRect(rect.origin, $0.visibleFrame) }) {
-        screen = lastScreen
-        NSScreen.log("MainWindowLastPosition \(rect.origin) matched", screen)
-      }
     }
 
     videoView.videoLayer.draw(forced: true)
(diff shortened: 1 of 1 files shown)


uiryuu:
I'm traveling and don't have another display to test out this PR. Just a few concerns. `determineScreenToUse` is only called in `MainWindowController.handleVideoSizeChange`, but is not called in `MiniPlayerWindowController.handlevideoSizeChange`, so the mini player will not opened in the correct remembered window. So maybe consider call `determineScreenToUse` in the super class (`PlayerWindowController`) and cache the determined size and window so that the subclass can use it.

Another thing that has bugged me for a while is that, IINA always tried to move the window to previous remembered window, and it seems we don't have an option to disable that (or do we?). This is sometimes annoying when you open the initial window on your main monitor, then open the media, the media will be opened on your secondary monitor. I think we should at least provide an option to always open the media in the current (focused) screen. But this can be added later.

low-batt:
I did not add `determineScreenToUse` to `PlayerWindowController` because the mini player has no support for remembering the screen as well as the other related settings. I wanted to keep this PR focused on fixing the regressions. We don't have an option to disable remembering the previous window position. Both that and the missing support in the mini player can be added in the future.

Focus on your trip!

uiryuu (approved):

PR #5497, opened 2025-05-03
Add CFBundleDocumentTypes for folder, volume (fixes #5488)

### LINKED ISSUE #5488: Ability to add IINA to "Open with..." in Path Finder...
In the Path Finder application (replacement for Finder) you can right-click a folder and choose "Open with..." which gives you the possibility to choose applications like "VLC" and "Quick Time Player".  But IINA isn't one of these applications... I have to go via "Open with..." -> "Applications" and find IINA amongst all my applications... every time... 
I asked the Path Finder application team, and they told me the list they pull is one that MacOS returns, so IINA will have to "attach" it self as an application that can open a folder...  I'm not an expert as to how this can be done, I have tried with different software packages with no luck so far...  but maybe it's just something that is easier to do from within IINA ?

svobs:
- [x] This implements/fixes issue #5488.

---

**Description:**

- Copied entries from VLC's `Info.plist`.
- Tested & confirmed that after the app is built with these entries, the `Open with` context menu lists IINA as an option.

### DIFF
--- a/iina/Info.plist
+++ b/iina/Info.plist
@@ -5,6 +5,18 @@
 	<key>CFBundleDocumentTypes</key>
 	<array>
 		<dict>
+			<key>CFBundleTypeName</key>
+			<string>Folder</string>
+			<key>CFBundleTypeRole</key>
+			<string>Viewer</string>
+			<key>LSHandlerRank</key>
+			<string>Default</string>
+			<key>LSItemContentTypes</key>
+			<array>
+				<string>public.folder</string>
+			</array>
+		</dict>
+		<dict>
 			<key>CFBundleTypeExtensions</key>
 			<array>
 				<string>mkv</string>


iina/Info.plist
```diff
@@ -4,6 +4,26 @@
 <dict>
 	<key>CFBundleDocumentTypes</key>
 	<array>
+		<dict>
+			<key>CFBundleTypeName</key>
+			<string>Folder</string>
+			<key>CFBundleTypeRole</key>
+			<string>Viewer</string>
```
low-batt, line 11:
```suggestion
			<string>Viewer</string>
			<key>LSHandlerRank</key>
			<string>Default</string>
```

iina/Info.plist
```diff
@@ -4,6 +4,26 @@
...
+			<key>LSItemContentTypes</key>
+			<array>
+				<string>public.folder</string>
+			</array>
+		</dict>
+		<dict>
+			<key>CFBundleTypeName</key>
+			<string>Volume</string>
+			<key>CFBundleTypeRole</key>
+			<string>Viewer</string>
```
low-batt:
```suggestion
			<string>Viewer</string>
			<key>LSHandlerRank</key>
			<string>Default</string>
```

low-batt (changes requested):
All of our other entires specify [LSHandlerRank](https://developer.apple.com/documentation/bundleresources/information-property-list/cfbundledocumenttypes/lshandlerrank), so I think these should as well to be consistent.

I tested with Path Finder and I was able to open a folder with IINA. **But** I was not able to open a volume with IINA using Path Finder or regular Finder. How do I test that change?

There is a _lot_ wrong with our `Info.plist` entries. I was working on it at one point, but got distracted by more serious problems.

iina/Info.plist
svobs: > I tested with Path Finder and I was able to open a folder with IINA. But I was not able to open a volume with IINA using Path Finder or regular Finder. How do I test that change?

I must confess I was being a bit lazy by copying the entry for `public.volume` from VLC. It looked important, but I didn't really test it.

I now did testing with various volume types, including a USB drive, a cloud drive mounted via MacFUSE, a Samba share, and even a CD. It doesn't look like there is a context menu in Path Finder for any type of volume. According to [this old post](https://stackoverflow.com/questions/4361530/accepting-drops-of-volumes-on-dock-icon), it looks like `public.volume` provided drag & drop support for volumes onto the Dock icon, at least for some versions of MacOS - reading the entire thread, it appears that the mechanism has changed at least once.

I am testing with Sequoia 15.4.1, and I found that IINA already supports drag & drop of volumes onto the Dock icon. So it seems that adding `public.volume` is just redundant. So I will remove it.

low-batt (approved):
Changes look good to me.

iina/Info.plist
low-batt: Sounds good!

PR #5502, opened 2025-05-06
Rename preferences to settings for user-facing strings

### LINKED ISSUE #4038: Rename "Preferences" to "Settings"
"Preferences" has been renamed to "Settings" throughout the system in macOS 13[^1]. If we build against macOS 13 SDK, the in-app menu will display "Settings":

image
[^1]: https://www.wwdcnotes.com/notes/wwdc22/10074/ 

Although we can decide display settings or preferences depending on the system version programmatically, I prefer to just rename to settings regardless the system version.

Another decision to be make is that should we also change the "Preferences" appears in filename/variable names to prevent further confusion (e.g. `PrefGeneralViewController`)? I'm open to this.

uiryuu:
- [x] This implements/fixes issue #4038 .

---

**Description:**
I tried to find all "preference" in the strings file and rename relevant ones to "settings". Strings in en-GB localization are also updated. I didn't tried to rename "preference" in non-user-facing strings (string keys, string comments, etc).

### DIFF
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -141,7 +141,7 @@
 "osd.playlist_loop" = "Loop Playlist";
 "osd.no_loop" = "Disable Looping";
 
-"preference.title" = "Preferences";
+"preference.title" = "Settings";
 "preference.general" = "General";
 "preference.ui" = "UI";
 "preference.video_audio" = "Video/Audio";
@@ -279,7 +279,7 @@
 
 "alert.extra_option.cannot_read" = "Cannot read user defined extra mpv options.";
 "alert.extra_option.empty" = "Key or value cannot be empty!";
-"alert.extra_option.error" = "Error setting option --%@=%@ with return value %d. Please check your extra option settings in Preference > Advanced.";
+"alert.extra_option.error" = "Error setting option --%@=%@ with return value %d. Please check your extra option settings in Settings > Advanced.";
 "alert.extra_option.config_folder" = "Error setting config directory \"%@\".";
 
 "alert.playlist.error_deleting" = "Error deleting: %@";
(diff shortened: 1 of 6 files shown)


low-batt (changes requested):
As IINA adheres to Apple standards we should follow what Apple does in this case and that is that it is only `Settings` when running under macOS Ventura or later. When running under earlier versions of macOS it is still `Preferences`. These strings need to be changed to substitute in the appropriate term for the version of macOS.

Here is one that was not changed:
preferences


I mentioned that to lhc along with some typos in plugin stuff quite a while ago. I think I caught him too busy at the time to deal with such stuff.

Another change along the same lines we might want to consider is maybe renaming `youtube-dl` to `yt-dlp` in the public facing places. Looks like `youtube-dl` is being actively maintained. But they aren't publishing releases and instead ask you to use their nightly builds. As we aren't testing `youtube-dl` I'm thinking our UI should reflect that IINA is expecting `yt-dlp`.

uiryuu:
> As IINA adheres to Apple standards we should follow what Apple does in this case and that is that it is only Settings when running under macOS Ventura or later. When running under earlier versions of macOS it is still Preferences. These strings need to be changed to substitute in the appropriate term for the version of macOS.

I don't think live substitution is really necessary for such a small change.
- This is mostly an one-time change; I don't see we will use such method in the future.
- We can only substitute the term for English (and probably en-GB). We don't know whether or not other language will adopt such a term change as well. This should be decided by translators.
- In this PR, we change to "Settings" in the strings to indicate the term change for the translators. If we don't do this now, we have to do this in the future when we drop support for macOS 12 and below. I think we should introduce this change to translators as soon as possible.

> Here is one that was not changed:

I'd consider the term change from "Preferences" to "Settings" only apply for the system settings panel. The "Preferences" here is for individual plugins. So I think it's okay to not change "Preferences" here? Ping @lhc70000 

> Another change along the same lines we might want to consider is maybe renaming youtube-dl to yt-dlp in the public facing places. Looks like youtube-dl is being actively maintained. But they aren't publishing releases and instead ask you to use their nightly builds. As we aren't testing youtube-dl I'm thinking our UI should reflect that IINA is expecting yt-dlp.

Good point. Will push another commit.

lhc70000:
> So I think it's okay to not change "Preferences" here?

I think we should use Settings to keep it consistent.

low-batt (approved):
Changes look good to me.

PR #5511, opened 2025-05-11
Make left/rightArrowLabel listen to mpv speed

### LINKED ISSUE #4052: OSC speed is not reinitialized when typing on Backspace [mpv Default shortcuts]
**System and IINA version:**

- macOS Monterey 12.6.1
- IINA 1.3.1

**Expected behavior:**
Given mpv Default shortcuts are selected and using "<<", ">>" buttons to increase/decrease speed on screen, when typing Backspace to get back to 1.00 speed, the speed is actualized on the OSC.

**Actual behavior:**
Given mpv Default shortcuts are selected and using "<<", ">>" buttons to increase/decrease speed on screen, when typing Backspace to get back to 1.00 speed, the speed is not actualized on the OSC and remains as previously.

Step 1: clicking on "<<" to reduce speed to 0.50
IINA 1 3 1 French Step 1 - decrease speed with OSC

Step 2: typing in "backspace" key, so the speed is back to 1.00, but still displayed 0.50 on the OSC ("Vitesse" means "Speed" in French):
IINA 1 3 1 French Step 2 - back to 1 00 speed useing backspace

uiryuu:
- [x] This implements/fixes issue #4052.

---

**Description:**
Currently the left/rightArrowLabel is only updated when clicking on the fast/slowfoward button, making it not always indicating the current playback speed and might cause confusion to users (see #4052). This PR makes it updates the labels when the playback speed changes. Other minor changes:
- Adjust the constraint of the rightArrowLabel to show more digits
- Make the rightArrowLabel always shows 3 significant digits
- Don't reset the labels on clicking play/pause button

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -2144,6 +2144,7 @@ class PlayerCore: NSObject {
     guard info.state.active else { return }
     info.playSpeed = speed
     sendOSD(.speed(speed))
+    mainWindow.updateSpeedLabel(speed: speed)
     needReloadQuickSettingsView()
     NowPlayingInfoManager.shared.updateInfo()
   }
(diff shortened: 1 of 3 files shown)


iina/PlayerCore.swift
```diff
@@ -2144,6 +2144,7 @@ class PlayerCore: NSObject {
     guard info.state.active else { return }
     info.playSpeed = speed
     sendOSD(.speed(speed))
+    mainWindow.updateSpeedLabel(speed: speed)
```
low-batt, line 2147: At some point we need to consider breaking all the direct connections between classes. With the addition of the plugin system we probably need to be making more use of events, as seen below:
```swift
postNotification(.iinaSecondSubVisibilityChanged)
```
But no need to do this for this PR. This is an issue for the future. First up should be `MPVController`. That low level class should not be directly calling all of the high level classes that it does today. Because it does so I can't reuse it to have a second mpv core that displays HDR OSC thumbnails.

low-batt (approved):
Pulled PR, built IINA, confirmed fix. Changes look good to me.

PR #5512, opened 2025-05-12
playback: change the behavior of prev media button, fix #4611

### LINKED ISSUE #4611: Make rewind button useful when listening to a single audio track (play track from the beginning)
**What you want IINA to do:**

1. I click on a mp3 file in Finder.
2. IINA is launched and starts playing the track.
3. I press `|<<` button in the IINA UI or the keyboard `<<` button.
4. I expect the track starts playing from the beginning.

**What IINA does currently:**

1. I click on a mp3 file in Finder.
2. IINA is launched and starts playing the track.
3. I press `|<<` button in the IINA UI or the keyboard `<<` button.
4. Absolutely nothing happens.

Because it should not be that hard to rewind current track with a hotkey or `<<` button.

Maybe VLS although it's doesn't do the same as I request. Instead, `<<` and `>>` buttons in VLC (even keyboard ones) do progressive rewind (they move the track a few seconds forwards or backwards).

uiryuu:
- [x] This implements/fixes issue #4611.

---

**Description:**
Restart from the beginning of the track when trying to navigate to the previous track in the playlist when there's no previous track.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -2861,7 +2861,7 @@ class MainWindowController: PlayerWindowController {
       }
 
     case .playlist:
-      player.mpv.command(left ? .playlistPrev : .playlistNext, checkError: false)
+      player.navigateInPlaylist(nextMedia: !left)
 
     case .seek:
       player.seek(relativeSecond: left ? -10 : 10, option: .relative)
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -1431,7 +1431,11 @@ class PlayerCore: NSObject {
   }
 
   func navigateInPlaylist(nextMedia: Bool) {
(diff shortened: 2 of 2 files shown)


low-batt (changes requested):
This is only a partial fix. This does not work with the main window because `MainWindowController.arrowButtonAction` does this:
https://github.com/iina/iina/blob/c2806d05b8becb32c21fbb5a0c69cd532175d1e8/iina/MainWindowController.swift#L2863-L2864

We should start trying to reduce the amount of code that by-passes `PlayerCore` and accesses `MPVController` directly.

uiryuu:
@low-batt nice catch! Fixed. I've searched for `.playlistPrev` and its not used anywhere else.

low-batt (approved):
Pulled PR, built IINA, tested, worked for me. Changes look good.

PR #5513, opened 2025-05-12
Handle show playlist/chapter panel menu item for mini player

### LINKED ISSUE #5412: Add show playlist/chapters panel for music mode
**What you want IINA to do:**
Please enhance IINA to support the `Show Playlist Panel` and `Show Chapters Panel` menu items in the `Playback` menu along with their keyboard shortcuts when in music mode.

**What IINA does currently:**
When in music mode the `Show Playlist Panel` and `Show Chapters Panel` menu items are disabled. ⇧⌘P and ⇧⌘C do nothing when IINA is in music mode.

Similar to the main window the music mode window has a panel with `PLAYLIST` and `CHAPTERS` tabs. Users that prefer the keyboard over the mouse should be able to use ⇧⌘P and ⇧⌘C to show and hide the corresponding tabs of the mini players panel, just like they can with the main window. But because the menu items are disabled the corresponding keyboard shortcuts do not work. There is no reason these menu items should not work when IINA is using the mini player window, along with their keyboard shortcuts.

This enhancement was requested in issue #5185.

uiryuu:
- [x] This implements/fixes issue #5412.

---

**Description:**
Handle `menuShowPlaylistPanel` and `menuShowChaptersPanel` for menu actions for the mini player.

P.S. We should really separate `playlistView` from the main window, which requires a pretty large refactor. Maybe consider doing it in 1.5.0?

### DIFF
--- a/iina/MenuController.swift
+++ b/iina/MenuController.swift
@@ -478,11 +478,10 @@ class MenuController: NSObject, NSMenuDelegate {
 
   private func updatePlaybackMenu() {
     let player = PlayerCore.active
-    let isDisplayingPlaylist = player.mainWindow.sideBarStatus == .playlist &&
-          player.mainWindow.playlistView.currentTab == .playlist
+    let playlistPanelVisible = player.isInMiniPlayer ? player.miniPlayer.isPlaylistVisible : player.mainWindow.sideBarStatus == .playlist
+    let isDisplayingPlaylist = playlistPanelVisible && player.mainWindow.playlistView.currentTab == .playlist
     playlistPanel?.title = isDisplayingPlaylist ? Constants.String.hidePlaylistPanel : Constants.String.playlistPanel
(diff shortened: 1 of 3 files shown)


low-batt (changes requested):
Partial fix. When the playlist is displayed in the mini player the `Playback` menu item is `Show Playlist Panel` instead of `Hide Playlist Panel`. This code in `MenuController.updatePlaybackMenu` needs to be changed as it is checking the main window, at least I suspect this is what is causing the problem:
https://github.com/iina/iina/blob/c2806d05b8becb32c21fbb5a0c69cd532175d1e8/iina/MenuController.swift#L479-L486

uiryuu:
Fixed!

low-batt (approved):
Pulled PR, built IINA, tested, worked for me. Changes look good.

PR #5518, opened 2025-05-17
Change the NSSwitch in the quick setting panel to mini size

uiryuu:

---

**Description:**

As per Apple's documentation of NSSwitch:
> Use a switch to toggle significant preferences, or preferences that provide access to other controls. Avoid creating lists or tables of switches; instead, for general-purpose toggles, use an instance of [NSButton](dash-apple-api://load?request_key=ls/documentation/appkit/nsbutton) to display a checkbox.

We should not use a normal sized NSSwitch to control one setting. This PR change the switch to mini size to match the look and feel of a switch in modern macOS settings:
image

Before:
image

After:
image

@low-batt I don't know whether or not mini size is available in macOS 10.15. Can you help me to test it out? If it doesn't work, then we have to refactor more code to adjust the constraints.


low-batt (approved):
Pulled PR, built IINA, tested under macOS 10.15 and 15.4.1. The mini switches work fine under 10.15. Changes look good to me.

Issue #5162, opened 2024-08-29
Suggestion: Adding a feature that may enable us to sort the playlist by names, dates, files types etc. 


G-CHUNG18:
**What you want IINA to do:**
IINA could add a new feature--allows users to sort their playlists by names, dates, files types etc., at best both ascending and descending--which will make iina the best of all. 

**What IINA does currently:**
This feature is one that iina doesn't have. 

**Why you think this should be added:**
Currently when selecting all files pinned with tags (macOS) and hit play, iina functions weirdly. It neither sorts the files by names, or dates, or other apparent factors, but by a strange, unspeakable order. In short, many files are sorted by names while some of the recent files are exhibit on top of else. 

**Examples of other projects that have something similar:**
Fig player, one player that has such feature, but is hard to use compare to iina.

vurtomatic:
+1 to this... seems pretty essential.

Issue #4095, opened 2022-11-26
Suggestion: Playlist support sort feature


PiFrancis:
**What you want IINA to do:**
```tex
Hope IINA playlist support sort(e.g.: sort by add time、sort by name、sort by artist name, and so on) feature in the future version.
```
**What IINA does currently:**
```tex
Don't support sort feature in the IINA  playlist  currently version.
```
**Why you think this should be added:**
```tex
I think playlist support sort is player basic function. By the way, What IINA playlist default sort order currently? 
```

PiFrancis:
Don't think about it?🤔

renanlet:
omg that's the only single thing missing for it to be perfect

vurtomatic:
+1 to this... seems pretty essential.

iammihirbhagat:
How about support for sort by Track no.? #2440

PR #5173, opened 2024-09-09
Add ability to sort playlist based on names both ascending and descending order

karan0046:
- [x] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)
- [x] This implements/fixes issue #5162 .

---

**Description:**
Added an new sort button that has ability to sort the playlist by their fileNames, in both ascending and descending order.

https://github.com/user-attachments/assets/48b20dc8-415b-4a89-be67-3f5c6e2d263a

Wording suggestions are welcomed

Issue #3357, opened 2021-03-13
Help search gets stuck when search in large playlist


najibla:
**System and IINA version:**

- macOS Big sur 11.2 - 2,2 GHz Quad-Core Intel Core i7 16GB RAM
- IINA 1.1.2 Build 128 mpv 0.32.0

** Test case **

Open a file that has a very large number of items in the playlist. For example, I have an m3u (IPTV) file that has about 50,000 entries.

Since there is no search in the playlist, I used suggestions from [this](https://github.com/iina/iina/issues/3251) to click Help => and start typing to get results from the playlist.

**Expected behavior:**

IINA displays results from the playlist that match the search 

**Actual behavior:**

IINA gets stuck. I wait for at least a minute and nothing happens. (just still being stuck) 

**How often does this happen?**

All the time.

PR #5522, opened 2025-05-25
Add a sort button for playlist

uiryuu:
- [x] This implements/fixes issue #5162, #4095.

---

**Description:**
Added a sort button, when clicking on it, a context menu will prompt the user to choose the sorting method. We could add more sorting methods in the future. For now, I only added sorting ascending and descending by file path.

The way I implemented the playlist reordering is a bit strange, that's because the only operations I have to manipulate the playlist are removing/adding item to the playlist, and moving an item to a specific index. If I use moving index, then the best time complexity I can achieve is O(n^2). So I remove all the playlist items (except the one is current playing) and add them back by the order we want.

This PR also refactors private playlist-related functions from player core to `MPVCommandWrapper`, which is an extension of `MPVController`. We should avoid directly call methods in `MPVController` outside of player core in the future.

### DIFF
--- a/iina/PlaylistViewController.swift
+++ b/iina/PlaylistViewController.swift
@@ -112,6 +112,7 @@ class PlaylistViewController: NSViewController, NSTableViewDataSource, NSTableVi
     shuffleBtn.toolTip = NSLocalizedString("mini_player.shuffle", comment: "shuffle")
     addBtn.toolTip = NSLocalizedString("mini_player.add", comment: "add")
     removeBtn.toolTip = NSLocalizedString("mini_player.remove", comment: "remove")
+    sortBtn.toolTip = NSLocalizedString("mini_player.sort", comment: "sort")
 
     hideTotalLength()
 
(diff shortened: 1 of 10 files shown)


iina/PlaylistViewController.swift
```diff
@@ -468,6 +474,28 @@ class PlaylistViewController: NSViewController, NSTableViewDataSource, NSTableVi
     subPopover.show(relativeTo: sender.bounds, of: sender, preferredEdge: .maxY)
   }
 
+  @IBAction func sortingBtnAction(_ sender: NSButton) {
+    let menu = NSMenu()
+    if #available(macOS 14.0, *) {
+      menu.addItem(.sectionHeader(title: "Sorting"))
```
low-batt: Needs localization.

iina/PlaylistViewController.swift
```diff
@@ -468,6 +474,28 @@ class PlaylistViewController: NSViewController, NSTableViewDataSource, NSTableVi
...
+    }
+    menu.addItem(withTitle: NSLocalizedString("playlist.sorting.path_ascending", comment: "File Path (A-Z)"), action: #selector(sortPathAscending), keyEquivalent: "")
```
low-batt: Files don't have to start with a letter, so A-Z isn't really appropriate. Maybe "Ascending/Descending"? I do like that (A-Z) is nice and short.

low-batt (changes requested):
Tuesday is my busy day, so just some quick possibly incoherent thoughts…

PR #5173, already on the 1.4.0-beta2 schedule, does something similar. This PR contains some improvements over that one.

The new button is missing a help tag, which the other two buttons have:
https://github.com/iina/iina/blob/7b4a7bda0a6c2a99c4ba98f874a1365f25a360cd/iina/PlaylistViewController.swift#L105-L106

Sorting based on file path is a bit confusing when titles are being displayed.

One concern I have about sorting is that apparently IPTV playlists can be very large. Will IINA lock up if someone tries to sort a large playlist? See issue #3357. I don't think this is a blocking factor because this isn't just a button you can accidentally hit as it pops up a menu. That makes it less of a concern.

I guess my biggest concern is that sorting by file path while displaying titles makes the feature seem like it is broken.

iina/PlaylistViewController.swift
uiryuu: Added.

iina/PlaylistViewController.swift
uiryuu: Changed to "Ascending/Descending", see whether or not it looks better to you.

uiryuu:
> PR https://github.com/iina/iina/pull/5173, already on the 1.4.0-beta2 schedule, does something similar.

I've added @karan0046 as coauthor in the followup commit to give them credit after this PR is merged.

> The new button is missing a help tag, which the other two buttons have:

Added.

> Sorting based on file path is a bit confusing when titles are being displayed.

I was afraid that the `filenameForDisplay` property will change depending on whether or not the user choose to show song title from metadata in the playlist. But `filenameForDisplay` is just the last path component for a file, so I was wrong. Added sort by filename in the latest commit.

> I guess my biggest concern is that sorting by file path while displaying titles makes the feature seem like it is broken.

Still want to keep sort by file path as maybe some users would like to sort based on the folder structure.

PR #5524, opened 2025-05-27
Fix top osc not responds to scroll wheel

### LINKED ISSUE #3863: Enable scroll wheel scrubbing with Controller UI in Top position

uiryuu:
- [x] This implements/fixes issue #3863.

---

**Description:**
Found a simple fix to this issue, just don't return when one of the override is true. @svobs @low-batt take a look

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -1041,7 +1041,6 @@ class MainWindowController: PlayerWindowController {
 
   override func scrollWheel(with event: NSEvent) {
     guard !isInInteractiveMode else { return }
-    guard !isMouseEvent(event, inAnyOf: [sideBarView, titleBarView, subPopoverView]) else { return }
 
     if isMouseEvent(event, inAnyOf: [fragSliderView]) && playSlider.isEnabled {
       seekOverride = true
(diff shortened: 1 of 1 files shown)


iina/MainWindowController.swift
```diff
@@ -1051,6 +1050,10 @@ class MainWindowController: PlayerWindowController {
       guard !isMouseEvent(event, inAnyOf: [currentControlBar]) else { return }
     }
 
+    if isMouseEvent(event, inAnyOf: [sideBarView, titleBarView, subPopoverView]) && !(seekOverride || volumeOverride) {
+      return
+    }
+
```
low-batt, line 1055: Why switch to a complex `if` over a `guard`? Can't this just be?:
```swift
    guard !isMouseEvent(event, inAnyOf: [sideBarView, titleBarView, subPopoverView])
            || seekOverride || volumeOverride else { return }
```

uiryuu: You are absolutely correct. Somehow I incorrectly thought converting to `if` would make it simpler.

low-batt (approved):
I confirmed this fixes the problem. Good to get this one fixed!

PR #5526, opened 2025-05-28
Close support cover artwork in Now Playing module, #5336

### LINKED ISSUE #5336: Support cover artwork in Now Playing module
**What you want IINA to do:**
Add support for setting [MPMediaItemPropertyArtwork](https://developer.apple.com/documentation/mediaplayer/mpmediaitempropertyartwork) in [MPNowPlayingInfoCenter.nowPlayingInfo](https://developer.apple.com/documentation/mediaplayer/mpnowplayinginfocenter/nowplayinginfo) so the macOS [Control Center](https://support.apple.com/guide/mac-help/quickly-change-settings-mchl50f94f8f/mac) Now Playing module can display the media's cover art.

**What IINA does currently:**
IINA does provide cover art to Now Playing so it only display's IINA's icon:
now-playing

low-batt:
This commit will:
- Move `NowPlayingInfoManager` out of `PlayerCore` and into its own file
- Change `PlayerCore.pauseChanged` to call `syncUITime` before changing the state so that `PlaybackInfo.videoPosition` is up to date when `NowPlayingInfoManager.updateInfo` is called
- Change updateInfo to set `MPNowPlayingInfoPropertyPlaybackRate` to zero when playback is paused as required by Now Playing
- Change updateInfo to set `playbackState` to `playing` before setting it to `paused` when playback is paused from the start to workaround a Now Playing problem
- Remove the state parameter from the `updateInfo` method and change the method to use the `state` property in `PlaybackInfo`
- Change `updateInfo` to remove `MPMediaItemPropertyAlbumTitle` and `MPMediaItemPropertyArtist` when not needed instead of setting them to the empty string
- Add a `cgImage` property and a `resized` method to the `NSImage` extension
- Add a `CGImage` extension
- Add a `CGSize` extension
- Add a `MPNowPlayingPlaybackState` extension that adopts the `CustomStringConvertible` property to allow values to be easily logged
- Add an `isImage` property to `MPVTrack`
- Add a `readArtwork` method to `FFmpegController`
- Add calls to `NowPlayingInfoManager.updateInfo` when thumbnails are ready
- Add support for setting `MPMediaItemPropertyArtwork` in `nowPlayingInfo` to `NowPlayingInfoManager`
- Add a new `enableNowPlayingArtwork` internal setting that can be used to disable showing artwork should a serious problem be discovered

This commit is builds upon the pioneering work done by @svobs to add support for showing artwork in Now Playing. NowPlayingInfoManager will first try and find a video track that represents cover art and use that for the Now Playing artwork. If that fails it will then ask the macOS Quick Look system to generate a thumbnail to use as Now Playing artwork. If that fails then it will try and use an OSC thumbnail.

This commit also corrects several existing problems with Now Playing found during testing.

- [x] This implements/fixes issue #5336.

---

**Description:**

### DIFF
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -8,6 +8,7 @@
 
 import Cocoa
 import CryptoKit
+import MediaPlayer
 
 extension NSSlider {
   /**
@@ -581,6 +582,11 @@ extension NSTextField {
 }
 
 extension NSImage {
+  var cgImage: CGImage? {
+    var rect = CGRect.init(origin: .zero, size: self.size)
+    return self.cgImage(forProposedRect: &rect, context: nil, hints: nil)
+  }
+
   func tinted(_ tintColor: NSColor) -> NSImage {
     guard self.isTemplate else { return self }
 
@@ -594,6 +600,14 @@ extension NSImage {
     image.isTemplate = false
 
     return image
+  }
+
+  /// `cornerRadius`: if greater than 0, round the corners by this radius
+  func resized(newWidth: Int, newHeight: Int, cornerRadius: CGFloat = 0) -> NSImage {
+    if let cgImageNew = cgImage?.resized(newWidth: newWidth, newHeight: newHeight, cornerRadius: cornerRadius) {
+      return NSImage(cgImage: cgImageNew, size: NSSize(width: newWidth, height: newHeight))
+    }
+    return self
   }
 
   func rounded() -> NSImage {
@@ -811,5 +825,108 @@ extension Process {
     process.waitUntilExit()
 
     return (process, stdout, stderr)
+  }
+}
+
+extension CGImage {
+  var nsImage: NSImage { NSImage(cgImage: self, size: size) }
+  var size: CGSize { CGSize(width: width, height: height) }
+
+  /// `cornerRadius`: if greater than 0, round the corners by this radius
+  func resized(newWidth: Int, newHeight: Int, cornerRadius: CGFloat = 0) -> CGImage {
+    guard newWidth != width || newHeight != height else {
+      return self
+    }
+
+    guard newWidth > 0, newHeight > 0 else {
+      Logger.fatal("NSImage.resized: invalid width (\(newWidth)) or height (\(newHeight)) - both must be greater than 0")
+    }
+
+    // Use raw CoreGraphics calls instead of their NS equivalents. They are > 10x faster, and only downside is that the image's
+    // dimensions must be integer values instead of decimals.
+    let newImage = CGImage.buildBitmapImage(width: Int(newWidth), height: Int(newHeight)) { cgContext in
+      let outputRect = CGRect(x: 0, y: 0, width: newWidth, height: newHeight)
+      if cornerRadius > 0.0 {
+        cgContext.beginPath()
+        cgContext.addPath(CGPath(roundedRect: outputRect, cornerWidth: cornerRadius, cornerHeight: cornerRadius, transform: nil))
+        cgContext.closePath()
+        cgContext.clip()
+      }
+      cgContext.draw(self, in: outputRect)
+    }
+
+    return newImage
+  }
+
+  /// Builds a bitmap image efficiently using CoreGraphics APIs.
+  ///
+  /// If it's found useful for any more situations, should put in its own class
+  static func buildBitmapImage(width: Int, height: Int, _ drawingCalls: (CGContext) -> Void) -> CGImage {
+    guard let compositeImageRep = CGImage.makeNewImgRep(width: width, height: height) else {
+      Logger.fatal("DrawImageInBitmapImageContext: Failed to create NSBitmapImageRep!")
+    }
+
+    guard let context = NSGraphicsContext(bitmapImageRep: compositeImageRep) else {
+      Logger.fatal("DrawImageInBitmapImageContext: Failed to create NSGraphicsContext!")
+    }
+
+    context.cgContext.interpolationQuality = .high
+    drawingCalls(context.cgContext)
+
+    return compositeImageRep.cgImage!
+  }
+
+  /// Creates RGB image with alpha channel
+  static func makeNewImgRep(width: Int, height: Int) -> NSBitmapImageRep? {
+    return NSBitmapImageRep(
+      bitmapDataPlanes: nil,
+      pixelsWide: width,
+      pixelsHigh: height,
+      bitsPerSample: 8,
+      samplesPerPixel: 4,
+      hasAlpha: true,
+      isPlanar: false,
+      colorSpaceName: NSColorSpaceName.calibratedRGB,
+      bytesPerRow: 0,
+      bitsPerPixel: 0)
+  }
+}
+
+extension CGSize {
+  var heightInt: Int { Int(height) }
+  var widthInt: Int { Int(width) }
+
+  /// Crops the current size to fit within a target aspect ratio, reducing either the width or height to match the aspect ratio of the
+  /// target rectangle.
+  /// - Parameter targetAspect: A rectangle or size structure that contains the desired aspect ratio.
+  /// - Returns: The cropped `NSSize` that fits within the given aspect ratio.
+  func crop(withAspect targetAspect: CGFloat) -> NSSize {
+    if aspect > targetAspect {  // self is wider, crop width, use same height
+      return NSSize(width: round(height * targetAspect), height: height)
+    } else {
+      return NSSize(width: width, height: round(width / targetAspect))
+    }
+  }
+
+  func getCropRect(withAspect aspect: CGFloat) -> NSRect {
+    let croppedSize = crop(withAspect: aspect)
+    let cropped = NSMakeRect(round((width - croppedSize.width) / 2),
+                             round((height - croppedSize.height) / 2),
+                             croppedSize.width,
+                             croppedSize.height)
+    return cropped
+  }
+}
+
+extension MPNowPlayingPlaybackState: @retroactive CustomStringConvertible {
+  public var description: String {
+    switch self {
+    case .unknown: return "unknown"
+    case .playing: return "playing"
+    case .paused: return "paused"
+    case .stopped: return "stopped"
+    case .interrupted: return "interrupted"
+    @unknown default: return String(self.rawValue)
+    }
   }
 }
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -556,6 +556,18 @@ extension NSMenuItem {
 
 
 extension URL {
+  /// A string representing the URL in the format mpv uses for
+  /// [playlist/N/filename](https://mpv.io/manual/stable/#command-interface-playlist/n/filename].
+  var mpvStr: String {
+    guard isFileURL else {
+      return absoluteString
+    }
+    guard #available(macOS 13.0, *) else {
+      return path
+    }
+    return path(percentEncoded: false)
+  }
+
   var creationDate: Date? {
     (try? resourceValues(forKeys: [.creationDateKey]))?.creationDate
   }
@@ -915,18 +927,5 @@ extension CGSize {
                              croppedSize.width,
                              croppedSize.height)
     return cropped
-  }
-}
-
-extension MPNowPlayingPlaybackState: @retroactive CustomStringConvertible {
-  public var description: String {
-    switch self {
-    case .unknown: return "unknown"
-    case .playing: return "playing"
-    case .paused: return "paused"
-    case .stopped: return "stopped"
-    case .interrupted: return "interrupted"
-    @unknown default: return String(self.rawValue)
-    }
   }
 }
--- a/iina/MPVTrack.swift
+++ b/iina/MPVTrack.swift
@@ -12,24 +12,29 @@ class MPVTrack: NSObject {
 
   /** For binding a none track object to menu, id = 0 */
   static let noneVideoTrack: MPVTrack = {
-    let track = MPVTrack(id: 0, type: .video, isDefault: false, isForced: false, isSelected: false, isExternal: false)
+    let track = MPVTrack(id: 0, type: .video, isDefault: false, isForced: false, isImage: false,
+                         isSelected: false, isExternal: false)
     track.title = NSLocalizedString("track.none", comment: "<None>")
     return track
   }()
   /** For binding a none track object to menu, id = 0 */
   static let noneAudioTrack: MPVTrack = {
-    let track = MPVTrack(id: 0, type: .audio, isDefault: false, isForced: false, isSelected: false, isExternal: false)
+    let track = MPVTrack(id: 0, type: .audio, isDefault: false, isForced: false, isImage: false,
+                         isSelected: false, isExternal: false)
     track.title = NSLocalizedString("track.none", comment: "<None>")
     return track
   }()
   /** For binding a none track object to menu, id = 0 */
   static let noneSubTrack: MPVTrack = {
-    let track = MPVTrack(id: 0, type: .sub, isDefault: false, isForced: false, isSelected: false, isExternal: false)
+    let track = MPVTrack(id: 0, type: .sub, isDefault: false, isForced: false, isImage: false,
+                         isSelected: false, isExternal: false)
     track.title = NSLocalizedString("track.none", comment: "<None>")
     return track
   }()
   /** For binding a none track object to menu, id = 0 */
-  static let noneSecondSubTrack = MPVTrack(id: 0, type: .secondSub, isDefault: false, isForced: false, isSelected: false, isExternal: false)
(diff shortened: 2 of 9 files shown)


low-batt:
This commit is builds upon the pioneering work done by @svobs to add support for showing artwork in Now Playing. Obtaining artwork to show is a bit involved because the inputs needed for the process such as video tracks and OSC thumbnails appear asynchronously after the media is opened and work to search for and construct an artwork image is done in the background.

This commit definitely needs a critical review!

One difference between mpv and IINA is visible in the following screenshots…

mpv 0.4.0 supplies an image that is not square and Now Playing shows it with grey borders:
mpv


IINA with this PR when possible crops the image to properly fit the space Now Playing provides:
iina


This has to do with the value that must be supplied for [MPMediaItemPropertyArtwork](https://developer.apple.com/documentation/mediaplayer/mpmediaitempropertyartwork), a [MPMediaItemArtwork](https://developer.apple.com/documentation/mediaplayer/mpmediaitemartwork) object. The [initializer](https://developer.apple.com/documentation/mediaplayer/mpmediaitemartwork/init(boundssize:requesthandler:)) takes a `boundsSize` parameter and a callback:
> **boundsSize**
>    The original size of the artwork.
>    
> **requestHandler**
> A handler that the system calls for the requested artwork.
>  &nbsp; &nbsp; &nbsp; &nbsp;**size**
>  &nbsp; &nbsp; &nbsp; &nbsp;The new size for the image.

Where `size` is a parameter to the `requestHandler` callback. Apple indicates:

> The request handler returns the image in the newly requested size. The requested size must be less than the `boundsSize` parameter.

Unfortunately this all the guidance I was able to find from Apple. 

mpv completely ignores the `size` parameter and returns the full sized image:

```swift
MPMediaItemArtwork(boundsSize: cover.size) { _ in return cover }
```

The implementation developed by @svobs crops and resizes the image to the `size` parameter. With verbose logging enabled you can see this:

```
17:27:06.051 [now-playing][v] Cropped and resized 2000x1125 artwork image to 768x768
```

In the above test case both IINA and mpv are displaying thumbnails generated by [QLThumbnailGenerator](https://developer.apple.com/documentation/quicklookthumbnailing/qlthumbnailgenerator). As a result the artwork matches up with the thumbnails shown in Finder:

finder

I have the Quick Look extension [QuickLook Video](https://github.com/Marginal/QLVideo) installed to gain thumbnail support for files not supported by Apple. It generated the thumbnails shown above.

If `QLThumbnailGenerator` fails to generate a thumbnail IINA will fall back to using an OSC thumbnail. The choice of playback position at which to obtain a thumbnail is similar to how `QuickLook Video` determines the position to use.

low-batt:
The mpv Now Playing implementation can be seen [here](https://github.com/mpv-player/mpv/blob/master/osdep/mac/remote_command_center.swift).

Note the loose coupling accomplished by using events. At some point IINA should consider making more use of events to break some of the tight coupling between IINA classes.

iina/MPVTrack.swift
```diff
@@ -121,11 +127,12 @@ class MPVTrack: NSObject {
 
   var decoderDesc: String?
 
-  init(id: Int, type: TrackType, isDefault: Bool, isForced: Bool, isSelected: Bool, isExternal: Bool) {
+  init(id: Int, type: TrackType, isDefault: Bool, isForced: Bool, isImage: Bool, isSelected: Bool, isExternal: Bool) {
```
lhc70000: I suggest adding a `convenience init` that matches the previous method signature and set `isImage` to false by default, so the above lines don't need to be changed.

iina/NowPlayingInfoManager.swift
```diff
@@ -0,0 +1,620 @@
...
+        info[MPMediaItemPropertyArtist] = artist
+      } else {
+        info[MPNowPlayingInfoPropertyMediaType] = MPNowPlayingInfoMediaType.video.rawValue
+        info[MPMediaItemPropertyTitle] = activePlayer.getMediaTitle(withExtension: false)
+        info.removeValue(forKey: MPMediaItemPropertyAlbumTitle)
+        info.removeValue(forKey: MPMediaItemPropertyArtist)
+      }
+    }
+
+    let duration = PlayerCore.lastActive.info.videoDuration?.second ?? 0
```
uiryuu:
```suggestion
    let duration = activePlayer.info.videoDuration?.second ?? 0
```

iina/NowPlayingInfoManager.swift
```diff
@@ -0,0 +1,620 @@
...
+    // When playback is paused Now Playing expects the playback rate to be set to zero. If this is
+    // not done the slider in Now Playing may incorrectly display a playback time of zero.
+    let paused = activePlayer.info.state == .paused
+    let speed = paused ? 0 : activePlayer.info.playSpeed
+
+    info[MPMediaItemPropertyPlaybackDuration] = duration
+    info[MPNowPlayingInfoPropertyElapsedPlaybackTime] = time
+    info[MPNowPlayingInfoPropertyPlaybackRate] = speed
+    info[MPNowPlayingInfoPropertyDefaultPlaybackRate] = 1
+
```
uiryuu, line 181: What about adding `MPNowPlayingInfoPropertyAssetURL` here as well? Didn't notice any difference but the system might use this value.
```suggestion
    info[MPNowPlayingInfoPropertyAssetURL] = activePlayer.info.currentURL

```

uiryuu: Also didn't notice any difference after adding `MPNowPlayingInfoPropertyChapterCount` and `MPNowPlayingInfoPropertyChapterNumber`, but those're also something we can add.

uiryuu (approved):
Nice work. During my tests I didn't found any problem. See my comments and you can decide whether or not to add those info.

uiryuu:
Just notice a problem, sometimes when you open a new file, the previous artwork still remains. I don't know what triggers this, but it seems to appear more often when you start the new file paused.

iina/NowPlayingInfoManager.swift
low-batt: For this PR I was trying to just add support for artwork and not cleanup the existing code. I ended cheating a bit on that and including some fixes because they were making it hard to test these changes. I'll go ahead and include these tweaks as well.

iina/NowPlayingInfoManager.swift
low-batt: I'll add the setting of `MPNowPlayingInfoPropertyAssetURL` and then correct it to not be set if `currentURL` is `nil`.

There is a _lot_ that still needs to be done with respects to this code. This PR is focused on the artwork for the Now Playing module. Some of the missing settings like `MPNowPlayingInfoPropertyChapterCount` I don't think matter to the Now Playing module. _However_ I am expecting they do matter if you try and use Siri to control IINA. Originally I was going to open that can of worms for 1.4.0. I'm guessing there are only a few things that need fixing. But currently I was thinking of postponing looking into this as I still have several issues I need to address for beta2. Should I look into this?

low-batt:
On the wrong artwork being displayed, that is exactly what I've been worried about due to all the asynchronous activity. The `artworkTicket` is supposed to prevent that from happening. I tested a lot by setting IINA to pause at startup and dragging a lot of files onto the IINA icon, but so far have not be able to reproduce the problem with the wrong artwork being displayed. I specifically added a lot of verbose logging due to the potential for such a problem. I'll probably improve those log messages a bit. So far I only found one problem, I was not resetting the QL retries. I'll keep trying to reproduce this problem.

iina/NowPlayingInfoManager.swift
uiryuu: Let's focus on b2 for now.

low-batt:
I added support for more properties and fixed a problem with missing code to reset `artworkQLGenerationRetries`. So far I've not been able to reproduce a problem with the wrong artwork being displayed. I looked through the code and didn't spot a way in which that can happen. I improved the logging so that if it can be reproduced we might have a clue as to what is happening if verbose logging is enabled. I'll look over the code some more tomorrow.

One possibility is that this is a Now Playing glitch. I hit problems when a file is opened and IINA is configured to pause. Now Playing would ignore the update. That is why there is a workaround that tells Now Playing the media is playing and then immediately sets it to paused. As updating the artwork happens after that maybe Now Playing sometimes ignores that. There is code in mpv:
```swift
    func makeCurrent() {
        infoCenter.playbackState = .paused
        infoCenter.playbackState = .playing
        updateInfoCenter()
    }
```

What is that setting of `playbackState` all about? Some sort of work around?

low-batt:
@svobs Thoughts on this PR?

iina/Extensions.swift
```diff
@@ -813,3 +827,106 @@ extension Process {
...
+    let croppedSize = crop(withAspect: aspect)
+    let cropped = NSMakeRect(round((width - croppedSize.width) / 2),
+                             round((height - croppedSize.height) / 2),
+                             croppedSize.width,
+                             croppedSize.height)
+    return cropped
+  }
+}
+
+extension MPNowPlayingPlaybackState: @retroactive CustomStringConvertible {
```
lhc70000: I think it's better to put this in NewPlayingInfoManager.swift

iina/NowPlayingInfoManager.swift
```diff
@@ -0,0 +1,620 @@
...
+    }
+    
+    // Because showing artwork is a complex operation there is an internal setting that can be
+    // changed to disable this feature should any problems with it be discovered.
+    if Preference.bool(for: .enableNowPlayingArtwork) {
+      // If the media item has changed then if artwork is being displayed it needs to be abandoned.
+      // NOTE this MUST be done before copying the nowPlayingInfo array below as abandonArtwork
+      // works directly on nowPlayingInfo.
+      if activePlayer.info.isNetworkResource {
+        abandonArtwork()
```
lhc70000: We have a setting `enableThumbnailForRemoteFiles` for thumbnail generation, maybe it should also be honored here.

iina/Extensions.swift
low-batt: Will do.

iina/MPVTrack.swift
low-batt: Will do.

iina/NowPlayingInfoManager.swift
low-batt: I'll test, but I believe that is already supported as that is for mounted volumes, not network. I'll double check that.

I was going to task about `enableThumbnailForRemoteFiles` as there are several issues where this has come up. Is there a reason why we don't have UI for that setting?

low-batt:
I am currently working on an update that fixes the problem @uiryuu found and cleans up some stuff. I'll also address the input from @lhc70000.

iina/MPVTrack.swift
low-batt: I added a default value of `false` for the new `isImage` parameter and removed the changes to the static constants.

iina/Extensions.swift
low-batt: This extension has been moved to `NowPlayingInfoManager`.

iina/NowPlayingInfoManager.swift
low-batt: The handling of `isNetworkResource` was incorrect. Streaming can have embedded artwork and there is a mpv option that allows you to specify artwork which will show up as a video track. The code now checks `isNetworkResource` and doesn't try and generate Quick Look thumbnails or search for OSC thumbnails when streaming, but will show any artwork found in tracks.

Where `enableThumbnailForRemoteFiles` comes into play is with mounted volumes. I could add code to check to see if the media is on a mounted volume and if so then not look for OSC thumbnails for it unless `enableThumbnailForRemoteFiles` is enabled. The main effect of that would only be changes in the log messages you see with verbose logging. Currently it just reports thumbnails are not ready, which is what happens unless `enableThumbnailForRemoteFiles` is enabled.

low-batt:
The review comments gave me ideas for other tests to run. I found a number of problems. One of them was caused because `PlayerCore` did not reset `PlaybackInfo` properties when opening a new file.

svobs:
> One difference between mpv and IINA is visible in the following screenshots…
> 
> mpv 0.4.0 supplies an image that is not square and Now Playing shows it with grey borders
> 
> IINA with this PR when possible crops the image to properly fit the space Now Playing provides

In defense of cropping the image (in case anyone is thinking about this), I consider it the least bad option, given the constraints imposed by Apple. Drawing the image with gray space around it (or any other color) just ends up looking clunky and shoehorned in, IMHO. Apple clearly wants a square image to be shown there. And their API implies that they may change their mind in the future and apps should be prepared to deliver any size they ask for. So I think this is an improvement from what mpv does.

The code looks good! Only a couple of things:
1. The multithreading looks well done and should avoid tying up the main thread. But I did find that it took a while for the artwork to appear when I opened a large 4K HDR movie file. I think this happened because `artworkQueue` uses `qos: .background`, which is 1 level lower priority than `thumbnailQueue`, which uses `qos: .utility`. So Now Playing may have been left pending until the the thumbnails finishing generating. I'm thinking `artworkQueue` should at least be `utility` as well?
2. I played with a few test cases to try to get it to show stale artwork. Couldn't trigger it with any videos. But if I reused the window to play to an audio file which had no artwork, I found that Now Playing still showed the previous file's artwork. In my fork when there's no video track or there's a failure to generate an image, I just use the default album art for `MPMediaItemArtwork`. But I suppose another option might be to just clear it.

low-batt:
I'll change the queue priority.  The code is supposed to delete the artwork when the url changes. I'll look into why that did not happen.

low-batt:
I think everyone is good with the cripping. Note that it does not eliminate all use of those grey borders Now Playing seems to want 768x768. If the video is tiny and all that is available is the tiny OSC thumbnails, then both the video size and the thumbnail size are too small and you will end up with grey borders.

svobs:
> The code is supposed to delete the artwork when the url changes. I'll look into why that did not happen.

Looks like it properly clears the art if changing to the audio file via playlist navigation. I needed to drag & drop the audio file onto the video.

svobs:
> Note that it does not eliminate all use of those grey borders Now Playing seems to want 768x768. If the video is tiny and all that is available is the tiny OSC thumbnails, then both the video size and the thumbnail size are too small and you will end up with grey borders.

Really? Seemed fine for me with one of my nano-size test videos:

https://github.com/user-attachments/assets/f05cfd9b-6fbb-4454-beeb-986174b4a296

low-batt:
My guess is that this is what happened in your test run:

```
16:16:35.067 [now-playing][d] Found Quick Look thumbnail for: /Users/low-batt/Documents/builds/iina-official/issues/5336/450632367-f05cfd9b-6fbb-4454-beeb-986174b4a296.mp4
16:16:35.067 [now-playing][v] Video size (320x240) is smaller than image, using image size for bounds
```

IINA detected the image size was larger than the video size and used the image size for the bounds. That was a change I made. This worked because even through we are asking `QLThumbnailGenerator` for a `lowQualityThumbnail` it provides an image bigger than Now Playing needs:
```text
16:16:35.275 [now-playing][v] Cropped and resized 2000x1500 artwork image to 768x768
```

To see the problem you need a file that `QLThumbnailGenerator` does not support so IINA has to use an OSC thumbnail.  If I add a `, false` in `generateQLThumbnail` like so:
```swift
        guard let image = thumbnail?.nsImage, false else {
```

It will force use of an OSC thumbnail and:
```
16:25:27.435 [now-playing][v] Artwork bounds size: 320x240

16:25:27.637 [now-playing][v] Cropped and resized 240x180 artwork image to 320x240
```

I'm thinking we need to set the bounds size to have a minimum size of 768 and reflect the aspect ratio of the video or image in order to cause the image to be sized up.

Gotta run. Tuesday is my busy day.

low-batt:
@svobs I've not been able to reproduce the problem. I disabled `Switch to "music mode" automatically when playing an audio file`. I started playing a video. I dragged a music file onto the main window. The music file started playing. The artwork in Now Playing changed to the IINA icon.

There was such a problem with the initial version of the PR. Make sure you have the latest commits.

svobs:
> There was such a problem with the initial version of the PR. Make sure you have the latest commits.

My bad! Sorry. Updated to the latest version and confirmed that I am not seeing stale artwork now.

> To see the problem you need a file that QLThumbnailGenerator does not support so IINA has to use an OSC thumbnail. If I add a , false in generateQLThumbnail like so:

I see it now. Interesting. Seems that the `CGSize` parameter of the `MPMediaItemArtwork` request handler actually tries to preserve the aspect ratio from `boundsSize` for lower resolutions... That says something... Up to you if you want to try to improve on that design decision, but you could be forgiven for strictly adhering to the API and leaving it as-is.

low-batt:
The latest commit makes sure `boundsSize` is always large enough to avoid the grey bars. I also changed the queue priority to `utility`.

See if the new changes look good to go.

Issue #5305, opened 2024-12-14
IINA requests access to Apple Music


low-batt:
### IINA Version

1.3.5

### Release Channel

Release

### macOS Version

15.2

### Steps to Reproduce

On a Mac running macOS Sequoia…
- Under the  menu click on `System Settings…`
- In the `System Settings` panel click on `Privacy & Security`
- On the right side of the panel lick on `Media & Apple Music`
- If IINA is listed click on it
- Click on the `-` button
- If prompted for admin credentials, enter them
- Using IINA, play a video that is in the `Movies` directory under your home directory
- A pop-up appears indicating IINA would like to access Apple Music

### Expected Behavior

IINA does not request access to Apple Music.

### Actual Behavior

The first time a video is played that is in the `Movies` directory or a song is played that is in the `Music` directory macOS Sequoia will display the following alert:
![alert](https://github.com/user-attachments/assets/9b74c94d-0499-46d0-90a7-280d093db2bf)

This is a violation of the [Privacy](https://developer.apple.com/design/human-interface-guidelines/privacy) section of Apple's [Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/), specifically:
> **Write copy that clearly describes how your app uses the ability, data, or resource you’re requesting.** The standard alert displays your copy (called a purpose string or usage description string) after your app name and before the buttons people use to grant or deny their permission.

The above alert is missing the "purpose string". To supply such a string IINA would need to add a [NSAppleMusicUsageDescription](https://developer.apple.com/documentation/bundleresources/information-property-list/nsapplemusicusagedescription) key with a string value such as "This app requires Apple Music library access to play music tracks." to the `Info.plist` file.

_However_ the HIG also says:
> **Request permission only when your app clearly needs access to the data or resource.**

Since IINA does not support accessing Apple Music it is odd that it is requesting access.

### How often can you reproduce the issue

Always, under macOS Sequoia

### Required Actions

- [X] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [X] I tested with the Nightly build and confirmed the issue is not fixed.
- [X] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

svobs:
@low-batt I see you self-assigned this so I'm curious if you think you have a lead on a fix. I dug into a rabbit hole and found:
[...]

PR #5527, opened 2025-05-30
Filter duplicated matched subtitles, fix #5399

### LINKED ISSUE #5399: Externally loaded subtitles not automatically selected
### IINA Version

1.3.5

### Release Channel

Release

### macOS Version

macOS 15.3

### Steps to Reproduce

- Download the YouTube video [Predictions on the USA - With many different Subtitles](https://www.youtube.com/watch?v=6QmwrtdqFJc)
- Download the 17 `.vtt` subtitle files associated with this video
- Start IINA
- The settings panel appears
- On the left side of the panel click on `Subtitle`
- Select `Subtitles containing media filename` for the `Auto Load` setting
- Play the above video
- Under the `Subtitles` menu open the submenu `Subtitle` 
- Notice `<None>` is selected
- Notice the list runs from `#18` to `#34`

### Expected Behavior

After loading external subtitle files IINA automatically selects one to be shown.

### Actual Behavior

After loading the external subtitles files subtitles are not displayed.

From a very quick look into this problem I'm thinking IINA is loading the same subtitle file more than once and in reaction to that `libmpv` is removing the track that was previously loaded. The code that selects a subtitle is found in PlayerCore.fileStarted:
https://github.com/iina/iina/blob/e015724a0cb0c8a03cad2b4f637e277d1d65ad28/iina/PlayerCore.swift#L1834

That code is setting the mpv [sid](https://mpv.io/manual/stable/#options-sid) option:
> **--sid=<ID|auto|no>**
> Display the subtitle stream specified by `<ID>`. auto selects the default, no disables subtitles.

This code assumes there is a subtitle track with an ID of 1, but `libmpv` is usually deleting the track with that ID as a duplicate.

At least that is what it seems like is happening from a very quick look into this.

_No response_

### Logs

[iina.log](https://github.com/user-attachments/files/18951553/iina.log)

[mpv.log](https://github.com/user-attachments/files/18951554/mpv.log)

### Sample Files

Download the YouTube video [Predictions on the USA - With many different Subtitles](https://www.youtube.com/watch?v=6QmwrtdqFJc) along with the associated subtitle files.

Usually.

### Required Actions

### Optional Actions

uiryuu:
- [x] This implements/fixes issue #5399.

---

**Description:**
After subtitles matching, filter out duplicated subtitles before loading them as external subtitles.

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -1866,7 +1866,11 @@ class PlayerCore: NSObject {
         // auto load matched subtitles
         if let matchedSubs = self.info.getMatchedSubs(path) {
           log("Found \(matchedSubs.count) subs for current file")
+          var loadedSubs = Set<URL>()
           for sub in matchedSubs {
+            // filter duplicated matched subtitles, see https://github.com/iina/iina/issues/5399
+            guard !loadedSubs.contains(sub) else { continue }
+            loadedSubs.insert(sub)
             try checkTicket(currentTicket)
             loadExternalSubFile(sub)
           }


low-batt (changes requested):
Only half fixed. Try setting `Also search subtitles in the following directories` to just `.`. With that setting the problem is back.

It is important that just `.` works due to issue #5305. The code changes I attempted to fix #5305 all failed. Just trying to read the Music settings to figure out which directory to avoid triggered the alert. Without a code fix you have to change `Also search subtitles in the following directories` from the default of `./*` to just `.` to avoid constantly triggering the alert while running development versions of IINA.

uiryuu:
With `Also search subtitles in the following directories` being set to `.` (or `./*`), I cannot reproduce the issue. Can you double check?

low-batt (approved):
I can't reproduce it either. I must have accidentally ran IINA from the wrong Xcode window after changing settings. Sorry!

Merging…

Issue #5482, opened 2025-04-25
UI Freeze on resize video window


xjbeta:
### Version
v1.4.0-beta1 a9e3bb0

### Reproduce
- Open a video
- During playback, use the shortcut keys to repeatedly switch between `normal size`(cmd+1) and `double size`(cmd+2). (Press the shortcut key after the window animation is completed)

### Behavior
`mac mini M4` Intermittent UI freezes (spinning wheel cursor)
`mbp m1 max` Part of the video screen is black when the window is resized

This issue also occurs during normal usage, but no reproducible steps have been identified.
(For example, drag the window size with the mouse)

[Behavior recording and .trace file](https://mega.nz/folder/BstW2ZhI#m1Xd9v4EWeeB4XBoP96dqw)

### Analyze

This could be a bug introduced by b0e8fc3
This can't be reproduced on v1.3.5 and d201aaf

@low-batt 

<img width="720" alt="Image" src="https://github.com/user-attachments/assets/6ecb3185-0281-44e7-ba65-849ef4e44807" />

low-batt:
I'm having trouble reproducing it reliably, but I definitely triggered it a few times.

I'm in the middle of some other changes I need to finish. Then I will turn my attention to this issue.

low-batt:
I'm still working on some other changes I need to finish, but wanted to give you some background on the changes in this area.

A bunch of changes deemed disruptive/risky had been postponed for a release that included a beta testing period. Many of those have gone into 1.4.0-beta1. Many fixes to thread data races detected by Xcode's [Thread Sanitizer](https://developer.apple.com/documentation/xcode/diagnosing-memory-thread-and-crash-issues-early) and other thread coordination problems have gone into 1.4.0. The commit b0e8fc3 contains some of those changes and has already caused one performance regression. See issue #5403.

In that case the main thread was having to wait for the `displayLock`. This was unexpected as the `mpvGLQueue` thread that is constantly locking the lock should not be holding the lock for very long. What exactly was happening with the locks was not determined because the root cause in that case was that the main thread was inappropriately making a high volume of calls to `VideoLayer.draw(forced:)`. The problem was solved by removing the needless excessive calls to `VideoLayer.draw(forced:)`. This case could be similar, although I was having trouble proving that in the quick tests I did.

[...]

PR #5550, opened 2025-06-16
Fix short hang upon starting a video, #5531

### LINKED ISSUE #5263: Quickly seeking / Pinching while media is playing causes IINA to lag
### IINA Version

IINA Nightly 6bf7d58

### Release Channel

Nightly

### macOS Version

macOS 15.1

### Steps to Reproduce

1. Launch IINA and play any video
2. Quickly tapping "Left" and "Right" on keyboard **while video is playing**
3. Seeking took longer time to finish, and spinning pinwheel appears

https://github.com/user-attachments/assets/ad2a8cca-9655-4f7d-b46e-df3eb29d9c26

(There are some obvious twitches around 00:13, possibly caused by some pending seek events before)

---

Similar things happen when window size changes quickly:
1. Set "Control" -> "Pinch to:" option to "Adjust window size"
2. Quickly pinching touchpad to resize IINA window **while video is playing**
3. Noticed similar lagging

---

I noticed that neither seeking nor pinching would cause any lag when video is paused.

https://github.com/user-attachments/assets/9817c12c-0b09-43e1-8281-eb6b80401e31

---

Also, providing additional `video-sync=display-resample` mpv option could fix this issue:

截屏2024-11-09 15 29 27

https://github.com/user-attachments/assets/ec8437fa-1525-47aa-b6ee-bd000f82ea03

### Expected Behavior

There should be no any lagging.

### Actual Behavior

Seeking and resizing have obvious lagging.

N/A

### Logs

N/A

### Sample Files

N/A

Always.

### Required Actions

### Optional Actions

### LINKED ISSUE #5531: IINA upon starting a video sometimes hangs for ~3 seconds before being able to interact with the file
### IINA Version

1.4.0-beta2 Build 153 (ba0bfa8)

### Release Channel

Nightly

### macOS Version

15.5 (24F74)

### Steps to Reproduce

- Open a video file (I've seen this with .mkv files cause that's the form factor I watch with mainly)

I've noticed it happens more often upon first opening IINA, but it's inconsistent to reproduce as sometimes it loads perfectly fine and other times it will hang as I am trying to show here

### Expected Behavior

- Video playback interaction is smooth and works as it should in IINA

### Actual Behavior

- Video file will have the Mac spinning wheel upon the start of video, being unable to interact with anything in IINA till it goes away

_No response_

### Logs

[iina.log](https://github.com/user-attachments/files/20580396/iina.log)

### Sample Files

_No response_

Sometimes, most often upon a fresh start up

### Required Actions

### Optional Actions

low-batt:
This commit will:
- Add a `forceDraw` method to `PlayerWindowController` that checks to see if drawing is needed before calling `VideoLayer`
- Change all code that was forcing drawing to use `PlayerWindowController.forceDraw`
- Change force drawing to use the `mpvGLQueue` queue instead of the main thread
- Rename `VideoLayer.draw(forced)` to `update(force)` to match mpv
- Rename the `VideoLayer` properties `needsMPVRender` and `forceRender` to be `needsFlip` and `forceDraw` to match mpv
- Add an `inLiveResize` property to `VideoLayer` and change live resize code to use it to match up with how mpv handles resizing
- Add `@Atomic` to some `VideoLayer` properties to correct TSAN errors

This corrects the momentary spinning beach ball of death sometimes seen when initially opening a file.

- [x] This implements/fixes issue #5531.

---

**Description:**

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -692,7 +692,9 @@ class MainWindowController: PlayerWindowController {
       NSScreen.log("NSScreen.screens[\(screen.offset)]" , screen.element)
     }
 
-    videoView.videoLayer.draw(forced: true)
+    // If a video is not actively playing then the initial drawing of the view needs to be forced.
+    // The forceDraw method will check to see if drawing is actually needed.
+    forceDraw("window loaded")
   }
 
   /// Returns the position in seconds for the given percent of the total duration of the video the percentage represents.
@@ -1324,7 +1326,7 @@ class MainWindowController: PlayerWindowController {
     videoViewConstraints.values.forEach { $0.constant = 0 }
     videoView.needsLayout = true
     videoView.layoutSubtreeIfNeeded()
-    videoView.videoLayer.draw(forced: true)
+    forceDraw("entered full screen mode")
 
     if Preference.bool(for: .blackOutMonitor) {
       blackOutOtherMonitors()
@@ -1434,7 +1436,7 @@ class MainWindowController: PlayerWindowController {
     videoViewConstraints.values.forEach { $0.constant = 0 }
     videoView.needsLayout = true
     videoView.layoutSubtreeIfNeeded()
-    videoView.videoLayer.draw(forced: true)
+    forceDraw("exited full screen mode")
 
     if Preference.bool(for: .pauseWhenLeavingFullScreen) && player.info.state == .playing {
       player.pause()
@@ -1447,7 +1449,7 @@ class MainWindowController: PlayerWindowController {
 
     resetCollectionBehavior()
     updateWindowParametersForMPV()
-    
+
     player.events.emit(.windowFullscreenChanged, data: false)
   }
 
@@ -1663,7 +1665,7 @@ class MainWindowController: PlayerWindowController {
   }
 
   func windowWillStartLiveResize(_ notification: Notification) {
-    videoView.videoLayer.isAsynchronous = true
+    videoView.videoLayer.inLiveResize = true
   }
 
   // resize framebuffer in videoView after resizing.
@@ -1671,7 +1673,7 @@ class MainWindowController: PlayerWindowController {
     // Must not access mpv while it is asynchronously processing stop and quit commands.
     // See comments in windowWillExitFullScreen for details.
     guard player.info.state.active else { return }
-    videoView.videoLayer.isAsynchronous = false
+    videoView.videoLayer.inLiveResize = false
     updateWindowParametersForMPV()
   }
 
@@ -2213,11 +2215,7 @@ class MainWindowController: PlayerWindowController {
       videoView.needsLayout = true
       videoView.layoutSubtreeIfNeeded()
       // force rerender a frame
-      videoView.videoLayer.mpvGLQueue.async {
-        DispatchQueue.main.sync {
-          self.videoView.videoLayer.draw()
-        }
-      }
+      forceDraw("interactive cropping")
     }
 
     let controlView = mode.viewController()
@@ -3110,14 +3108,11 @@ extension MainWindowController: PIPViewControllerDelegate {
 
     addVideoViewToWindow()
 
-    // Similarly, we need to run a redraw here as well. We check to make sure we
-    // are paused, because this causes a janky animation in either case but as
-    // it's not necessary while the video is playing and significantly more
-    // noticeable, we only redraw if we are paused.
-    let currentTrackIsAlbumArt = player.info.currentTrack(.video)?.isAlbumart ?? false
-    if player.info.state == .paused || currentTrackIsAlbumArt {
-      videoView.videoLayer.draw(forced: true)
-    }
+    // Similarly, we need to run a redraw here as well. We check to make sure we are paused, because
+    // this causes a janky animation in either case but as it's not necessary while the video is
+    // playing and significantly more noticeable, we only redraw if we are paused. The forceDraw
+    // method checks to make sure drawing is required.
+    forceDraw("exiting PiP")
 
     updateTimer()
 
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -1164,7 +1164,7 @@ class MainWindowController: PlayerWindowController {
       if recognizer.state == .began {
         // began
         lastMagnification = recognizer.magnification
-        videoView.videoLayer.isAsynchronous = true
+        videoView.videoLayer.inLiveResize = true
         frameWhenStartedPinching = window.frame
       } else if recognizer.state == .changed {
         // changed
@@ -1181,7 +1181,7 @@ class MainWindowController: PlayerWindowController {
         lastMagnification = recognizer.magnification
       } else if recognizer.state == .ended {
         updateWindowParametersForMPV()
-        videoView.videoLayer.isAsynchronous = false
+        videoView.videoLayer.inLiveResize = false
       }
     }
   }
--- a/iina.xcodeproj/project.pbxproj
+++ b/iina.xcodeproj/project.pbxproj
@@ -161,6 +161,8 @@
 		51096C0B2CD88BEC00702C31 /* MPVOptionDefaults.swift in Sources */ = {isa = PBXBuildFile; fileRef = 51096C0A2CD88BEC00702C31 /* MPVOptionDefaults.swift */; };
 		5111B18B2C8BCEAC00DEC0AC /* TouchBarSettings.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5111B18A2C8BCEAC00DEC0AC /* TouchBarSettings.swift */; };
 		5114FA942DC704CE00D02435 /* NowPlayingInfoManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5114FA932DC704CE00D02435 /* NowPlayingInfoManager.swift */; };
+		51165FBC2E1D91290060B817 /* ReadWriteAtomic.swift in Sources */ = {isa = PBXBuildFile; fileRef = 51165FBB2E1D91290060B817 /* ReadWriteAtomic.swift */; };
+		51165FC02E1D91940060B817 /* ReadWriteLock.swift in Sources */ = {isa = PBXBuildFile; fileRef = 51165FBF2E1D91940060B817 /* ReadWriteLock.swift */; };
 		512B8FDD2BC2376E00AF41BF /* OutlineView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 512B8FDC2BC2376E00AF41BF /* OutlineView.swift */; };
 		513A4FFA29B53F8100A8EA7D /* Atomic.swift in Sources */ = {isa = PBXBuildFile; fileRef = 513A4FF929B53F8100A8EA7D /* Atomic.swift */; };
 		515B5E5A2A579903001FCD49 /* iina-plugin in Copy Executables */ = {isa = PBXBuildFile; fileRef = E38558872A484A2D0083772D /* iina-plugin */; settings = {ATTRIBUTES = (CodeSignOnCopy, ); }; };
@@ -1189,6 +1191,8 @@
 		51096C0A2CD88BEC00702C31 /* MPVOptionDefaults.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = MPVOptionDefaults.swift; sourceTree = "<group>"; };
 		5111B18A2C8BCEAC00DEC0AC /* TouchBarSettings.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TouchBarSettings.swift; sourceTree = "<group>"; };
 		5114FA932DC704CE00D02435 /* NowPlayingInfoManager.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = NowPlayingInfoManager.swift; sourceTree = "<group>"; };
+		51165FBB2E1D91290060B817 /* ReadWriteAtomic.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ReadWriteAtomic.swift; sourceTree = "<group>"; };
+		51165FBF2E1D91940060B817 /* ReadWriteLock.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ReadWriteLock.swift; sourceTree = "<group>"; };
 		512B8FDC2BC2376E00AF41BF /* OutlineView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = OutlineView.swift; sourceTree = "<group>"; };
 		513A4FF929B53F8100A8EA7D /* Atomic.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = Atomic.swift; sourceTree = "<group>"; };
 		51537C7229FA26DD00F9A472 /* Nightly.xcconfig */ = {isa = PBXFileReference; lastKnownFileType = text.xcconfig; path = Nightly.xcconfig; sourceTree = "<group>"; };
@@ -2414,6 +2418,7 @@
 				84A886E21E24F2D3008755BB /* Sub */,
 				8497A4831D2FF573005F504F /* iina-Bridging-Header.h */,
 				513A4FF929B53F8100A8EA7D /* Atomic.swift */,
+				51165FBF2E1D91940060B817 /* ReadWriteLock.swift */,
 				51854CD92A1C489300C40B78 /* FFmpegLogger.swift */,
 				84EB1F061D2F5E76004FA5A1 /* Utility.swift */,
 				8461C52F1D462488006E91FF /* VideoTime.swift */,
@@ -2446,6 +2451,7 @@
 				51C1BA39291CA76700C1208A /* InfoDictionary.swift */,
 				51DE55C82A6646710050AD06 /* Sysctl.swift */,
 				519CCABC2BFFAEF10079DCAF /* HardwareDecodeCapabilities.swift */,
+				51165FBB2E1D91290060B817 /* ReadWriteAtomic.swift */,
 			);
 			name = Utils;
 			sourceTree = "<group>";
@@ -3194,6 +3200,7 @@
 				840AF5821E732C8F00F4AF92 /* JustXMLRPC.swift in Sources */,
 				E3513AF820EF79C500F8C347 /* PreferenceWindowController.swift in Sources */,
 				841A599D1E1FF5800079E177 /* SleepPreventer.swift in Sources */,
+				51165FBC2E1D91290060B817 /* ReadWriteAtomic.swift in Sources */,
 				84ED99FF1E009C8100A5159B /* PrefAdvancedViewController.swift in Sources */,
 				840AF5841E73CE3900F4AF92 /* OpenSubSubtitle.swift in Sources */,
 				E32B157D21BC27B100CDBCEA /* JavascriptAPIMenu.swift in Sources */,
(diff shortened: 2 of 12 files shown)


low-batt:
At the moment I'm thinking it might be better to not make the changes in this PR and instead move to using the main thread for all drawing in order to align with mpv. I am worried about these changes since the last time drawing was changed to only use the `mpvGLQueue` queue there were problems.

How mpv handles this can be seen [here](https://github.com/mpv-player/mpv/blob/master/video/out/mac/gl_layer.swift). Always using the main thread simplifies the code.

You can also see that the mpv code does a bunch of stuff that IINA probably needs to do as well. I was expecting we'd work on that in the next release. Or do you want me to port over more changes?

There are subtle differences that might be significant. In IINA's `createContext` you will see this:
```swift
    // Enable multi-threaded GL engine.
    CGLEnable(ctx, kCGLCEMPEngine)
```

I believe mpv suspected there were defects in the multi-threaded GL engine and removed that at some point. These are the kind of differences that made me hesitate on making IINA fully match the mpv behavior.

Thoughts?

iina/MainWindowController.swift
```diff
@@ -2213,11 +2215,7 @@ class MainWindowController: PlayerWindowController {
       videoView.needsLayout = true
       videoView.layoutSubtreeIfNeeded()
       // force rerender a frame
-      videoView.videoLayer.mpvGLQueue.async {
-        DispatchQueue.main.sync {
-          self.videoView.videoLayer.draw()
-        }
-      }
```
low-batt, line 2220: I found this code very confusing. The comment says it is forcing a rendering, but the forced parameter was not passed to draw. I also don't understand the synchronized dispatch to main. That is usually a no-no. This looks like we wanted to block the `mpvGLQueue` while drawing? I tested this code and the change didn't seem to break functionality, but it worries me I don't understand the need for the current code.

lhc70000:
If the window can still be resized smoothly when using the main thread for all drawing, we can go for it. This was the main reason why we didn't draw on main thread when IINA was created.

low-batt:
I will make that change and test resizing and other animations.

low-batt:
Exhausted after fighting with this all day. I hit two problems.

The first problem I've seen before. It was so area I never was able to reproduce it. IINA would hang upon loading a video in `PlayerCore.fileLoaded` calling into mpv to set a property. For some reason drawing makes that happen a lot. From a quick check of mpv sources mpv does not do the kind of work IINA is doing upon getting the file loaded event. I did not have time to check where they do that kind of work. I just hacked around this problem to be able to see how well playback works.

Playback seemed ok. Resizing seemed fine. I've not tested enough because I hit a problem. I had trouble grabbing the window. If you mouse down and move the cursor the window fails to move. Then all of a sudden it jumps to where the cursor is and from that point follows the mouse without hesitation. As if it is initially stuck on some lock that then times out. I'm still trying to figure out what is causing this.

Using Instruments it does not look like drawing will overload the main thread. This was on a 2019 MacBook Pro. I did not have time to test on my MacBook Air.

I'll have to think on how to track down what is causing dragging of the window to initially lock up.

svobs:
> The first problem I've seen before. It was so area I never was able to reproduce it. IINA would hang upon loading a video in PlayerCore.fileLoaded calling into mpv to set a property.

You're sure it's not one of those properties which can deadlock the mpv core if called before `mpv_render_context_create` & `mpv_render_context_set_update_callback` have finished, right?

> If the window can still be resized smoothly when using the main thread for all drawing, we can go for it. This was the main reason why we didn't draw on main thread when IINA was created.

Should be fine now that `videoLayer.isAsynchronous` is being set to true before all types of window resize (#5013).

low-batt:
> You're sure it's not one of those properties which can deadlock the mpv core if called before mpv_render_context_create & mpv_render_context_set_update_callback have finished, right?

I believe you have brought up before in some other issue that there are some race conditions having to do with window creation. This may have something to do with those race conditions, but it isn't a case of rendering not being initialized. Rendering is initialized in `MainWindowController.windowDidLoad` when it calls `PlayerCore.initVideo` which then calls `MPVController.mpvInitRendering`. I added debug logging and I am seeing rendering having been initialized before the `MPV_EVENT_FILE_LOADED` is received when the hang happens.

The mpv code that handles  `MPV_EVENT_FILE_LOADED` does very little. I need to dig into the mpv code and see where they do all the work IINA is doing upon receiving that event.

In the following process sample the main thread looks like this:
```text
    +                                       6947 PlayerCore.fileLoaded()  (in IINA.debug.dylib) + 1028  [0x10994fc74]  PlayerCore.swift:1950
    +                                         6947 MPVController.getInt(_:)  (in IINA.debug.dylib) + 226  [0x1099d2882]  MPVController.swift:843
    +                                           6947 mpv_get_property  (in libmpv.2.dylib) + 101  [0x10d8059c5]
	    +                                             6947 mp_dispatch_lock  (in libmpv.2.dylib) + 205  [0x10d7e980d]
```

[Process Sample fileLoaded.txt](https://github.com/user-attachments/files/20783681/Process.Sample.fileLoaded.txt)

The hang occurs on a call to either set or get a mpv property around this range of `PlayerCore.fileLoaded`
https://github.com/iina/iina/blob/3673eb8e04576167a90c477d25aedcad22fa5712/iina/PlayerCore.swift#L1942-L1953

I suspect this is a rare race condition that switching to drawing on the main thread brings out. It needs to be investigated, but I've been focusing on whether we can use the main thread for drawing like mpv does. I now believe the answer is no, based on mpv issue [#14173](https://github.com/mpv-player/mpv/issues/14173).

I have a lot more to say on this, but Tuesday is my busy day and I'm out of time. Later…

low-batt:
I had time to do more investigation into the significant lagging when IINA has been changed to use the main thread for drawing. I found the severe lagging when dragging or resizing the window while the video was playing was caused by [BetterTouchTool](https://folivora.ai/). When I quit BTT the behavior significantly improved. However the sliding animation to show and hide the side panels was _very_ sluggish. Dragging the floating OSC around was jerky. I've concluded drawing on the main thread is not viable for IINA.

I believe issue #5531 is a must fix for beta2 because changes in this area of IINA can introduce problems that are rare and thus may not show up until a lot of users test IINA. Changes are needed in this area in order to address data races detected by Xcode's [Thread Sanitizer](https://developer.apple.com/documentation/xcode/diagnosing-memory-thread-and-crash-issues-early) .

I think we have to go with the changes in this PR that use `mpvGLQueue` for all drawing.

Please review!

On this:
> Should be fine now that videoLayer.isAsynchronous is being set to true before all types of window resize (#5013).

This PR changes that slightly. I ported changes from mpv that use a `inLiveResize` property to control `isAsynchronous`. This ensures there is one more draw at the end of resizing.

lhc70000:
My several thoughts:

1. If we have always been good drawing on background threads, we don't need to switch to the main thread. 

2. According to my test, when `isAsynchronous = true`, although the docs say the system will call display() periodically, it actually only calls it ONCE on the main thread. If we continue drawing on the background, all draws will still happen on the background thread.
    - However, with this PR, if you disable drawing on the background thread during live resize, all drawing will be on the main thread, basically in line with the docs.
    - It's not a good idea to do so, since (1) the timing is controlled by the system rather than mpv, and (2) the resizing animation will become noticeably laggy.

4. I think the `isUninited` lock blocks too many things, since we use it in the CVDisplayLink callback (report swap), mpvGLQueue and on the main thread. It's also the cause of #5531. We should not limit all these operations to be one at a time. I tried to change it to a Read Write Lock that allows simultaneous reading. The bottom line is that mpv has its internal lock, so even there are multiple draws happening at the same time, it should be safe.
    - Furthermore, we now have many
       ```swift
      videoView.$isUninited.withLock() { isUninited in
        guard !isUninited else { return }
        // do other stuff
      }
      ```
      Are these necessary? Can we change them to
      ```swift
      videoView.$isUninited.withLock() { isUninited in
        guard !isUninited else { return }
      }
      // do other stuff
      ```

5. I saw that @low-batt changed the `drawLock` in #5490.  Why moved it from `display()` to `draw(forced:)`? Previously it was `displayLock` that prevented concurrent calls to `display()`, which makes sense since calling `display()` from a background thread is an undocumented behavior. But now it's in `draw(forced:)`, this seems unnecessary since we already have the isUninited lock, and unsafe since the system can call `display()` on the main thread when `isAsynchronous = true`.

I have pushed my changed related to (4) and (5) to the `test-read-write-lock` branch. Please have a test.

low-batt:
On this:
> If we have always been good drawing on background threads, we don't need to switch to the main thread.

I looked into this in depth and IINA can not switch to using the main thread.

The main push for this was to match up with mpv. This also would avoid the need for locking. I tested this and ran into trouble with significant lagging. Turns out mpv ran into trouble with lagging as well, when using OpenGL. It appears they decided to live with it because mpv performs better with the Metal layer, which is now the default.

Using the main thread also significantly disrupted the sliding animated transition of the side bar. So using the main thread is definitely not viable for IINA.

On this:
> According to my test, when isAsynchronous = true, although the docs say the system will call display() periodically, it actually only calls it ONCE on the main thread. …

That came from PR #5013, so we need to bring @uiryuu into this discussion.

> I think the isUninited lock blocks too many things…

When accessing mpv from a background thread the `isUninited` lock **must** be held. If not, IINA can crash when quitting. mpv has a similar lock.

> I saw that @low-batt changed the drawLock in Fix UI freeze on resize video window, #5482 #5490.

That was an attempt to address trouble with the main thread blocking. The lock should be in `display`. That is what mpv uses. I've been trying to figure out what is difference about mpv that it does not show problems with holding up the main thread.

> I have pushed my changed related to (4) and (5) to the test-read-write-lock branch. Please have a test.

I will take a look at it.

low-batt:
When in the test-read-write-lock branch In Xcode edit the scheme and enable `Thread Sanitizer` and `Undefined Behavior Sanitizer`. Then test. I'm having trouble reproducing it, but TSAN sometimes is detecting data race conditions:

[TSAN-errors.txt](https://github.com/user-attachments/files/21093947/TSAN-errors.txt)

That means something is wrong with the locking in that branch.

low-batt:
The test-read-write-lock branch was made off of the `develop` branch, not this PR's branch. Definitely in the `develop` branch the `isUninited` lock is used for more than coordinating shutdown. To avoid the need for additional locks that lock is held when _modifying_ other state in `ViewLayer`. So the reason for the above failures is that effectively a _read_ lock is being held when writing. Using a read/write lock might be viable. I will look into this tomorrow.

lhc70000:
> On this:
>
>> According to my test, when isAsynchronous = true, although the docs say the system will call display() periodically, it actually only calls it ONCE on the main thread. …

>
> That came from PR https://github.com/iina/iina/pull/5013, so we need to bring @uiryuu into this discussion.

Strangely, even though it only draws once on the main thread according to my log, it still works to prevent the jittering when resizing the window. So `isAsynchronous` is still needed and we don't want to change this.

low-batt:
I'm confused as well. In a quick test with the changes in this PR and adding some logging in `canDraw` and `draw` I'm seeing drawing switching to the main thread during a resize:

<details>
<summary><b>Test run:</b></summary>

```text
19:24:35.993 [iina][d] com.colliderli.iina.mpvgl draw
19:24:36.034 [iina][d] com.colliderli.iina.mpvgl canDraw true
19:24:36.034 [iina][d] com.colliderli.iina.mpvgl draw
19:24:36.071 [window0][d] ********* windowWillStartLiveResize
19:24:36.071 [iina][d] com.colliderli.iina.mpvgl canDraw true
19:24:36.071 [iina][d] com.colliderli.iina.mpvgl draw
19:24:36.111 [iina][d] com.apple.main-thread canDraw true
19:24:36.112 [iina][d] com.apple.main-thread draw
19:24:36.174 [iina][d] com.apple.main-thread canDraw true
19:24:36.174 [iina][d] com.apple.main-thread draw
19:24:36.199 [iina][d] com.apple.main-thread canDraw true
19:24:36.199 [iina][d] com.apple.main-thread draw
19:24:36.237 [iina][d] com.apple.main-thread canDraw true
19:24:36.238 [iina][d] com.apple.main-thread draw
19:24:36.278 [iina][d] com.apple.main-thread canDraw true
19:24:36.278 [iina][d] com.apple.main-thread draw
19:24:36.320 [iina][d] com.apple.main-thread canDraw true
19:24:36.320 [iina][d] com.apple.main-thread draw
19:24:36.360 [iina][d] com.apple.main-thread canDraw true
19:24:36.360 [iina][d] com.apple.main-thread draw
19:24:36.397 [iina][d] com.apple.main-thread canDraw true
19:24:36.397 [iina][d] com.apple.main-thread draw
19:24:36.438 [iina][d] com.apple.main-thread canDraw true
19:24:36.438 [iina][d] com.apple.main-thread draw
19:24:36.479 [iina][d] com.apple.main-thread canDraw true
19:24:36.479 [iina][d] com.apple.main-thread draw
19:24:36.521 [iina][d] com.apple.main-thread canDraw true
19:24:36.521 [iina][d] com.apple.main-thread draw
19:24:36.553 [iina][d] com.apple.main-thread canDraw true
19:24:36.553 [iina][d] com.apple.main-thread draw
19:24:36.592 [window0][d] ********* windowDidEndLiveResize
19:24:36.592 [mpv0][v] Set property: window-scale=1.6290322580645162
19:24:36.592 [iina][d] com.colliderli.iina.mpvgl canDraw true
19:24:36.594 [iina][d] com.colliderli.iina.mpvgl draw
19:24:36.635 [iina][d] com.colliderli.iina.mpvgl canDraw true
19:24:36.636 [iina][d] com.colliderli.iina.mpvgl draw
```
</details>

I'm looking into the read/write lock changes…

low-batt:
I figured out what was wrong with the read write lock. I have that working. Before adding those changes to the PR I'd like to reproduce the problems you are hitting that the read write lock addresses.

I'm not seeing any lagging or hangs reported by Instruments. Playing a downloaded copy of the video [New York 8k | 4k HDR](https://vimeo.com/196168570) I do see glitches when entering and exiting full screen mode. Hard to describe. Like a flicker.

What video are you playing that shows lagging?

low-batt:
This is the change I made to the property wrapper:
```swift
  func withReadLock<R>(_ body: (Value) throws -> R) rethrows -> R {
    return try lock.read {
      return try body(value)
    }
  }

  func withWriteLock<R>(_ body: (inout Value) throws -> R) rethrows -> R {
    return try lock.write {
      return try body(&value)
    }
  }
```

`VideoView.uninit` needs to use `withWriteLock`.

lhc70000:
>  I do see glitches when entering and exiting full screen mode. Hard to describe. Like a flicker.

This is an old issue and needs further investigation. It draws an extra frame with wrong size after the animation. I will check this out after fixing this issue.

> Before adding those changes to the PR I'd like to reproduce the problems you are hitting that the read write lock addresses.

So, what I mean is that, I tried to use read write lock to resolve #5531 directly.

I think the problem in #5531 is caused by both the main thread and the background thread trying to acquire the uninit lock, which is actually unnecessary. In instruments, sometimes I also observe micro hang when the display link callback tries to acquire the uninit lock. 

- From our previous discussion, we still need to have `isAsynchronous` and we don't want to stop drawing on the background queue during live resize.
  - So we need to face the problem when the main thread and background thread draw at the same time.
    - So we can't simply move all drawing to the mpv GL queue as in this PR.
  - This is actually the same problem in 5531.
  - I start from the develop branch and tried to find another solution.
    - I observed that these hangs are caused by multiple threads trying to acquire the uninit lock.
      - So I believe simply changing the uninit lock to a read write lock can resolve this.

lhc70000:
I did many tests on the `test-read-write-lock` branch and found that the `drawLock` can still cause unresponsiveness. This lock cannot be a read write lock since we don't want `display()` to be called by multiple threads simultaneously. But there are solutions to minimize the cases when we have both the main thread and background thread trying to draw:

- When live resizing: I found that we can simply skip all drawing calls from the main thread, and this can still prevent jittering.
- On file start (5531): I think we can move all force drawing calls to the mpv GL queue, as in this PR.

Note that we still need the uninit lock to be a read write lock, since the display link thread also acquire this lock.

I have updated the `test-read-write-lock` branch to address this. When running Instruments I can see the least unresponsiveness I've ever seen on the main thread.

Do you think this branch alone can solve all the problems?

low-batt:
> Do you think this branch alone can solve all the problems?

Not in its current form. Testing the latest update on that branch, there are data races because it is missing the changes in this PR that are needed to use `isUninited` as a read/write lock. In quick tests I'm experiencing jitter.

Normally I will break up changes into separate PRs, but since we are so slow to review and merge this PR combines a bunch of changes that are needed. If we merge the test-read-write-lock branch I will need to apply some of the fixes in this PR afterwards. So I've added the read/write lock to this PR along with the change to ignore certain drawing calls. I also added back in `displayLock`. From a quick test I'm still experiencing jitter with these changes.

You brought up what I was going to ask about. I'm questioning the need for any forced drawing. When playback is in progress shouldn't there be no need for the main thread runloop to draw the video view? In other words maybe set
[layerContentsRedrawPolicy](https://developer.apple.com/documentation/appkit/nsview/layercontentsredrawpolicy-swift.property) to [never](https://developer.apple.com/documentation/appkit/nsview/layercontentsredrawpolicy-swift.enum/never)? And when playback is not in progress shouldn't drawing always be forced so that the current calls to force drawing would not be needed because the normal main thread runloop flush would do the drawing. Or am I confused?

Tuesday's are my busy day, so I made the changes in a rush and haven't had time to run Instruments and see how things behave with the latest commit.

As the changes in this PR centralize forced drawing it is easy to change that to the main thread if you think that is needed.

low-batt:
I suspect I put the main thread check in the wrong place. That will need to be moved. No time at the moment. Gotta run.

low-batt:
I've moved the check for the main thread into `display`. Please give the latest commit a try.

lhc70000:
You still disable drawing on the background queue during live resize. This will make live resizing laggy (you can feel it when quickly pinching with the trackpad). We shouldn't draw on the main thread (continuously) at any time.

- Moving that check to `display()` does nothing, because CALayer calls `canDraw` and `draw` directly when isAsynchronous. 
-  Drawing shouldn't be skipped on the background queue during live resize.

`startup-hang`:
image

`test-read-write-lock` (skip main thread drawing during live resize):
image

> Testing the latest update on that branch, there are data races because it is missing the changes in this PR that are needed to use isUninited as a read/write lock. 

Could you point out which line causes the data race?

>  In quick tests I'm experiencing jitter.

Do you mean image jittering during live resize?

low-batt:
I'm super confused…

On this:
> Moving that check to display() does nothing, because CALayer calls canDraw and draw directly when isAsynchronous.

I moved it to `display` because that is where it is in the `test-read-write-lock` branch. Do you have local changes not pushed to that branch?

The `test-read-write-lock` branch has this code which uses the `isUnited` to coordinate writing to the flags. For that to work you must use `withWriteLock`:
```swift
  func draw(forced: Bool = false) {
    videoView.$isUninited.withReadLock() { isUninited in
      // The properties forceRender and needsMPVRender are always accessed while holding
      // isUninited's lock. This avoids the need for separate locks to avoid data races with these
      // flags. No need to check isUninited at this point.
      needsMPVRender = true
      if forced { forceRender = true }
    }
```

The better solution is to start with the changes in this PR where a bunch of this has been corrected and `isUninited` is only used to coordinate shutdown.

> Do you mean image jittering during live resize?

Effectively yes. I was testing entering and exiting full screen mode.

low-batt:
I originally had the check in `canDraw`. I then remembered it was in `display` in the `test-read-write-lock` branch and added a commit that moved the check into `display` to match the code in `test-read-write-lock`. I just deleted that commit so the check is back in `canDraw`.

low-batt:
The latest commit corrects the problem with not continuing to draw using the background thread when resizing. And it changes the two flags to use thread local storage.

low-batt:
Just pushed a minor change to avoid constantly creating the keys used with `threadDictionary`.

lhc70000 (approved):

PR #5558, opened 2025-06-21
chore: sleep prevention optimization & cleanup

svobs:

---

**Description:**
* Rename `PlayerCore.playing` to `PlayerCore.nonIdle` to avoid confusion with playing/paused states
* Move `checkStatusForSleep` into `SleepPreventer.swift`, rename to `updateSleepPrevention` to better indicate that it will change state
* Simplify logic
* Add yet more comments

### DIFF
--- a/iina/MenuController.swift
+++ b/iina/MenuController.swift
@@ -743,7 +743,7 @@ class MenuController: NSObject, NSMenuDelegate {
   }
 
   private func updateOpenMenuItems() {
-    if PlayerCore.playing.count == 0 {
+    if PlayerCore.nonIdle.count == 0 {
       open.title = stringForOpen
       openAlternative.title = stringForOpen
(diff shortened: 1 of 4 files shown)


svobs:
Side note: seems like there should be a standard way to describe the type of update which re-evaluates the state then reapplies it. Here I changed it to `update*` but there is also `refreshSyncUITimer` which does the same kind of thing. I never much liked the word "refresh" (though I often use it), since it sounds to me like one of those kludgey Microsoft-isms. Maybe `reapply*`? Seems like the React people should have figured this out, but seems like (based on my limited familiarity with React) the closest equivalent there would be "reduce", which seems slightly misleading. Not the biggest deal really.

uiryuu (approved):

uiryuu:
Thanks!

PR #5562, opened 2025-06-25
Change Logger.log to take a closure for the message

low-batt:
This commit will:
- Change the type of the message parameter of `Logger.log` to be `() -> String`
- Add a `log` method to `Logger` that takes a message parameter of type `@autoclosure () -> String` and calls the existing `log `method
- Change the type of the message parameter of utility log methods supply the subsystem of the associated class to be `@autoclosure () -> String`

Passing the message as a closure instead of a string is useful when the message includes string interpolations as it allows the evaluation of the expressions and construction of the string to be delayed until it is known that the message will be written to the log file and not discarded due to logging either being disabled or configured to not emit messages at the required level.

---

**Description:**

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -1307,7 +1307,7 @@ class RemoteCommandController {
 
   // MARK: - Private Functions
 
-  private func log(_ message: String, level: Logger.Level = .debug) {
+  private func log(_ message: @autoclosure () -> String, level: Logger.Level = .debug) {
     Logger.log(message, level: level, subsystem: Logger.Sub.nowPlaying)
   }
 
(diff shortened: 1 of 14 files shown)


svobs:
`@autoclosure`. Brilliant. 😅

uiryuu (approved):

low-batt:
I was looking for a way to do this and noticed this is how Apple handles this as seen in [assert](https://developer.apple.com/documentation/swift/assert(_:_:file:line:)) and other methods. Seemed so wasteful to prepare a complex log message with embedded dynamic information only to have that string discarded in the usual case where logging is not enabled. This seems like a good solution to that.

PR #5585, opened 2025-07-13
Fix buffering 100% UI appears when seeking, #5580

### LINKED ISSUE #5580: Buffering 100% UI appears when seeking HTTP streams even if fully cached
### IINA Version

1.4.0-beta1(build 152)

### Release Channel

Beta

### macOS Version

15.5

### Steps to Reproduce

1. Play a direct HTTP video link such as:
   - Local example: `http://192.168.2.4:8096/emby/videos/1193802/original.mp4?...&Static=true`
   - Public example（please test this）: `https://devstreaming-cdn.apple.com/videos/wwdc/2016/510ndmh9wkcvzneegv2/510/510_sd_validating_http_live_streams.mp4?dl=1`
2. Wait until the entire video is fully cached .
3. Seek to any part of the video or press left/right to fast-forward/backward.

### Expected Behavior

When the video is already fully cached — or when seeking within a part of the video that has already been buffered — the buffering UI should not appear.

The buffering UI should only appear if actual buffering is needed, and it should reflect the real progress (e.g. 20%, 75%), not just "100%".

Or provide a setting to disable this display.

### Actual Behavior

Even when the entire video is fully cached, the UI still shows a "Buffering 100%, 0B" message when seeking.

This also happens when seeking into uncached parts — it only shows 100%, which is misleading and not useful.
（This issue does **not occur** when playing the exact same video via a WebDAV mount added through Finder in the local network （I have not tested this via WebDAV over the internet.））

_No response_

### Logs

_No response_

### Sample Files

_No response_

every time

### Required Actions

### Optional Actions

low-batt:
This commit will:
- Add a new `Seeking…` throbber
- Change `MainWindowController.updateNetworkState` to distinguish between buffering and seeking and not display `cache-buffering-state` when seeking
- Change the formatting of `cacheUsed` to use the correct units
- Change `PlayerCore.syncUI` to use the correct type for `fw-bytes`
- Change `syncUI` to use `demuxer-cache-state/raw-input-rate`, which was obtained when getting the value of `fw-bytes`, instead of getting `cache-speed` from mpv

This corrects multiple problems:
- The percent buffered was being displayed when seeking, but that value is only valid when `paused-for-cache` is `true`
- The number of bytes buffered was being converted to an `Int`, but mpv returns an `Int64`, resulting in the value always being displayed as zero
- The number of bytes buffered was being formatted as if it was in KiB instead of a byte count

This commit corrects the defects reported in #5580, however it does provide the ability to suppress the throbber while seeking. That will be addressed in a separate PR.

- [x] This implements/fixes issue #5580.

---

**Description:**

### DIFF
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -37,6 +37,7 @@
 // MainWindowController.swift
 "main.buffering_indicator" = "Buffering… %d%%";
 "main.opening_stream" = "Opening stream…";
+"main.seeking_indicator" = "Seeking…";
 "osc.precision" = "Precision";
 "osc.1s" = "1 second";
 "osc.100ms" = "100 milliseconds";
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -2755,19 +2755,30 @@ class MainWindowController: PlayerWindowController {
     }
   }
 
+  /// Update the state of the throbber indicating buffering or seeking is occurring.
+  /// - Important: The mpv
+  ///     [cache-buffering-state](https://mpv.io/manual/stable/#command-interface-cache-buffering-state)
+  ///     property is only valid when
+  ///     [paused-for-cache](https://mpv.io/manual/stable/#command-interface-paused-for-cache) is `true`
+  ///     and can not be used to provide an indication of progress when seeking.
   func updateNetworkState() {
-    let needShowIndicator = player.info.pausedForCache || player.info.isSeeking
-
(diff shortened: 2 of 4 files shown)


low-batt:
See the discussion in the issue regarding trying to detect if seeking is occurring within the cache and not putting up the throbber in that case. Need feedback on whether to make that change.

uiryuu (approved):
Tried this PR with a online video, LGTM

low-batt:
Thanks for reviewing!

Issue #5452, opened 2025-04-12
Attempt to change sub background color closes side panel


low-batt:
### IINA Version

1.4.0-beta1

### Release Channel

Beta

### macOS Version

macOS 13.7.1

### Steps to Reproduce

- Start IINA playing a video with text subtitles
- Click on `Show Subtitles Panel` under the `Subtitles` menu
- Scroll to the bottom of the side panel
- In the `Text style` section under `BACKGROUND` click on the color icon
- Try and select the transparent color by clicking on it
- Instead of selecting the color the panel closes 

![Image](https://github.com/user-attachments/assets/da22bce2-a3f7-41e7-86e6-0da9f5128f93)

### Expected Behavior

The background of text subtitles can be changed to be transparent using the subtitles side panel.

### Actual Behavior

Using IINA 1.4.0-beta1 attempts to set the subtitle background color to transparent close the panel instead of selecting a transparent background.

### How often can you reproduce the issue

Every time.

### Required Actions

- [x] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [x] I tested with the Nightly build and confirmed the issue is not fixed.
- [ ] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

lhc70000:
This is the same issue as #4322.

low-batt:
Agree, same issue. I had totally forgotten about that problem. The colorwells on the subtitle panel will also have to be changed to the expanded colorwells.

yuxiqian:
Just curious, why NSColorWells in Apple's first-party apps could enable alpha support with `.minimal` style, like Preview.app and Mail.app?

<img width="466" alt="Image" src="https://github.com/user-attachments/assets/ed44e6b5-2168-48ad-8632-5d61ff35d2ef" />

<img width="409" alt="Image" src="https://github.com/user-attachments/assets/9e95b374-2b9f-4825-b861-98781a99eeff" />

uiryuu:
> Just curious, why NSColorWells in Apple's first-party apps could enable alpha support with `.minimal` style, like Preview.app and Mail.app?
[...]

PR #5604, opened 2025-07-21
Only close the side panel when mouse up detected in video VideoView

uiryuu:
- [x] This implements/fixes issue #5452.

---

**Description:**
Added a variable in `VideoView` to record the last `mouseUp` event which lands on the `VideoView`. If the event doesn't reach to `VideoView`, then it should not trigger a sidebar closure, for instance when the event is in the `NSColorWell`'s popover (see #5452).


low-batt (approved):
Pulled PR, built IINA, tested.

This corrects the problem with closing the side panel. _However_ it does not fully address #5452. I'll explain that in the issue.

PR #5610, opened 2025-07-25
Fix window is resizable in legacy full screen mode, #5609

### LINKED ISSUE #5609: Window is resizable in legacy full screen mode
### IINA Version

1.4.0-beta1

### Release Channel

Beta

### macOS Version

macOS 15.5

### Steps to Reproduce

- Start IINA
- The settings panel appears
- In the enable `Use legacy full screen` in the `Behavior` section
- Play a video
- Click on `Enter Full Screen` under the `Video` menu
- Move the cursor to the right edge of the screen
- Notice the cursor switches to the resize-left [cursor](https://developer.apple.com/documentation/appkit/nscursor)
- Click and drag right and notice the window shrinks

### Expected Behavior

Window resizing is disabled when in full screen mode.

### Actual Behavior

When `Use legacy full screen mode` is enabled and IINA is in full screen mode it is possible to resize the window.

Image

_No response_

### Logs

_No response_

### Sample Files

_No response_

Every time.

### Required Actions

### Optional Actions

low-batt:
This commit will:
- Change `legacyAnimateToFullscreen` to remove `resizable` from the window `styleMask`
- Change `legacyAnimateToWindowed` to insert `resizable` into the `styleMask`
- Change `legacyAnimateToWindowed` to use `disableAnimations` instead of `legacyFullScreenAnimation`
- Remove `legacyFullScreenAnimation` from `Preferences`
- Change the observer for `didChangeScreenParametersNotification` to not call `setWindowFrameForLegacyFullScreen` if the screen count did not change

The change to remove `resizable` from the window `styleMask` will prevent the window from being resized while in full screen mode.

This commit will also replace the `legacyFullScreenAnimation` setting for which there was no UI with the new `disableAnimations` setting which can be set using IINA's settings panel.

When entering/exiting full screen mode IINA changes the app presentation options to enable/disable hiding of the menu bar and dock. As the menu bar and dock are sliding away the observer for
`didChangeScreenParametersNotification` because the moving of the menu bar and dock result in changes to the dimensions of the visible frame. The change to only call `setWindowFrameForLegacyFullScreen` when the screen count changes eliminates extra drawing that disrupts the full screen transition animations.

- [x] This implements/fixes issue #5609.

---

**Description:**

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -644,18 +644,19 @@ class MainWindowController: PlayerWindowController {
     addObserver(to: .default, forName: NSApplication.didChangeScreenParametersNotification) { [unowned self] _ in
       // This observer handles a situation that the user connected a new screen or removed a screen
       let screenCount = NSScreen.screens.count
-      if self.fsState.isFullscreen && Preference.bool(for: .blackOutMonitor) && self.cachedScreenCount != screenCount {
-        self.removeBlackWindow()
-        self.blackOutOtherMonitors()
+      let countChanged = cachedScreenCount != screenCount
+      if fsState.isFullscreen && Preference.bool(for: .blackOutMonitor) && countChanged {
+        removeBlackWindow()
+        blackOutOtherMonitors()
       }
       // Update the cached value
-      self.cachedScreenCount = screenCount
-      self.videoView.updateDisplayLink()
+      cachedScreenCount = screenCount
+      videoView.updateDisplayLink()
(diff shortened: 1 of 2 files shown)


uiryuu:
Do you also want to disable the animation for legacy fs for Tahoe in this PR? The legacy fs animation is still laggy on this commit.

low-batt:
In the latest commit I've disabled animations for legacy full screen under macOS Tahoe for now.

PR #5616, opened 2025-07-27
Fix invalid entries in audio device list

low-batt:
This commit will:
- Add a new struct `MPVAudioDevice` to represent an audio device
- Change `PlayerCore.getAudioDevices` to return a list of `MPVAudioDevice` objects and filter out devices based on audio output driver setting
- Change `MPVController` to observe the `current-ao` property and inform `PlayerCore` when it changes
- Add a `currentAoChanged` function to `PlayerCore` that attempts to adjust the current selected audio device when the audio output driver changes
- Change `PrefCodecViewController` to update the list of audio devices when the audio output driver changes and attempt to adjust the selected preferred audio device

These changes avoid placing the burden on the user to change the audio device after changing the audio output driver (coreaudio/avfoundation).

See the `PlayerCore.getAudioDevices` documentation comment for a full discussion.

---

**Description:**

### DIFF
--- a/iina.xcodeproj/project.pbxproj
+++ b/iina.xcodeproj/project.pbxproj
@@ -177,6 +177,7 @@
 		51AC1CAB2A9FBF3700DF7079 /* OpenSubClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = 51AC1CAA2A9FBF3700DF7079 /* OpenSubClient.swift */; };
 		51C1BA3A291CA76700C1208A /* InfoDictionary.swift in Sources */ = {isa = PBXBuildFile; fileRef = 51C1BA39291CA76700C1208A /* InfoDictionary.swift */; };
 		51C889D52C42FF7F007B2584 /* PlayerState.swift in Sources */ = {isa = PBXBuildFile; fileRef = 51C889D42C42FF7F007B2584 /* PlayerState.swift */; };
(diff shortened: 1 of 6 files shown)


low-batt:
While we are on this topic, why does IINA use this format for audio devices:

`[LG UltraFine Display Audio] avfoundation/AppleUSBAudioEngine:LG Electronics Inc.:LG UltraFine Display Audio:3241000:2`

As the first part is intended to be human readable text, I would have expected this:

`LG UltraFine Display Audio (avfoundation/AppleUSBAudioEngine:LG Electronics Inc.:LG UltraFine Display Audio:3241000:2)`

Or better just:

`LG UltraFine Display Audio`

Which matches what the user sees in the macOS Sound system settings.

Did we encounter cases where there were two devices with the same human readable text? If so, we could detect that and only include the audio API-specific ID for the entries that could not be distinguished without it.

uiryuu:
> Did we encounter cases where there were two devices with the same human readable text? If so, we could detect that and only include the audio API-specific ID for the entries that could not be distinguished without it.

Yes, exactly. I wanted to display more information to distinguish audio device with same name. But I found even in the system dialog there exist device with same name, so I guess we can also only display the human readable text as well.

image

low-batt:
I will make that change in a separate PR once we get this one merged.

Issue #5665, opened 2025-09-15
IINA 1.4.0 fails to honor the "Resize the window to fit video size" setting.


lundejd:
### IINA Version

1.4.0

### Release Channel

Release

### macOS Version

Tahoe 26.0

### Steps to Reproduce

1. Upgraded to 1.4.0
2. Play video, size window as I like it.
3. Play another video, it snaps back to small window. (not even the size and position I have set for initial play)
4. Tried changing the setting back and forth
5. Tried using advanced commands but those all fail (like no-keepaspect-window)
6. Installed previous version, works like it should again.

### Expected Behavior

Once I size a window it should stay that size and not change, unless of course it's going from horizontal to vertical which is a whole other issue but thats another story. 

### Actual Behavior

Playing a new video causes the window to snap small.

### How often can you reproduce the issue

Every time. Every new video snaps it back even after I re-size.

### Required Actions

- [x] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [ ] I tested with the Nightly build and confirmed the issue is not fixed.
- [ ] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

lhc70000:
Thanks, we are looking into this and aim to resolve it in a few days.

low-batt:
This was fixed in IINA 1.4.1.

Issue #5669, opened 2025-09-16
Window resizes despite "Resize window to fit video size" being set to "Disabled"


Sanseviaria:
### IINA Version

1.4.0

### Release Channel

Release

### macOS Version

15.6.1

### Steps to Reproduce

- Set Initial window size (e.g., Width: 300pt).
- Set Initial window position (e.g., 0 right, 0 top).
- Set "Resize the window to fit video size" to Disabled.
- Open any local or online video playlist.

### Expected Behavior

The video should open at 300px width and be placed in the top-right corner of the screen. When clicking another video in the playlist, the window should remain at 300px width, as it did in versions prior to v1.4.0.

### Actual Behavior

The video opens at 300px width and is placed in the top-right corner of the screen. However, when clicking another video in the playlist, the window enlarges to fit the video's width, ignoring the "Resize window to fit video size: Disabled" setting.

This behavior did not occur in versions prior to v1.4.0.

Additionally, using autofit and autofit-larger in the Additional MPV options has no effect — they are not respected either.

### How often can you reproduce the issue

Always

### Required Actions

- [x] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [ ] I tested with the Nightly build and confirmed the issue is not fixed.
- [ ] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

low-batt:
This was fixed in IINA 1.4.1.

PR #5674, opened 2025-09-16
Fix window resize behavior

lhc70000:
- [x] This implements/fixes issue #5665, #5669.

---

**Description:**

Let's define the expected behavior for window resizing.

- **Initial position**: Only applies when the window is opened for the first time.
  - Internally this is synced with mpv's `geometry`, and we may have different behavior with mpv. That's acceptable since IINA is GUI-focused.
  - Even if the user set `geometry` in their conf, it still only applied once. If the user wants a fixed position, they should also set _Resize to fit_ to disabled.  They can also write plugins / user scripts to further control the window size.
  - The previous screen should still be remembered, and we put the window on that screen, if available.
- **Resize to fit**:
  - **Always**: Every time when a new video is played, resize the window.
  - **When opened manually**: Only when the video is opened manually (From menu, drag & drop, etc.). Navigating through the playlist, whether automatically or manually through shortcuts, should NOT resize the video.
  - **Disabled**: Should NOT resize the video in any cases, other than the first time when we apply the initial position. If the window aspect ratio changed, make minimal change and keep the window's center position.
- When resizing, _minimal_ change should be made. The window's center position should not change, and it should not move to another screen in any circumstances. We can have more enhancements later.

@low-batt @uiryuu Please have a look and test these scenarios:

|                 | Window opened | Open new video in window | Navigate in playlist |
|-----------------|---------------|--------------------------|----------------------|
| Always          | Initial       | Y                        | Y                    |
| Opened manually | Initial       | Y                        | N                    |
| Disabled        | Initial       | N                        | N                    |

Initial = Apply initial position when it's configured, otherwise center on the last used screen
Y = Resize, keep window center
N = No resize unless aspect ratio changed

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -2205,6 +2205,9 @@ class PlayerCore: NSObject {
     // restart even while paused. See issue #5337.
     syncUI(.time)
     reloadSavedIINAfilters()
+    
+    info.justOpenedFile = false
+    info.justStartedFile = false
 
     NowPlayingInfoManager.shared.updateInfo()
 
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -2206,6 +2206,10 @@ class PlayerCore: NSObject {
     syncUI(.time)
     reloadSavedIINAfilters()
     
+    // The new video's size is guaranteed to be available. Reset the flags used for window resizing.
+    // We can't put this in MPV_EVENT_VIDEO_RECONFIG because it can be emitted with the old video's size
+    // after switching to a new video.
+    // We should keep these flags until a MPV_EVENT_VIDEO_RECONFIG with the new video's size.
     info.justOpenedFile = false
     info.justStartedFile = false
 
--- a/iina/Extensions.swift
+++ b/iina/Extensions.swift
@@ -270,11 +270,11 @@ extension NSRect {
     if newOrigin.y < biggerRect.origin.y {
       newOrigin.y = biggerRect.origin.y
     }
-    if newOrigin.x + width > biggerRect.origin.x + biggerRect.width {
-      newOrigin.x = biggerRect.origin.x + biggerRect.width - width
+    if newOrigin.x + newSize.width > biggerRect.origin.x + biggerRect.width {
+      newOrigin.x = biggerRect.origin.x + biggerRect.width - newSize.width
     }
-    if newOrigin.y + height > biggerRect.origin.y + biggerRect.height {
-      newOrigin.y = biggerRect.origin.y + biggerRect.height - height
+    if newOrigin.y + newSize.height > biggerRect.origin.y + biggerRect.height {
+      newOrigin.y = biggerRect.origin.y + biggerRect.height - newSize.height
     }
     return NSRect(origin: newOrigin, size: newSize)
   }
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -2499,6 +2499,10 @@ class MainWindowController: PlayerWindowController {
   /// - Parameter window: Window to determine the screen for.
   /// - Returns: Screen to use for the given window.
   private func determineScreenToUse(_ window: NSWindow) -> NSScreen {
+    // If the window is currently showing on a screen, use this screen
+    if window.isOnActiveSpace, let currentScreen = window.screen {
+      return currentScreen
+    }
     guard let rectString = UserDefaults.standard.value(forKey: "MainWindowLastPosition") as? String else {
       return window.selectDefaultScreen()
     }
@@ -2542,15 +2546,14 @@ class MainWindowController: PlayerWindowController {
       // - Resize the window to fit video size
       // - Use physical resolution on Retina displays
       // - Direct use of the mpv geometry option
-      let geometrySet = !(player.mpv.getString(MPVOption.Window.geometry) ?? "").isEmpty
       let resizeTiming = Preference.enum(for: .resizeWindowTiming) as Preference.ResizeWindowTiming
       switch resizeTiming {
       case .always:
         needResizeWindow = true
       case .onlyWhenOpen:
-        needResizeWindow = player.info.justOpenedFile || geometrySet || shouldApplyInitialWindowSize
+        needResizeWindow = player.info.justOpenedFile || shouldApplyInitialWindowSize
       case .never:
-        needResizeWindow = geometrySet || shouldApplyInitialWindowSize
+        needResizeWindow = shouldApplyInitialWindowSize
       }
     } else {
       // video size changed during playback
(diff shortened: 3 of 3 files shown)


iina/PlayerCore.swift
```diff
@@ -2205,6 +2205,9 @@ class PlayerCore: NSObject {
     // restart even while paused. See issue #5337.
     syncUI(.time)
     reloadSavedIINAfilters()
+    
+    info.justOpenedFile = false
+    info.justStartedFile = false
```
low-batt, line 2214: This is in `PlayerCore.playbackRestarted`, which is frequently called. Definitely not where I would expect it to be. If nothing else I think a comment is needed as to why it belongs here.

I am concerned whether there are any race conditions. These flags are not for use by `PlayerCore` / mpv. This is needed by `MainWindowController`, which is operating independently from mpv. @svobs has in the past mentioned problems with the order of window events vs. mpv events.

In the case at hand, how do we know for certain that mpv does not emit `MPV_EVENT_PLAYBACK_RESTART` before the window has opened? Yes, the window should normally have opened before this, but is that deterministic?

I see this is also tied into the mpv emitting `MPV_EVENT_VIDEO_RECONFIG`.

I'm lost trying to figure out if there are any race conditions in how this is handled. Would be nice to have a comment somewhere regarding the assumptions about the expected order of window and mpv events.

low-batt:
> The previous screen should still be remembered, and we put the window on that screen, if available.

This part did not work for me. I opened a video manually. It opened on my external screen positioned based on the `Initial window position`. I dragged the window to the other screen. I then closed the window. I then opened the video again manually. The window opened back on the original screen. Did not matter how `Resize the window to fit video size` was set.

I disabled `Initial window position` and it still did not remember the screen I dragged the window to. Again did not matter how `Resize the window to fit video size` was set.

I then shutdown IINA, restarted and reopened the video. This time it placed the video at the bottom of the external screen. It was on the builtin screen when I quit IINA.

Running the tests in the table seemed to work for me. With `Initial window position` set and  `Resize the window to fit video size` set to `Always`, navigating in the playlist I ended up with this:

always

Otherwise running the tests in the table all worked for me.

On this:
> Internally this is synced with mpv's geometry, and we may have different behavior with mpv. That's acceptable since IINA is GUI-focused.

Not a good time to change anything about this, but it is very broken. IINA's own settings should definitely override any use of the mpv geometry option. But there is no reason why IINA should interpret the geometry option value differently from mpv. Obviously IINA can't support some types of mpv configuration, but in general IINA should strive to be as compatible as possible with mpv configuration. 

I looked into IINA's support of the geometry option a very long time ago as a user entered an issue about it. IINA's interpretation of the geometry option value is definitely broken. I briefly played around with fixes, but never posted anything because other problems have been more important. Also, if I am recalling properly  support depends upon removing IINA's current keepaspect restriction. Anyway, just for the record, eventually we should consider matching mpv's behavior.

low-batt (commented):
Seemed like there is a problem remembering the screen.

low-batt:
I frequently reopen videos by right clicking on IINA's running icon in the dock and then clicking on one of the files in the pop-up menu. I keep encountering the behavior where the menu closes and then nothing happens. Naturally when I focus and go slow it doesn't reproduce. So maybe my mistake. But keep an eye out for any trouble opening a file.

low-batt:
I confirmed IINA 1.3.5 properly remembers the screen if you quit IINA. The wrong screen is used with this PR. But IINA 1.3.5 also does not remember the screen if I leave IINA running, close the video and then reopen it.

low-batt:
I will dig into the code and report what is going on tomorrow.

uiryuu:
Can confirm that #5669 is fixed by this PR.

> Even if the user set geometry in their conf, it still only applied once.

This is also the behavior of mpv when setting the geometry in the `mpv.conf`.

During testing, I found an issue where window will jump to an offscreen position when the new video is larger than the screen under some circumstances, but it appears that the issue is not caused by this PR, rather a existing issue triggered by this PR. I've filled #5675 for that.

Otherwise this PR looks fine to me.

lhc70000:
@low-batt 

> This part did not work for me. 
> I confirmed IINA 1.3.5 properly remembers the screen if you quit IINA. The wrong screen is used with this PR.

I couldn't reproduce it stably. It _might_ be caused by an issue that I fixed in the new commit 2a961606a090067305bb71e669cf8fb694d3abb1. `visibleScreen` returns the screen area without the notch, and if the window's origin was there, it will fallback to the default screen.

I don't see this problem on my machine with this branch now.

iina/PlayerCore.swift
lhc70000: Confirmed with testing, sometimes `MPV_EVENT_VIDEO_RECONFIG` can be emitted multiple times after switching to a new video: e.g. first time for the previous video with previous video size, second time for the new video. In this case we need to resize according to the new video. So we need to reset the flag only after confirming that the new video has started.

According to mpv's source code there should be at least one `VIDEO_RECONFIG` before `PLAYBACK_RESTART`. We use this flag inside the handler for `VIDEO_RECONFIG`. Also since the handler runs on the main thread, it should only run after window being opened. So I don't see a problem here, and everything is on the main thread. 

I will add some comments.

lhc70000:
> This is also the behavior of mpv when setting the geometry in the mpv.conf.

> Anyway, just for the record, eventually we should consider matching mpv's behavior.

Actually, mpv's behavior is: every time you start a new file, the geometry is applied. Say you set the `geometry` to put the window at the top left corner. Then you drag the window somewhere else. Then when it continues to play the next video in the playlist, the window is moved back to the top left corner. I couldn't think of any benefit of this behavior, especially for a GUI software user.

We already have many behaviors that are different from mpv. mpv doesn't resize the window to fit video size. So, to me, the conflict is inevitable and we should apply `geometry` only when the window is first opened. We can also unlink the initial window size with `geometry` and make it only configurable in the conf.

Furthermore, the user can still write a simple code snippet with the plugin system to enforce `geometry` if they indeed what this, and even more advanced control.

low-batt:
Caught me having Xcode problems trying to build 1.3.5 to debug how screens used to be handled. I confirmed it is not this PR that is causing the problem with screen selection. I will investigate if it is the commit you mentioned.

The new comment is helpful. I'm still a bit concerned about the interaction between mpv events and window events. In other words I would expect `MainWindowController` to be the one that clears these flags once they have been used. That is still not clear to me.

Anyway, my concern about the wrong screen being used is not related to this PR. So the changes are working for me as well.

On the mpv geometry option, the problem is that IINA is totally miss-interpreting the geometry value and placing the window in the wrong place on the screen. At least that is what I remember. It has been a few years. I found several problems in the code that parses and interprets that option value. I probably could dig up my notes, but I don't think we want deal with this problem now. Maybe after we remove the keepaspect restriction.

low-batt:
Pulled the latest commits. Screen selection problem still present. I will continue to investigate.

low-batt:
I know what is happening with the screen and how to fix it. I will writeup an issue with details and prepare a PR. Gotta run some errands. Will get this done today.

low-batt:
For the wrong display problem see issue #5680. I've posted a fix.

BUT IINA is centering the window instead of using the exact old position. That is happening in `MainWindowController.handleVideoSizeChange`. That seems wrong, yes?

lhc70000:
> BUT IINA is centering the window instead of using the exact old position. That is happening in MainWindowController.handleVideoSizeChange. That seems wrong, yes?

Seems wrong, but acceptable to me. Let's fix this later.

PR #5675, opened 2025-09-17
Fix sometimes window goes offscreen/random position when changing demensions

uiryuu:
- [ ] This implements/fixes issue #(?).

---

**Description:**
The calculations should be using `newSize` instead of `size`. According to blame, this appears to be a 9-year-old problem...

I vaguely remember that there's an issue (or multiple) complaining that the window jumps to a random position when advancing to the next video in the playlist. Hopefully this PR can resolve that/those issues.


lhc70000:
Cherry picked to [fix-resize-behavior](https://github.com/iina/iina/tree/fix-resize-behavior).

Issue #5694, opened 2025-09-20
Ability to properly see the current progress


adeftu:
**What you want IINA to do:**

I'd like to press a key and see the current timestamp where the video is at. 
Also, I'd like the text size of this display to be customized.

Alternatively, I'd like the size of the On screen controller toolbar size to be customized so that the timestamp is not as tiny.

**What IINA does currently:**

As far as I know there are currently 2 ways to find the timestamp:
* In the On screen controller toolbar. The problem with this is that the text is absolutely tiny and I can barely see the timestamp and time left. If this could be customized, that would be great.
* When I pause/resume the video, in the top left corner of the screen below Pause/Resume. The size of this text can be indeed customized, but I don't really want to pause the video just to see where I'm at.

**Examples of other projects that have something similar:**

In mpv I can can press O key and see the current progress.

AMRITESH240304:
I tried to make it simple now when you press the O key, it shows the current timestamp with progress bar on the screen.

https://github.com/user-attachments/assets/f04e1808-1ae8-407a-a21a-020fa425bb35

uiryuu:
Fixed in https://github.com/iina/iina/commit/1e706fdfa1a7ff86b897318a1c77264413fac48a

sitenoise:
Wow! I don't know how I got cc:'d on this and didn't know it was something someone else had asked for (I don't know how to make sense of all these "commit" & "PR" things) 

This makes me so HAPPY Thank you @uiryuu and @low-batt !

PR #5750, opened 2025-10-10
feat: show current timestamp and progress bar on 'O' key press

AMRITESH240304:
- [x] This implements/fixes issue #5694 .
.

---

**Description:**
when you press the O key, it shows the current timestamp and progress bar on the screen.
changes in the input.conf by adding the o keybinding

### DIFF
--- a/iina/PlayerWindowController.swift
+++ b/iina/PlayerWindowController.swift
@@ -283,6 +283,10 @@ class PlayerWindowController: NSWindowController, NSWindowDelegate {
       case MPVCommand.screenshot.rawValue:
         return player.screenshot(fromKeyBinding: keyBinding)
         
+      case MPVCommand.showProgress.rawValue:
+        showTimestampWithPosition()
+        returnValue = 0
+        
       default:
         returnValue = player.mpv.command(rawString: keyBinding.rawAction)
       }
@@ -700,6 +704,15 @@ class PlayerWindowController: NSWindowController, NSWindowDelegate {
     default:
       break
     }
+  }
+
+  internal func showTimestampWithPosition() {
+    guard player.info.state.loaded else { return }
+    let osdText = (player.info.videoPosition?.stringRepresentation ?? Constants.String.videoTimePlaceholder) + " / " +
+    (player.info.videoDuration?.stringRepresentation ?? Constants.String.videoTimePlaceholder)
+    let percentage = (player.info.videoPosition / player.info.videoDuration) ?? 0
+    
+    player.sendOSD(.seek(osdText, percentage))
   }
 
   internal func isMouseEvent(_ event: NSEvent, inAnyOf views: [NSView?]) -> Bool {
--- a/iina/config/iina-default-input.conf
+++ b/iina/config/iina-default-input.conf
@@ -84,6 +84,7 @@ ENTER set fullscreen yes
 q quit
 
 p cycle pause                          # toggle pause/playback mode
+o show-progress                        # show playback progress with position bar
 
 . frame-step                           # advance one frame and pause
 , frame-back-step                      # go back by one frame and pause
--- a/iina/config/input.conf
+++ b/iina/config/input.conf
@@ -92,7 +92,7 @@ SPACE cycle pause                      # toggle pause/playback mode
 ENTER playlist-next                    # skip to the next file
 < playlist-prev                        # skip to the previous file
 O no-osd cycle-values osd-level 3 1    # toggle displaying the OSD on user interaction or always
-o show-progress                        # show playback progress
+o show-progress                        # show playback progress with position bar
 P show-progress                        # show playback progress
 i script-binding stats/display-stats   # display information and statistics
 I script-binding stats/display-stats-toggle # toggle displaying information and statistics


AMRITESH240304:
Hey @low-batt , can you review the changes and let me know if anything needs improvement?

iina/PlayerWindowController.swift
```diff
@@ -283,6 +283,10 @@ class PlayerWindowController: NSWindowController, NSWindowDelegate {
       case MPVCommand.screenshot.rawValue:
         return player.screenshot(fromKeyBinding: keyBinding)
         
+      case MPVCommand.showProgress.rawValue:
+        showTimestampWithPosition()
+        returnValue = 0
+
```
low-batt, line 289: This line contains a bunch of spaces, probably Xcode being helpful. When getting ready to commit do a `git diff` and look for red. Git will point out extraneous trailing spaces.

iina/config/iina-default-input.conf
```diff
@@ -84,6 +84,7 @@ ENTER set fullscreen yes
 q quit
 
 p cycle pause                          # toggle pause/playback mode
+o show-progress                        # show playback progress with position bar
```
low-batt, line 87: The `IINA Default` key bindings were derived from the mpv key bindings. When we are updating IINA's `mpv Default` key bindings we review what changes the mpv project has made to their bindings and determine what changes might be applicable to IINA's bindings.

To make it easier to see the differences between IINA's bindings and mpv's bindings it is desirable to try and match up with the mpv source, when possible.

In the case at hand:
- ` with position bar` has been added to the key binding comment. Better to exactly match the shorter comment used by mpv.
- The new binding was added after `p cycle pause`. In the mpv source it appears after `, frame-back-step` and before `m cycle mute`. So I would move this to be after `, frame-back-step`.
- IINA's `mpv Default` also contains `P show-progress                        # show playback progress`. For whatever reason mpv defines two keys for this action. I would be consistent with mpv and add the binding to capital P to IINA's bindings as well.

iina/config/input.conf
```diff
@@ -92,7 +92,7 @@ SPACE cycle pause                      # toggle pause/playback mode
 ENTER playlist-next                    # skip to the next file
 < playlist-prev                        # skip to the previous file
 O no-osd cycle-values osd-level 3 1    # toggle displaying the OSD on user interaction or always
-o show-progress                        # show playback progress
+o show-progress                        # show playback progress with position bar
```
low-batt, line 95: For the `mpv Default` key bindings we definitely want to match up with mpv. Altering the comment makes it harder to compare IINA's mpv bindings to the mpv source.

iina/PlayerWindowController.swift
```diff
@@ -702,6 +706,15 @@ class PlayerWindowController: NSWindowController, NSWindowDelegate {
     }
   }
 
+  internal func showTimestampWithPosition() {
+    guard player.info.state.loaded else { return }
+    let osdText = (player.info.videoPosition?.stringRepresentation ?? Constants.String.videoTimePlaceholder) + " / " +
+    (player.info.videoDuration?.stringRepresentation ?? Constants.String.videoTimePlaceholder)
+    let percentage = (player.info.videoPosition / player.info.videoDuration) ?? 0
+
```
low-batt, line 714: Blank line contains spaces.

low-batt:
@AMRITESH240304 Sorry I'm slow to get to this PR. Super busy these days. Thanks for poking me about it. Also I am a _junior_ IINA developer and normally don't get involved with key bindings. Generally other IINA developers handle that. I was hoping others would get to this PR before me.

You did a good job on this PR. When I see that I provide a super picky review. When you see all the feedback I'm going to provide remember I think this is good work and IINA very much appreciates the submission.

Also remember I don't usually cover this area. You might want to wait to hear from the other IINA developers to see if they agree with me before making changes to address my comments. Use your own judgement about that.

You won't be able to see it, but I will be adding this PR to our work list for IINA 1.5.0. I will also ask the other developers to take a look at it.

I will be posting my review comments shortly…

----

For the other IINA developers…

This PR is about adding support for the [show-progress](https://mpv.io/manual/stable/#command-interface-show-progress) mpv command:
> **show-progress**
> Show the progress bar, the elapsed time and the total duration of the file on the OSD. no-osd has no effect on this command.

In general IINA has not made an effort to support mpv configuration where that configuration applies to the mpv UI. I'm thinking IINA should try and support more mpv UI configuration where it is fairly generic and could apply to IINA's UI, like the case at hand.

In response to the `show-progress` command mpv shows an OSD message as well as the progress bar:

mpv


With this PR IINA only shows the OSD, which makes sense to me as the OSD message contains a progress bar:

iina

iina/PlayerWindowController.swift
```diff
@@ -702,6 +706,15 @@ class PlayerWindowController: NSWindowController, NSWindowDelegate {
     }
   }
 
+  internal func showTimestampWithPosition() {
+    guard player.info.state.loaded else { return }
+    let osdText = (player.info.videoPosition?.stringRepresentation ?? Constants.String.videoTimePlaceholder) + " / " +
+    (player.info.videoDuration?.stringRepresentation ?? Constants.String.videoTimePlaceholder)
+    let percentage = (player.info.videoPosition / player.info.videoDuration) ?? 0
+    
+    player.sendOSD(.seek(osdText, percentage))
```
low-batt, line 715: Two things bother me about this code. We are using the existing `seek` OSD message. But no seeking has occurred. We are _reusing_ this OSD message for a situation it was not intended for because the message it shows is exactly what we want to display in this case. This happens a lot with software. The danger is that in the future someone alters this OSD message to be specific to seeking. Like adding "Seeked to" to the display.

One way to solve this would be to rename the `seek` OSD message enum constant to be a more generic `position`.

The other thing that bothers me is that this duplicates the code to form the message, found in `MainWindowController`:

https://github.com/iina/iina/blob/23f663ab1c615522b6db61f659d2b39c03a9ce65/iina/MainWindowController.swift#L2935-L2938

It is not much code. But in this case since `MainWindowController` is a subclass of `PlayerWindowController` it is easy to move this code into a little utility method and have both classes use that method. Maybe `createPositionOSDMessage`? Assuming the `seek` message is renamed…

low-batt (commented):
Some feedback to think about…

iina/PlayerWindowController.swift
uiryuu: @AMRITESH240304 FYI, you should enable these 2 options in Xcode's settings:

Image

uiryuu: Only the main window will react to this action, so this function should only be implemented in `MainWindowController` instead of `PlayerWindowPlayer`, which will be used for the main window and the mini player. Therefore I suggest moving this case (together with `showTimerstampWithPosition()`) to `MainWindowController`.

iina/PlayerWindowController.swift
uiryuu: What about we create a new OSD message for this? We can also add some additional information like high-precision time stamp, current frame number to make it more helpful?

uiryuu:
Nice and clean commit. Definitely something we'd like to add in the following release. See my comment above.

iina/PlayerWindowController.swift
low-batt: That would be nice.

lhc70000:
Thanks for the PR, it looks good to me if the comments can be addressed.

uiryuu:
Merged by https://github.com/iina/iina/commit/1e706fdfa1a7ff86b897318a1c77264413fac48a

Issue #3210, opened 2020-11-14
Add vapoursynth support?


varenc:
**What you want IINA to do:**
Use vapoursynth filters in IINA without resorting to building libmpv locally and replacing the lib instead of IINA

**What IINA does currently:**
Currently IINA does not use an mpv compiled with vapoursynth support.

**Why you think this should be added:**
Besides me, there seems to be sufficient interest in it.

Signs of interests:
- There's a guide on the IINA wiki on how to add in vapoursynth support: https://github.com/iina/iina/wiki/Use-SVP-with-IINA  (it's outdated)
- The SVP project has its own guide on how to add support to IINA: https://www.svp-team.com/wiki/SVP:IINA
- Users often have trouble with this, resulting in issues filed on this project even though the user's main problem is just understanding mpv compliation ([#2875 for example](https://github.com/iina/iina/issues/2875))

It looks like IINA used to have vapoursynth support, but this was removed in added463f7e5eca4d141687956bb1cf129017710.  Looking at the corresponding homebrew forumla update (https://github.com/iina/homebrew-mpv-iina/commit/4c06d1e047f53073f860447c451a8df8d805237a), it says this was done because "vapoursynth dep because it cannot be compiled targeted 10.11"?  Is that still true?

I have no idea of the support burden involved, so I can't be confident it's worth it. But I do know it'll make a lot of people happy!

HarukaMa:
On 2020-11-15 08:57, Chris Varenhorst wrote:
> 
> 
> 
> **What you want IINA to do:**
> Use vapoursynth filters in IINA without resorting to building libmpv
> locally and replacing the lib instead of IINA
> 
> 
> **What IINA does currently:**
> Currently IINA does not use an mpv compiled with vapoursynth support.
> 
> 
> **Why you think this should be added:**
> Besides me, there seems to be sufficient interest in it.
> 
> Signs of interests:
> - There's a guide on the IINA wiki on how to add in vapoursynth
> support: https://github.com/iina/iina/wiki/Use-SVP-with-IINA  (it's
> outdated)
> - The SVP project has its own guide on how to add support to IINA:
> https://www.svp-team.com/wiki/SVP:IINA
> - Users often have trouble with this, resulting in issues filed on
> this project even though the user's main problem is just understanding
> mpv compliation ([#2875 for
> example](https://github.com/iina/iina/issues/2875))
> 
> 
> It looks like IINA used to have vapoursynth support, but this was
[...]

Issue #5443, opened 2025-04-09
Ship `ffmpeg` and `ffprobe` command-line tools


Ecco:
### Feature request

Judging by the content of the `Frameworks` directory, it feels like IINA is using ffmpeg (or libav, whatever) internally.

It would be really cool if IINA would also include the `ffmpeg` and `ffprobe` command-line tools in its `MacOS` folder. Those are really useful for converting and probing media files.

### Notes

Yes, installing ffmpeg with Homebrew is possible, but it takes forever to install a gazillion dependencies. And it makes updating Homebrew very slow. Plus it's a giant waste since those dependencies are already in IINA. Long story short : it'd be a lot more convenient to have ffmpeg in IINA directly.

low-batt:
Yes, IINA includes _some_ [FFmpeg](https://ffmpeg.org/) libraries. They are used by IINA and by `libmpv`.  The FFmpeg build is _very_ configurable:

<details>
<summary><b>Build Configuration Options:</b></summary>

```text
low-batt@gag ffmpeg (master $%=)$ ./configure --help
Usage: configure [options]
Options: [defaults in brackets after descriptions]

Help options:
  --help                   print this message
  --quiet                  Suppress showing informative output
  --list-decoders          show all available decoders
  --list-encoders          show all available encoders
  --list-hwaccels          show all available hardware accelerators
  --list-demuxers          show all available demuxers
  --list-muxers            show all available muxers
  --list-parsers           show all available parsers
  --list-protocols         show all available protocols
  --list-bsfs              show all available bitstream filters
  --list-indevs            show all available input devices
  --list-outdevs           show all available output devices
  --list-filters           show all available filters

Standard options:
  --logfile=FILE           log tests and output to FILE [ffbuild/config.log]
  --disable-logging        do not log configure debug information
  --fatal-warnings         fail if any configure warning is generated
  --prefix=PREFIX          install in PREFIX [/usr/local]
  --bindir=DIR             install binaries in DIR [PREFIX/bin]
  --datadir=DIR            install data files in DIR [PREFIX/share/ffmpeg]
  --docdir=DIR             install documentation in DIR [PREFIX/share/doc/ffmpeg]
  --libdir=DIR             install libs in DIR [PREFIX/lib]
  --shlibdir=DIR           install shared libs in DIR [LIBDIR]
  --incdir=DIR             install includes in DIR [PREFIX/include]
  --mandir=DIR             install man page in DIR [PREFIX/share/man]
[...]

Issue #5739, opened 2025-10-04
Crashing on MacOS Catalina


Molasar49:
### IINA Version

IINA 1.4.0 / 1.4.1 

### Release Channel

Release

### macOS Version

Catalina 10.15.7

### Steps to Reproduce

- Install IINA
- Right click
- Open

### Expected Behavior

Run normally (like v1.3.5)

### Actual Behavior

Crash upon trying to run

### Crash Report or Process Sample

[IINA_2025-10-04-164842_Michaels-iMac.log](https://github.com/user-attachments/files/22704592/IINA_2025-10-04-164842_Michaels-iMac.log)

### How often can you reproduce the issue

Every time

### Required Actions

- [x] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [ ] I tested with the Nightly build and confirmed the issue is not fixed.
- [ ] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

Molasar49:
Tried deleting:

~/Library/Preferences/com.colliderli.iina.plist
~/Library/Application Support/com.colliderli.iina 

no joy.

low-batt:
The crash is in the library that runs [Lua](https://www.lua.org/) scripts:
```text
Thread 11 Crashed:: lua/ytdl_hook
0   libluajit-5.1.2.dylib         	0x000000010ad11597 0x10ad00000 + 71063
```

This is the mpv thread that runs the Lua script that interfaces to [yt-dlp](https://github.com/yt-dlp/yt-dlp).

I too have Macs still running macOS 10.15.7. I've been running IINA 1.4.1 and have not encountered this crash.

I am confused about what could be triggering this on your Mac and not on mine. I will do some more testing under 10.15.7.

Molasar49:
Give me a holler if I can be of help...

Molasar49:
Ok, did this....

Copied the libluajit-5.1.2.dylib (Contents/Frameworks) from v1.3.5 to v1.4.1 (overwrite) and it ran correctly.... ???

No crash.

Molasar49:
Just had to do this:

sudo xattr -cr /Applications/IINA\ copy.app
sudo xattr -r -d com.apple.quarantine /Applications/IINA\ copy.app
sudo codesign --force --deep --sign - /Applications/IINA\ copy.app

and it ran no problem...

low-batt:
[...]

Issue #5149, opened 2024-08-19
Pls compile with libbs2b


IfSheDiesSheDies:
so we can use af=bs2b pls pls pls pls

low-batt:
The request is for IINA to include support for [bs2b](https://ffmpeg.org/ffmpeg-filters.html#bs2b) filtering by building FFmpeg with `--enable-libbs2b`.

uiryuu:
Fixed in https://github.com/iina/iina/pull/6261. Will be included in the next release

PR #5754, opened 2025-10-14
Proof-of-concept-ish Vapoursynth Support

florius0:
- [x] This implements/fixes issue #3210 #5443.

---

**Description:**

This pull request implements new build process for the IINA using nix to achieve the following goals:

1. Add vapoursynth support – #3210
2. Enhance plugin development experience – #5443
3. Simplify the build process by describing it in a single `flake.nix` file
4. Make builds (semi-)reproducible by depending only on host's Xcode and commonly available libraries from `nixpkgs` rather than custom-managed IINA dependencies to simplify maintenance.
    Please note that overriding the dependencies to suite custom needs of IINA team that the PR's author wasn't aware of should be possible by modifying `flake.nix`
5. Keep to the macOS ideology of self-contained `.app`s

## General build strategy

To ease the review and (hopefully) further collaboration on this, I include this description of the build process:

For all supported systems (defined in `outputs.systems`):

1. Build custom `ffmpeg`
2. Build custom `mpv`
3. Gather the external dependencies in `depsInc`, `depsLib`, `depsIndirect`, `depsExecutable`
4. Fetch `SwiftPM` packages into seprate derivation to cache that step
5. Build `IINA.app` for that platform:
    1. Copy `depsInc` and `depsLib` as per readme.md
    2. Rewrite their rpaths for proper linkage
    3. Run `xcodebuild`
6. Patch the resulting `IINA.app` to:
    1. Include `depsIndirect`
    2. Copy recursively other dependencies from `/nix/store` that might be referenced 
    3. Package `python3` and `vapoursynth`'s `site-packages`
    4. Rewrite rpaths for proper loading at runtime so the `.app` remains self-contained
    5. Restore symlinks across `dylib`'s versions
    6. Add necessary environment variables as `LSEnvironment` entries into `Info.plist`
    7. Codesign the resulting `IINA.app`
8. Stich the builds for each platform into universal app using `lipo` and also
    1. Ensure rpaths for proper loading at runtime so the `.app` remains self-contained
    2. Restore symlinks across `dylib`'s versions
    3. Codesign the resulting `IINA.app`

## Testing

The resulting universal build was tested on:

1. M4 Max machine
2. M4 Pro machine
3. Intel-based machine

The testing strategy included the following:

1. Running `IINA.app`
2. Playing h264 and av1 videos as is.
3. Exposing `/tmp/iinasocket` to test SVP integration

While this strategy is very barebones, no issues were found.

## Running the build

To run the build:

1. Ensure `nix` is installed on your system
2. Run `nix build` in the repository root
3. This will produce symlinked `result` directory to the universal binary

## Vapoursynth support

As far as I understand from #3210 the main challenge of supporting `vapoursynth` was the packaging of `python3` as it is required by `vapoursynth`.

While solving it I've discovered that `vapoursynth` requires for some environment variables to be set to function properly. To do this at a packaging stage outside of Xcode the following solution was implemented:

1. Bundle everything using `nix`
2. Write the variables as `LSEnvironment` entries into `Info.plist`
3. Expand the paths on application start to keep the `.app` properly self-contained.

## Enhancing plugin development experience by bundling of `ffmpeg`, `mpv` and `python3` executables

While gathering information on `vapoursynth` support I've also came across #5443, and decided to implement it since the proposed solution already includes libraries required by these tools, so overall bundle size impact is negligible.

## Other modifications

While implementing it, I've disabled linking against `libgcc_s.1.1` and `libstdc++.6` on x86 platforms since it seems these are unnecessary.

## Other considerations

1. SVP Support – If we make `/tmp/iinasocket` available by default, IINA will support tools like SVP out-of-the box.
2. Build size – the build size was noticeably increased from ~160mb to ~500mb, as far as I understand major contributor to it might be ffmpeg, since I didn't strip it down as was mentioned in #5443. Do we need to address that?
3. `.xcconfig` support – currently the build is done using Release configuration. Should we expose it as a parameter?
4. Potential redundancy in the Xcode build pipeline – since the proposed solution does some of the dylib packaging by itself, do we need to keep copying from `deps/lib` in the Xcode build pipeline
5. Proper code signing support – We may want to implement non-adhoc codesigning to enable this solution to be used for releasing IINA to the AppStore.
6. Fixing compatibility step in the platform IINA build – I wasn't able to figure out why Xcode wants to link against these specific versions, since other packages shouldn't depend on it. As a workaround these libraries are simply copied.
7. `libc++` linkage – as far as I understand, `libc++` is provided by macOS, but I chose to link against `nix`-provided `libc++`. Do we need to link agains macOS-provided version explicitly

## Resulting build (not an official IINA build)

I would also like to include resulting build for those willing to try it.
Please understand this is **not an official build** done by IINA team.
If you encounter any problems, it is very likely IINA team has nothing to do with them, so please don't open issues in the IINA repo. If you would like to provide feedback, do it here as a comment.

Also note that the build needs to be signed on your machine to run: `sudo /usr/bin/codesign --force --deep --sign - path/to/IINA.app`.

[Download link](https://filebin.net/arx7us0fo5rduw0a)

PS

I'm not very familiar with the development of native macOS apps, so please consider this PR as a proof-of-concept rather than production-ready solution. I would very much like to collaborate on this with IINA team to achieve the goals stated at the beginning.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -114,7 +114,7 @@ browser/Firefox_Open_In_IINA/*
 *.xcarchive
 /.vscode
 
-deps/include
-deps/lib
-deps/executable
-deps/plugins
+deps/
+.home/
+.spm/
+.tmp/
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,9 @@ Temporary Items
 build/
 DerivedData/
 
+## nix build result symlink
+result
+
 ## Xcode generated
 */project.xcworkspace/xcshareddata/
 
@@ -113,4 +116,3 @@ browser/Firefox_Open_In_IINA/*
 
 deps/executable/youtube-dl
 deps/plugins
-
--- a/.gitignore
+++ b/.gitignore
@@ -114,5 +114,7 @@ browser/Firefox_Open_In_IINA/*
 *.xcarchive
 /.vscode
 
-deps/executable/youtube-dl
+deps/include
+deps/lib
+deps/executable
 deps/plugins
--- a/Configs/iina.xcconfig
+++ b/Configs/iina.xcconfig
@@ -22,7 +22,7 @@ LIBRARY_SEARCH_PATHS = $(inherited) $(SRCROOT)/deps/lib
 // The following libraries are not needed for arm64, so they are not universal libraries. To avoid
 // triggering build warnings they are not included in the Link Binary With Libraries phase. Instead
 // they are added here where we can add them only for the x86_64 build.
-OTHER_LDFLAGS[arch=x86_64] = $(inherited) -l gcc_s.1.1 -l stdc++.6
+OTHER_LDFLAGS[arch=x86_64] = $(inherited)
 PRODUCT_BUNDLE_IDENTIFIER = com.colliderli.$(TARGET_NAME)
 PRODUCT_NAME = IINA
 SWIFT_EMIT_LOC_STRINGS = YES
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1759439645,
+        "narHash": "sha256-oiAyQaRilPk525Z5aTtTNWNzSrcdJ7IXM0/PL3CGlbI=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "879bd460b3d3e8571354ce172128fbcbac1ed633",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-25.05",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "nixpkgs": "nixpkgs"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1759439645,
-        "narHash": "sha256-oiAyQaRilPk525Z5aTtTNWNzSrcdJ7IXM0/PL3CGlbI=",
+        "lastModified": 1763049705,
+        "narHash": "sha256-A5LS0AJZ1yDPTa2fHxufZN++n8MCmtgrJDtxFxrH4S8=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "879bd460b3d3e8571354ce172128fbcbac1ed633",
+        "rev": "3acb677ea67d4c6218f33de0db0955f116b7588c",
         "type": "github"
       },
       "original": {
--- a/flake.nix
+++ b/flake.nix
@@ -19,7 +19,7 @@
           pkgs = import nixpkgs { inherit system; };
 
           # Pull system's xcode in
-          xcode = pkgs.runCommandNoCC "system-xcode" { } ''
+          xcode = pkgs.runCommand "system-xcode" { } ''
             mkdir -p $out/bin
             ln -sf /Applications/Xcode.app/Contents/Developer/usr/bin/xcodebuild $out/bin/xcodebuild
           '';
@@ -146,6 +146,7 @@
                   pkgs.libass
                   pkgs.libb2
                   pkgs.libbluray
+                  pkgs.libbs2b
                   pkgs.libidn2
                   pkgs.libjpeg_turbo
                   pkgs.libjxl
@@ -661,7 +662,7 @@
 
               # Rewrite SwiftPM workspace-state.json to fix absolute paths
               if [ -f .spm/workspace-state.json ]; then
-                old_prefix=$(grep -Eo "/tmp/nix-build-[^/]+/source" .spm/workspace-state.json | head -n1)
+                old_prefix=$(grep -Eo "/nix/var/nix/builds/nix-[^/]+/source" .spm/workspace-state.json | head -n1)
                 echo "Patching workspace-state.json: replacing $old_prefix → $PWD"
                 sed -i -E "s|$old_prefix|$PWD|g" .spm/workspace-state.json
               fi
--- a/flake.nix
+++ b/flake.nix
@@ -419,50 +419,74 @@
                 done
               }
 
+              declare -Ag BUNDLED_DEPS=()
+
               bundle_dep() {
-                local dep="$1"
-                local depbase
-                depbase=$(basename "$dep")
-                local dest="$frameworks/$depbase"
+                local dep="$1"   # path as found in otool -L (may be libbs2b.0.dylib)
 
-                if [ -f "$dest" ]; then
-                  echo "✅ Already bundled: $depbase"
+                # Canonical path for cycle detection
+                local dep_real
+                dep_real=$(realpath "$dep" 2>/dev/null || echo "$dep")
+
+                # Break cycles by *real* path
+                if [[ -n "''${BUNDLED_DEPS["$dep_real"]:-}" ]]; then
+                  echo "🔁 Already processed dep: $dep_real"
+                  return
+                fi
+                BUNDLED_DEPS["$dep_real"]=1
+
+                local request_base
+                request_base=$(basename "$dep")
+                local real_base
+                real_base=$(basename "$dep_real")
+                local dest="$frameworks/$request_base"
+
+                if [[ "$dep_real" == /usr/lib/* ]] || [[ "$dep_real" == /System/* ]] || [[ "$real_base" == libffi-trampoline.dylib ]]; then
+                  echo "🚫 Skipping system/non-target dep: $dep_real"
                   return
                 fi
 
-                if [[ "$dep" == /usr/lib/* ]] || [[ "$dep" == /System/* ]] || [[ "$depbase" == libffi-trampoline.dylib ]]; then
-                  echo "🚫 Skipping system/non-target dep: $dep"
+                if [ ! -f "$dep_real" ]; then
+                  echo "⚠️  Dep missing on disk, skipping: $dep_real"
                   return
                 fi
 
-                if [ ! -f "$dep" ]; then
-                  echo "⚠️  Dep missing on disk, skipping: $dep"
+                # Ensure the real payload file exists in Frameworks under some canonical name
+                local canonical="$frameworks/$real_base"
+                if [ ! -f "$canonical" ]; then
+                  echo "📥 Copying dep → Frameworks: $dep_real → $canonical"
+                  cp -L -p "$dep_real" "$canonical" || { echo "❌ Copy failed for $dep_real"; return; }
+                fi
+
+                # Ensure the *requested* name exists and points to the payload
+                if [ ! -e "$dest" ]; then
+                  echo "🔗 Copying $canonical → $dest"
+                  cp -L -p "$canonical" "$dest"
+                else
+                  echo "✏️ Normalizing already-bundled dep alias: $request_base"
+                fi
+
+                ensure_writable "$canonical"
+
+                if ! is_macho "$canonical"; then
+                  echo "🧾 Non-Mach-O dep (no patching): $real_base"
                   return
                 fi
 
-                echo "📥 Copying dep → Frameworks: $dep → $dest"
-                cp -L -p "$dep" "$dest" || { echo "❌ Copy failed for $dep"; return; }
-                ensure_writable "$dest"
+                echo "✏️ Setting install_name id on $real_base"
+                install_name_tool -id "@rpath/$request_base" "$canonical" || true
 
-                if ! is_macho "$dest"; then
-                  echo "🧾 Copied non-Mach-O dep (no patching): $depbase"
-                  return
-                fi
+                ensure_rpath "$canonical"
+                rewrite_deps_to_rpath "$canonical"
 
-                echo "✏️ Setting install_name id on $depbase"
-                install_name_tool -id "@rpath/$depbase" "$dest" || true
-
-                ensure_rpath "$dest"
-                rewrite_deps_to_rpath "$dest"
-
-                # Recurse into transitive deps
-                otool -L "$dest" | awk 'NF && $1 !~ /:$/ {print $1}' | while read -r sub; do
+                # Recurse into transitive deps based on their /nix/store path
+                otool -L "$canonical" | awk 'NF && $1 !~ /:$/ {print $1}' | while read -r sub; do
                   case "$sub" in
                     /nix/store/*)
                       local subbase
                       subbase=$(basename "$sub")
-                      echo "  🔗 Repointing subdep for $depbase: $sub → @rpath/$subbase"
-                      install_name_tool -change "$sub" "@rpath/$subbase" "$dest" || true
+                      echo "  🔗 Repointing subdep for $real_base: $sub → @rpath/$subbase"
+                      install_name_tool -change "$sub" "@rpath/$subbase" "$canonical" || true
                       bundle_dep "$sub"
                       ;;
                     *) : ;;
--- a/flake.nix
+++ b/flake.nix
@@ -632,15 +632,6 @@
               cp -RL ${depsLib}/.               deps/lib
               cp -RL ${depsExecutable}/.        deps/executable/
 
-              echo "🔃 Fixing compatibility"
-              cp deps/executable/yt-dlp             deps/executable/youtube-dl
-              cp deps/lib/librubberband.3.dylib     deps/lib/librubberband.2.dylib
-              cp deps/lib/libplacebo.349.dylib      deps/lib/libplacebo.338.dylib
-              cp deps/lib/libjpeg.62.dylib          deps/lib/libjpeg.8.dylib
-              cp deps/lib/libjxl_cms.0.11.dylib     deps/lib/libjxl_cms.0.10.dylib
-              cp deps/lib/libjxl_threads.0.11.dylib deps/lib/libjxl_threads.0.10.dylib
-              cp deps/lib/libjxl.0.11.dylib         deps/lib/libjxl.0.10.dylib
-
               echo "✏️ Rewriting install names to use @rpath"
               find deps/lib -type f \( -perm -111 -o -name "*.dylib" -o -name "*.so" \) | while read -r dep; do
                 echo "✏️ Patching install names in $dep"
--- a/flake.nix
+++ b/flake.nix
@@ -27,8 +27,15 @@
           # Override ffmpeg to use our version of libs
           ffmpeg = pkgs.ffmpeg_7.override {
             withSoxr = true;
+            soxr = pkgs.soxr;
+
+            withRubberband = true;
             rubberband = pkgs.rubberband;
+
+            withPlacebo = true;
             libplacebo = pkgs.libplacebo;
+
+            withJxl = true;
             libjxl = pkgs.libjxl;
           };
 
--- a/flake.nix
+++ b/flake.nix
@@ -270,7 +270,11 @@
                 )
                 [
                   ffmpeg
-                  mpv
+                  # Grab the real mpv binary instead of the /bin wrapper script
+                  (pkgs.runCommand "iina-mpv-executable" { } ''
+                    mkdir -p $out/bin
+                    cp -p ${mpv}/Applications/mpv.app/Contents/MacOS/mpv $out/bin/mpv
+                  '')
                   pkgs.vapoursynth
                   pkgs.python3
                   pkgs.yt-dlp
--- a/flake.nix
+++ b/flake.nix
@@ -29,6 +29,9 @@
             withSoxr = true;
             soxr = pkgs.soxr;
 
+            withBs2b = true;
+            libbs2b = pkgs.libbs2b;
+
             withRubberband = true;
             rubberband = pkgs.rubberband;
 
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,862 @@
+{
+  description = "IINA – The modern video player for macOS.";
+
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-25.05";
+
+  outputs =
+    { self, nixpkgs }:
+    {
+      systems = [
+        "aarch64-darwin"
+        "x86_64-darwin"
+      ];
+
+      archApps = builtins.map (system: self.packages.${system}.iina) self.systems;
+
+      packages = nixpkgs.lib.genAttrs self.systems (
+        system:
+        let
+          pkgs = import nixpkgs { inherit system; };
+
+          # Pull system's xcode in
+          xcode = pkgs.runCommandNoCC "system-xcode" { } ''
+            mkdir -p $out/bin
+            ln -sf /Applications/Xcode.app/Contents/Developer/usr/bin/xcodebuild $out/bin/xcodebuild
+          '';
+
+          # Override ffmpeg to use our version of libs
+          ffmpeg = pkgs.ffmpeg_7.override {
+            withSoxr = true;
+            rubberband = pkgs.rubberband;
+            libplacebo = pkgs.libplacebo;
+            libjxl = pkgs.libjxl;
+          };
+
+          # Override mpv with vapoursynth support
+          mpv = pkgs.mpv-unwrapped.override {
+            ffmpeg = ffmpeg;
+            lua = pkgs.luajit;
+            vapoursynth = pkgs.vapoursynth;
+
+            # enable features we want
+            vapoursynthSupport = true;
+            javascriptSupport = true;
+            cmsSupport = true;
+            rubberbandSupport = true;
+            archiveSupport = true;
+            bluraySupport = true;
+            openalSupport = true;
+            vulkanSupport = true;
+            zimgSupport = true;
+
+            # disable linux-only bits
+            alsaSupport = false;
+            jackaudioSupport = false;
+            pipewireSupport = false;
+            x11Support = false;
+            waylandSupport = false;
+            vaapiSupport = false;
+            vdpauSupport = false;
+            sdl2Support = false;
+            cddaSupport = false;
+            dvbinSupport = false;
+            sixelSupport = false;
+          };
+
+          # Collect include deps as per readme.md
+          depsInc = pkgs.linkFarm "iina-deps-inc" [
+            {
+              name = "mpv";
+              path = "${pkgs.lib.getDev mpv}/include/mpv";
+            }
+            {
+              name = "libavcodec";
+              path = "${pkgs.lib.getDev ffmpeg}/include/libavcodec";
+            }
+            {
+              name = "libavdevice";
+              path = "${pkgs.lib.getDev ffmpeg}/include/libavdevice";
+            }
+            {
+              name = "libavfilter";
+              path = "${pkgs.lib.getDev ffmpeg}/include/libavfilter";
+            }
+            {
+              name = "libavformat";
+              path = "${pkgs.lib.getDev ffmpeg}/include/libavformat";
+            }
+            {
+              name = "libavutil";
+              path = "${pkgs.lib.getDev ffmpeg}/include/libavutil";
+            }
+            {
+              name = "libpostproc";
+              path = "${pkgs.lib.getDev ffmpeg}/include/libpostproc";
+            }
+            {
+              name = "libswresample";
+              path = "${pkgs.lib.getDev ffmpeg}/include/libswresample";
+            }
+            {
+              name = "libswscale";
+              path = "${pkgs.lib.getDev ffmpeg}/include/libswscale";
+            }
+          ];
+
+          # Collect lib deps as per readme.md
+          depsLib = pkgs.linkFarm "iina-deps-lib" (
+            pkgs.lib.flatten (
+              map
+                (
+                  pkg:
+                  let
+                    libdir = "${pkgs.lib.getLib pkg}/lib";
+                  in
+                  builtins.map (file: {
+                    name = baseNameOf file;
+                    path = "${libdir}/${file}";
+                  }) (builtins.attrNames (builtins.readDir libdir))
+                )
+                [
+                  ffmpeg
+                  mpv
+                  (pkgs.libhwy.overrideAttrs (old: {
+                    cmakeFlags = (old.cmakeFlags or [ ]) ++ [ "-DBUILD_SHARED_LIBS=ON" ];
+                  }))
+                  pkgs.brotli
+                  pkgs.dav1d
+                  pkgs.fontconfig
+                  pkgs.freetype
+                  pkgs.fribidi
+                  pkgs.gettext
+                  pkgs.glib
+                  pkgs.gmp
+                  pkgs.gnutls
+                  pkgs.graphite2
+                  pkgs.harfbuzz
+                  pkgs.lcms2
+                  pkgs.libarchive
+                  pkgs.libass
+                  pkgs.libb2
+                  pkgs.libbluray
+                  pkgs.libidn2
+                  pkgs.libjpeg_turbo
+                  pkgs.libjxl
+                  pkgs.libplacebo
+                  pkgs.libpng
+                  pkgs.libsamplerate
+                  pkgs.libsodium
+                  pkgs.libtasn1
+                  pkgs.libuchardet
+                  pkgs.libunibreak
+                  pkgs.libunistring
+                  pkgs.libwebp
+                  pkgs.luajit
+                  pkgs.lz4
+                  pkgs.mujs
+                  pkgs.nettle
+                  pkgs.p11-kit
+                  pkgs.pcre2
+                  pkgs.python3
+                  pkgs.rubberband
+                  pkgs.shaderc
+                  pkgs.snappy
+                  pkgs.soxr
+                  pkgs.speex
+                  pkgs.vapoursynth
+                  pkgs.vid-stab
+                  pkgs.vulkan-loader
+                  pkgs.xorg.libX11
+                  pkgs.xorg.libXau
+                  pkgs.xorg.libxcb
+                  pkgs.xorg.libXdmcp
+                  pkgs.xz
+                  pkgs.zeromq
+                  pkgs.zimg
+                  pkgs.zstd
+                ]
+            )
+          );
+
+          # Collect indirect deps
+          depsIndirect = pkgs.linkFarm "iina-deps-indirect" (
+            pkgs.lib.flatten (
+              map
+                (
+                  pkg:
+                  let
+                    libdir = "${pkgs.lib.getLib pkg}/lib";
+                    files = builtins.attrNames (builtins.readDir libdir);
+                  in
+                  pkgs.lib.concatMap (
+                    file:
+                    if pkgs.lib.hasSuffix ".dylib" file then
+                      [
+                        {
+                          name = file;
+                          path = "${libdir}/${file}";
+                        }
+                      ]
+                    else
+                      [ ]
+                  ) files
+                )
+                [
+                  pkgs.bzip2
+                  pkgs.expat
+                  pkgs.lame
+                  pkgs.libaom
+                  pkgs.libcaca
+                  pkgs.libcxx
+                  pkgs.libdovi
+                  pkgs.libdvdcss
+                  pkgs.libdvdnav
+                  pkgs.libffi
+                  pkgs.libjxl
+                  pkgs.libogg
+                  pkgs.libopenmpt
+                  pkgs.libopus
+                  pkgs.libplacebo
+                  pkgs.librist
+                  pkgs.libssh
+                  pkgs.libtheora
+                  pkgs.libvdpau
+                  pkgs.libvmaf
+                  pkgs.libvorbis
+                  pkgs.libxml2
+                  pkgs.llvmPackages.openmp
+                  pkgs.ocl-icd
+                  pkgs.openal
+                  pkgs.openjpeg
+                  pkgs.rubberband
+                  pkgs.SDL2
+                  pkgs.sdl3
+                  pkgs.srt
+                  pkgs.svt-av1
+                  pkgs.vapoursynth
+                  pkgs.x264
+                  pkgs.x265
+                  pkgs.zlib
+                  pkgs.zstd.out
+                  pkgs.zvbi
+                ]
+            )
+          );
+
+          # Collect executables to expose them for plugins
+          depsExecutable = pkgs.linkFarm "iina-deps-executable" (
+            pkgs.lib.flatten (
+              map
+                (
+                  pkg:
+                  let
+                    bindir = "${pkgs.lib.getBin pkg}/bin";
+                  in
+                  builtins.map (file: {
+                    name = baseNameOf file;
+                    path = "${bindir}/${file}";
+                  }) (builtins.attrNames (builtins.readDir bindir))
+                )
+                [
+                  ffmpeg
+                  mpv
+                  pkgs.vapoursynth
+                  pkgs.python3
+                  pkgs.yt-dlp
+                ]
+            )
+          );
+
+          # Collect SwiftPM deps as separate derivation for them to be cached
+          spmDeps = pkgs.stdenv.mkDerivation {
+            pname = "iina-spm-deps";
+            version = if self ? rev then builtins.substring 0 8 self.rev else "dev";
+
+            # Only include SwiftPM-related files as input
+            src = pkgs.lib.cleanSourceWith {
+              src = ./.;
+              filter =
+                path: type:
+                let
+                  relPath = pkgs.lib.removePrefix (toString ./. + "/") (toString path);
+                in
+                pkgs.lib.hasSuffix "Package.resolved" relPath
+                || pkgs.lib.hasSuffix "Package.swift" relPath
+                || pkgs.lib.hasPrefix "iina.xcodeproj" relPath;
+            };
+
+            dontFixup = true;
+
+            nativeBuildInputs = [
+              xcode
+              pkgs.git
+              pkgs.gnused
+              pkgs.unzip
+              pkgs.zip
+            ];
+
+            buildPhase = ''
+              export HOME=$PWD/.home
+              export CFFIXED_USER_HOME="$HOME"
+              export __XPC_CFFIXED_USER_HOME="$HOME"
+              export TMPDIR="$PWD/.tmp"; mkdir -p "$TMPDIR"
+              export DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer
+
+              APPLE_BIN="$DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/bin"
+              export PATH="$APPLE_BIN:$DEVELOPER_DIR/usr/bin:/usr/bin:/bin"
+
+              export TOOLCHAINS=XcodeDefault
+              export SDKROOT=macosx
+
+              mkdir -p .spm .spm-cache build
+
+              xcodebuild \
+                -workspace iina.xcodeproj/project.xcworkspace \
+                -scheme iina \
+                -resolvePackageDependencies \
+                -derivedDataPath "$PWD/build" \
+                -clonedSourcePackagesDirPath "$PWD/.spm" \
+                -packageCachePath "$PWD/.spm-cache" \
+                -disablePackageRepositoryCache \
+                -IDEPackageSupportDisableManifestSandbox=YES \
+                -IDEPackageSupportDisablePluginExecutionSandbox=YES \
+                ARCHS="$(uname -m)" ONLY_ACTIVE_ARCH=YES \
+                SWIFT_ENABLE_EXPLICIT_MODULES=NO
+            '';
+
+            # Copy everything — keep full structure (SPM state, caches, workspace, etc.)
+            installPhase = ''
+              mkdir -p $out
+              cp -R . $out/
+            '';
+          };
+
+          # hared normalizer script
+          normalizer = pkgs.writeShellApplication {
+            name = "iina-normalize-app";
+            runtimeInputs = [
+              pkgs.coreutils
+              pkgs.findutils
+              pkgs.gawk
+              pkgs.gnugrep
+              pkgs.file
+            ];
+            text = ''
+              set -euo pipefail
+
+              if [[ $# -lt 1 ]]; then
+                echo "usage: iina-normalize-app /path/to/IINA.app" >&2
+                exit 2
+              fi
+
+              app="$1"
+              frameworks="$app/Contents/Frameworks"
+
+              echo "🔧 iina-normalize-app: normalizing $app"
+              mkdir -p "$frameworks"
+
+              echo "📝 Making app contents writable"
+              find "$app" -type d -exec chmod u+rwx {} \;
+              find "$app" -type f -exec chmod u+rw  {} \;
+
+              # ✏️ Helpers
+              is_macho() {
+                # true for thin/fat Mach-O; false for scripts/text
+                file -b "$1" 2>/dev/null | grep -Eq 'Mach-O (universal binary|64-bit|arm64|x86_64)'
+              }
+
+              ensure_writable() { chmod u+rw "$1" 2>/dev/null || true; }
+
+              ensure_rpath() {
+                local f="$1"
+
+                if ! is_macho "$f"; then
+                  echo "🧾 Non-Mach-O, skipping rpath: $f"
+                  return
+                fi
+
+                if otool -l "$f" | grep -q '@executable_path/../Frameworks'; then
+                  echo "✅ LC_RPATH present → $f"
+                  return
+                fi
+                
+                echo "➕ LC_RPATH @executable_path/../Frameworks → $f"
+                install_name_tool -add_rpath "@executable_path/../Frameworks" "$f" || true
+              }
+
+              rewrite_deps_to_rpath() {
+                local bin="$1"
+
+                if ! is_macho "$bin"; then
+                  echo "🧾 Non-Mach-O, skipping dep rewrite: $bin"
+                  return
+                fi
+
+                echo "✏️ Rewriting deps → @rpath for: $bin"
+
+                # /nix/store/* → @rpath/<basename>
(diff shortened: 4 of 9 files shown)


SoprachevAK:
Useful. Tested it on an Intel Mac, works.

low-batt:
Hello from a _junior_ IINA developer!

Normally I wait to comment until I have fully studied an issue / pull request so my comments will be somewhat coherent. But this PR caught me totally overloaded. I have only had time to skim over it. However I did not want to leave you waiting for a response, so here are some possibly very incoherent and dumb comments…

Note that my comments mostly don't matter. The senior IINA developers handle this aspect of IINA.

My initial impression is one of excitement. I have found the current way IINA builds the libraries it is dependent upon to be problematic.

It is clear you know a lot about IINA, so you may already be aware of some of what I'm about to cover…

First I want to explain the current state of IINA development and why I might be hard to get the attention of the senior developers at this moment in time.

The IINA 1.4.0 development was planed to split between two beta releases. Beta releases were considered essential for this release because of the need to address long delayed technical debt in crticial areas of IINA that was considered risky to change. The risky changes went into the first beta. Problems with those changes were found by beta testers. Most of those problems had been fixed with only a couple remaining when darkness descended upon IINA in the form of macOS Tahoe 26 which badly broke IINA causing multiple problems including buttons malfunctioning.

The problems when running under macOS Tahoe were so bad that release plans were changed. Fixes for _some_ of the macOS Tahoe problems were hastily added to IINA and 1.4.0 was rushed out the door with some fixes in pull requests planed for 1.4.0-beta2 left sitting on the dock as the 1.4.0 ship sailed away.

Thus what went out as 1.4.0 was supposed to be 1.4.0-beta2. From the issues that have poured in I believe it is clear IINA could have greatly benefited from a second beta. IINA 1.4.1 was a quick release that only addressed a few of the regressions.

For 1.4.0 IINA upgraded the versions of the dependent libraries IINA uses. That upgrade took about a month and a half as regressions were found in the libraries and IINA had to report those problems to the associated project, obtain fixes and then apply patches and rebuild the libraries. One of the questions on this PR will be how we can apply patches to sources outside of IINA.

Unfortunately more than one regression in the libraries IINA is dependent upon are causing serious problems in 1.4.0. The most reported one is a crash playing AV1 encoded videos on Mac's with Apple Silicon chips. The plan had been to speed up the release IINA 1.5.0 which includes updating, pushing most planned work to future releases in order to focus 1.5.0 on the library upgrade and fixing regressions. But that plan has run into trouble.

For many people in the world the cost of Apple hardware means they cannot afford to upgrade their Apple hardware on the schedule decreed by Apple. IINA is sensitive to this and tries to support old versions of macOS running on old versions of hardware. For IINA 1.3.5 the libraries for Intel Macs were built to support macOS 10.11+ running on Intel Core 2 processors. Not clear that support for macOS 10.11 was successful as crashes were reported.

IINA 1.4.0 was forced to upgrade the minimum supported macOS version to 10.15 due to mpv raising its minimum supported macOS and removing code that supported earlier versions of macOS. The current way IINA builds dependent libraries abuses [brew](https://brew.sh/). The Homebrew developers are very clear about it not being a build system and not supporting cross compilling. To be able to build libraries for old versions of macOS and old hardware using, brew must be patched. This is far from ideal as the patches become out of date and fail to apply when brew is updated. When the IINA 1.4.0 were built the patch for forcing brew to believe it was running on a Core 2 processor did not work. One of the regressions in IINA 1.4.0 is that it is crashing on old hardware apparently due to use of x86 instructions that are missing from Core 2 hardware.

I see a big advantage of nix is that it officially supports [cross-compilation](https://nix.dev/tutorials/cross-compilation). I see that the build includes both Apple Silicon and Intel. I did not spot where older versions of macOS and older hardware is handled. I may have missed it in my quick look.

Back to the current situation with IINA…

The attempt to build the latest versions of the libraries for macOS 10.15 has run into trouble. One of the libraries that is not building is [libjxl](https://github.com/libjxl/libjxl). I'm not doing this work, so I'm not familiar with the problems. As a result it is looking like IINA 1.5.0 will need to move the minimum supported macOS version to macOS Big Sur 10.11. There are some important advantages for IINA in not supporting 10.15, but I still consider this a bummer.

As it is desirable to leave users stuck on macOS 10.15 with a better release than IINA 1.4.1, the team is now working on an IINA 1.4.2 release with additional fixes for problems reported against 1.4.0 / 1.4.1. It appears at least one serious regression is due to mpv 0.38.0. For that reason I believe IINA will need to still keep IINA 1.5.0 focused on upgrading the libraries and fixing problems.

On this:
> do we need to keep copying from deps/lib in the Xcode build pipeline

Because if the bad history of finding new regressions when we update libraries the senior developers _may_ want to continue to build libraries separately. If not, then we'd need to be able to "pin" the versions of the libraries, which I believe nix can do. If the build is changed to build all of the libraries, then yes, there will be no need for deps/lib. Will have to see what the other developers think.

On this:
> Proper code signing support – We may want to implement non-adhoc codesigning to enable this solution to be used for releasing IINA to the AppStore.

For security reasons the IINA build does not include the official signature used for IINA release and beta builds.

An AppStore release is a non-trivial topic. There is an issue for regarding that. For example, IINA is using a private Apple framework for PiP support. Pretty sure that violates an AppStore rule.

I've run out of time and this comment is way too long.

florius0:
Hello and thank you for your fast response!

> It is clear you know a lot about IINA, so you may already be aware of some of what I'm about to cover…

I'm actually not that familiar with the current state of affairs so thank you for your detailed overview.

> One of the questions on this PR will be how we can apply patches to sources outside of IINA.

I believe this case is supported by nix as well, for example one could write:

```nix
# Create an overlay that modifies libjxl
overlay = final: prev: {
  libjxl = prev.libjxl.overrideAttrs (old: {
    patches = (old.patches or []) ++ [ ./fix-libjxl.patch ];
  });
};

pkgs = import nixpkgs {
  inherit system;
  overlays = [ overlay ];
};
```

So you can patch denendecy source code w/out forking it, and have this patch live in your repo.

> Unfortunately more than one regression in the libraries IINA is dependent upon are causing serious problems in 1.4.0. The most reported one is a crash playing AV1 encoded videos on Mac's with Apple Silicon chips. The plan had been to speed up the release IINA 1.5.0 which includes updating, pushing most planned work to future releases in order to focus 1.5.0 on the library upgrade and fixing regressions. But that plan has run into trouble.

Looks like this build supports av1 on Apple Silicon, I've just tested it by playing [this video](https://test-videos.co.uk/vids/sintel/mp4/av1/360/Sintel_360_10s_1MB.mp4)

> I did not spot where older versions of macOS and older hardware is handled

There indeed is no explicit handling of older hardware/macOS in this PR. Perhaps you can point out where in the current build process it is handled so I can re-implement it in nix?

> Because if the bad history of finding new regressions when we update libraries the senior developers may want to continue to build libraries separately. If not, then we'd need to be able to "pin" the versions of the libraries, which I believe nix can do. If the build is changed to build all of the libraries, then yes, there will be no need for deps/lib. Will have to see what the other developers think.

If my understanding is correct, this build essentially already does it, since the libraries come from nixpkgs, which are set to specific version.

It is also possible to make some/all libraries to be built from  source and pin the source code version if we want to be extra-explicit about it.

> For security reasons the IINA build does not include the official signature used for IINA release and beta builds.

Understood.
I thought the process here is similar to how the iOS apps are built & signed, so the secrets can be stored as GitHub secrets to enable CI builds.

> An AppStore release is a non-trivial topic. There is an issue for regarding that. For example, IINA is using a private Apple framework for PiP support. Pretty sure that violates an AppStore rule.

Thank you for bringing this up. As I said I'm new to native macOS development, yet I've know some iOS apps use private Apple APIs for things like application icon changes, so my understanding is that apple is not that strict.

I'm very much looking forward to collaborate further on this one!

uiryuu:
@florius0 First of all, thanks for your time and expertise on this. I didn't have enough time to look into this but this already looks pretty interesting and promising to me. Just to clarify some of the current deps building process regarding to old systems. 

We are using Homebrew to compile every dependencies of mpv (and FFmpeg). Because we have a specific target minimum macOS version to support, previously we patch Homebrew compiling script to inject compilation flags during compiling. 

But later we found it was increasingly annoying and unstable since different packages use different building system, then we have to inject different flags for each package semi-manually. Plus the fact that Homebrew is more and more unwilling to let developers to hack it, which make it harder to inject flags. So instead, we now compile all the packages *on the minimum system that we want to support*. Therefore I suppose if nix can compile the package locally, then there's no problem to support older system versions.

florius0:
> Because we have a specific target minimum macOS version to support, previously we patch Homebrew compiling script to inject compilation flags during compiling

> Plus the fact that Homebrew is more and more unwilling to let developers to hack it, which make it harder to inject flags. So instead, we now compile all the packages on the minimum system that we want to support. Therefore I suppose if nix can compile the package locally, then there's no problem to support older system versions.

If the main modification to the packages is changing compilation flags I suppose it's even easier in nix:

```nix
libjxl.overrideAttrs (old: {
  # Flags for CMAKE
  cmakeFlags = (old.cmakeFlags or []) ++ [
    "-DJPEGXL_ENABLE_TOOLS=OFF"
    "-DJPEGXL_ENABLE_BENCHMARK=OFF"
  ];
  # Flags for all GCC/Clang invocations
  NIX_CFLAGS_COMPILE = (old.NIX_CFLAGS_COMPILE or "") + " -O2 -march=native";
});
```

> But later we found it was increasingly annoying and unstable since different packages use different building system

This problem will likely remain since nix doesn't force particular build systems.
It should be possible to start with the source code and write the package build by hand to force similar build processes across the packages, but I doubt that it will actually simplify things.

Maybe you can provide an example of what flags do you use for some particular library?

As far as I gathered from the issues, x86 libraries are build using `-march=core2`, is that correct?

I'll try to look into how each dependency is built.

low-batt:
@uiryuu is the expert, so I will limit my comments…

On this:
> As far as I gathered from the issues, x86 libraries are build using -march=core2, is that correct?

The IINA 1.3.5 libraries were built that way. As @uiryuu touched on there was trouble getting this to work for the 1.4.0 libraries resulting in the crash reported in issue #5739. I'll defer to @uiryuu for what the x86 chip requirements will be for the IINA 1.5.0 libraries. My only concern is that we document what IINA supports.

On this:
> Looks like this build supports av1 on Apple Silicon, I've just tested it by playing [this video](https://test-videos.co.uk/vids/sintel/mp4/av1/360/Sintel_360_10s_1MB.mp4)

We think the latest version of [libdav1d](https://code.videolan.org/videolan/dav1d) fixes the problem. But this appears to be one of those thread coordination problems where on some Mac the problem is very rare and hard to reproduce and on other Macs it occurs frequently. It seemed like running under macOS Tahoe made it more likely to occur. I've not been able to crash IINA playing AV1 videos with the latest version of `libdav1d`. But it was always very hard for me to trigger this problem.

On this:
> I thought the process here is similar to how the iOS apps are built & signed, so the secrets can be stored as GitHub secrets to enable CI builds.

I believe you can do that, if you are willing to trust having credentials in a "secure" cloud.

On this:
> some iOS apps use private Apple APIs for things like application icon changes, so my understanding is that apple is not that strict.

Yes, in the past IINA made it past that check and into the App Store. Illegally. See issue #420. I tracked down the App Store issue I was remembering, it is #2152. There are some other [App Review Guidelines](https://developer.apple.com/app-store/review/guidelines/) IINA does not meet. Further discussion should be in issue #2152.

florius0:
@low-batt @uiryuu Hi!

Apologies for the delay.

> The IINA 1.3.5 libraries were built that way

I've checked the resulting binaries, including ones provided by nixpkgs, and it looks like they are already compatible with core2 macs.

```sh
$ otool -hv result/Applications/IINA.app/Contents/MacOS/IINA
result/Applications/IINA.app/Contents/MacOS/IINA (architecture x86_64):
Mach header
      magic  cputype cpusubtype  caps    filetype ncmds sizeofcmds      flags
MH_MAGIC_64   X86_64        ALL  0x00     EXECUTE   138      12360   NOUNDEFS DYLDLINK TWOLEVEL BINDS_TO_WEAK PIE
result/Applications/IINA.app/Contents/MacOS/IINA (architecture arm64):
Mach header
      magic  cputype cpusubtype  caps    filetype ncmds sizeofcmds      flags
MH_MAGIC_64    ARM64        ALL  0x00     EXECUTE   139      12112   NOUNDEFS DYLDLINK TWOLEVEL BINDS_TO_WEAK PIE

$ otool -hv result/Applications/IINA.app/Contents/Frameworks/ffmpeg
result/Applications/IINA.app/Contents/Frameworks/ffmpeg (architecture x86_64):
Mach header
      magic  cputype cpusubtype  caps    filetype ncmds sizeofcmds      flags
MH_MAGIC_64   X86_64        ALL  0x00     EXECUTE    84       5920   NOUNDEFS DYLDLINK TWOLEVEL PIE
result/Applications/IINA.app/Contents/Frameworks/ffmpeg (architecture arm64):
Mach header
      magic  cputype cpusubtype  caps    filetype ncmds sizeofcmds      flags
MH_MAGIC_64    ARM64        ALL  0x00     EXECUTE    83       5864   NOUNDEFS DYLDLINK TWOLEVEL PIE

$ otool -hv result/Applications/IINA.app/Contents/Frameworks/libdav1d.7.dylib
result/Applications/IINA.app/Contents/Frameworks/libdav1d.7.dylib (architecture x86_64):
Mach header
      magic  cputype cpusubtype  caps    filetype ncmds sizeofcmds      flags
MH_MAGIC_64   X86_64        ALL  0x00       DYLIB    16       1672   NOUNDEFS DYLDLINK TWOLEVEL NO_REEXPORTED_DYLIBS
result/Applications/IINA.app/Contents/Frameworks/libdav1d.7.dylib (architecture arm64):
Mach header
      magic  cputype cpusubtype  caps    filetype ncmds sizeofcmds      flags
MH_MAGIC_64    ARM64        ALL  0x00       DYLIB    16       1752   NOUNDEFS DYLDLINK TWOLEVEL NO_REEXPORTED_DYLIBS
```

I defer to your expertise here, from my understanding it is correct, since the `cpusubtype` lists `ALL`. Is there any way we can test it?

Regarding other matters:

> I believe you can do that, if you are willing to trust having credentials in a "secure" cloud.

If my memory serves me well, one can produce an app-specific set of credentials to revoke them later in a case of some security breach.
That's up to the holder of the credentials, of course.

> Yes, in the past IINA made it past that check and into the App Store

That explains why I've seen it in the App Store at some point and assumed this was an official build, henceforth all my questions on App Store support. From the context you've provided I gather it isn't required as of now and in forceable future.

And the most important part:

What else do we need for this PR to be accepted? Is there any other requirements we have not discussed yet, e.g. CI Support, debug builds, using different xcconfigs etc?

low-batt:
On this:
> I defer to your expertise here, from my understanding it is correct, since the `cpusubtype` lists `ALL`. Is there any way we can test it?

That was my first thought as well when I saw `ALL`. I tested `otool` and it reported `ALL` for a library that I'm pretty sure does not support Core 2. Confusing answers from the net, but from a quick look I think these are the choices for `cpusubtype` when the `cputype` is `CPU_TYPE_X86_64`:
https://github.com/hdoc/llvm-project/blob/a38b25fa77bdf1437c690494ae6d61179b3bb4f8/llvm/include/llvm/BinaryFormat/MachO.h#L1489-L1492

Seems like people gave up on keeping up with all the different x86 instruction sets? Or maybe it was never intended to represent that? It certainly would have been nice to include such information in a library. But I don't think `cpusubtype` provides what we want. A true test would require running on actual hardware. I don't have anything that could test that Core 2 is supported.

The `otool` output I'm interested in seeing is whether the minimum supported macOS versions are correct. This is me checking the `libjxl` library:
```text
low-batt@gag lib (develop *$%=)$ otool -l libjxl.0.10.dylib | grep -E -A4 '(LC_VERSION_MIN_MACOSX|LC_BUILD_VERSION)' | grep minos
    minos 10.15
    minos 12.0
low-batt@gag lib (develop *$%=)$ 
```

I've heard there is trouble building the latest `libjxl` for macOS 10.15. What macOS version were you able to build for?

I'll pass on commenting about storing credentials in the cloud as that is an issue for the senior developers.

On the App Store…

From what I've read there was interest at one point, but I think a lot things need to change before that is possible. Some of those things are slowing being addressed.

On the important part…

Remember, I'm the _junior_ IINA developers. What matters is the comments from the senior developers. The focus at the moment is on fixing some regressions and releasing IINA 1.4.2. Once that is done then we are back to work on 1.5.0 which includes the library upgrades and the potential for changing how we build the libraries. So expect to hear from them after 1.4.2 is out. Hopefully that happens soon.

I was hoping to have time to try this out myself and dig into the code, but have been too busy fixing problems. From just quickly looking at the code I'm unsure what `expandAppEnvironmentVariables` is doing and why it is needed. I think a comment would be helpful.

florius0:
> A true test would require running on actual hardware. I don't have anything that could test that Core 2 is supported.

Neither do I, will try to ask around.

> I've heard there is trouble building the latest libjxl for macOS 10.15. What macOS version were you able to build for?

I didn't build it, the binary comes from `nixpkgs`. If you are referring to the macOS version in general, I didn't change any macOS version-related build settings on the XCode side of things. The `.app` I've attached was built on macOS `15.7.2` & XCode `26.0.1`.

> So expect to hear from them after 1.4.2 is out. Hopefully that happens soon.

Thank you for clarifying timelines.

> what expandAppEnvironmentVariables is doing and why it is needed

It rewrites environment variables' values, replacing dylib-style paths with the absolute paths in order to set proper paths for the python & vapoursynth libs/executables that are bundled into the `.app`.
My understanding is that the paths here must be absolute, since there is no automatic expansion due to the fact they aren't evaluated by a shell.

low-batt:
I had time to try out this PR today and got stuck on the first step, installing nix. I found all kinds of different advice on how to install nix under macOS. Some of it had to do with the normal nix install not providing a way to uninstall? A lot of it was about moving from brew to nix. I don't want to do that. I want to install nix so I can test out this PR.

How should I install nix?

florius0:
> A lot of it was about moving from brew to nix. I don't want to do that

You don't have to. By itself nix is just another package manager, it doesn't force you into using it exclusively or managing your entire system with it.

> How should I install nix?

There are two popular ways to do that – [official nix install script](https://nixos.org/download/) and [DeterminateSystems' nix installer](https://github.com/DeterminateSystems/nix-installer). I recommend the latter since it provides a cleaner way to manage the installation via `/nix/nix-installer`, including uninstalling it (`sudo /nix/nix-installer uninstall`).

PS

Uninstalling the official installation is of course also possible, it's just a [little bit involved](https://nix.dev/manual/nix/2.21/installation/uninstall)

low-batt:
I went with your recommendation and used the Determinate Systems installer:
```text
low-batt@gag iina (florius0/develop $)$ nix --version
nix (Determinate Nix 3.12.0) 2.32.1
```

The build ran for a long time and then failed. Terminal output and nix log below. Looks like a permission problem? Thoughts?

<details>
<summary><b>Terminal output:</b></summary>

```text
low-batt@gag iina (florius0/develop $)$ nix build
error: Cannot build '/nix/store/rzjcmn3i3jg98pcwx7ggmj4vsqbn154r-iina-d32e0774.drv'.
       Reason: builder failed with exit code 1.
       Output paths:
         /nix/store/r6scc63ssl0kvqbx37g7mmkz6dj2i14j-iina-d32e0774
       Last 25 log lines:
       > ✏️ Patching install names in deps/lib/libpcre2-posix.dylib
       > ✏️ Patching install names in deps/lib/libpostproc.58.3.100.dylib
       > ✏️ Patching install names in deps/lib/libxcb-xv.la
       > error: install_name_tool: input file: deps/lib/libxcb-xv.la is not a Mach-O file
       > ✏️ Patching install names in deps/lib/libharfbuzz-subset.dylib
       > ✏️ Patching install names in deps/lib/libhwy_test.dylib
       > ✏️ Patching install names in deps/lib/libxcb-screensaver.0.0.0.dylib
       > ✏️ Patching install names in deps/lib/libxcb-record.0.0.0.dylib
       > ✏️ Patching install names in deps/lib/libturbojpeg.0.dylib
       > ✏️ Patching install names in deps/lib/libsodium.la
       > error: install_name_tool: input file: deps/lib/libsodium.la is not a Mach-O file
       > ✏️ Patching install names in deps/lib/libxcb-xv.0.dylib
       > ✏️ Patching install names in deps/lib/libxcb-present.0.0.0.dylib
       > ✏️ Patching install names in deps/lib/libhwy_contrib.dylib
       > ✏️ Patching install names in deps/lib/libgmpxx.dylib
       > ✏️ Patching install names in deps/lib/libgio-2.0.0.dylib
       > ✏️ Patching install names in deps/lib/libgettextpo.dylib
       > ✏️ Patching install names in deps/lib/libxcb-xvmc.la
       > error: install_name_tool: input file: deps/lib/libxcb-xvmc.la is not a Mach-O file
       > ✏️ Patching install names in deps/lib/libhwy_contrib.1.0.7.dylib
       > ✏️ Patching install names in deps/lib/libxcb-shm.0.0.0.dylib
       > ✏️ Patching install names in deps/lib/libhwy_contrib.1.dylib
       > ✏️ Patching install names in deps/lib/libjxl_threads.0.11.dylib
       > ✏️ Patching install names in deps/lib/libX11-xcb.dylib
       > 📦 Copying SPM deps
       For full logs, run:
         nix log /nix/store/rzjcmn3i3jg98pcwx7ggmj4vsqbn154r-iina-d32e0774.drv
error: Cannot build '/nix/store/jy05jwb98gnp0qrjk7v80blj8gaxanw4-iina-universal-d32e0774.drv'.
       Reason: 1 dependency failed.
       Output paths:
         /nix/store/dlif8540srwg32v7mkvd4i429crl76ja-iina-universal-d32e0774
low-batt@gag iina (florius0/develop $)$ 
```
</details>

[nix.log](https://github.com/user-attachments/files/23294915/nix.log)

florius0:
Looks like the build failed [here](https://github.com/florius0/iina/blob/d32e0774816a73d02cd41ce9488c64e0eed43ccb/flake.nix#L652).
Yet it's quite strange that there is no error in logs.

I suspect one of the four reasons:

1. Your system doesn't have `rsync` installed, in that case you should see something like `bash: rsync: command not found` in the logs. If it is indeed the case, we should use rsync managed by nix. To confirm:

    ```shell
    nix develop
    which rsync
    ```

2. `spmDeps` build failed which caused `${spmDeps}` path to be missing. To confirm:

    ```shell
    ls /nix/store/*iina-spm-deps*
    ```

3. Perhaps, there is a problem with nix sandboxing. I've noticed that it is disabled on my system. Check if it is on yours via 

    ```shell
    nix show-config | grep -i sandbox
    ```

    If you see `sandbox = true` run the build with `--option sandbox false` to confirm it was the issue:
    
    ```shell
    nix build --option sandbox false --option debug 1 --keep-failed --print-build-logs
    ```

5. It is possible that rsync is silently OOM-killed, in that case we should replace it with smth more memory efficient. Check the memory consumption during build to confirm.


If nothing helps, can you run the build again with the following flags `nix build --option debug 1 --keep-failed --print-build-logs`, which directs nix to do additional logging and keep the temporary build directories in case of a failure, which will allow me to inspect them later.

low-batt:
I do have `rsync`:
```text
low-batt@gag iina (florius0/develop $%)$ which rsync
/usr/bin/rsync
```

Debugging the build it seems to be this line that is failing:
```text
old_prefix=$(grep -Eo "/tmp/nix-build-[^/]+/source" .spm/workspace-state.json | head -n1)
```

The `workspace-state.json` file is present, but it does not contain any paths that start with `/tmp`. Here is the file:

[workspace-state.json.zip](https://github.com/user-attachments/files/23377176/workspace-state.json.zip)


I did confirm `sandbox` is set to `false`:

```text
low-batt@gag iina (florius0/develop $%)$ nix show-config | grep -i sandbox
warning: 'show-config' is a deprecated alias for 'config show'
darwin-log-sandbox-violations = false
sandbox = false
sandbox-fallback = true
sandbox-paths = /System/Library/Frameworks /System/Library/PrivateFrameworks /bin/bash /bin/sh /private/tmp /private/var/tmp /usr/lib
low-batt@gag iina (florius0/develop $%)$ 
```

florius0:
> Debugging the build it seems to be this line that is failing:
> 
> ```
> old_prefix=$(grep -Eo "/tmp/nix-build-[^/]+/source" .spm/workspace-state.json | head -n1)
> ```
> 
> The `workspace-state.json` file is present, but it does not contain any paths that start with `/tmp`

Looks like build dir was changed in newer nix versions, and I didn't account for it.
Lets compare the versions to confirm:

```shell
$ nix --version
nix (Nix) 2.28.4
```

low-batt:
Mine is newer:
```text
low-batt@gag iina (florius0/develop *$%)$ nix --version
nix (Determinate Nix 3.12.0) 2.32.1
```

svobs:
I too tried to build using the latest version of Determinate Nix, and encountered the same issue. First time using Nix myself, and I'm not familiar enough with SwiftPM to have any leads on a direct solution. But it stumbled upon a post which states that [Determinate Nix just dropped support for macOS Intel](https://github.com/DeterminateSystems/nix-src/issues/224). So maybe sticking to a fixed older version of Nix would be the prudent solution.

It took me a while stumbling around in the dark but I got it to build by downgrading my copy of Nix. I learned that each version of `nix-installer` is pegged to a particular version of `nix`, but trying to find which one was closest to Nix 2.28.4 was harder than I expected. Oddly I wasn't able to find any table or set of release notes online which showed the two versions side by side. I took a pot shot based on the 2.28 release date in April, and ended up downloading v3.4.1 of `nix-installer`. This installed Nix 2.28.2, which was close enough (for now).

Here are the steps I used to set up Nix from scratch:

0. If Determinate Nix is already installed, first run the uninstaller to remove it: `/nix/nix-installer uninstall`
1. Download `nix-installer-aarch64-darwin` from [nix-installer release v3.4.1](https://github.com/DeterminateSystems/nix-installer/releases?page=4) because my Mac uses Apple Silicon.
2. In a terminal, add execute permission to the downloaded file: `chmod +x nix-installer-aarch64-darwin`
3. Try to execute it via terminal: (e.g.,  from its directory,`./nix-installer-aarch64-darwin`). It's not blessed by Apple, so do the usual song & dance in System Settings > Privacy & Security to get it out of quarantine & give it permission to run, and run it again.
4. When it's done installing, run `. /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh` (as it tells you) to start the daemon.
5. Verify the correct version using `nix --version`:
    ```
    msvoboda|01:02:05|iina-2$ nix --version
    nix (Determinate Nix 3.4.1) 2.28.2
    ```

6. Important: when building, I kept getting this error:
    ```
    warning: download buffer is full; consider increasing the 'download-buffer-size' setting
    ```
    I fixed this by appending the following line to `/etc/nix/nix.custom.conf`:<p>
    ```
    download-buffer-size = 800000000  # 800 MB
    ```
    (I saw people on forums using 500 MB, but that was not enough to get rid of the error.)

7. Then `cd` to the IINA project directory and run: `nix build`.

So incredible that there weren't more problems getting this working. Thank you @florius0 for what must have been a lot of work! I for one hope to find this very useful and will work to dive in to it further.

florius0:
> I'm not familiar enough with SwiftPM to have any leads on a direct solution

The actual problem is not with SwiftPM, but with the way it is cached & restored later.

> Important: when building, I kept getting this error:

Interesting, I didn't encounter it at all.

> So incredible that there weren't more problems getting this working.

Thank you, I hope it'll be useful.

Apologies for the delay, there was a lot of things to be done lately, I hope I'll be able to update it to support modern nix versions this weekend.

low-batt:
I've been waiting for an update before proceeding further. I'm unsure of the next step. IINA currently follows Apple's mindset which appears to be "YOU WILL USE THE LATEST VERSIONS OF EVERYTHING". This is why the IINA [README](https://github.com/iina/iina?tab=readme-ov-file#using-the-pre-compiled-libraries) says:
> 2. Open iina.xcodeproj in the [latest public version of Xcode](https://apps.apple.com/app/xcode/id497799835). _IINA may not build if you use any other version._

That Determinate Nix has dropped support for building on Macs with Intel chips definitely causes a problem for IINA. Will need to hear the senior developers thoughts on relying upon an older version of Nix to be able to build on Intel Macs.

I am going to try and follow @svobs instructions and see if I can get a build to complete.

svobs:
@low-batt:

> IINA currently follows Apple's mindset which appears to be "YOU WILL USE THE LATEST VERSIONS OF EVERYTHING". 

@florius0 can correct me if I'm way off but in my understanding, Nix is just a tool and environment manager, so it shouldn't be tied to an particular version of Xcode. So it should be fine to use an old version of Nix to manipulate the latest version of Xcode. Moreover, it might be a good general strategy to keep the official build locked to a specified Nix release, so as not to be blindsided by any side effects of some future Nix release which could potentially break things again.

@florius0:

> > Important: when building, I kept getting this error:
> 
> Interesting, I didn't encounter it at all.

Yeah I'm not entirely sure what to make of it. If I interpreted the status text correctly doing a full `nix build` from a fresh install of Nix resulted in ~1.3 GB of data downloaded, but I can't imagine any single file being larger than ~ 100MB. I see that Determinate Nix advertises that they have been parallelizing a lot of tasks. I wonder if that includes concurrent downloads.

> Apologies for the delay, there was a lot of things to be done lately, I hope I'll be able to update it to support modern nix versions this weekend.

No problem! You've given plenty to chew on already 👌

florius0:
> ...can correct me if I'm way off but in my understanding, Nix is just a tool and environment manager...

@svobs yes you are correct.

> resulted in ~1.3 GB of data downloaded

If I remember correctly, I had similar download size.
Perhaps there is difference here between upstream Nix and Determinate Nix.

> That Determinate Nix has dropped support for building on Macs with Intel chips definitely causes a problem for IINA

From the issue https://github.com/DeterminateSystems/nix-src/issues/224 @svobs mentioned I gather that it only affects Determinate Nix, and it looks like upstream Nix has no plans like this since I was unable to find any architecture-specific discussions.

So we can (and perhaps should) use upstream Nix.

low-batt:
> So we can (and perhaps should) use upstream Nix.

I think we have to do that. I'm certain the senior developers will consider it too soon to kill off the ability to build IINA on Macs with Intel chips.

Determinate Nix already has a complicated install in order to deal with macOS security features. An old version might not be compatible with a new macOS version that tightens security. We don't want to be locked into old versions of build tools.

grahamc:
> I think we have to do that. I'm certain the senior developers will consider it too soon to kill off the ability to build IINA on Macs with Intel chips.

Hi, in case it is useful: Determinate Nix supports building _for_ Intel Macs as long as it is building _on_ an Apple Silicon Mac. Upstream Nixpkgs has also announced deprecation of support [starting next year](https://github.com/NixOS/nixpkgs/pull/415566).

Ultimately, we found only 0.02% or so of our users were on Intel Macs, and about 95% of that 0.02% was in CI where it is easy to switch to an Apple Silicon based Mac.

I hope that information is useful to you. I'm sorry for the disturbance and deprecation timeline. 

I love the project!

low-batt:
@grahamc Thanks so much for pointing that out.

I will see if I can get a senior IINA developer to comment on IINA's plans for being able to build on Macs with Intel chips. Won't be long before Macs with Intel chips won't be able to run the latest version of Xcode that IINA prefers to use. So Apple will also be forcing a decision on using Intel Macs for building.IINA. Once that happens it will become painful to debug Intel specific problems.

low-batt:
Following the instructions from @svobs I was able to successfully complete a build.

Yeah!

Running `nix build` seems to have created a `result` link. In that directory there is an `Applications` directory. In `Applications` I found `IINA.app`.

Big picture question first…

How do I develop IINA? How can I debug from Xcode? How can I make a change and take advantage of Xcode's fast incremental build? In other words, why is Nix needed for more than building the 3<sup>rd</sup> party libraries? I would have expected the Nix build to just populate `deps/lib` and `deps/executable` and Xcode would continue to be used to build the IINA application? I'm expecting this is due this goal:
> 3. Simplify the build process by describing it in a single `flake.nix` file

I too like that. But we must be able to build, debug and analyze in Xcode. I'm lost on how to use Xcode with this.

The resulting `IINA.app` crashed for me under macOS 10.15:
```text
Termination Reason:    DYLD, [0x1] Library missing

Application Specific Information:
dyld: launch, loading dependent libraries

Dyld Error Message:
  Library not loaded: /usr/lib/swift/libswiftUniformTypeIdentifiers.dylib
  Referenced from: /Volumes/transfer/IINA.app/Contents/Frameworks/libmpv.2.dylib
  Reason: image not found
```
[Crash Report.txt](https://github.com/user-attachments/files/23565120/Crash.Report.txt)

I also noticed FFmpeg 7 is being used. With the upgrade to mpv 0.40.0 we are also upgrading to FFmpeg 8. I could not figure out how to get the Nix build to use FFmpeg 8.

We want to make sure FFmpeg is built with libbs2b. See issue #5149.

All generated files have a modified date of Dec 31, 1969!

florius0:
@low-batt thank you for trying it out!

> How do I develop IINA? How can I debug from Xcode? How can I make a change and take advantage of Xcode's fast incremental build? In other words, why is Nix needed for more than building the 3rd party libraries? I would have expected the Nix build to just populate `deps/lib` and `deps/executable` and Xcode would continue to be used to build the IINA application? I'm expecting this is due this goal:

> > 3. Simplify the build process by describing it in a single flake.nix file

> I too like that. But we must be able to build, debug and analyze in Xcode. I'm lost on how to use Xcode with this.

It is a perfectly valid concern, so let's address it.

> why is Nix needed for more than building the 3rd party libraries

Initially my idea was to only package the libraries, yet when implementing vapoursynth support I've discovered that I also need to package not only the libs it requires, but the python interpreter as well. Since I'm not very familiar with XCode build system, I chose to do it in [flake.nix](https://github.com/florius0/iina/blob/d32e0774816a73d02cd41ce9488c64e0eed43ccb/flake.nix#L702-L741) as well.
I also found out that it needs other dependencies not directly listed anywhere in the original source code, so I've included them as well (`depsInderect`).
And since I wanted to address #5443, I included some executables as well (`depsExecutable`).

> I would have expected the Nix build to just populate `deps/lib` and `deps/executable`

In a sense it already does it, and more, but in the sandboxed build environment.

Going forward I see two options (they don't exclude each other):

1. Nix has a feature called "nix shell" that enables development environment setup to be used outside of flake. That probably will be the easiest option to support development processes you've mentioned.
   The caveat here is that currently we're doing a bunch of additional things in the flake after `xcodebuild` step, and ideally we should move them into XCode build system, so the flake builds and plain XCode builds during development are as similar as possible. And it shouldn't be that hard since from my brief overview of the build process, XCode can execute shell commands/scripts as part of it, so we simply need to move & adjust them.
   If we decide to simply copy the libs in this scenario, the difference between the builds will be the following:

   1. XCode build will be linked against some libraries in `/nix/store` (or all, depending on the implementation)
   2. XCode build won't have vapoursynth/python packaged
   3. XCode build won't have executables packaged
   
   Perhaps we could accept it at least for a starting point, then the solution you've described (just copy the dependencies) will be trivial.

2. Make the development process more flake-centric. Correct me if I'm wrong, but it should be possible to attach debugger/analysis tools to already packaged build. On the Nix side of things it should be possible to implement incremental builds to give you the speed you want.

> The resulting IINA.app crashed for me under macOS 10.15:

Regarding the crash, did it occur at the application start?
I've noticed I don't have this library on my system, yet the .app I've built doesn't crash with similar message.
Can you share the resulting `.app`?

> I also noticed FFmpeg 7 is being used. With the upgrade to mpv 0.40.0 we are also upgrading to FFmpeg 8. I could not figure out how to get the Nix build to use FFmpeg 8.

Yep, I've noticed 1.4.1 is built with FFmpeg 7, so I've used the same FFmpeg version.
FFmpeg 8 is available on `unstable` nixpkgs channel, I'll switch them.

> We want to make sure FFmpeg is built with libbs2b. See issue https://github.com/iina/iina/issues/5149.

```nix
# Override ffmpeg to use our version of libs
ffmpeg = pkgs.ffmpeg_7.override {
  withSoxr = true;
  withBs2b = true;
  rubberband = pkgs.rubberband;
  libplacebo = pkgs.libplacebo;
  libjxl = pkgs.libjxl;
};
```

It's as simple as that, will do.

> All generated files have a modified date of Dec 31, 1969!

Perhaps it is some nix quirk, is that an issue?

florius0:
Hi, I've implemented some changes we've discussed:

1. Modern nix support. The flake now runs on latest Determinate Nix:

    ```shell
    nix --version
    nix (Determinate Nix 3.13.1) 2.32.4
    ```

    To update, run the Determinate System's Nix installer, it will likely pick up on existing Nix installation. If it plans to re-create volumes, **DON'T**, you will likely remain with dangling umnounted APFS volume. Instead, either uninstall nix by running `sudo /nix/nix-installer uninstall`, or purge the existing installation metadata by running
    
    ```shell
    sudo rm -f /nix/receipt.json
    sudo rm -rf /nix/var/nix/profiles/default/Library/LaunchDaemons/org.nixos.nix-daemon.plist
    ```
    
    If still wants to create volumes, use `.pkg` installer, it is "smarter" about working w/ existing installation.
    If you're not using `nix-darwin` and/or `home-manager` (I assume you don't), it should be perfectly save to delete existing installation.

    If "download buffer is full" error is still present, I refer you to @svobs [comment 6th step](https://github.com/iina/iina/pull/5754#issuecomment-3529493670).

2. FFmpeg is now built w/ libbs2b.

3. I've looked into devshells, it will require slight refactoring on the flake side of things, since right now the cleanest solution requires IINA to be built, which may not always be possible during development.

4. I've removed "fixing compatibility step" (some libraries were copied around to conform to versions pinned in the xcode project) and instead update the xcode project itself to reference newer versions.

5. I've tried upgrading to ffmpeg 8 and encountered an issue. There's a clash between Apple's libiconv and GNU's. They both have same names, but different ABIs.
   I've modified rpaths manually and managed to get a working build (can share it if you'd like), however I believe such issues should be resolved automatically since they are extremely time-consuming to work with.
   Right now, normalizer script distinguishes between different libraries based only on their name, which leads to incorrect linking (other libraries expect Apple's and get GNU's vice versa). I'll likely will opt into changing that to full path comparisons.

low-batt:
Busy day for me today. I will upgrade Nix and test once I get a chance.

On this:
> I've removed "fixing compatibility step" (some libraries were copied around to conform to versions pinned in the xcode project) and instead update the xcode project itself to reference newer versions.

That is exactly what we have to do when we upgrade libraries. No need to match up with the versions referenced by the Xcode project. There are a bunch of changes that have to happen when we upgrade the 3<sup>rd</sup> party libraries. You can see some of them in commit 4a58c2e on the `mpv-0.40.0` branch. I believe that is work in progress and not finished.

On this:
> I've tried upgrading to ffmpeg 8 and encountered an issue. There's a clash between Apple's libiconv and GNU's. They both have same names, but different ABIs.

I have a version of IINA running with mpv 0.40.0 and FFmpeg 8 without `libiconv`. Maybe it is using Apple's `libiconv`? Why is GNU's needed?

low-batt:
In a quick look at the build failure I found this failure: 
```text
ld: library 'rubberband.3' not found
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

florius0:
> I believe that is work in progress and not finished.

I think it would be best to wait until it finished and then update the Nix side of things, what is your opinion on this?

> I have a version of IINA running with mpv 0.40.0 and FFmpeg 8 without libiconv. Maybe it is using Apple's libiconv? Why is GNU's needed?

It is possible that your version is linked against `libiconv` provided by the system itself.
I also suspect it may be due to the difference in other dependencies versions or in how ffmpeg is built.

I found this libraries require Apple's libiconv (`/usr/lib/libiconv.2.dylib`)

- `libarchive.13.dylib`
- `libass.9.dylib`
- `libavcodec.62.11.100.dylib`
- `libavformat.62.3.100.dylib`
- `libglib-2.0.0.dylib`
- `libintl.8.dylib`
- `libmpv.2.dylib`

While this libraries require GNU's libiconv

- `libavdevice.62.1.100.dylib`
- `libavfilter.11.4.100.dylib`
- `libdovi.3.3.1.dylib`
- `libgnutls.30.dylib`
- `libidn2.0.dylib`
- `libunistring.5.dylib`
- `libxml2.16.dylib`
- `libzvbi.0.dylib`

By require I mean that in such linking configuration the App doesn't crash at startup, I may be wrong about actual requirements.

> In a quick look at the build failure I found this failure:
> 
> ```
> ld: library 'rubberband.3' not found
> clang: error: linker command failed with exit code 1 (use -v to see invocation)
> ```

I expect CI to fail, since XCode Project was updated to reference newer versions.

low-batt:
On this:
> I think it would be best to wait until it finished and then update the Nix side of things, what is your opinion on this?

The first step is to build the libraries IINA needs. After we have the libraries in hand we start making the changes needed for IINA to use the upgraded libraries. We need to get the Nix build upgraded to FFmpeg 8.

That `libavfilter.11.4.100.dylib` is using GNU's libiconv and other FFmpeg libraries aren't is interesting. Not sure what to make of that. We need to track down what is pulling in GNU's libiconv and figure out if we can break that dependency.

Even with FFmpeg 7.1.1 we are pulling in GNU's libiconv.

On this:
> I expect CI to fail, since XCode Project was updated to reference newer versions.

My bad. I should have realized that would be the case.

On this:
> Regarding the crash, did it occur at the application start?

Yes. The app icon appeared and then IINA crashed.

On this:
> I've noticed I don't have this library on my system, yet the .app I've built doesn't crash with similar message.

And that would be because people in the software industry keep finding new ways to introduce confusing behaviors that are intended to cause you to spend hours desperately searching the Internet and smashing your face on your keyboard in frustraction until you finally find the following under `Known issues` in the [macOS Big Sur 11.0.1 Release Notes](https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_0_1-release-notes):
> - New in macOS Big Sur 11.0.1, the system ships with a built-in dynamic linker cache of all system-provided libraries. As part of this change, copies of dynamic libraries are no longer present on the filesystem. Code that attempts to check for dynamic library presence by looking for a file at a path or enumerating a directory will fail. Instead, check for library presence by attempting to dlopen() the path, which will correctly check for the library in the cache. (62986286)

This is new `libmpv` behavior. The `libmpv` IINA 1.4.1 is using does not reference this library. I will investigate.

Attempting to run the mpv found in `IINA.app/Contents/MacOS/` fails as seen here when not on the Mac I built IINA on:
```text
$ ./mpv --version
-bash: ./mpv: /nix/store/5sq6n9cpwp3jlf8jvfa7apd4wh8wm4a5-bash-5.2p37/bin/bash: bad interpreter: No such file or directory
```

The problem can be seen at the start of the `mpv_identify.sh`:
```text
low-batt@gag MacOS$ cat mpv_identify.sh 
#!/nix/store/6m28anjby1y83l8jsi0kw1zlnjswhk5k-bash-interactive-5.2p37/bin/sh
```

Running `ffmpeg --version` shows a popup that says:
> **Fatal error! Cannot continue!**
> 
> Failed loading SDL 3 library.

Before aborting / crashing.

I've not heard a response from the other IINA developers yet on my questions about the direction we should take.

florius0:
> The first step is to build the libraries IINA needs. After we have the libraries in hand we start making the changes needed for IINA to use the upgraded libraries. We need to get the Nix build upgraded to FFmpeg 8.

Got it. Essentially, the last step to upgrading the flake to ffmpeg 8 is to figure out `libiconv` clash.

> That libavfilter.11.4.100.dylib is using GNU's libiconv and other FFmpeg libraries aren't is interesting. Not sure what to make of that. We need to track down what is pulling in GNU's libiconv and figure out if we can break that dependency.

Yes, it sounds strange to me also, I'll try to look in to it.

> copies of dynamic libraries are no longer present on the filesystem

It explains why despite having no library, the app doesn't crash on my machine on macOS 26.
The strange thing here is it looks like when building on yours it links to it despite you not having it.

> Attempting to run the mpv found in IINA.app/Contents/MacOS/ fails as seen here when not on the Mac I built IINA on

Looks like small packaging error, some scripts of MPV still referencing /nix/store

> Running `ffmpeg --version` shows a popup that says ...

I assume you mean `IINA.app/Contents/MacOS/ffmpeg --version`, will look into it as well.

low-batt:
I tracked down the crash under macOS 10.15. We will need to patch mpv with this commit: https://github.com/mpv-player/mpv/commit/065af28bc2901a25058e041de0be8ea1d5cf05c0

The patch file:

[065af28.patch](https://github.com/user-attachments/files/23686304/065af28.patch)

There is a question as to whether IINA 1.5.0 will support macOS 10.15 or not. Might depend upon whether we can show Nix can build libraries that work on macOS 10.15.

I still haven't heard back from the senior developers on my questions about this PR. I will try again to catch them this weekend.

florius0:
> We will need to patch mpv with this commit https://github.com/mpv-player/mpv/commit/065af28bc2901a25058e041de0be8ea1d5cf05c0

Shouldn't it already be included into [mpv 0.40](https://github.com/mpv-player/mpv/compare/master...release/0.40)?

On this:

> Looks like small packaging error, some scripts of MPV still referencing /nix/store

It was the case indeed, mpv was a shell script referencing `/nix/store/...mpv.../bin/Applications...`, now we package the real executable.

florius0:
> 1. Nix has a feature called "nix shell" that enables development environment setup to be used outside of flake. That probably will be the easiest option to support development processes you've mentioned.
>    The caveat here is that currently we're doing a bunch of additional things in the flake after `xcodebuild` step, and ideally we should move them into XCode build system, so the flake builds and plain XCode builds during development are as similar as possible. And it shouldn't be that hard since from my brief overview of the build process, XCode can execute shell commands/scripts as part of it, so we simply need to move & adjust them.
>    If we decide to simply copy the libs in this scenario, the difference between the builds will be the following:
>    
>    1. XCode build will be linked against some libraries in `/nix/store` (or all, depending on the implementation)
>    2. XCode build won't have vapoursynth/python packaged
>    3. XCode build won't have executables packaged
>    
>    Perhaps we could accept it at least for a starting point, then the solution you've described (just copy the dependencies) will be trivial.


@low-batt I've added development shell as we discussed above. To try it out, run `nix develop` – this will symlink all the dependencies into `deps/` directory, keeping to the original `deps/` layout and adding some more.

This should allow us to run development builds in xcode.
How do we test it?

low-batt:
Remember I'm a _junior_ IINA developer. This part of IINA is controlled by the senior developers, so you do not have to listen to my concerns about this PR. I've been trying to get some guidance from the other developers, but have not been able to get their attention. So I'm going to share my own thoughts.

I prefer as much as possible to keep a PR focused on one change. Not only does it make it easier to review a PR, but it is nicer to have smaller commits if you have to review commit history looking for the change that introduced a problem. This PR combines quite a few major changes. It changes how the libraries IINA uses are built. That is a big change right there. It changes the IINA build to use Nix. It adds vapoursynth support. It includes mpv and FFmpeg command line programs in IINA. I do not recall the senior developers ever mentioning a preference for not combining changes in one PR. But I have seen that big PRs that combine lots of changes end up being passed over during reviewing and merging sessions.

I would have started with a PR that uses Nix to build the libraries IINA is currently using (upgraded to the latest releases) and populate the current directories under `deps` with files (not links). The senior devs could then upload those files to IINA's server as is done today.

With this change anyone wanting to build IINA can continue to use `other/download_libs.sh` and it will download the Nix built libraries. IINA developers who want to test a newer version of a library can install Nix, modify the Nix build and build their own copies of the libraries. This would allow IINA to abandon the problematic brew based library build. With this change Nix no longer supporting Intel Macs is a non-issue. It only means that if you want to build the libraries you have to have an Apple Silicon based Mac.

Due to the current state of IINA, the IINA 1.5.0 release is going to be focused on fixing problems and is unlikely to include major changes to support functionality. However as upgrading to mpv 0.40.0 and FFmpeg 8 is a key goal, the libraries must be rebuilt. I think the senior developers would consider merging a simple PR as described above for IINA 1.5.0. See: [foot in the door](https://en.wikipedia.org/wiki/Foot-in-the-door_technique).

The next PR would be to add support for `vapoursynth`. I do not see why that would require Nix to build IINA. Why can't Nix build all of the additional binaries required and populate them under `deps`? The Xcode IINA build can then be changed to include the needed files in IINA.app. The Xcode build already does this today, including files like `yt-dlp`.

On this:
> Shouldn't it already be included into mpv 0.40?

Well it _should_ have been included in mpv 0.40.0, but it wasn't. That fix occurred after mpv 0.40.0 was released. I'm not clear on whether IINA 1.5.0 will support macOS 10.15 or not. They were having trouble building for macOS 10.15. This might be one of the reasons why. I still think we should try and get IINA running under macOS 10.15 with the Nix built libraries, which means applying that patch. But there is a possibility IINA will drop macOS 10.15 for IINA 1.5.0. A disadvantage of doing so is that I can test IINA under 10.15, but I don't have a good setup to test IINA under macOS Big Sur.

At this point the priority for proving Nix can be used should be on getting FFmpeg 8 to work. That means solving the `libiconv` problem.

I tried `nix develop`. That seemed to have worked. With a quick try I had trouble getting the Xcode build to work. We need to keep the libraries needed by IINA separate from the libraries needed by other things. I think the `lib` directory has `.dylib` files needed for other stuff?

low-batt:
After much work I was able to build IINA.app using Xcode with the files provided by `nix develop`. Using links is a problem as the full path ends up in the Xcode project file, so the project can't be shared between users. The `deps` directory must be populated with actual files, not links.

But we should not be focused on that. `flake.nix` needs to be updated to apply the patch to mpv and to use FFmpeg 8. Then we need to sort out any problems with libraries introduced by FFmpeg 8. Likely we need to add a dependency on `libiconv`, so the Nix supplied one is always used.

low-batt:
In addition to switching to FFmpeg 8, we now need to switch to mpv 0.41.0. That will eliminate the need for the patch to make mpv work under macOS Catalina as 0.41.0 has the fix.

low-batt:
@florius0 With the release of IINA 1.4.2 the attention of the developers has turned to IINA 1.5.0. A primary goal is to upgrade mpv, FFmpeg and the other dependencies. This is an ideal time to switch to Nix for building the dependencies. We'd want to create a PR specifically for just that purpose, providing a Nix build that just populates `deps/lib` with the required libraries, just like `other/download_libs.sh` currently does. [NixOS 25.11](https://nixos.org/blog/announcements/2025/nixos-2511/) was released with mpv 0.41.0 and FFmpeg 8.0, so it should be possible to get Nix to build the required libraries.

Are you interested in working on this? Or overloaded with other work?

PR #5758, opened 2025-10-16
(re)name bundled `yt-dlp` as `iina-ytdl` instead of `youtube-dl`

pantheraleo-7:

---

**Description:**
I am working on the official IINA plugin `Online Media` and found out that there's no way in the plugin API to distinguish between executables having the same name on the system `$PATH` and IINA's executable directory - in my case - it is `youtube-dl`.

The `JavascriptAPIUtils.findInPath` searches in, both, system `$PATH` and IINA's executable directory (`IINA.app/Contents/MacOS/`).

More than overcoming this limitation of the API, I think prefixing the executable name with "iina-" to give it a specialised name is a more appropriate solution.

NOTE: (I am forced to make this PR as I need this distinction for the plugin work I am doing) I don't know the Swift language, nor `libmpv`. This PR could be flat out wrong and broken, and I am extremely sorry for that if that's the case. Please bear with me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,6 +111,6 @@ browser/Firefox_Open_In_IINA/*
 *.xcarchive
 /.vscode
 
-deps/executable/youtube-dl
+deps/executable/iina-ytdl
 deps/plugins
 
--- a/iina.xcodeproj/project.pbxproj
+++ b/iina.xcodeproj/project.pbxproj
@@ -231,7 +231,7 @@
 		845FB0C71D39462E00C011E0 /* ControlBarView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 845FB0C61D39462E00C011E0 /* ControlBarView.swift */; };
 		8460FBA91D6497490081841B /* PlaylistViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8460FBA71D6497490081841B /* PlaylistViewController.swift */; };
 		846121BD1F35FCA500ABB39C /* DraggingDetect.swift in Sources */ = {isa = PBXBuildFile; fileRef = 846121BC1F35FCA500ABB39C /* DraggingDetect.swift */; };
-		8461BA681E4237C7008BB852 /* youtube-dl in Copy Executables */ = {isa = PBXBuildFile; fileRef = 8461BA671E4237C2008BB852 /* youtube-dl */; settings = {ATTRIBUTES = (CodeSignOnCopy, ); }; };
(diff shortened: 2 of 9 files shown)


lhc70000:
Thanks for the pull request. However, I don’t think it’s a good idea, mainly for two reasons:

- IINA still tries to maintain compatibility with mpv, and mpv’s yt-dlp support will look for youtube-dl or yt-dlp in the path. Changing the binary name will break this.
- Replacing a file inside the app bundle might cause problems, as it invalidates the signature. macOS might block IINA from launching.

I still think the current implementation is better (see my comments in https://github.com/iina/plugin-online-media/pull/8). I was forced to download the unpacked version because the standalone binary’s launch time is unstable and significantly harms user experience. 

So we need to provide a fallback youtube-dl/yt-dlp (if the user doesn’t have the plugin), and if the plugin is installed, because it cannot modify the bundled one, it can only maintain another binary.

pantheraleo-7:
> mpv’s yt-dlp support will look for youtube-dl or yt-dlp in the path. Changing the binary name will break this.

https://github.com/iina/iina/pull/5758/files#diff-f46415959e0cf478aed9734f1e0c884ee66df898da449d4749580eb96020dd8fR542

?? I'm confused. The exectuable path that mpv looks up is configurable, and this line (from the PR) does exactly that.

> Replacing a file inside the app bundle might cause problems, as it invalidates the signature. macOS might block IINA from launching.

I thought so too. But locally (at least on macOS 26) I have tested with no issues whatsoever. I theorised that maybe since yt-dlp updates itself (and not some 3rd entity), macOS is allowing it? haha I don't know.

> the standalone binary’s launch time is unstable and significantly harms user experience.

Oh so that's why whenever the plugin fallback to the bundled standalone binary, everything became so slow.

This is all the more reason to remove the bundled binary apart from redundancy. @\low-batt said we can't because github is as good as unavailable in some countries, so we can't just download yt-dlp at first launch.

How about this approach:
bundle the _zip_ file instead of the standalone binary with the app, the plugin (or maybe the app itself) then finds it and extracts it at first launch, and if macOS allows it, delete the zip file. As the plugin comes preinstalled with the app, this shouldn't be a problem?

If this approach works and is viable, it will have the following benefit:
- no network operations (i.e. download) required
- performance benefit as we will be using the (un)zipped yt-dlp in all scenarios
- (pending research) eliminate redundancy, as we will now have one and only one copy of yt-dlp/youtube-dl

pantheraleo-7:
the binary is going to be removed in IINA v1.5 (right?), so I think this PR could be closed.

> mpv’s yt-dlp support will look for youtube-dl or yt-dlp

this can be overridden with `ytdl_path` hook. which this PR did:

https://github.com/iina/iina/pull/5758/changes#diff-f46415959e0cf478aed9734f1e0c884ee66df898da449d4749580eb96020dd8fR542

> Replacing a file inside the app bundle might cause problems, as it invalidates the signature

this change was at build time.

Issue #2658, opened 2019-11-28
Is there any possible that IINA support LRC


kuangshenyan:
now iina can play music, but it doesn't support lrc, so, is there any possible to support local lrc or lrc from a url ?

DroidFreak32:
I second this as well.
Coming from PotPlayer on windows, I really miss its .lrc file support.

xjbeta:
MPV / IINA can load lrc file as subtitle.
Menu - Subtitles - Load External Subtitle...
Select .lrc file.

![891FDD63-56A6-4756-819D-2B231BD3FE89](https://user-images.githubusercontent.com/11794321/70623079-3d567f00-1c58-11ea-8da2-c9eee85647a9.png)

kuangshenyan:
I want to have a mechanism where I can switch to music player mode, after which IINA can automatically load IRC files from a specified directory, and then match the current song and display.

kuangshenyan:
> MPV / IINA can load lrc file as subtitle.
> Menu - Subtitles - Load External Subtitle...
> Select .lrc file.
> 
> ![891FDD63-56A6-4756-819D-2B231BD3FE89](https://user-images.githubusercontent.com/11794321/70623079-3d567f00-1c58-11ea-8da2-c9eee85647a9.png)

lhc70000:
I think it's possible to implement this, at least as a plugin.

Please wait a few more days before I can resume the development.

kuangshenyan:
thank you very much.

francoestrubia:
I just commented this on other threat. 
IINA actually renders lrc files (don't remember filepaths).
Is just matter of make it look for lrc files, as do for regular subs. 
(and add it to music mode render pipeline).
LRC files has allways the same audiofile name.

DroidFreak32:
I agree with @francoestrubia , IINA plays the .lrc file just fine.
The only missing thing is IINA does not proactively search for a .lrc file with the same name and play it automatically.

francoestrubia:
Exactly. Proactively search can be such a task to write, but here is already coded.
Is all working, except you need to manually load the file.

TianyiShi2001:
The above PR adds `.lrc` to the array of supported subtitle file extensions. After merging it, an additional change will be required: currently subtitle autoload only searches subtitles for videos, not audios. Specifically, changes will be required near

[...]

Issue #4599, opened 2023-08-27
Show src lyrics when playing music


jsonvot:
I hope to display the format in the embedded audio file or separate src format lyrics with the same name as the song

**What IINA does currently:**

**Why you think this should be added:**

**Examples of other projects that have something similar:**

Issue #4317, opened 2023-04-02
synced lyrics lrc file could not load.


sviion:
when I play mp3, I could not load "lrc" file, even manual select

Further more, how can it be auto loaded?

Issue #2383, opened 2019-03-29
[Feature Request] Play audio with subtitles


johan456789:
**System and IINA version:**

- macOS 10.14.4 (18E226)
- IINA v1.0.2 (105)

**Expected behavior:**
Play with subtitles after .srt file dragged onto the window.

**Actual behavior:**
Nothing happens.

- [x] MPV does not have this problem.

Can play audio with subtitles using this command:

`mpv --no-audio-display --force-window --sub-file="<subfile>" "<inputfile>"`

![image](https://user-images.githubusercontent.com/14802181/55255252-1eb08380-5295-11e9-9414-4ea4364677e4.png)

hsyi:
I have the same request as well.

uiryuu:
Duplicate of #1138

johan456789:
@uiryuu I don't think it's a duplicate issue. #1138 requests support for lyrics of music files. This issue is a request for supporting adding subtitles for any audio files. The use case I have is showing transcript for podcast playback.

uiryuu:
From an implementation perspective, they are the same. I can leave this open though.

johan456789:
I see. I just have concerns the focus of the other issue will be retrieving lyrics from the Internet. Thank you for leaving this open!

PR #5790, opened 2025-10-30
Auto load timecoded lyrics from LRC or metadata

ViRb3:
- [x] This implements/fixes issue #2658, #2383, #1138, #4317, #4599.

---

**Description:**

MPV and IINA already support LRC timecoded lyrics, but you need to manually load them from `Subtitles > Load External Subtitle...` for each audio file. This PR makes a few improvements on that front:
- Auto load `.lrc` files if they have the same name as the input file
- Parse out `lyrics` metadata from audio files and if it matches LRC format (it is timecoded), load it as such

There is one minor issue with the current implementation - if the subtitles are extracted from metadata, they are saved to a temporary `.lrc` file so it can be loaded with mpv command. The file is immediately deleted after. I couldn't find a way to load from memory, but if there is, please let me know.

The lyrics code currently runs on any input file, including video, but it causes no issues if it doesn't find LRC file or relevant metadata. If this is unwanted for videos, please let me know how to refactor it.

I tested MP3 and M4A (AAC) with embedded lyrics as well as the same files with external LRC lyrics. All scenarios worked for me. If anybody wants to test without compiling, I am uploading a build here:

https://files.catbox.moe/3c2iin.zip


ViRb3:
Hi @low-batt, would it be possible to merge this?

PR #5795, opened 2025-11-02
Fix wrong History table header color in MacOS Tahoe's dark mode

svobs:
- [xj] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)

---

**Description:**
Looks like MacOS Tahoe's `NSOutlineView` adds improved support for custom background color, but this revealed that the background color was set incorrectly. It's not clear why only the table header was drawn with the wrong color (and not the remaining rows).

This patch changes the table's background to the default color.

Before:
Before

After:
After


low-batt (approved):
I reproduced the problem under macOS Tahoe and confirmed the fix. I also confirmed that light mode still works and confirmed the change does not break anything under macOS Sequoia.

low-batt:
Glad you pointed this one out. I'm always in light mode and had not noticed this problem.

PR #5820, opened 2025-11-24
Change getTrackInfo to use getNode

low-batt:
This commit will:
- Change the `getTrackInfo` method in the `PlayerCore` class to get all tracks at once using `getNode`
- Add a `nodeValueAsInt` method to `MPVController` for extracting an integer from a node map

Instruments identified `getTrackInfo` as sometimes being responsible for short hangs. This is because `mpv_get_property` sometimes has to wait to acquire a mpv lock. The current `getTrackInfo` implementation makes 20 calls to `mpv_get_property` for each track. By changing `getTrackInfo` to make one call to `getNode`, obtaining the entire list in a single call to `mpv_get_property`, the time waiting for `mpv_get_property` is reduced.

---

**Description:**

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -2818,36 +2818,16 @@ class PlayerCore: NSObject {
       let raw = mpv.getNode(MPVProperty.trackList)
       guard let list = raw as? [[String: Any]] else {
         // Internal error, should not occur.
-        log("Cast of mpv node failed: \(String(describing: raw))", level: .error)
+        log("Cast of mpv node failed while getting track list: \(String(describing: raw))",
+            level: .error)
         return
       }
       for dict in list {
-        guard let type = dict["type"] as? String else { continue }
-        guard let isDefault = dict["default"] as? Bool, let isForced = dict["forced"] as? Bool,
-              let isImage = dict["image"] as? Bool, let isSelected = dict["selected"] as? Bool,
-              let isExternal = dict["external"] as? Bool else {
+        guard let track = MPVTrack(dict) else {
           // Internal error, should not occur.
           log("Unable to construct MPVTrack from mpv node map: \(dict)", level: .error)
           continue
         }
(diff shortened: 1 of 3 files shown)


iina/PlayerCore.swift
```diff
@@ -2816,30 +2681,38 @@ class PlayerCore: NSObject {
...
+        guard let type = dict["type"] as? String else { continue }
+        guard let isDefault = dict["default"] as? Bool, let isForced = dict["forced"] as? Bool,
+              let isImage = dict["image"] as? Bool, let isSelected = dict["selected"] as? Bool,
+              let isExternal = dict["external"] as? Bool else {
+          // Internal error, should not occur.
+          log("Unable to construct MPVTrack from mpv node map: \(dict)", level: .error)
+          continue
+        }
+        let id = MPVController.nodeValueAsInt(dict["id"])
+        let track = MPVTrack(id: id, type: MPVTrack.TrackType(rawValue: type)!,
```
uiryuu: It would be cleaner to write a new constructor for `MPVTrack` that accept a dict[Stirng: Any]. That way we can move all `MPVTrack` initialization logic to its class.

uiryuu (changes requested):
Rebased and tested, no problem found. See comments below. Other changes look good to me.

iina/PlayerCore.swift
low-batt: Agreed. Will work on that.

From what I recall I decided not to do that because I wanted to minimize the code changes as I was hoping this change would go into 1.4.2. But as this is critical code and the change was not required to fix a defect, lhc wanted to postpone this one until 1.5.0 where it can get more testing.

low-batt:
Updated.

In general, In this area of IINA (including `MPVController`) the code is loose, rather than tight/strict. Calls into mpv do not check the return code for an error. Properties that are not of the expected type quietly result in `nil` values. I did not try and tighten up the code and followed the existing conventions.

Issue #4487, opened 2023-07-10
Zoom, Zoom Zoom Baby! It is now time for this to hit No1 needed Feature!


TBuTcher71:
I have sugested it before a number of years ago, cannot find the search feature option so please add this to the existing request if needed. THANK YOU 

ZOOM! For love of everything holy please include zoom ability!

you can make full screen but you cannot zoom in 2x - 10x etc

Allows you to zoom into a smaller section of a 4k video or even 1080P and view it fullscreen, great for security playback and other videos where you need to pixel peek .

Open VLC
Window/Video Effects/Geometry/ (Magnification/zoom) enable

then in a video you have a slider to zoom slider scale you adjust it and a reference screen

low-batt:
On this:
> I have sugested it before a number of years ago, cannot find the search feature option so please add this to the existing request if needed.

If we are talking about searching the playlist, issue #4404 is tracking that.

IINA already has a search feature, but I suspect most users do not know about it. Have a look at [this comment](https://github.com/iina/iina/issues/4392#issuecomment-1535514245) for how to search using the Help menu.

low-batt:
This feature was requested yet again in discussion #5563. Related issues are #4477 and #3208. Time to at least document what works now and what needs to change.
 
Here is a quick writeup on what IINA currently can do and what is missing. Please comment on what I got wrong or missed.
 
**What you want IINA to do:**

Support zooming, panning and rotating.

IINA should support the standard gestures specified in Apple's Human Interface Guidelines [here](https://developer.apple.com/design/human-interface-guidelines/gestures#Standard-gestures) . These gestures can be seen in the Apple support article [Use Multi-Touch gestures on your Mac](https://support.apple.com/en-us/102482).

These features should also be accessible using the mouse and the keyboard.

These features should work both in windowed and full screen mode.

These features should provide the OSD messages showing the current setting after a change.

**What IINA does currently:**

As can be seen in this screenshot, IINA can zoom, pan and rotate:

![Image](https://github.com/user-attachments/assets/205e3054-0762-4c18-ab3e-1a070156b0ff)

_However_ the keyboard must be used to access these features and there are _significant restrictions_ on their use. And no OSD messages are shown when a setting is changed.

For audio/video and subtitle rendering support IINA uses a library from the [mpv](https://github.com/mpv-player/mpv) project. IINA must use the zoom, pan and rotate features provided by `libmpv`.

### Rotate

[...]

PR #5798, opened 2025-11-06
RFC. Command (allow to assign a hotkey) for toggling auto-hiding of On Screen Controller

beyondcompute:
- [x] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)
- [x] This implements/fixes issue #5566 (somewhat)

---

**Description:**
There was an [idea to add a hotkey](https://github.com/iina/iina/issues/5566) to briefly show the UI (without the need to move the mouse, for example). But perhaps it would be also nice to have a hotkey to enable-disable the UI auto-hiding itself. The screenshot below shows the setting that we toggle via a hotkey basically.

Screenshot of Iina app settings view, with an arrow pointing at the checkbox next to the “Auto hide after” setting

Here’s a screenshot of a debug build of the implementation:
Screenshot of Iina keyboard shortcuts settings, show adding a new shortcut with a “Toggle On Screen Controller auto-hiding” option highlighted

<details><summary>Screenshots of the player window with OSD messages</summary>
Screenshot of Iina player window with an on-screen alert message saying “On Screen Controller always visible”

Screenshot of Iina player window with an on-screen alert message saying “Auto-hiding On Screen Controller”

</details>

This PR is just a suggestion (hence titled as an “RFC”), feel free to close or to request style changes. Thank you! 🙏 

> [!warning]
> This is an AI-assisted work 🤷‍♂️

Issue #5566, opened 2025-06-26
Support binding a key to a command that shows the OSC


low-batt:
**What you want IINA to do:**
Support binding a key to a command that causes the on screen controller to be displayed if it is hidden.

**What IINA does currently:**
Currently when the on screen controller is hidden you must move the mouse for the OSC to be displayed or perform another action that causes the OSC to be displayed.

**Why you think this should be added:**
It would be nice to be able to just “peek” at the UI while using the keyboard without having to move your hand to the mouse.

I am mainly interested in the playback progress, seeing the progress bar and the time left (in my case). It would be great if I could assign a hotkey to some “display UI” action.

This feature was proposed by @beyondcompute in discussion #5564.

low-batt:
@beyondcompute The straight forward way to implement this would be to add an IINA command that triggers showing of the OSC in the same way moving the mouse does. The OSC and title bar would then be hidden again after the time configured in IINA's `UI` settings just as it does after moving the mouse.

But since you mentioned "peek" I'm wondering if you are thinking this feature should not be tied to the OSC `Auto hide after` setting and the OSC should only be shown while the key is held down? Or maybe the command would have an option to select between these two behaviors.

@svobs knows the key binding code better than me. Lets see what he thinks about this proposed feature.

beyondcompute:
I think, a command having the same effect as simply moving the mouse, would certainly suffice! 💯

svobs:
Given @low-batt's suggestion:

> @beyondcompute The straight forward way to implement this would be to add an IINA command that triggers showing of the OSC in the same way moving the mouse does. The OSC and title bar would then be hidden again after the time configured in IINA's UI settings just as it does after moving the mouse.

And @beyondcompute's response:

> I think, a command having the same effect as simply moving the mouse, would certainly suffice! 💯

Not complicated to add.

The OSC and title bar are grouped together in the code as a set (well, an Array, but it should be a Set) called `fadeableViews`.  For this discussion I'll shorten that and call these "fadeables", because than some of the other terms used aren't as precise.

I volunteered to decouple the visible states of the title bar & OSC in #4553, but then and now there doesn't seem to be a large public desire for doing that. And it is easier not to add that complexity.

Just spewing out some things which might be useful to implement this:
1. In the code, a call to the `showUI()` method will show fadeables (unless prohibited via certain conditions)
2. Similarly, there exists a `hideUI()` method which will hide all the fadeables.
[...]

PR #5822, opened 2025-11-25
RFC. Zoom in full screen!

beyondcompute:
- [x] This implements/fixes issue #4487

---

**Description:**

A suggestion and a conversation-starter. An AI-assisted work again 🤖, so the code might need some love.

* Zoom works when the setting `Control → Pinch to` is set to _“Adjust window size“_.  (Maybe a new setting could be introduced to keep legacy behavior if many users wish so)
* If we are not zoomed-in (so the size is 100%), the zoom starts around the cursor position (clamping to the nearest point in the video when the cursor is outside of the video)
* If we zoom several times (“consecutive zooms”), zooming still happens around the point where we started when we were at 100%. This avoids “jumps” that naïve panning would cause in such a case. And it feels natural anyway
* But perhaps a more easing-in mode of panning for consecutive zooms and zooms after panning (see below) could be introduced 
* Only when zoom is above 100%, we can pan (with two fingers on a touchpad, for example). After we have panned, zoom happens around the center of the screen. (This could be a little confusing if your cursor is far from the center. But if we zoom around the cursor again, that would cause a “jump“ of fast panning as well; see above.)
* Panning seems to play along fine with the `Control → Scroll vertically to` and `Scroll horizontally to` actions. When we are zoomed above 100%, we get panning, at 100% we get our configured actions
* Max zoom level is hardcoded in the code
* The behavior of the zoom in windowed mode should not be affected
* If `General → Resume last playback position` setting is enabled, zoom and pan state is also remembered. I think that‘s fine

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -174,6 +174,15 @@ class MainWindowController: PlayerWindowController {
 
   var lastMagnification: CGFloat = 0.0
   var frameWhenStartedPinching = NSRect()
+  var pinchOriginInWindow: NSPoint?
+  var pinchOriginInVideo: NSPoint?
+  var pinchOriginInVideoUnit: NSPoint?
+  var pinchScale: CGFloat = 1.0
+  var pinchInitialZoom: Double = 1.0
+  private let pinchZoomMultiplier: Double = 1.0
+  private let pinchMinZoom: Double = 1.0
+  private let pinchMaxZoom: Double = 4.5
+  private let pinchMaxPan: Double = 1.0
 
   /** Views that will show/hide when cursor moving in/out the window. */
   var fadeableViews: [NSView] = []
@@ -584,7 +593,7 @@ class MainWindowController: PlayerWindowController {
     addVideoViewToWindow()
 
     // gesture recognizer
-    cv.addGestureRecognizer(magnificationGestureRecognizer)
+    videoView.addGestureRecognizer(magnificationGestureRecognizer)
 
     // Work around a bug in macOS Ventura where HDR content becomes dimmed when playing in full
     // screen mode once overlaying views are fully hidden (issue #3844). After applying this
@@ -1173,6 +1182,7 @@ class MainWindowController: PlayerWindowController {
   @objc func handleMagnifyGesture(recognizer: NSMagnificationGestureRecognizer) {
     guard pinchAction != .none else { return }
     guard !isInInteractiveMode, let window = window, let screenFrame = NSScreen.main?.visibleFrame else { return }
+    let isFullscreen = fsState.isFullscreen
 
     switch pinchAction {
     case .none:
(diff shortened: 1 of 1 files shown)


low-batt:
A while back I too started on changes to add pan, rotate and zoom. I did not take that code past the just playing around stage because the priority is to first remove the restriction that prevents these features from being used except in full screen mode. I've heard changes were made in that area for IINA 1.4.1, but haven't had time to look into that.

Originally the plan was for IINA 1.5.0 to remove that restriction and add other features. But Apple breaking all kinds of things in macOS Tahoe disrupted IINA's plans. The IINA 1.4.0 release was supposed to be 1.4.0-beta2. Because IINA was so broken in macOS Tahoe, quick fixes were added for _some_ of the macOS 26 problems and IINA 1.4.0 was rushed out the door. Some important fixes that were planned for 1.4.0 were left sitting on the dock as IINA 1.4.0 sailed away. IINA 1.4.1 was a quick release to fix some more of the problems Apple introduced in macOS 26. More remain.

Worse a couple of severe problems have been found in the upgraded libraries IINA is built on top of. Due to the way IINA builds libraries it can be difficult to patch and rebuild libraries when they are no long the current released version. Because of this the plan was to make IINA 1.5.0 focused on upgrading mpv, FFmpeg and the other libraries, fixing remaining macOS Tahoe problems and other regressions.

That plan has run into some problems building libraries for the oldest version of macOS IINA supports (Catalina). Because of that we are currently working on an IINA 1.4.2 release with the latest version of [libdav1d](https://code.videolan.org/videolan/dav1d), more fixes for macOS Tahoe problems and fixes for other regressions. Once that release is out we will go back to working on IINA 1.5.0.

Because of all this turmoil I'm unsure what the other developers are thinking with regards to features going into IINA 1.5.0. It is looking like IINA has to upgrade to mpv 0.40.0 and FFmpeg 8 to fix some of the problems. That adds significant pressure to quickly complete the development of IINA 1.5.0 and greatly limit the inclusion of any features.

Another consideration has to do IINA starting to address technical debt and other long postponed work. The way the IINA UI is coded (XIBs) is now two generations behind. Work has been underway to introduce use of [SwiftUI](https://developer.apple.com/documentation/swiftui) starting with the settings panel. That work can be seen on the [new-pref-ui](https://github.com/iina/iina/tree/new-pref-ui) branch. Major rework like this comes with the usual problem of ongoing changes to the existing code base causing additional work as those changes then have to be reworked for the new UI code. This makes finishing and merging the new UI code a priority. Given the extra work not merging the new UI code generates I'm expecting this might be the other change included in IINA 1.5.0.

Remember I'm a _junior_ IINA developer and not always up to date on our plans. But it is looking to me like near term IINA will be very focused on a couple of releases to fix defects before the focus turns back to features. If you don't see activity on this PR in the near term, it is likely due to the priorities I described above.

beyondcompute:
Hello, @low-batt! Thank you taking the time to provide me with a context and a glimpse into the IINA’s team‘s plans! 🙏 I meant those PRs exactly as a more material alternative to the [Discussions](https://github.com/iina/iina/discussions) section and also as a learning opportunity for myself.

_I was also strongly expecting that I will have to run my own “fork” of IINA locally (and to delay the upgrade to ’26 as long as possible 😄)._

svobs:
Hi @beyondcompute, thank you so much for this! The code is nearly perfect. I merged it into my fork and expanded it a bit so that:
- It can be enabled/disabled via `Settings…` > `Control`, and the maximum zoom can be changed (up to 10x).
- An OSD message is displayed when `video-zoom` changes.
- It also starts zooming when the window is maximized on screen when the pinch is started. Subsequent pinches will continue zooming / panning until a pinch ends with 1x zoom, and then the next pinch will reduce the window again. (If the user reduces the window size via some other method while zoomed, the zoom auto-resets to 1x).

You can take a look and download a notarized binary release [here](https://github.com/svobs/iina-advance/releases/tag/v1.4). In my source tree the code lives in [MagnificationGestureHandler.swift](https://github.com/svobs/iina-advance/blob/v1.4/iina/MagGestureHandler.swift).

beyondcompute:
Thank you, @svobs. TIL about [IINA Advance](https://github.com/svobs/iina-advance)! ℹ️ And wow, that looks so cool! I like the chapter markers, the control bar, etc. Also the ability to restore windows after restarting the machine and so on, is something I noticed but haven’t realized how cool it would be to have it! (Does something like https://github.com/iina/iina/pull/5798 have a chance of ending up there? 😄)

A small piece of feedback. Top margin looks a bit tight on this view:
Screenshot 2025-12-24 at 05 45 06

Also I can’t import my current keyboard config (even after I removed my custom command from it). ⌨️ (OK, that got fixed here: https://github.com/svobs/iina-advance/pull/16)

<hr>

Merry Christmas to everyone who celebrates! 🎄🎅

svobs:
> A small piece of feedback. Top margin looks a bit tight on this view:

> Also I can’t import my current keyboard config (even after I removed my custom command from it). ⌨️ (OK, that got fixed here: https://github.com/svobs/iina-advance/pull/16)

Thanks for the feedback/help! Looks like the top margin was a regression for pre-Tahoe Macs. The broken import looks...much older, oops. I'll soon roll out a 1.4.2 update which will include these fixes.

> Does something like https://github.com/iina/iina/pull/5798 have a chance of ending up there? 😄

I've had plans to develop a generalized mechanism for making all the IINA commands bindable via input bindings configuration (you may have seen #5566), and I've had some fragments of ideas of doing the same for IINA preference keys generally (or at least a large subset of them). I'll see what I can come up with.

beyondcompute:
> I've had plans to develop a generalized mechanism for making all the IINA commands bindable via input bindings configuration (you may have seen #5566), and I've had some fragments of ideas of doing the same for IINA preference keys generally (or at least a large subset of them). I'll see what I can come up with.

That’s cool! I’ll be looking forward to those. 😌

Issue #5803, opened 2025-11-09
Large files loaded in RAM, 86 GB in use by IINA


httpstorm:
### IINA Version

1.4.1 Build 160

### Release Channel

Release

### macOS Version

15.7.2 (24G325)

### Steps to Reproduce

- Open a very large movie. Dune is a good choice 86 GB
Dune.2021.2160p.BluRay.REMUX.HEVC.DTS-HD.MA.TrueHD.7.1.Atmos-FGT.mkv
d7b21ca3a5556b8012a65b37b9484f2e0c0bdda5

- IINA tries to load the entire movie in RAM, and total havoc breaks loose.
- My laptop has only 64 GB of RAM, so it starts paging excessively. MBP 2019 16" Intel.
- Constant disk reads and writes exceeding 1 GB/s during the entire movie
- It takes just a few seconds for the swap use to reach 46 GB.
- Playback freezes from time to time.
- Loud fan noise due to excessive stress on the system.
- Disabling Network cache makes no difference.

- I tried opening the same movie in MPV: no issues

### Expected Behavior

There's no sane reason to use that much RAM or load the entire movie in RAM

### Actual Behavior

An insane amount of RAM is used

### Logs

[iina.log](https://github.com/user-attachments/files/23440693/iina.log)
[mpv.log](https://github.com/user-attachments/files/23440694/mpv.log)

### Sample Files

<img width="1026" height="748" alt="Image" src="https://github.com/user-attachments/assets/d66476d9-d5c5-48ed-ae93-8faffd8e623f" />
<img width="1026" height="748" alt="Image" src="https://github.com/user-attachments/assets/06d49959-7283-4b3b-ac38-1cf13f16cd9a" />

Dune.2021.2160p.BluRay.REMUX.HEVC.DTS-HD.MA.TrueHD.7.1.Atmos-FGT.mkv
magnet hash:
d7b21ca3a5556b8012a65b37b9484f2e0c0bdda5

### How often can you reproduce the issue

Every time

### Required Actions

- [x] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [x] I tested with the Nightly build and confirmed the issue is not fixed.
- [x] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

uiryuu:
I guess this is because the file is loaded into memory when generating thumbnails. Can you try to uncheck "Enable thumbnail preview" to see whether the problem still exists?

[...]

PR #5835, opened 2025-12-05
Change constructImage to reduce memory use

low-batt:
This commit will change `NowPlayingInfoManager.constructImage` to use `CGImageSourceCreateThumbnailAtIndex` instead of an `NSImage` constructor to reduce memory use.

Apple recommends using `CGImageSourceCreateThumbnailAtIndex` to minimize memory use when a full image is not required.

---

**Description:**

### DIFF
--- a/iina/NowPlayingInfoManager.swift
+++ b/iina/NowPlayingInfoManager.swift
@@ -234,6 +234,17 @@ class NowPlayingInfoManager {
   }
 
   /// Construct an image for the artwork represented by the given video track.
+  ///
+  /// A track may represent artwork embedded in the file being played, or an external artwork file specified using the mpv
+  /// [cover-art-auto](https://mpv.io/manual/stable/#options-cover-art-auto) option or the mpv
+  /// [cover-art-files](https://mpv.io/manual/stable/#options-cover-art-files) option.
+  /// - Important: IINA intentionally does not use
+  ///     [NSImage(contentsOfFile:)](https://developer.apple.com/documentation/appkit/nsimage/init(contentsoffile:))
+  ///     to construct an image from an external artwork file. This constructor starts by reading the entire file into memory. If mpv
+  ///     selects a high resolution file this will require a large amount of memory. Since only a thumbnail is needed IINA uses the
+  ///     Apple recommended method
+  ///     [CGImageSourceCreateThumbnailAtIndex](https://developer.apple.com/documentation/imageio/cgimagesourcecreatethumbnailatindex(_:_:_:)
+  ///     that minimizes memory use.
   /// - Parameters:
   ///   - url: The URL of the media item.
   ///   - track: Video track representing the image.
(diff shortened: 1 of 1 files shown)


low-batt:
This pull request partially addresses [this comment](https://github.com/iina/iina/issues/5803#issuecomment-3569341594) from @httpstorm:
> There's a rule that says: never trust any data you don't control. It could be from the network, user controlled or in this case an external library that may have bugs.
>
> If the size is known, check it, before constructing the NSImage. Else allocate memory and start reading. If a certain safe size is exceeded, fail. If all is good, I think the NSImage can be constructed from the Data just read.

That comment is on issue #5803 where, thanks to a lot of investigative work by @httpstorm, we know that `libmpv` 0.38.0 incorrectly identified a track as an image (fixed in 0.40.0) causing IINA to attempt construct an image from the huge video file using [NSImage(contentsOf:)](https://developer.apple.com/documentation/appkit/nsimage/init(contentsof:)). An Instruments trace provided by @httpstorm showed that `NSImage` constructor results in a call to the [NSData(contentsOfFile:)](https://developer.apple.com/documentation/foundation/nsdata/init(contentsoffile:)) constructor causing an attempt to malloc enough contiguous memory to read in the entire file.

That problem is fixed by PR #5818, however some side issues were raised in the comments on #5803. The above comment is one of them. This PR addresses one place where IINA might be tricked into attempting to construct a `NSImage` from a huge file. The proposed fix uses [CGImageSourceCreateThumbnailAtIndex](https://developer.apple.com/documentation/imageio/cgimagesourcecreatethumbnailatindex(_:_:_:)) instead of [NSImage(contentsOf:)](https://developer.apple.com/documentation/appkit/nsimage/init(contentsof:)). Using this method is recommended by Apple in the WWDC 2018 video [iOS Memory Deep Dive](https://developer.apple.com/videos/play/wwdc2018/416/).

The additional use of [CGImage](https://developer.apple.com/documentation/coregraphics/cgimage) means this part of the Now Playing support should be refactored to maintain the image as a `CGImage` object until it must be converted to a [NSImage](https://developer.apple.com/documentation/appkit/nsimage) object. I will address that in some future PR. This PR solves the problem at hand.

The other place where the [NSImage(contentsOf:)](https://developer.apple.com/documentation/appkit/nsimage/init(contentsof:)) constructor is used is in `PlayerCore.createImage`. That code is used when the IINA setting `Show previews after taking screenshots` is enabled. As that code is reading a screenshot image file that was just created by `libmpv` it is not as vulnerable to attempting to read a massive file into memory. _However_, that code also does not need a full image, only a thumbnail sized appropriately for IINA's screenshot preview. So this code too should be refactored to use [CGImageSourceCreateThumbnailAtIndex](https://developer.apple.com/documentation/imageio/cgimagesourcecreatethumbnailatindex(_:_:_:)).  I will address this as well in some future PR.

httpstorm:
Thanks! Please let me know when a new release is available to test. I'll be unavailable for a few days during the Easter holiday.

low-batt:
IINA development is in the phase where the `develop` branch is frozen, meaning no new changes are being merged, while the development version undergoes testing, looking for problems that must be fixed before a release can be cut.

So far testing has uncovered a couple of problems. They are currently being investigated. Once more is known then a decision will be made as to whether to consider them release blockers, in which case the release will be delayed for a fix, or whether it is acceptable to postpone fixing them until the next release. Depending upon the outcome of testing, a release might appear soon.

This next release is focused on fixing defects. It will not include upgrading mpv, FFmpeg and other dependencies. There are some problems with mpv that need to be solved. After this next release is out, attention will turn to upgrading dependencies.

PR #5845, opened 2025-12-13
Reset Scroll on Switching Tabs in Settings

rnihesh:

---

**Description:**
- I noticed an issue where the tab switching in the settings page doesn't reset the scroll we do in other tabs
- Added a reset to make sure this would be good UX

### DIFF
--- a/iina/PreferenceWindowController.swift
+++ b/iina/PreferenceWindowController.swift
@@ -303,6 +303,11 @@ class PreferenceWindowController: NSWindowController {
     detailViewBottomConstraint?.isActive = !isScrollable
     prefDetailScrollView.verticalScrollElasticity = .none
 
+    // Reset scroll position to top when switching tabs
+    if let documentView = prefDetailScrollView.documentView {
+      documentView.scroll(NSPoint(x: 0, y: documentView.bounds.maxY))
+    }
+
     // find label
     if let title = title, let label = findLabel(titled: title, in: vc.view) {
       maskView.perform(#selector(maskView.highlight(_:)), with: label, afterDelay: 0.25)
--- a/iina/PreferenceWindowController.swift
+++ b/iina/PreferenceWindowController.swift
@@ -305,7 +305,7 @@ class PreferenceWindowController: NSWindowController {
 
     // Reset scroll position to top when switching tabs
     if let documentView = prefDetailScrollView.documentView {
-      documentView.scroll(NSPoint(x: 0, y: documentView.bounds.maxY))
+      documentView.scroll(NSPoint(x: 0, y: 0))
     }
 
     // find label


rnihesh:
Hii, @low-batt & @uiryuu , this fix might be a good one in terms of UX

low-batt:
IINA follows Apple's [Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/), so the first task is to figure out what Apple thinks the proper behavior is in this case. In a quick look I did not find explicit guidelines for what the UI should do in this case. I could have missed something.

In the [Scroll views](https://developer.apple.com/design/human-interface-guidelines/scroll-views) section under [Best practices](https://developer.apple.com/design/human-interface-guidelines/scroll-views#Best-practices) it discusses when to scroll automatically. This part may be relevant:
> In all cases, automatically scroll the content only as much as necessary to help people retain context. For example, if part of a selection is visible, you don’t need to scroll the entire selection into view.

I read that as an app should err on the side of not changing the scroll position.

When it is unclear what the correct behavior should be after reading the [Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/) the next thing to look at is how Apple's own apps handle the situation. **However** I have found that Apple apps sometimes fail to follow their own standards and there are inconstancies between Apple's apps.

In the case at hand it is hard to find examples since Apple apps frequently do not offer enough settings on an individual tab to need scrolling. In changing tabs in Safari's settings it _seemed_ to be remembering the scroll position for a tab. At least when I went back to the `Websites` tab it was still scrolled to be bottom. I'm thinking that is a desirable behavior.

The last check I will do is to search the net to see if I can get a feeling for how users expect the UI to operate. In poking around the the net I found people complaining loudly about apps that reset the scroll position to the top when switching tabs.

I'm thinking we at least need to preserve the scroll position on a tab basis.

Currently there is another consideration with respects to changes in this area of IINA…

With IINA 1.4.0 the project has started to address technical debt. The way IINA's UI is coded (XIB files) is now 2 generations behind. Work is underway to update the UI to SwiftUI, starting with settings. You can see that work in the [new-pref-ui](https://github.com/iina/iina/tree/new-pref-ui) branch. Any changes to the current settings code will have to be adapted and added to the new code.

I am a junior IINA developer and always defer to the senior developers on questions about proper UI behavior. Will have to see what the other developers think about such a change.

rnihesh:
As you mentioned you go back to the `Websites` it would be in the last scrolled position, but it doesn't take the position of the previous `Website Tab`. 

But adding this would be better here.
> I'm thinking we at least need to preserve the scroll position on a tab basis.


I guess the context that we are talking here goes wrong, the problem that I think I'm solving is, the tabs shouldn't take the scroll position of the previous tabs.
> In poking around the the net I found people complaining loudly about apps that reset the scroll position to the top when switching tabs.


And I'm fine with doing the preservation on tab basis.

low-batt:
Need to hear what @uiryuu thinks. It is his opinion that matters.

PR #5863, opened 2025-12-28
Font Picker window improvements and bug fixes

svobs:

---

**Description:**

- Update UI on window open to select current family & face. This requires a search through all fonts & typefaces, but in practice this is nearly instantaneous. And it should be much more useful/intuitive to the user to see the existing selection. Obviously if the current font is not found in the list, it is simply not selected in the tables, and will be shown in the manual font entry at the bottom.
- Update manual font entry text when font selected in table by user. (This should be much more intuitive to the user when selecting fonts. Though currently the updates still do not go in the opposite direction - i.e., typing the font name does not select it in the table in real time).
- Fix layout so that changing the preview font doesn't result in a change to window height
- Minor changes to modernize layout (got rid of a horizontal line...)

### DIFF
--- a/iina/AppData.swift
+++ b/iina/AppData.swift
@@ -82,6 +82,7 @@ struct Constants {
     static let dot = "●"
     static let blackRightPointingTriangle = "▶︎"
     static let blackLeftPointingTriangle = "◀"
+    static let mpvDefaultFont = "sans-serif"
     static let videoTimePlaceholder = "--:--:--"
     static let trackNone = NSLocalizedString("track.none", comment: "<None>")
     static let chapter = "Chapter"
--- a/iina/Base.lproj/FontPickerWindowController.xib
+++ b/iina/Base.lproj/FontPickerWindowController.xib
@@ -191,7 +191,7 @@ Gw
                     </textField>
                     <textField focusRingType="none" verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="Qur-08-eHn">
                         <rect key="frame" x="12" y="52" width="434" height="24"/>
-                        <textFieldCell key="cell" controlSize="large" scrollable="YES" lineBreakMode="clipping" selectable="YES" editable="YES" sendsActionOnEndEditing="YES" state="on" borderStyle="bezel" placeholderString="sans-serif" drawsBackground="YES" id="Kt3-me-43R">
+                        <textFieldCell key="cell" controlSize="large" scrollable="YES" lineBreakMode="clipping" selectable="YES" editable="YES" sendsActionOnEndEditing="YES" state="on" borderStyle="bezel" drawsBackground="YES" id="Kt3-me-43R">
                             <font key="font" metaFont="system"/>
                             <color key="textColor" name="controlTextColor" catalog="System" colorSpace="catalog"/>
                             <color key="backgroundColor" name="textBackgroundColor" catalog="System" colorSpace="catalog"/>
--- a/iina/Base.lproj/FontPickerWindowController.xib
+++ b/iina/Base.lproj/FontPickerWindowController.xib
@@ -6,7 +6,7 @@
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <objects>
-        <customObject id="-2" userLabel="File's Owner" customClass="FontPickerWindowController" customModule="IINA_Advance" customModuleProvider="target">
+        <customObject id="-2" userLabel="File's Owner" customClass="FontPickerWindowController" customModule="IINA" customModuleProvider="target">
             <connections>
                 <outlet property="faceTableView" destination="igd-1P-vQT" id="vk2-bn-1ZM"/>
                 <outlet property="familyTableView" destination="av5-Km-BXV" id="Ttw-xq-6IC"/>
--- a/iina/Base.lproj/FontPickerWindowController.xib
+++ b/iina/Base.lproj/FontPickerWindowController.xib
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="14313.18" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="24506" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="14313.18"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="24506"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <objects>
-        <customObject id="-2" userLabel="File's Owner" customClass="FontPickerWindowController" customModule="IINA" customModuleProvider="target">
+        <customObject id="-2" userLabel="File's Owner" customClass="FontPickerWindowController" customModule="IINA_Advance" customModuleProvider="target">
             <connections>
                 <outlet property="faceTableView" destination="igd-1P-vQT" id="vk2-bn-1ZM"/>
                 <outlet property="familyTableView" destination="av5-Km-BXV" id="Ttw-xq-6IC"/>
@@ -22,27 +22,26 @@
             <windowStyleMask key="styleMask" titled="YES" closable="YES" miniaturizable="YES"/>
             <windowPositionMask key="initialPositionMask" leftStrut="YES" rightStrut="YES" topStrut="YES" bottomStrut="YES"/>
             <rect key="contentRect" x="196" y="240" width="458" height="524"/>
-            <rect key="screenRect" x="0.0" y="0.0" width="2560" height="1417"/>
+            <rect key="screenRect" x="0.0" y="0.0" width="1496" height="937"/>
             <view key="contentView" wantsLayer="YES" id="se5-gp-TjO">
-                <rect key="frame" x="0.0" y="0.0" width="458" height="502"/>
+                <rect key="frame" x="0.0" y="0.0" width="458" height="496"/>
                 <autoresizingMask key="autoresizingMask"/>
                 <subviews>
                     <scrollView focusRingType="none" autohidesScrollers="YES" horizontalLineScroll="19" horizontalPageScroll="10" verticalLineScroll="19" verticalPageScroll="10" usesPredominantAxisScrolling="NO" translatesAutoresizingMaskIntoConstraints="NO" id="Wus-vA-7bo">
-                        <rect key="frame" x="12" y="167" width="218" height="268"/>
+                        <rect key="frame" x="12" y="160" width="218" height="268"/>
                         <clipView key="contentView" focusRingType="none" id="Zfn-kp-cP8">
-                            <rect key="frame" x="1" y="0.0" width="216" height="267"/>
+                            <rect key="frame" x="1" y="1" width="216" height="266"/>
                             <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                             <subviews>
-                                <tableView focusRingType="none" verticalHuggingPriority="750" allowsExpansionToolTips="YES" columnAutoresizingStyle="lastColumnOnly" columnReordering="NO" multipleSelection="NO" autosaveColumns="NO" headerView="hpt-ox-hcu" id="av5-Km-BXV">
-                                    <rect key="frame" x="0.0" y="0.0" width="216" height="244"/>
-                                    <autoresizingMask key="autoresizingMask"/>
+                                <tableView focusRingType="none" verticalHuggingPriority="750" allowsExpansionToolTips="YES" columnAutoresizingStyle="none" columnReordering="NO" multipleSelection="NO" autosaveColumns="NO" headerView="hpt-ox-hcu" id="av5-Km-BXV">
+                                    <rect key="frame" x="0.0" y="0.0" width="220" height="243"/>
+                                    <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                                     <size key="intercellSpacing" width="3" height="2"/>
                                     <color key="backgroundColor" name="controlBackgroundColor" catalog="System" colorSpace="catalog"/>
                                     <color key="gridColor" name="gridColor" catalog="System" colorSpace="catalog"/>
                                     <tableColumns>
-                                        <tableColumn editable="NO" width="213" minWidth="40" maxWidth="1000" id="2Ld-md-U2v">
+                                        <tableColumn editable="NO" width="200" minWidth="200" maxWidth="200" id="2Ld-md-U2v">
                                             <tableHeaderCell key="headerCell" lineBreakMode="truncatingTail" borderStyle="border" title="Family">
-                                                <font key="font" metaFont="smallSystem"/>
                                                 <color key="textColor" name="headerTextColor" catalog="System" colorSpace="catalog"/>
                                                 <color key="backgroundColor" name="headerColor" catalog="System" colorSpace="catalog"/>
                                             </tableHeaderCell>
@@ -51,7 +50,7 @@
                                                 <color key="textColor" name="controlTextColor" catalog="System" colorSpace="catalog"/>
                                                 <color key="backgroundColor" name="controlBackgroundColor" catalog="System" colorSpace="catalog"/>
                                             </textFieldCell>
-                                            <tableColumnResizingMask key="resizingMask" resizeWithTable="YES" userResizable="YES"/>
+                                            <tableColumnResizingMask key="resizingMask" resizeWithTable="YES"/>
                                         </tableColumn>
                                     </tableColumns>
                                 </tableView>
@@ -60,35 +59,34 @@
                         <constraints>
                             <constraint firstAttribute="height" constant="268" id="03m-Ug-zYu"/>
                         </constraints>
-                        <scroller key="horizontalScroller" hidden="YES" wantsLayer="YES" verticalHuggingPriority="750" horizontal="YES" id="x8g-DZ-V9f">
-                            <rect key="frame" x="1" y="251" width="208" height="16"/>
+                        <scroller key="horizontalScroller" wantsLayer="YES" verticalHuggingPriority="750" horizontal="YES" id="x8g-DZ-V9f">
+                            <rect key="frame" x="1" y="250" width="216" height="17"/>
                             <autoresizingMask key="autoresizingMask"/>
                         </scroller>
                         <scroller key="verticalScroller" hidden="YES" wantsLayer="YES" verticalHuggingPriority="750" horizontal="NO" id="biG-DG-2Qv">
                             <rect key="frame" x="-15" y="0.0" width="16" height="0.0"/>
                             <autoresizingMask key="autoresizingMask"/>
                         </scroller>
-                        <tableHeaderView key="headerView" id="hpt-ox-hcu">
-                            <rect key="frame" x="0.0" y="0.0" width="216" height="23"/>
+                        <tableHeaderView key="headerView" wantsLayer="YES" id="hpt-ox-hcu">
+                            <rect key="frame" x="0.0" y="0.0" width="220" height="23"/>
                             <autoresizingMask key="autoresizingMask"/>
                         </tableHeaderView>
                     </scrollView>
                     <scrollView focusRingType="none" autohidesScrollers="YES" horizontalLineScroll="19" horizontalPageScroll="10" verticalLineScroll="19" verticalPageScroll="10" usesPredominantAxisScrolling="NO" translatesAutoresizingMaskIntoConstraints="NO" id="Smi-Ny-cCV">
-                        <rect key="frame" x="229" y="167" width="217" height="268"/>
+                        <rect key="frame" x="229" y="160" width="217" height="268"/>
                         <clipView key="contentView" focusRingType="none" id="2NA-x9-5pt">
-                            <rect key="frame" x="1" y="0.0" width="215" height="267"/>
+                            <rect key="frame" x="1" y="1" width="215" height="266"/>
                             <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                             <subviews>
-                                <tableView focusRingType="none" verticalHuggingPriority="750" allowsExpansionToolTips="YES" columnAutoresizingStyle="lastColumnOnly" columnReordering="NO" multipleSelection="NO" emptySelection="NO" autosaveColumns="NO" headerView="qva-E8-4Pz" id="igd-1P-vQT">
-                                    <rect key="frame" x="0.0" y="0.0" width="215" height="244"/>
-                                    <autoresizingMask key="autoresizingMask"/>
+                                <tableView focusRingType="none" verticalHuggingPriority="750" allowsExpansionToolTips="YES" columnAutoresizingStyle="none" columnReordering="NO" multipleSelection="NO" emptySelection="NO" autosaveColumns="NO" headerView="qva-E8-4Pz" id="igd-1P-vQT">
+                                    <rect key="frame" x="0.0" y="0.0" width="220" height="243"/>
+                                    <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                                     <size key="intercellSpacing" width="3" height="2"/>
                                     <color key="backgroundColor" name="controlBackgroundColor" catalog="System" colorSpace="catalog"/>
                                     <color key="gridColor" name="gridColor" catalog="System" colorSpace="catalog"/>
                                     <tableColumns>
-                                        <tableColumn editable="NO" width="212" minWidth="40" maxWidth="1000" id="fez-Z3-dnL">
+                                        <tableColumn editable="NO" width="200" minWidth="200" maxWidth="200" id="fez-Z3-dnL">
                                             <tableHeaderCell key="headerCell" lineBreakMode="truncatingTail" borderStyle="border" title="Typeface">
-                                                <font key="font" metaFont="smallSystem"/>
                                                 <color key="textColor" name="headerTextColor" catalog="System" colorSpace="catalog"/>
                                                 <color key="backgroundColor" name="headerColor" catalog="System" colorSpace="catalog"/>
                                             </tableHeaderCell>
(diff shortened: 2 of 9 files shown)


svobs:
Did a minor update to remove use of mpv `DispatchQueue` for the get request

iina/Base.lproj/FontPickerWindowController.xib
```diff
@@ -1,12 +1,12 @@
...
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="24506" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="14313.18"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="24506"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <objects>
-        <customObject id="-2" userLabel="File's Owner" customClass="FontPickerWindowController" customModule="IINA" customModuleProvider="target">
+        <customObject id="-2" userLabel="File's Owner" customClass="FontPickerWindowController" customModule="IINA_Advance" customModuleProvider="target">
```
low-batt:
```suggestion
        <customObject id="-2" userLabel="File's Owner" customClass="FontPickerWindowController" customModule="IINA" customModuleProvider="target">
```

iina/FontPickerWindowController.swift
```diff
@@ -18,24 +18,37 @@ class FontPickerWindowController: NSWindowController, NSTableViewDelegate, NSTab
...
+  private var enableSelectionChangeListener = true
+
   override var windowNibName: NSNib.Name {
     get {
       return NSNib.Name("FontPickerWindowController")
     }
   }
 
+  init() {
+    self.chosenFace = ""
```
low-batt: Why initialize `chosenFace` here and not in the variable declaration above?

iina/FontPickerWindowController.swift
```diff
@@ -50,6 +63,47 @@ class FontPickerWindowController: NSWindowController, NSTableViewDelegate, NSTab
...
+  private func updateTablesFromOtherFieldValue() {
+    let selectedFace = otherField.stringValue
+    chosenFace = selectedFace
+
+    // Search for font. Unfortunately this requires checking all typefaces in the system.
+    // But it's still quite fast.
+    // If there is a filter string already set don't try to change it, even if this means not finding a match.
+    let fontNamesDisplayed = isSearching ? filteredFontNames : fontNames
+    for (familyIndex, family) in fontNamesDisplayed.enumerated() {
+      if let typefaces = FixedFontManager.typefaces(forFontFamily: family.name) as? [[Any]] {
```
low-batt: Could use guard/else/continue for the two if statements and avoid deep indentation. Not sure what the senior developers would prefer in this case.

iina/FontPickerWindowController.swift
```diff
@@ -50,6 +63,47 @@ class FontPickerWindowController: NSWindowController, NSTableViewDelegate, NSTab
...
+        for (typefaceIndex, typeface) in typefaces.enumerated() {
+          if let faceName = typeface[0] as? String, faceName == selectedFace {
+            enableSelectionChangeListener = false
+
+            chosenFamily = family
+            chosenFontMembers = typefaces
```
low-batt: Should these two variables be cleared if no match is found?

iina/FontPickerWindowController.swift
```diff
@@ -76,24 +130,29 @@ class FontPickerWindowController: NSWindowController, NSTableViewDelegate, NSTab
...
+    if !chosenFace.isEmpty {
+      otherField.stringValue = chosenFace
     }
+    updatePreview()
   }
 
   // - MARK: NSTextField delegate
 
-  func controlTextDidChange(_ obj: Notification) {
+  // Type-to-filter updates
```
low-batt:
```suggestion
  /// Type-to-filter updates
```

low-batt (changes requested):
The `customModule` definitely needs to be fixed. I will be adding a comment that shows two problems I noticed during testing.

low-batt:
The font picker definitely needs this improvement.

The following feels like a regression to me. When I first bring up the font picker in 1.4.1, not having changed the default the `Or enter the font name` field shows a dimmed `sans-serif`:



But in the PR the text box has `sans-serif` filled in:

pr

Odd, but probably ok. 

In 1.4.1 if I set the font to something else and then open the font picker and click on `OK` without having selected a font or entering a font name then the font is set back to the `sans-serif` default.

The PR code changes the situation as it automatically selects the font that was set and fills in `Or enter the font name`. If I then erase the font name in `Or enter the font name` then it shows a dim `sans-serif`. If I click `OK` then even though `sans-serif` (dimmed) shows in  `Or enter the font name` the font is not changed. I'm expecting that is because the font is still selected in the table. To get back to the `sans-serif` default I actually have to type `sans-serif` in `Or enter the font name`.

I'm thinking that maybe when you change the selected font name in `Or enter the font name` the selection in the table should be adjusted accordingly including not selecting anything if you clear the text box. Maybe something like that? Need some easy way to get back to the `sans-serif` default other than making the user type it in.

The following is an existing inconsistency. No need to address this in this PR. This screenshot shows the `SUBTITLES` side panel has disabled the `Font…` button. I assume that is because the select subtitle is in ASS format. But the `Font…` menu item under the `Subtitles` menu allowed me to bring up the font picker. One of the problems with disabling controls is that it confuses users. I think we need help tags that provide a hint as to why the control is disabled. Anyway, a problem to address in the future…
disabled

low-batt:
On this:
> Did a minor update to remove use of mpv `DispatchQueue` for the get request

A while back I noticed IINA Advance was dispatching some mpv requests to the `DispatchQueue` in `MPVController`. Been meaning to ask you what prompted that change.

I finally documented the non-obvious reason why we want to get to a point where `MPVController.queue` is only used for reading events from mpv:

https://github.com/iina/iina/blob/0282a30d50f2e7cbc2698186ffac67fd4ab0d873/iina/MPVController.swift#L98-L108

We could use some other `DispatchQueue` for mpv gets, but we want `MPVController.queue` dedicated to reading events.

In correcting problems with locking, work in response to a mpv event that had been done on the `MPVController.queue` was dispatched to the main thread. That resulted in more gets from mpv being done on the main thread. Instruments showed that sometimes the main thread was stuck waiting on a mpv get that was waiting for a mpv lock. Most of the time the wait was short. But one case stood out, getting the track list. That is the reason for PR #5820 which improves the performance of that operation.

What has been your experience?

iina/Base.lproj/FontPickerWindowController.xib
svobs: Whoops. Guess that is something I need to remember to check for. Fixed in new commit

iina/FontPickerWindowController.swift
svobs: 🤷🏼‍♂️went ahead and changed this.

iina/FontPickerWindowController.swift
svobs: Yeah unclear to me whether much is gained by using `guard` statements. To me the `continue`s + `return` seem harder to read. Don't care enough to debate though; went ahead and changed this.

iina/FontPickerWindowController.swift
svobs: These variables don't even need to be stored. Refactored to make them computed.

iina/FontPickerWindowController.swift
svobs: Done

svobs:
> But in the PR the text box has sans-serif filled in:

> Need some easy way to get back to the sans-serif default other than making the user type it in.

Ahh, I did not quite understand the desired default behavior before. My latest commit should add proper support for defaulting to "sans-serif" when the field is empty.

> I'm thinking that maybe when you change the selected font name in` Or enter the font name` the selection in the table should be adjusted accordingly including not selecting anything if you clear the text box.

Yeah I agree. I only omitted this because my initial attempt to do this failed and I wanted to pick a stopping point rather than try to make things perfect in one go 😅 Will take a look at that...

> One of the problems with disabling controls is that it confuses users.

Yes. It drives me nuts when Apple does it 😂 I suppose it's inevitable to start falling down the rabbit hole of subtitles issues...

svobs:
> A while back I noticed IINA Advance was dispatching some mpv requests to the DispatchQueue in MPVController. Been meaning to ask you what prompted that change.

At the time, it was mostly to help with my mental model, because the upstream code still had numerous race conditions and had not yet incorporated your changes which dispatch all events to the `main` queue. It made more sense to me to make all the mpv requests on the same queue as the event callbacks, under the general principle that we do not know how long each mpv request would take to complete, and thus any request could turn out to be long-running. And thus, it would be good to avoid tying up the main queue, as per Apple's general guidelines, and instead launch an async task on a non-UI thread, which is what the mpv queue already provides.

I think it has turned out to be a good decision overall. I think the main advantages have been:
1. It avoids the known deadlock danger for the handful of mpv APIs which can do that on the main thread
2. It reduces the danger of a spinning beach ball cursor generally. In practice, I have seen that loading a new file from a slow network drive can take seconds to return, so it has been good to keep this off the main thread, if only to avoid seeing the scary ball.
3. This is a bit of a hand-wavy argument, but generally I take comfort by being able to assume that when I make an mpv request, since it's on the same queue that processes received events, that there is no ambiguity about the timing of received events compared to the requests I make to mpv. For example, if I have just begun executing a task on the `mpv` queue, and in this task I immediately make a request to set the `video-params/rotate` property, I know it's a safe assumption that the next `video-params/rotate` event received in the mpv event queue will be from that update, because any previous events of that type can be expected to have been processed prior to the queue task on which I'm running. If I had made the request from the `main` queue, I could not be nearly so sure of the timing. It's not 100% but it's certainly more orderly, and I feel that this helps me a lot in tracking down problems.

Obviously, one theoretical downside is that there is some danger of additional lag caused by waiting for the mpv queue to finish processing its previous tasks before getting to this new async task, then (usually) followed by yet another async hop back to the `main` DQ to update the UI to respond. In practice this has only turned out to be an issue for a couple of very timing-critical cases. For those I resorted to using `mpv.queue.sync` from `main`, which has yielded "good enough" results.

> I finally documented the non-obvious reason why we want to get to a point where MPVController.queue is only used for reading events from mpv

This does make sense, and this does sound a bit scary. But I haven't seen this situation occur yet. I'm not too worried either:
1. I'm still not using the mpv queue for any "heavy" work (disk or network access, or any kind big big-O computation over large datasets). Anything of that sort is put onto other queues. So the danger of overflow from the "demand side", so to speak, should be pretty low.
2. I'm not subscribing to too many mpv events which would risk causing an overflow while waiting to be processed, so the danger from the "supply side" is not too great either. The 2 biggest generators of events from mpv are (1) at file load, and (2) `MPV_EVENT_LOG_MESSAGE` events. I make effort to process the former very quickly, avoiding duplicate work, and the latter do not pose too much risk as long as the log level of the subscription is not made too verbose.
3. Generally it has seemed that when there is a bottleneck, it's on mpv's side, not IINA's. It does have to do the job of processing huge files, after all. In practice, like I mentioned, this only seems to be a noticeable issue on slow remote drives. Under such circumstances, there seems to be very little risk of an event buffer overflow because it's internally blocked / too busy to send them out.

I don't want to claim that using the mpv queue in this way is definitively superior to using the main queue. Mainly I've committed to it at this point and it seems to work pretty well. Of course I'll tear it out if some major issue is uncovered, but haven't seen that yet.

iina/FontPickerWindowController.swift
low-batt: Confirmed fixed.

iina/FontPickerWindowController.swift
low-batt: Confirmed resolved.

I think the senior developers tend to like an `if` over a `guard` when it is a simple single `if` statement. In a case like this I find it easier to read and understand the loops without the deeper nesting. I think, but I'm not sure, the senior developers might agree with the use of `guard`.

iina/FontPickerWindowController.swift
low-batt: Confirmed resolved.

iina/AppData.swift
```diff
@@ -82,6 +82,7 @@ struct Constants {
     static let dot = "●"
     static let blackRightPointingTriangle = "▶︎"
     static let blackLeftPointingTriangle = "◀"
+    static let mpvDefaultFont = "sans-serif"
```
low-batt, line 85: No need to change anything as this is an improvement over hardcoding `sans-serif` in many places, but note that software engineering practices would say the better solution would avoid any hardcoding and instead ask mpv for the default value. There is now a `MPVOptionDefaults` class (made for other reasons). I suggest we **do not** use that to obtain the default font.

The complication of using `MPVOptionDefaults` that would occur in this case has to do with class / struct hierarchy.  The struct `Preference` must be one of the lowest in the hierarchy. This is required because it is needed by `Logger`. And we want `Logger` to be very low in the hierarchy as otherwise the classes and structs it is dependent upon cannot use logging. At least not during initialization. Every time we add an external dependency to these classes / structs we should be thinking of this. In this case referencing `Constants` seems fine to me and preferable to the changes needed to obtain the default from mpv.

There are other places in the code that could make use of this new constant. Search for "sans-serif".

iina/Base.lproj/FontPickerWindowController.xib
low-batt: Confirmed fixed.

iina/PrefSubViewController.swift
```diff
@@ -87,9 +87,9 @@ class PrefSubViewController: PreferenceViewController, PreferenceWindowEmbeddabl
   }
 
   @IBAction func chooseSubFontAction(_ sender: AnyObject) {
-    Utility.quickFontPickerWindow { font in
-      Preference.set(font ?? "sans-serif", for: .subTextFont)
-      UserDefaults.standard.synchronize()
+    let subFont = Preference.string(for: .subTextFont)
+    Utility.quickFontPickerWindow(selecting: subFont) { font in
+      Preference.set(font, for: .subTextFont)
```
low-batt, line 92: The removal of `??` and fallback to the default font has caused a regression. With the latest fixes if I open the font window and erase the font the window now correctly shows a dim "sans-serif". All good. But when I click on `OK` the setting panel shows nothing at all for the font:

pr-settings

This is because it was set to the empty string:
```text
low-batt@gag ~$ defaults read com.colliderli.iina | grep subTextFont
    subTextFont = "";
```

This is working correctly in IINA 1.4.1:

1 4 1-settings

```text
low-batt@gag ~$ defaults read com.colliderli.iina | grep subTextFont
    subTextFont = "sans-serif";
```

There might be other places in the code that do this.

low-batt (changes requested):
Nearly all fixed in the latest commit. Yeah! The selecting behavior is a very welcome improvement!

The only problem I found that needs fixing is the setting of the preference to the empty string causing nothing to be displayed for the font in the settings panel.  That is a regression and needs to be fixed.

low-batt:
On this:
> It avoids the known deadlock danger for the handful of mpv APIs which can do that on the main thread

I'm very concerned about this as very rarely I have seen mpv lockup on just a get call from the main thread. I don't see anything in the mpv client documentation that would require the client to not use the main thread, so I'm thinking this is a mpv defect. We need to upgrade to mpv 0.41.0 and see if this still happens and if so, consult with the mpv group. But first we need to release IINA 1.4.2 with fixes before upgrading mpv.

> It reduces the danger of a spinning beach ball cursor generally. In practice, I have seen that loading a new file from a slow network drive can take seconds to return, so it has been good to keep this off the main thread, if only to avoid seeing the scary ball.

What mpv operation is IINA pending on in this case?

> This is a bit of a hand-wavy argument, but generally I take comfort by being able to assume that when I make an mpv request, since it's on the same queue that processes received events, that there is no ambiguity about the timing of received events compared to the requests I make to mpv.

This only makes it more likely the event received is for the mpv request, it is not deterministic. IINA must not assume this is the case.

The queuing to the main thread was one part of the effort to eliminate all the data races being reported by the [Thread Sanitizer](https://developer.apple.com/documentation/xcode/diagnosing-memory-thread-and-crash-issues-early). Last I checked running with TSan enabled did not report any data races. I'm trying to remember to occasionally test with TSan and the address checked to make sure regressions don't sneak in.

This does not preclude performing mpv operations in a thread other than the main thread. But it may require the background operation to queue a task to the main thread that stores the results in `PlaybackInfo`.

On this:
> The 2 biggest generators of events from mpv are (1) at file load, and (2) `MPV_EVENT_LOG_MESSAGE` events.

It was a PR that increased the mpv log level that got me to document why `MPVController.MPVLogLevel` must not be raised above `warn` and this danger of overflowing the mpv event queue. Even that is no guarantee that mpv doesn't have a problem that results in a warning message event storm.

To really stress test this we need to setup a test with both `Add files in the same folder automatically` and subtitles auto load enabled and have IINA play a file in a directory that has a huge number of video files and subtitle files. As the background task for those features add files it generates mpv events. There are some issues having to do with performance of huge playlists. I've never had time to investigate those.

The problem is that if the mpv event queue overflows we currently have no way of knowing that was the source of whatever bad behavior IINA exhibits (which could be very bad, such as hanging). This is how you end up with random mysterious problems that are impossible to debug when it only happens for users.

To avoid that situation IINA needs to listen for `MPV_EVENT_QUEUE_OVERFLOW` events. The question is what IINA should do when that happens. Definitely needs to be logged as an error, but for rare problems that doesn't help as users usually don't have logs turned on. So this probably has to be an alert. Maybe a fatal one? That alert probably has to support being suppressed just in case it reveals that some users do encounter this.

As always, I am very interested in your opinions on this.

iina/AppData.swift
svobs: I was able to remove one occurrence, but I'm not comfortable removing the remaining occurrences. A few look unrelated and are part of long HTML strings. One is in a XIB and gets overwritten anyway but I think it helps improve the readability of the XIB preview.

iina/PrefSubViewController.swift
svobs: Gotcha. Should be fixed now.

svobs:
> I'm very concerned about this as very rarely I have seen mpv lockup on just a get call from the main thread. I don't see anything in the mpv client documentation that would require the client to not use the main thread, so I'm thinking this is a mpv defect.

I was actually referring to the case where doing certain (render-related?) requests on the main thread could cause deadlock if called while `mpv_render_context_create` is executing, which I believe @krackers explained...somewhere. As usual I'm not able to locate the thread where this was mentioned, so unfortunately I'm going from memory. Given that various menu items and other app-wide things make references the "last active player", "idle or active player", et al, and use that for various "get" requests, could there be a logic hole where one such request is allowed to execute at the wrong time? By staying off the main thread for requests that doesn't need to be worried about. (There could also be bugs/regressions in mpv's code too, of course).

> What mpv operation is IINA pending on in this case?

Ah, sorry - I was talking a bit loosely in this case 😇 I had previously identified such cases in IINA Advance, and found that these were fixed by (as described before) putting all mpv requests in async tasks on the mpv queue. I was just assuming that IINA would show a beach ball when it makes an mpv request on my main thread while the mpv core is blocked. If the request fails to return in a timely manner, it can be assumed that, after a certain point, MacOS will decide that the main thread is "not responding" and will start spinning the beach ball. And I know that I've seen cases where mpv requests can be slow to respond - notably when playing files over SMB, which can sometimes slow down quite a bit unpredictably.

I tried doing just that right now. Unfortunately my Mac's Samba driver seems to be pretty snappy at the moment... I tried opening multiple windows, each for a separate directory on my NAS, and started playing all of them, then navigating through the playlists of each. When frantically loading file after file in quick succession, I was able to get the occasional beach ball while waiting for a file to load. Briefly, maybe 1 out of 10 files.  I couldn't tell you which API was causing it without more debugging, but it may not matter really - if the load is high enough, eventually, everything will back up. I was able to see that the UI still had a tendency to lock up momentarily after double-clicking on a new file in the playlist and before the file started playing. For a fraction of a second I would be unable to move the window or perform some other interaction. Although I'm not sure these cases would meet Apple's definition of "microhangs", I would describe them as such. I believe such episodes would trigger the beach ball if only they were longer in duration. But even if not, these hiccups create a subpar user experience if the user is trying to do something and notices them.

By contrast, in IINA Advance I enqueue all the mpv requests asynchronously (or at least nearly all...) , including the "get requests", so that even the requests which were triggered via user actions in the UI are (by definition) enqueued in that queue. So if the mpv queue is blocked by some previous work and takes a long time to get to it, during this time from the user's standpoint it's almost like their request was forgotten. But during that time they can still do other UI things which don't require mpv requests, like resizing or or moving the window, or maybe opening sidebars, etc. In theory none of those things should block either because they are also enqueuing async requests to mpv - maybe they won't populate or refresh right away, but the average user will care less about the UI being slow to process their action than they would about being forced to wait before making an action. 

I dabbled a bit in trying to formally convert either `MPVController` or `PlayerCore` into a Swift actor to enforce isolation of the "core" from the main thread. But it looks like Swift 6 will be needed to do that well (and supporting Swift 6 opens a Pandora's box of issues...)

> It was a PR that increased the mpv log level that got me to document why MPVController.MPVLogLevel must not be raised above warn and this danger of overflowing the mpv event queue. Even that is no guarantee that mpv doesn't have a problem that results in a warning message event storm.

Right - I remember this. That was very useful information, thanks!

iina/AppData.swift
low-batt: I had not looked into those references. I agree that we should not try and refactor those references in this PR.

iina/PrefSubViewController.swift
low-batt: Tested and confirmed fixed.

low-batt (approved):
Tested latest commit, working fine for me.

Changes look good to me.

low-batt:
On this:
> I was actually referring to the case where doing certain (render-related?) requests on the main thread could cause deadlock if called while `mpv_render_context_create` is executing

Thanks, that might be key information. I'm going to look into this. I will also see if I can find that comment from @krackers.

On moving all mpv requests to a background thread…

At one point I did look into this using Instruments. More testing is needed, but the main offender was getting the track list. The way it is coded now it is an O(n) operation. From what I recall, when IINA is adding subtitles the track list is changing requiring IINA to frequently refetch the track list. PR #5820 changes this to be an O(1) operation. The improvement that change makes is very evident in Instruments.

The problem with moving many of these mpv operations to a background queue is that the results from gets usually need to be stored in objects that do not have locks because they are intended to only be accessed by the main thread. That means when the get completes the background thread would have to queue a task to the main thread to update objects such as `PlaybackInfo`. As some of the gets are inline with other processing that would involve refactoring to separate out the mpv operations. With Instruments not showing long hangs once getting the track list was changed, I did not look further into making such changes. Of course that was without investigating IINA under a stress test designed to bring out problems waiting on mpv. Such a test may prove the worth of moving mpv requests to the background. But I would like to first be able to see the hangs happening in Instruments and identify the offending operations. At the moment there are other priorities.

Currently the focus is on IINA 1.4.2 which needs to be about fixing regressions and not introducing any new problems. Odd "interesting" macOS Tahoe problems are still being discovered. See my analysis in issue #5879 of one of the recently discovered problems. Definitely not the time for major refactoring.

IINA 1.5.0 was supposed to be about adding functionality, but because of problems in the libraries IINA is dependent upon, the focus of that release has changed to be about upgrading dependencies and limiting other changes to avoid that release taking too long.

One coding level design problem IINA has is that the classes are very tightly coupled. That is harder to avoid with high level classes, but it is desirable to at least try and achieve loose coupling in lower level classes. To me `MPVController` is a lower level class where we should be striving to reduce its knowledge / dependencies on higher level classes. Because that class is so intertwined with higher level classes it can not be reused to display OSC thumbnails, which is one of the solutions proposed to be able to properly display thumbnails of HDR videos. At some point I plan on creating a reusable `MPVClient` that is solely focused on providing a Swift mpv client, hiding all the special code required to use the C based mpv client API. Something like that. Too many higher priorities so I have not given this a lot of thought, but this might be where this area of IINA code is heading.

I'll try and find time to create a test that might exhibit main thread hangs. But at the moment I need to focus on IINA 1.4.2.

krackers:
> (render-related?) requests on the main thread could cause deadlock if called while mpv_render_context_create is executing

`mpv_render_context_create` call should have been fully completed before libmpv calls your render frame callback, right? So I'm not sure what you mean by "render-related requests" here. I don't think I ever commented on `render_context_create` in particular, but the deadlock situations before in IINA were related to not calling `mpv_render_update` in a timely fashion when requested.

There are also other deadlock situations that can result when you make use of certain vo properties that access video/gpu context from the render-thread itself (e.g. querying `vo-passes`). But I'm not sure if there would be an issue with one thread querying vo properties at the same time as `mpv_render_context_create` is called in another thread. 

To my knowledge there shouldn't be any issue calling `mpv_render_context_create` on one thread then receiving callbacks on another thread (this is called out as acceptable in API docs). I haven't looked at all this in a few years though, so take what I say with a grain of salt.

See https://github.com/iina/iina/pull/4830
https://github.com/iina/iina/issues/4822

svobs:
@low-batt:

> The problem with moving many of these mpv operations to a background queue is that the results from gets usually need to be stored in objects that do not have locks because they are intended to only be accessed by the main thread. That means when the get completes the background thread would have to queue a task to the main thread to update objects such as PlaybackInfo.

Right. My solution to this over time has been to ensure that most of these objects are only read or written to while in the mpv DispatchQueue. I suppose I should use TSan at some point to see what I missed, because it has been a piecemeal process. Generally I've been trying to refactor all the model objects into read-only structs. When one var needs to be read on a separate DQ, I first make a local copy of it on their "home" DQ before sending it to the DQ which needs it. In some cases such as the Playlist table, where I need to determine diffs across data for the sake of animation, I actually have a separate copies of the data which has the possibility of being briefly out of data. Doing things this way ends up actually using less memory & is far faster (at least everywhere so far) than using mutable classes, avoids data races, and will make the data easier to plug into SwiftUI in the future. But as you mentioned, this does involve extensive refactoring which carries its own set of risks.

> To me MPVController is a lower level class where we should be striving to reduce its knowledge / dependencies on higher level classes

Yeah I agree. But it's not completely unmanageable (e.g., something like a case where mpv & IINA requests nested in a braided fashion several layers deep). It seems to be the shared state variables which is the biggest problem.

There is also the blurry boundary between `PlayerCore` and `PlayerWindowController` to worry about. Seems like it would be nice to at least isolate all the mpv stuff to `MPVController` and `PlayerCore` at first? PWC & `MainWindowController` should ideally be in charge of ordering AppKit around, and leave the player-related stuff to the other two classes. I have been trying to support running the app in a "non-interactive" mode, so that scripts like `thumbfast` can launch instances of it in the background to generate thumbnails, and there should be no need to create a window controller because there will be no window displayed...yet I still need to do that because of all the entangled logic.

@krackers:

> `mpv_render_context_create` call should have been fully completed before libmpv calls your render frame callback, right? So I'm not sure what you mean by "render-related requests" here. I don't think I ever commented on `render_context_create` in particular, but the deadlock situations before in IINA were related to not calling `mpv_render_update` in a timely fashion when requested.

Ah, yes, maybe `mpv_render_update`  is what I was remembering. So this puts my mind at ease.

> There are also other deadlock situations that can result when you make use of certain vo properties that access video/gpu context from the render-thread itself (e.g. querying `vo-passes`) But I'm not sure if there would be an issue with one thread querying vo properties at the same time as `mpv_render_context_create` is called in another thread.

Right. I don't believe the main thread is being used for rendering, while `mpv_render_context_create` is being called on the main thread, so that shouldn't be an issue.

> I haven't looked at all this in a few years though, so take what I say with a grain of salt.

Same 😅 Thanks.

low-batt:
@krackers In the hang I encountered IINA was permanently blocked on a call to `mpv_set_property`. This happened when IINA was processing a `MPV_EVENT_FILE_LOADED` event. That is handled by the `PlayerCore.fileLoaded` method which was making this call to pause playback:
```swift
mpv.setFlag(MPVOption.PlaybackControl.pause, true, level: .verbose)
```

IINA was always pausing playback as a workaround for some ancient defect where volume would always start at 100%. As this workaround was breaking gapless playback it has been removed. 

This hang seem to start happening with IINA 1.4.0. The two interesting changes were upgrading to mpv 0.38.0 and handling the `MPV_EVENT_FILE_LOADED` event on the main thread instead of a background thread.

I've also seen similar hangs calling `mpv_get_property`. Possibly related to one thread retrieving the track list while another was adding tracks.  As a side note, in PR #5820 I have proposed changing the code that gets the track list to obtain the entire list in one call using `MPV_FORMAT_NODE`.

I did not look into these hangs mainly because they are really rare and hard to reproduce, but also because IINA code in this area is changing and IINA is using such an old `libmpv`. The plan is to release an IINA 1.4.2 with fixes for some regressions and problems under macOS Tahoe. And then work on an IINA 1.5.0 that will upgrade mpv and FFmpeg. Once IINA is up to date with mpv I was going to keep an eye out for this kind of hang and aggressively pursue it if it happens again.

@svobs The above problem was why I was so interested your efforts to move mpv requests off the main thread. I was a little worried the above hang had something to do with making mpv calls on the main thread.

On this:
> My solution to this over time has been to ensure that most of these objects are only read or written to while in the mpv DispatchQueue

I'm confused? A lot of the information in `PlaybackInfo` is needed by the UI, which is going to be running on the main thread. I had not considered moving to structs. That could be an effective technique. Definitely do some testing with TSan.

On this:
> There is also the blurry boundary between `PlayerCore` and `PlayerWindowController` to worry about. Seems like it would be nice to at least isolate all the mpv stuff to `MPVController` and `PlayerCore` at first?

Yes, I have some concerns about the `PlayerCore` and `PlayerWindowController` relationship, but have not thought much about that.

In our CONTRIBUTING document in the [Current structure](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md#current-structure) section indicates:
> - `PlayerCore` encapsulates general playback functions.
>  	- Setting options/properties directly through `MPVController` is discouraged.

But that is violated in quite a few places as the `MPVController` object in `PlayerCore` is not private and code is accessing that property directly.

However that is not the refactoring I was looking into. Where it makes sense I think it is desirable for IINA to line up with mpv code. Where that especially makes sense is in the `VideoView` and `ViewLayer` code. The in mpv code the rendering code is grouped together. In IINA there is a bunch of it in `MPVController`. As that class is quite large I was looking into creating a `MPVRenderer` class and moving rendering code into that class, similar to how mpv structures its code.

On this:
> I don't believe the main thread is being used for rendering

Most rendering occurs in the background on the `VideoLayer.mpvGLQueue`, but some rendering is performed by AppKit on the main thread. There is trouble in this area with main thread lock starvation.

In mpv, all rendering is now performed using the main thread. I believe mpv is able to do this due to the better performance of Metal. If you force mpv back to using OpenGL by specifying `--vo=libmpv` then mpv will exhibit UI issues due to the main thread being overloaded. Since IINA uses OpenGL, a background thread must be used for most rendering.

PR #5868, opened 2025-12-31
Expand file open customization settings

### LINKED ISSUE #5865: Expand file open customization settings
The way IINA handle opening (multiple) files from the Finder has changed a bit this year, and some people have responded with requests for more options. This PR aims to rework IINA's open settings just enough to support all the needs that have been expressed.

# Proposal

We would replace this setting:

- [ ] Always open media in new window

With these:

- [X] Allow multiple player windows
	- [X] Group media in playlist when opening multiple files at once
	- [ ] Allow opening the same file in multiple windows

The two nested settings would only be enabled if the parent setting is checked. The default values, as shown, would match standard macOS behavior.

I'm happy to implement this, but wanted to ask for approval first!

# Context

- #5071 made it so that selecting multiple files in the Finder, and opening them all at once, doesn't create a playlist anymore, when “Always open media in new window” is selected. Instead, it opens each in its own window.
	- This matches system behavior, and likely matches most people's expectations
	- You can still create playlists by unchecking the setting; or by dragging files on an existing player
	- But, that means it's now harder to decide on-the-fly between opening a group of files in a single window, or multiple (before, you could make that choice by either opening each file separately, or opening one big selection all at once)
		- Requests: https://github.com/iina/iina/issues/5574#issuecomment-3522807735, https://github.com/iina/iina/issues/5574#issuecomment-3621579409, https://github.com/iina/iina/issues/5574#issuecomment-3693760584, https://github.com/iina/iina/issues/5488#issuecomment-3313495250
- That same PR also made it impossible to open the same file in multiple windows.
	- This matches system behavior, and likely matches most people's expectations
	- But, that means you can no longer open two windows of the same file to e.g. compare different times of the video
		- Request: #5689
	- This change has already been completely reverted in #5689. I proposed instead adding a setting for it in #5762. That PR would be replaced by the changes proposed here.

So, since no single behavior will satisfy everyone, it could be a good idea to find the right set of settings to add. These are the small, easiest-to-understand set I've been able to come up with.

Cykelero:
- [X] This implements/fixes issue #5865.

---

This PR modifies the `alwaysOpenInNewWindow` setting, and adds `groupSimultaneousOpensInPlaylist` and `allowDuplicatePlayers`. See issue for details.

The openURLs() method loses two arguments:
- `preferredCore` wasn't used anywhere
- `inverseOpenInNewWindowPref` was used to allow temporarily inverting the open behavior, by holding Option while choose File > Open. This was hard to discover, and would be harder to design predictably, now that three distinct options affect the opening behavior.

I tested all the scenarios that seemed relevant (opening one or more files in the Finder, that are already open or not, etc) for all configurations of the new options.

- The PR adds a comment in Preference.swift about the fact that “alwaysOpenInNewWindow” now has a slightly different meaning; not sure if there's a better way to document this change.
- In the settings window, the two new sub-settings get disabled (grayed out) when the parent setting is unchecked. It would be even nicer if they got unchecked, to communicate that they don't apply anymore; but that didn't seem like reasonable time and complexity for a UI that's about to get completely rewritten!

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -932,7 +932,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, SPUUpdaterDelegate {
         }
       }
       let isAlternative = (sender as? NSMenuItem)?.tag == AlternativeMenuItemTag
-      if PlayerCore.openURLs(panel.urls, inverseOpenInNewWindowPref: isAlternative) == 0 {
+      if PlayerCore.openURLs(panel.urls) == 0 {
         Utility.showAlert("nothing_to_open")
       }
     }
--- a/iina/Base.lproj/PrefGeneralViewController.xib
+++ b/iina/Base.lproj/PrefGeneralViewController.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="23090" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="24506" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="23090"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="24506"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <objects>
@@ -24,19 +24,19 @@
             <point key="canvasLocation" x="-1607" y="64"/>
         </customView>
         <customView id="6zR-96-iKB" userLabel="Prefs &gt; General &gt; Behavior">
-            <rect key="frame" x="0.0" y="0.0" width="480" height="574"/>
+            <rect key="frame" x="0.0" y="0.0" width="486" height="577"/>
             <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMinY="YES"/>
             <subviews>
-                <textField identifier="SectionTitleBehavior" horizontalHuggingPriority="251" verticalHuggingPriority="750" allowsCharacterPickerTouchBarItem="YES" translatesAutoresizingMaskIntoConstraints="NO" id="38y-2I-MGs">
-                    <rect key="frame" x="-2" y="550" width="66" height="16"/>
+                <textField identifier="SectionTitleBehavior" focusRingType="none" horizontalHuggingPriority="251" verticalHuggingPriority="750" allowsCharacterPickerTouchBarItem="YES" translatesAutoresizingMaskIntoConstraints="NO" id="38y-2I-MGs">
+                    <rect key="frame" x="-2" y="553" width="66" height="16"/>
                     <textFieldCell key="cell" sendsActionOnEndEditing="YES" alignment="left" title="Behavior:" id="uQR-Fx-oEm">
                         <font key="font" metaFont="systemBold"/>
                         <color key="textColor" name="labelColor" catalog="System" colorSpace="catalog"/>
                         <color key="backgroundColor" name="controlColor" catalog="System" colorSpace="catalog"/>
                     </textFieldCell>
                 </textField>
-                <textField horizontalHuggingPriority="251" verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="OXi-vL-mzp">
-                    <rect key="frame" x="118" y="550" width="65" height="16"/>
+                <textField focusRingType="none" horizontalHuggingPriority="251" verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="OXi-vL-mzp">
+                    <rect key="frame" x="118" y="553" width="65" height="16"/>
                     <textFieldCell key="cell" scrollable="YES" lineBreakMode="clipping" sendsActionOnEndEditing="YES" title="At launch:" id="gqT-CB-Puw">
                         <font key="font" metaFont="system"/>
                         <color key="textColor" name="labelColor" catalog="System" colorSpace="catalog"/>
@@ -44,7 +44,7 @@
                     </textFieldCell>
                 </textField>
                 <popUpButton verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="bV8-LP-cwH">
-                    <rect key="frame" x="186" y="543" width="179" height="25"/>
+                    <rect key="frame" x="189" y="549" width="188" height="24"/>
                     <popUpButtonCell key="cell" type="push" title="Show welcome window" bezelStyle="rounded" alignment="left" lineBreakMode="truncatingTail" state="on" borderStyle="borderAndBezel" imageScaling="proportionallyDown" inset="2" selectedItem="z9T-8h-1T0" id="6y1-KS-0tJ">
                         <behavior key="behavior" lightByBackground="YES" lightByGray="YES"/>
                         <font key="font" metaFont="message"/>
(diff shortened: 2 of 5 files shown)


low-batt:
With `Group media in playlist when opening multiple files at one` _disabled_, if I drag-n-drop a group of 3 video files onto the open IINA window playing another video, the video being played is replaced by one of the files in the group and all 3 of the files are added to the playlist. In the case of drag-n-drop  `Group media in playlist when opening multiple files at one` is being applied when it is _disabled_. 

Also, the title of the single IINA window still shows the title of the previous video being played under the `Window` menu and in IINA's `Dock menu`. The title in the title bar of IINA's window correctly reflects the title of the new video being played. I vaguely recall some sort of AppKit defect in this area. I will investigate this behavior.

Ideally Xcode would not report any problems with IINA XiBs. That is not true today. Xcode will identify some problem in IINA's XiBs. My own rule is to not add new XiB warnings/errors. I checked the current version of `PrefGeneralViewController.xib` and Xcode did not report any errors. With the version in the PR Xcode is reporting problems with constraints that need to be corrected:

errors

The [Localizations](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md#localizations) section of IINA's [CONTRIBUTING](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md) guidelines says this:
> We use [Crowdin](https://crowdin.com/project/iina) for localization. When there is a UI text change, only add/update the strings in `Base.lproj` and `en.lproj`. Please do not include any localization changes for other languages in the pull request. New strings on the develop branch from the `en.lproj` folder will be synced automatically to Crowdin, and we will fetch the latest translations from Crowdin before releasing the next version.

That probably should be expanded to provide more details. The need to have English text in both `Base.lproj` and `en.lproj` has to do with the integration with Crowdin.

A while back switching to a [string catalog](https://developer.apple.com/documentation/xcode/localizing-and-varying-text-with-a-string-catalog) was investigated, but that caused issues with the Crowdin integration.

In the case at hand, because localizable text in `PrefGeneralViewController.xib` has been changed the file `iina/en.lproj/PrefGeneralViewController.strings` needs to be updated. 

The way I update the en.lproj strings file for a XIB is to use [ibtool](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/MaintaingYourOwnStringsFiles/MaintaingYourOwnStringsFiles.html#//apple_ref/doc/uid/10000171i-CH19-SW6) like so:
```text
ibtool iina/Base.lproj/PrefGeneralViewController.xib --generate-strings-file temp.strings
```

And then run a good diff tool. Except the good diff tool I was using no longer working under macOS. That makes things painful, so I went ahead and identified the changes needed.

These two new entries need to be added to `iina/en.lproj/PrefGeneralViewController.strings`:
```text
/* Class = "NSButtonCell"; title = "Group media in playlist when opening multiple files at once"; ObjectID = "WZE-pS-9ne"; */
"WZE-pS-9ne.title" = "Group media in playlist when opening multiple files at once";

/* Class = "NSButtonCell"; title = "Allow opening the same file in multiple windows"; ObjectID = "Z7U-cf-ezr"; */
"Z7U-cf-ezr.title" = "Allow opening the same file in multiple windows";
```

It would also be good to correct this _existing_ inconsistency in the XiB. The title of this button in the XiB should be corrected:
```xml
                   <buttonCell key="cell" type="check" title="Allow multiple player windows" bezelStyle="regularSquare" imagePosition="left" state="on" inset="2" id="M0z-bI-88d">
```

That text is not seen in the UI because the correct text is in the `iina/en.lproj/PrefGeneralViewController.strings` file:
```text
/* Class = "NSButtonCell"; title = "Always open media in new window"; ObjectID = "M0z-bI-88d"; */
"M0z-bI-88d.title" = "Always open media in new window";
```

After a PR with text changes is merged we run scripts that update Crowdin. We have other scripts that pull translations from Crowdin and merge them into the `develop` branch.

I've run out of time today. I will do some more testing when I get a chance.

Cykelero:
Thank you for testing the changes!

> In the case of drag-n-drop `Group media in playlist when opening multiple files at one` is being applied when it is _disabled_.

Good observation. I think this is desirable behavior: drag-and-dropping files onto a player window (as opposed to just double-clicking them, for instance) is a pretty deliberate action. I feel like it'll be a natural way for people to indicate they want to add to the window's playlist. If they want to open separate windows, then they'll probably know to simply double-click the files.

This is especially useful, because it means people can disable the “Group media in playlist” option, and still have an easy way of assembling playlists on the fly, when needed. Does that sound right to you?

> problems with IINA XiBs

I fixed the warning (the “localization issue” missing trailing constraint), but my Xcode isn't reporting the content priority ambiguity seen in your screenshot, strangely, even after reopening the file.

> These two new entries need to be added to iina/en.lproj/PrefGeneralViewController.strings:

Done.

> The title of this button in the XiB should be corrected:

And done!

low-batt:
First, thank you for the excellent work. I'm always jumping right into commenting on the work which is usually about problems and forgetting to mention that I think this is great work and the contribution is appreciated.

On drag-and-dropping not obeying `Group media in playlist when opening multiple files at once`. I am unsure what the proper behavior should be. For UI questions I defer to the senior developers. Let's not make any changes now and see what they have to say.

I will point out that if the user wants to add to the playlist they could open the playlist panel and drag and drop the files into the playlist. Of course that is not as convenient as just drag and dropping into the main window.

What really troubles me is that the original changes were made because IINA had a setting that wasn't obeyed in some cases. With the current drag and drop behavior it feels like that is still the case. Again, let's see what the senior developers think.

On this:
> but my Xcode isn't reporting the content priority ambiguity seen in your screenshot, strangely, even after reopening the file.

That would be because Xcode is stinky, buggy and frequently refuses to obey. Worse, you always have to be ready for it to "helpfully" totally mess up your XiB, moving everything around, requiring you to drop all the local changes and start over. This is especially true if you ever say yes to its prompts to fix a XiB problem for you.

Also, when it reports a XiB problem, sometimes it correctly points out the source of the problem, but sometimes it is totally incorrect about the nature of the problem. Like a lot of computer error reporting be ready for the error reported only meaning that something is wrong somewhere.

A lot of the time the nice behavior you see with source code errors where Xcode incrementally compiles the code after you make a change and removes the compilation error it flagged does not happen with XiBs. Typically I have to close the XiB. Click on `Clean Build Folder…` under `Product`. Click on `Build` under `Product`. And then reopen the XiB file to see all the XiB warnings and errors. A real pain.

I tracked down the source of this problem and how to fix it. This is one of those cases where the error reported by Xcode is a side effect of the actual problem.

As seen here Xcode seems to be complaining about `Prevent screen saver from starting while playing`:

prevent-screen

But that is not where the error is. When I am making changes and hitting problems like this I examine closely the constraints on the controls I added and compare them to the constraints on existing controls. In this case that showed that the constraints on `Allow opening the same file in multiple windows` are incorrect:

incorrect

After correcting those two constraints (not touching priorities), having Xcode helpfully change everything in the XiB, throwing away local changes and fixing the constraints again, Xcode is happy with the XiB:

correct

Exhausted from fighting with Xcode. I still need to walk through the code changes and test some more. Will come back to this review soon.

Cykelero:
Thank you for the nice words, and the investigation.

I've fixed the constraint you describe—hopefully your Xcode will agree too!
In the end there's one less horizontal constraint than in your suggestion, because the checkbox below (“Quit after all windows are closed”) already has a leading edge constraint with “Allow multiple players windows”; so no need to also constrain it to “Allow opening the same file in multiple windows”:

image

low-batt:
Xcode 26.1.1 is not still happy. If I add the extra constraint then Xcode is happy. Not making sense to me as the error is complaining about vertical hugging priority, so how does adding a horizontal constraint solve that?

What version of Xcode are you using?

I will try my other development machine which has a newer macOS/Xcode. Stay tuned.

Cykelero:
Yeah, errors in XIBs are really not reliable. You can both get nonsensical errors that disappear on their own, or have errors show up that previously weren't there. Hard to tell which one we're looking at, here.

I'm running Xcode 26.2. I tried 26.1.1 like you, and it still doesn't show the errors.

If your Xcode keeps showing the error, feel free to submit a patch for it (you might be able to push to this branch) and we'll be done with it. It's not worth spending much more time on this, since these XIBs are soon to be replaced!

low-batt:
I had a chance to test this on my other development Mac today an Intel Mac running macOS Tahoe 26.2 and Xcode 26.2. Xcode 26.2 is happy with the current XiB. My conclusion is that this is an error in Xcode 26.1.1, although I'm really confused as to why that version of Xcode does not show the error for you. I am running Xcode 26.1.1 on my M1 Mac under macOS Sequoia 15.7.2.

I'm a bit surprised that Apple would be changing anything in Xcode that would change the XiB behavior. Maybe there were changes to code involving constraints due to all the UI changes in macOS Tahoe? We are still finding features broken in macOS Tahoe by changes Apple made to AppKit.

I still need to do more testing and look at the code changes. Juggling a bunch of work at the moment. Should be able to do that soon.

PR #5933, opened 2026-02-26
[Feat] Configure system media keys

### LINKED ISSUE #1932: Global hotkeys?
Hi!

I was wondering if there's any way to enable global hotkeys in IINA. I usually work in my main software and let videos run in IINA. I would like to forward/backward skip without switching the main app to IINA, but rather just using any shortcut so I could remain in my main app and not disrupt my workflow.

Thanks!

### LINKED ISSUE #2502: Rebindable media keys
**Actual behavior:**
Related to  #1864, I think it would be beneficial to be able to rebind the media keys in the usual fashion, even if they're abstracted away by some macOS remote control API. It would make it possible to e.g. skip backwards/forwards 10 s with the prev/next keys even when in PiP instead of having to find the window and bring it into focus.

If at all possible it would be nice to have them listed as e.g. "macOS media control: next".

I couldn't find an open request about this, but I suspect there might be one buried...

### LINKED ISSUE #3512: Can I set the system prev./next button to rw/ff function?
**What you want IINA to do:**
the customisable shortcut is fantastic. I'm using system media control which is macOS fn keys to pause/play to sort out media data, so it requires many rewind and fast forward. How can I set the system prev./next button to rw/ff function?

**What IINA does currently:**
When I press prev./next aka F7/F9 above MacBook keyboard, it gets prev./next song as if it were a song player.

I don't know if I could add it in shortcut. As I press the functions it doesn't recognise it.

n/a, hope IINA be the first player for media data sorting.

### LINKED ISSUE #4631: Seek with F7/F9
**What you want IINA to do:**
I would like IINA to seek with F7/F9 media keys instead of control the playlist.

**What IINA does currently:**
F7/F9 media keys control the playlist.

There are no shortcuts to control seeking by one button press without focusing IINA or hovering the player and scrolling.

Personally, I haven't ever found `playlist-next`/`prev` to be useful when viewing videos, but find myself seeking frequently. If I am viewing a video and typing notes, it is easier to press F7 to rewind and relisten to something I missed than to change window focus or move the mouse.

benniekiss:
- [x] This implements/fixes issue #2502, #4631, #3512, #1932

---

**Description:**

This PR implements user-configured keybindings for the system media keys (PREV/NEXT, PLAYPAUSE, etc.).

This enables customizing the behavior of the media keys using the regular keybinding config.

The implementation first tries to find a configured keybinding, and if one is found, it dispatches the command to MPV. If no command was configured, it falls back to default values.

This is my first attempt at writing swift, so I would appreciate any feedback. I have tested this locally and it works great for me.

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -1280,11 +1280,13 @@ class RemoteCommandController {
     guard RemoteCommandController.useSystemMediaControl, !isEnabled else { return }
     let remoteCommand = MPRemoteCommandCenter.shared()
     remoteCommand.playCommand.addTarget { _ in
-      PlayerCore.lastActive.resume()
+      let command = PlayerCore.keyBindings["PLAY"]
+      _ = PlayerCore.lastActive.mpv.command(rawString: command!.rawAction)
       return .success
     }
     remoteCommand.pauseCommand.addTarget { _ in
-      PlayerCore.lastActive.pause()
+      let command = PlayerCore.keyBindings["PAUSE"]
+      _ = PlayerCore.lastActive.mpv.command(rawString: command!.rawAction)
       return .success
     }
     remoteCommand.togglePlayPauseCommand.addTarget { _ in
@@ -1296,11 +1298,13 @@ class RemoteCommandController {
       return .success
     }
     remoteCommand.nextTrackCommand.addTarget { _ in
-      PlayerCore.lastActive.navigateInPlaylist(nextMedia: true)
+      let command = PlayerCore.keyBindings["NEXT"]
+      _ = PlayerCore.lastActive.mpv.command(rawString: command!.rawAction)
       return .success
     }
     remoteCommand.previousTrackCommand.addTarget { _ in
-      PlayerCore.lastActive.navigateInPlaylist(nextMedia: false)
+      let command = PlayerCore.keyBindings["PREV"]
+      _ = PlayerCore.lastActive.mpv.command(rawString: command!.rawAction)
       return .success
     }
     remoteCommand.changeRepeatModeCommand.addTarget { _ in
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -1317,7 +1317,7 @@ class RemoteCommandController {
         log("""
             Seek flag was not one of \(allowedFlags.map({ "'\($0)'" }).joined(separator: ", ")), not setting seek interval:
                 \(key) \(keyBinding.rawAction)
-            """
+            """)
         return []
       }
     }
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -1280,7 +1280,7 @@ class RemoteCommandController {
   ///
   /// IINA allows the user to bind a mpv command to the `FORWARD` and `REWIND` media keys. This method must:
   /// - Determine if there is a key binding for the given key and if not return the default of 15 seconds
-  /// - Determine if the key is bound to an IINA command and if so return an empy array indicating the property is not applicable
+  /// - Determine if the key is bound to an IINA command and if so return an empty array indicating the property is not applicable
   /// - Determine if the key is bound to the mpv
   ///     [seek](https://mpv.io/manual/stable/#command-interface-seek-%3Ctarget%3E-[%3Cflags%3E]) command
   ///     and if not, return an empty array
@@ -1301,23 +1301,23 @@ class RemoteCommandController {
     guard !keyBinding.isIINACommand else { return [] }
     let action = keyBinding.action
     guard action.count > 1, action[0] == MPVCommand.seek.rawValue else { return [] }
-    guard let target = Int(action[1]) else {
+    guard let target = Double(action[1]) else {
       log("""
-          Unable to parse seek target as an Integer in key binding:
+          Unable to parse seek target as a Double in key binding:
               \(key) \(keyBinding.rawAction)
           """, level: .error)
       return []
     }
     if action.count > 2 {
-      let validFlags: Set<String> = ["exact", "keyframes", "relative"]
+      let allowedFlags: Set<String> = ["exact", "keyframes", "relative"]
       // Multiple flags can be composed using `+`, and each one must be valid
       let flags = action[2].split(separator: "+")
 
-      guard flags.allSatisfy({ validFlags.contains(String($0)) }) else {
+      guard flags.allSatisfy({ allowedFlags.contains(String($0)) }) else {
         log("""
-            Unable to parse seek flag as one of "exact", "keyframes", or "relative":
+            Seek flag was not one of \(allowedFlags.map({ "'\($0)'" }).joined(separator: ", ")), not setting seek interval:
                 \(key) \(keyBinding.rawAction)
-            """, level: .error)
+            """
         return []
       }
     }
@@ -1450,15 +1450,18 @@ class RemoteCommandController {
       guard isEnabled else { return }
       // The user has modified the key bindings, possibly changing the mpv commands associated with
       // the FORWARD and REWIND media keys. The preferredIntervals values are set based on the mpv
-      // commands assigned to those keys. When expanded the macOS Now Playing module may display
-      // the interval in the seek backward and seek forward buttons. The user may have changed the
-      // number of seconds to seek causing the value displayed in the Now Playing module buttons to
-      // to be out of date. Merely updating the preferredIntervals values is insufficient to get the
-      // Now Playing module to update its buttons. Support for media keys and remote commands must
-      // be disabled and then re-enabled in a separate task.
+      // commands assigned to those keys. The Now Playing module may display the interval in the
+      // seek backward and seek forward buttons causing the value displayed in the Now Playing
+      // module buttons to to be out of date. Merely updating the preferredIntervals values is
+      // insufficient to get the Now Playing module to update its buttons. Support for media keys
+      // and remote commands must be disabled and then re-enabled.
       log("Restarting support for remote commands due to changes to key bindings")
       disable()
-      DispatchQueue.main.async { self.enable() }
+      // Immediately re-enabling media keys and remote commands only partially worked. The Now
+      // Playing module would update buttons if the module was expanded, but the buttons shown in
+      // the module's small form would still display the old interval. Work around this curious
+      // Now Playing behavior by delaying the re-enabling.
+      DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + 1) { self.enable() }
     }
   }
 }
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -1282,48 +1282,48 @@ class RemoteCommandController {
 
     // For each command, apply a configured keybinding or fallback to default values.
     remoteCommand.playCommand.addTarget { _ in
-      if let action = PlayerCore.keyBindings["PLAY"]?.rawAction {
-        _ = PlayerCore.lastActive.mpv.command(rawString: action)
+      if let action = PlayerCore.keyBindings["PLAY"] {
+        PlayerCore.lastActive.mainWindow.handleKeyBinding(action)
       } else {
         PlayerCore.lastActive.resume()
       }
       return .success
     }
     remoteCommand.pauseCommand.addTarget { _ in
-      if let action = PlayerCore.keyBindings["PAUSE"]?.rawAction {
-        _ = PlayerCore.lastActive.mpv.command(rawString: action)
+      if let action = PlayerCore.keyBindings["PAUSE"] {
+        PlayerCore.lastActive.mainWindow.handleKeyBinding(action)
       } else {
         PlayerCore.lastActive.pause()
       }
       return .success
     }
     remoteCommand.togglePlayPauseCommand.addTarget { _ in
-      if let action = PlayerCore.keyBindings["PLAYPAUSE"]?.rawAction {
-        _ = PlayerCore.lastActive.mpv.command(rawString: action)
+      if let action = PlayerCore.keyBindings["PLAYPAUSE"] {
+        PlayerCore.lastActive.mainWindow.handleKeyBinding(action)
       } else {
         PlayerCore.lastActive.togglePause()
       }
       return .success
     }
     remoteCommand.stopCommand.addTarget { _ in
-      if let action = PlayerCore.keyBindings["STOP"]?.rawAction {
-        _ = PlayerCore.lastActive.mpv.command(rawString: action)
+      if let action = PlayerCore.keyBindings["STOP"] {
+        PlayerCore.lastActive.mainWindow.handleKeyBinding(action)
       } else {
         PlayerCore.lastActive.stop()
       }
       return .success
     }
     remoteCommand.nextTrackCommand.addTarget { _ in
-      if let action = PlayerCore.keyBindings["NEXT"]?.rawAction {
-        _ = PlayerCore.lastActive.mpv.command(rawString: action)
+      if let action = PlayerCore.keyBindings["NEXT"] {
+        PlayerCore.lastActive.mainWindow.handleKeyBinding(action)
       } else {
         PlayerCore.lastActive.navigateInPlaylist(nextMedia: true)
       }
       return .success
     }
     remoteCommand.previousTrackCommand.addTarget { _ in
-      if let action = PlayerCore.keyBindings["PREV"]?.rawAction {
-        _ = PlayerCore.lastActive.mpv.command(rawString: action)
+      if let action = PlayerCore.keyBindings["PREV"] {
+        PlayerCore.lastActive.mainWindow.handleKeyBinding(action)
       } else {
         PlayerCore.lastActive.navigateInPlaylist(nextMedia: false)
       }
@@ -1331,8 +1331,8 @@ class RemoteCommandController {
     }
     remoteCommand.skipForwardCommand.preferredIntervals = [15]
     remoteCommand.skipForwardCommand.addTarget { event in
-      if let action = PlayerCore.keyBindings["FORWARD"]?.rawAction {
-        _ = PlayerCore.lastActive.mpv.command(rawString: action)
+      if let action = PlayerCore.keyBindings["FORWARD"] {
+        PlayerCore.lastActive.mainWindow.handleKeyBinding(action)
       } else {
         PlayerCore.lastActive.seek(relativeSecond: (event as! MPSkipIntervalCommandEvent).interval, option: .exact)
       }
@@ -1340,8 +1340,8 @@ class RemoteCommandController {
     }
     remoteCommand.skipBackwardCommand.preferredIntervals = [15]
     remoteCommand.skipBackwardCommand.addTarget { event in
-      if let action = PlayerCore.keyBindings["REWIND"]?.rawAction {
-        _ = PlayerCore.lastActive.mpv.command(rawString: action)
+      if let action = PlayerCore.keyBindings["REWIND"] {
+        PlayerCore.lastActive.mainWindow.handleKeyBinding(action)
       } else {
         PlayerCore.lastActive.seek(relativeSecond: -(event as! MPSkipIntervalCommandEvent).interval, option: .exact)
       }
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -1269,6 +1269,65 @@ class RemoteCommandController {
 
   private var isEnabled = false
 
+  /// Returns the value to use for the [preferredIntervals](https://developer.apple.com/documentation/mediaplayer/mpskipintervalcommand/preferredintervals) property.
+  ///
+  /// The [MPRemoteCommandCenter](https://developer.apple.com/documentation/MediaPlayer/MPRemoteCommandCenter)
+  /// expects the media keys tied to the  [seekBackwardCommand](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/seekbackwardcommand) and the [seekForwardCommand](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/seekforwardcommand) to seek backward and
+  /// forward in the current media track. The
+  /// [MPSkipIntervalCommand](https://developer.apple.com/documentation/mediaplayer/mpskipintervalcommand)
+  /// property [preferredIntervals](https://developer.apple.com/documentation/mediaplayer/mpskipintervalcommand/preferredintervals) provides the number of
+  /// seconds pressing the key will skip.
+  ///
+  /// IINA allows the user to bind a mpv command to the `FORWARD` and `REWIND` media keys. This method must:
+  /// - Determine if there is a key binding for the given key and if not return the default of 15 seconds
+  /// - Determine if the key is bound to an IINA command and if so return an empy array indicating the property is not applicable
+  /// - Determine if the key is bound to the mpv
+  ///     [seek](https://mpv.io/manual/stable/#command-interface-seek-%3Ctarget%3E-[%3Cflags%3E]) command
+  ///     and if not, return an empty array
+  /// - Parse the `target` value of the `seek` command as an integer, if it cannot be parsed log an error and  return an empty array
+  /// - If present, parse the `seek` command flags and if any flags other than `exact`, `keyframes` and `relative` are
+  ///     present then return an empty array as this is not a normal seek
+  /// - When all the above checks pass the key has been bound to a normal seek command and the absolute value of the seek
+  ///     command target parameter can be used as the interval
+  ///
+  /// To see the `preferredIntervals` value open
+  /// [Control Center](https://support.apple.com/guide/mac-help/quickly-change-settings-mchl50f94f8f/mac)
+  /// and double click on the Now Playing module with IINA playing media. The expanded Now Playing module will contain seek
+  /// backward and seek forward buttons. The interval may be shown inside the button icons.
+  /// - Parameter key: Media key the value is for.
+  /// - Returns: Value to use for` preferredIntervals`.
+  private func formPreferredIntervalsValue(_ key: String) -> [NSNumber] {
+    guard let keyBinding = PlayerCore.keyBindings[key] else { return [15] }
+    guard !keyBinding.isIINACommand else { return [] }
+    let action = keyBinding.action
+    guard action.count > 1, action[0] == MPVCommand.seek.rawValue else { return [] }
+    guard let target = Int(action[1]) else {
+      log("""
+          Unable to parse seek target as an Integer in key binding:
+              \(key) \(keyBinding.rawAction)
+          """, level: .error)
+      return []
+    }
+    if action.count > 2 {
+      let validFlags: Set<String> = ["exact", "keyframes", "relative"]
+      // Multiple flags can be composed using `+`, and each one must be valid
+      let flags = action[2].split(separator: "+")
+
+      guard flags.allSatisfy({ validFlags.contains(String($0)) }) else {
+        log("""
+            Unable to parse seek flag as one of "exact", "keyframes", or "relative":
+                \(key) \(keyBinding.rawAction)
+            """, level: .error)
+        return []
+      }
+    }
+    // The seek command target may be negative to indicate seeking backwards, however the remote
+    // command dictates the direction and requires that the interval to be positive.
+    let seconds = abs(target)
+    log("Seek interval for the \(key) key is \(seconds) s")
+    return [NSNumber(value: seconds)]
+  }
+
   func disable() {
     guard isEnabled else { return }
     commands.forEach { $0.removeTarget(nil) }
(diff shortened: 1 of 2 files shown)


iina/AppDelegate.swift
```diff
@@ -1312,16 +1356,6 @@ class RemoteCommandController {
...
-    remoteCommand.skipForwardCommand.preferredIntervals = [15]
-    remoteCommand.skipForwardCommand.addTarget { event in
-      PlayerCore.lastActive.seek(relativeSecond: (event as! MPSkipIntervalCommandEvent).interval, option: .exact)
-      return .success
-    }
-    remoteCommand.skipBackwardCommand.preferredIntervals = [15]
-    remoteCommand.skipBackwardCommand.addTarget { event in
-      PlayerCore.lastActive.seek(relativeSecond: -(event as! MPSkipIntervalCommandEvent).interval, option: .exact)
-      return .success
-    }
```
low-batt, line 1324: Looks like this code was moved to be in front of the code for `changeRepeatModeCommand`?

low-batt:
Hello from a _junior_ IINA developer. Thank you for preparing this PR!

I'll try and provide some feedback, but I usually leave this area of IINA to others. I do have a few concerns…

I'm expecting this odd assignment:
```swift
_ = PlayerCore.lastActive.mpv.command(rawString: action)
```

Is to avoid a "Result of call is unused" compilation warning. The better way to handle this is to apply the [discardableResult](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/attributes/#discardableResult) attribute to the method being called. That tells the compiler that is it ok for a caller to not use the value returned. _However_ in this case that attribute has not been applied to the `command` method as we don't want callers to ignore the error code.

When configuring key bindings the user is allow to directly enter text. At that time there is no validation that they have entered a valid mpv command. They will press the key and nothing happens. Possibly IINA should put up an alert when that happens, but that is not the current convention when a command fails. Right now IINA just logs the error as seen here from `PlayerWindowController.handleKeyBindings`:
https://github.com/iina/iina/blob/bd16c8412570ed5332530fd158dcb8de2464880a/iina/PlayerWindowController.swift#L293

If you look through the method `handleKeyBindings` you will see another concern. There are a couple of commands that we don't want to pass directly to mpv and require special handling by IINA.

The code to handle stop:
```swift
    remoteCommand.stopCommand.addTarget { _ in
      if let action = PlayerCore.keyBindings["STOP"]?.rawAction {
        _ = PlayerCore.lastActive.mpv.command(rawString: action)
      } else {
        PlayerCore.lastActive.stop()
      }
      return .success
    }
```

Will call `PlayerCore.stop` if no key binding is found. Eventually that method sends a `stop` command to mpv:
https://github.com/iina/iina/blob/bd16c8412570ed5332530fd158dcb8de2464880a/iina/PlayerCore.swift#L893

But if you look at that method you will see there is a lot of other stuff that IINA needs to do when stopping playback.

**NOTICE** we just found a defect in the existing code. The method `PlayerWindowController.handleKeyBindings` should also be checking for the `stop` command and calling the `PlayerCore` method.

Have a look at the various default key bindings configuration files. Notice `STOP` is bound to `quit`, **NOT** `stop`!

This is unfortunate. The key bindings were copied from mpv. Maybe because mpv only supports one window, they consider `STOP` to be `quit`? I guess as IINA supports playing multiple videos IINA wanted `STOP` to mean just stop playing the current video. This means we found another existing defect. The default bindings need to be corrected.

But it gets worse. Users may have copied the defaults to create their own bindings. With this change IINA will now use the command in the bindings. That means the behavior of the `STOP` button will change for such users. Does this mean we should migrate/correct user's bindings? Messy. Need the senior developers to weigh in on that.

Those are my quick thoughts.

@svobs knows way more about this area of IINA. Hopefully he will comment on this PR.

benniekiss:
> If you look through the method handleKeyBindings you will see another concern. There are a couple of commands that we don't want to pass directly to mpv and require special handling by IINA.

Thank you for pointing out this method! I'll take a look through it and start thinking about ways to hook the system media keys into the logic.

iina/AppDelegate.swift
benniekiss: Yes, exactly. I reorganized the commands just to put all of the configurable ones next to each other

benniekiss:
So it looks like the `PlayerWindowController` class is accessible on `PlayerCore` via the `lastActive.mainWindow` attribute, which  would provide access to the `handleKeyBindings` for system media keys. Does this sound like the right way to move forward?

iina/AppDelegate.swift
low-batt: Sounds good.

iina/AppDelegate.swift
```diff
@@ -1279,28 +1279,72 @@ class RemoteCommandController {
...
     remoteCommand.previousTrackCommand.addTarget { _ in
-      PlayerCore.lastActive.navigateInPlaylist(nextMedia: false)
+      if let action = PlayerCore.keyBindings["PREV"] {
+        PlayerCore.lastActive.mainWindow.handleKeyBinding(action)
+      } else {
+        PlayerCore.lastActive.navigateInPlaylist(nextMedia: false)
+      }
+      return .success
+    }
+    remoteCommand.skipForwardCommand.preferredIntervals = [15]
```
low-batt: This line of code assumes this key will seek forward by 15 seconds. With IINA playing open [Control Center](https://support.apple.com/guide/mac-help/quickly-change-settings-mchl50f94f8f/mac) and double click on the Now Playing module. In macOS Sequoia this will give you an expanded Now Playing module where you can see that the `FORWARD` button moves `15` seconds ahead:

Image

As we are giving the user the ability to change what this key does, the number of seconds has to match what the user configured, assuming the user configured this key to seek. If they configured it do to something else then we should not set this.

This means we have to parse the action, see if it is the [seek](https://mpv.io/manual/stable/#command-interface-seek-%3Ctarget%3E-[%3Cflags%3E]) command. If it is the `seek` command then we have to check if the user added flags. If the user used any flags other than these:
- relative
- keyframes
- exact

Then we don't set [preferredIntervals](https://developer.apple.com/documentation/mediaplayer/mpskipintervalcommand/preferredintervals).

If those checks pass then we have to parse the number of seconds and use that for the value of `preferredIntervals`.

This also means there is another error in the default key binding files:
https://github.com/iina/iina/blob/bd16c8412570ed5332530fd158dcb8de2464880a/iina/config/iina-default-input.conf#L113-L115

Not only is `STOP` bound to `quit` instead of `stop`, but the `FORWARD` and `REWIND` keys are bound to a 60 second seek, not 15. If we don't change these assignments the default behavior of IINA will be changed by this PR. This is true of all of the key binding configuration files.

iina/AppDelegate.swift
```diff
@@ -1279,28 +1279,72 @@ class RemoteCommandController {
...
+    remoteCommand.skipForwardCommand.addTarget { event in
+      if let action = PlayerCore.keyBindings["FORWARD"] {
+        PlayerCore.lastActive.mainWindow.handleKeyBinding(action)
+      } else {
+        PlayerCore.lastActive.seek(relativeSecond: (event as! MPSkipIntervalCommandEvent).interval, option: .exact)
+      }
+      return .success
+    }
+    remoteCommand.skipBackwardCommand.preferredIntervals = [15]
```
low-batt: And of course this statement also has the problem discussed above.

low-batt (changes requested):
I had time to do more testing and found a problem I should have noticed the last time. Sorry I missed it!

low-batt:
Thank you for tracking down the issues this PR addresses. Very helpful! I've attached this PR to those issues.

On this:
> Does this sound like the right way to move forward?

Yes, I believe so, although we really need to hear from @svobs and @lhc70000 who normally work on this code.

The next IINA release will be 1.4.2 which is focused on fixing defects. I'm thinking this is a small enough enhancement that it should be included in that release. You can't see it, but I've added it to the 1.4.2 project. The decision on what actually makes it into a release is made by the senior developers. Of course first the problems I noted above need to be addressed.

iina/AppDelegate.swift
benniekiss: Hm, this does seem quite intrusive. 

Other than adding in the helpers to parse/sanitize the command, I would be open to dropping the configuration for these keys, and only apply the custom keybindings for `nextTrackCommand` (`NEXT`),  `previousTrackCommand`, (`PREV`), and `togglePlayPauseCommand` (`PLAYPAUSE`), as those are the bindings mapping to the keyboard media keys F7, F8, and F9, and that is what my main motivation is -- both for useability, and so that I can stop accidentally skipping to the next media item in a playlist when I mean to press pause.

low-batt: I'm thinking the senior developers might reject this PR if it only partially provides support for media keys. Partial support could cause users to report something is broken when they try and override the keys for which support is missing.

I did notice you mentioned this is your first attempt at writing Swift. I did not comment on that because you are doing a great job. Do not hesitate to ask me if you need help with Swift coding.

As frequently seems the case with software the task at hand that seemed easy has proven to require more work to handle all the possible scenarios. Here is a start on what I think is needed:

```swift
  /// Returns the value to use for the [preferredIntervals](https://developer.apple.com/documentation/mediaplayer/mpskipintervalcommand/preferredintervals) property.
  ///
  /// The [MPRemoteCommandCenter](https://developer.apple.com/documentation/MediaPlayer/MPRemoteCommandCenter)
  /// expects the media keys tied to the  [seekBackwardCommand](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/seekbackwardcommand) and the [seekForwardCommand](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/seekforwardcommand) to seek backward and
  /// forward in the current media track. The
  /// [MPSkipIntervalCommand](https://developer.apple.com/documentation/mediaplayer/mpskipintervalcommand)
  /// property [preferredIntervals](https://developer.apple.com/documentation/mediaplayer/mpskipintervalcommand/preferredintervals) provides the number of
  /// seconds pressing the key will skip.
  ///
  /// IINA allows the user to bind a mpv command to the `FORWARD` and `REWIND` media keys. This method must:
  /// - Determine if there is a key binding for the given key and if not return the default of 15 seconds
  /// - Determine if the key is bound to an IINA command and if so return an empy array indicating the property is not applicable
  /// - Determine if the key is bound to the mpv
  ///     [seek](https://mpv.io/manual/stable/#command-interface-seek-%3Ctarget%3E-[%3Cflags%3E]) command
  ///     and if not, return an empty array
  /// - Parse the `target` value of the `seek` command as an integer, if it cannot be parsed log an error and  return an empty array
  /// - If present, parse the `seek` command flags and if any flags other than `exact`, `keyframes` and `relative` are
  ///     present then return an empty array as this is not a normal seek
  /// - When all the above checks pass the key has been bound to a normal seek command and the absolute value of the seek
  ///     command target parameter can be used as the interval
  ///
  /// To see the `preferredIntervals` value open
  /// [Control Center](https://support.apple.com/guide/mac-help/quickly-change-settings-mchl50f94f8f/mac)
  /// and double click on the Now Playing module with IINA playing media. The expanded Now Playing module will contain seek
  /// backward and seek forward buttons. If the interval is between 10 and 90 it will be shown inside the button icons,
  /// - Parameter key: Media key the value is for.
  /// - Returns: Value to use for` preferredIntervals`.
  private func formPreferredIntervalsValue(_ key: String) -> [NSNumber] {
    guard let keyBinding = PlayerCore.keyBindings[key] else { return [15] }
    guard !keyBinding.isIINACommand else { return [] }
    let action = keyBinding.action
    guard action.count > 1, action[0] == MPVCommand.seek.rawValue else { return [] }
    guard let target = Int(action[1]) else {
      log("""
          Unable to parse seek target as an Integer in key binding:
              \(key) \(keyBinding.rawAction)
          """, level: .error)
      return []
    }
    if action.count > 2 {
      // TODO! Flags are present, confirm only exact, keyframes and relative are being used.
    }
    // The seek command target may be negative to indicate seeking backwards, however the remote
    // command dictates the direction and requires that the interval to be positive.
    let seconds = abs(target)
    log("Seek interval for the \(key) key is \(seconds) s")
    return [NSNumber(value: seconds)]
  }
```

Notice I left checking to see if the user included flags in the [seek](https://mpv.io/manual/stable/#command-interface-seek-%3Ctarget%3E-[%3Cflags%3E]) command for you to implement. Follow that link to the mpv manual to see the flags that might be present. This will require you to learn about the Swift [String](https://developer.apple.com/documentation/swift/string) structure. It has some "interesting" behavior.

There is one more thing that is needed and that is dynamically changing the interval if the user edits the key bindings. I will look into what is needed for that.

benniekiss: > Partial support could cause users to report something is broken when they try and override the keys for which support is missing.

This makes total sense to me!

> As frequently seems the case with software the task at hand that seemed easy has proven to require more work to handle all the possible scenarios. Here is a start on what I think is needed:

Thank you so much! I will finish this snippet out and test on my end. I'll be sure to let you know if I'm stuck on anything in particular. 

> There is one more thing that is needed and that is dynamically changing the interval if the user edits the key bindings. I will look into what is needed for that.

I'll see what I can find, as well. I've been meaning to get familiar with Apple docs and swift anyway.

low-batt: Good luck! I'll be around if you have Swift questions.

I looked into dynamically handling changes to key bindings yesterday. Add the following code to the end of the `init` method:

```swift
    NotificationCenter.default.addObserver(forName: .iinaGlobalKeyBindingsChanged, object: nil,
                                           queue: .main) { [unowned self] _ in
      guard isEnabled else { return }
      // The user has modified the key bindings, possibly changing the mpv commands associated with
      // the FORWARD and REWIND media keys. The preferredIntervals values are set based on the mpv
      // commands assigned to those keys. When expanded the macOS Now Playing module may display
      // the interval in the seek backward and seek forward buttons. The user may have changed the
      // number of seconds to seek causing the value displayed in the Now Playing module buttons to
      // to be out of date. Merely updating the preferredIntervals values is insufficient to get the
      // Now Playing module to update its buttons. Support for media keys and remote commands must
      // be disabled and then re-enabled in a separate task.
      log("Restarting support for remote commands due to changes to key bindings")
      disable()
      DispatchQueue.main.async { self.enable() }
    }
```

I tested this by changing the seek interval between `15` and `60` and then checking the Now Playing module's buttons to see that the displayed value reflected the current bindings. I was testing under macOS Sequoia. Are you working under macOS Tahoe? I know Now Playing has been updated for macOS 26. Need to make sure all this is working under the latest macOS release.

This comment in the code I gave you is defective!:
```swift
  /// backward and seek forward buttons. If the interval is between 10 and 90 it will be shown inside the button icons,
```

Besides the typo of using a coma instead of a period, after more testing the statement proved to not be true, at least under macOS Sequoia. Not clear to me why Now Playing sometimes shows the interval and sometimes does not. For me `15` and `60` are shown, but `20` is not. So replace the above text with this:
```swift
   /// backward and seek forward buttons. The interval may be shown inside the button icons.
```

I think we are close to a fully working implementation.

low-batt: Don't for get we need to deal with the default key bindings files.

The bindings in question originate from mpv. See [this file](https://github.com/mpv-player/mpv/blob/af9c81fa1ecbd0d3d7b6b946a7eaea42f1986e63/etc/input.conf#L163-L165).

As seen here those same bindings are in the [iina-default-input.conf](https://github.com/iina/iina/blob/develop/iina/config/iina-default-input.conf) file:
https://github.com/iina/iina/blob/bd16c8412570ed5332530fd158dcb8de2464880a/iina/config/iina-default-input.conf#L113-L115

Those bindings are currently ignored, due to the missing support for them and do not match IINA's current behavior, which is:

| Key | Current Binding | Current Behavior |
| --- | --- | -- |
| STOP | quit | stop |
| FORWARD | seek 60 | seek 15 |
| REWIND | seek -60 | seek -15 |

If we don't update `iina-default-input.conf` media key behavior for users using the `IINA Default` key bindings would change. I think those 3 bindings should be updated to match IINA's current behavior.

I don't think we should change the `mpv Default` key bindings in the [input.conf](https://github.com/iina/iina/blob/develop/iina/config/input.conf) file. I'm thinking that even though seeking will go from 15 s to 60 s that should be considered a bug fix. The `mpv Default` key bindings will be more aligned with mpv player behavior. If users don't like this they can always duplicate and modify the key bindings.

I tested [VLC](https://www.videolan.org/vlc/) and it also seeks 60 s. So I would not change [vlc-default-input.conf](https://github.com/iina/iina/blob/develop/iina/config/vlc-default-input.conf).

I don't know anything about [Movist](https://movistprime.com/faqs.html).

I'm thinking we only update the [iina-default-input.conf](https://github.com/iina/iina/blob/develop/iina/config/iina-default-input.conf) file to preserve the current IINA behavior.

When it comes to users who duplicated one of the default key bindings and customized it I think we note the change in behavior in the release notes and leave it to the users to update their customized bindings.

benniekiss:
So this is currently working quite well except for one issue  I noticed on macOS 26. The commit message above outlines this as well, but I'll repeat here: 

The control center widget only shows the initial interval values, and updating the keybindings does not changes those that are displayed. The player window has to be closed and re-opened in order for the new values to be visible. However, when clicking on the widget to expand the Now Playing  section, the correct intervals _are_ displayed, and these values do update dynamically when the keybindings are changed.

I'm not sure where the widget values are sourced from, but I'll do some more digging.

iina/AppDelegate.swift
```diff
@@ -1269,6 +1269,65 @@ class RemoteCommandController {
...
+  /// The [MPRemoteCommandCenter](https://developer.apple.com/documentation/MediaPlayer/MPRemoteCommandCenter)
+  /// expects the media keys tied to the  [seekBackwardCommand](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/seekbackwardcommand) and the [seekForwardCommand](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/seekforwardcommand) to seek backward and
+  /// forward in the current media track. The
+  /// [MPSkipIntervalCommand](https://developer.apple.com/documentation/mediaplayer/mpskipintervalcommand)
+  /// property [preferredIntervals](https://developer.apple.com/documentation/mediaplayer/mpskipintervalcommand/preferredintervals) provides the number of
+  /// seconds pressing the key will skip.
+  ///
+  /// IINA allows the user to bind a mpv command to the `FORWARD` and `REWIND` media keys. This method must:
+  /// - Determine if there is a key binding for the given key and if not return the default of 15 seconds
+  /// - Determine if the key is bound to an IINA command and if so return an empy array indicating the property is not applicable
```
low-batt: The code I gave you has a typo in "return an empy array". Sorry!

low-batt:
Was busy today. From a quick look at the changes, they look good. I too will test tomorrow.

Glad your eyes are better than me! The button in the Now Playing manager is so small and blurry for me that I did not see that it contains the interval value. It is not updating under macOS Sequoia either. 

That the expanded module correctly uses the updated values and the regular module continues to use a stale value seems rather buggy. Possibly ending the Now Playing session and restarting it will get macOS to notice the changes. I will think on whether that would have any bad side effects.

iina/AppDelegate.swift
```diff
@@ -1269,6 +1269,65 @@ class RemoteCommandController {
...
+  /// and double click on the Now Playing module with IINA playing media. The expanded Now Playing module will contain seek
+  /// backward and seek forward buttons. The interval may be shown inside the button icons.
+  /// - Parameter key: Media key the value is for.
+  /// - Returns: Value to use for` preferredIntervals`.
+  private func formPreferredIntervalsValue(_ key: String) -> [NSNumber] {
+    guard let keyBinding = PlayerCore.keyBindings[key] else { return [15] }
+    guard !keyBinding.isIINACommand else { return [] }
+    let action = keyBinding.action
+    guard action.count > 1, action[0] == MPVCommand.seek.rawValue else { return [] }
+    guard let target = Int(action[1]) else {
```
low-batt: The [seek](https://mpv.io/manual/stable/#command-interface-seek-%3Ctarget%3E-[%3Cflags%3E]) command target does not have to be an Integer, a floating point value can be specified. My bad!
```suggestion
    guard let target = Double(action[1]) else {
```

iina/AppDelegate.swift
```diff
@@ -1269,6 +1269,65 @@ class RemoteCommandController {
...
+      log("""
+          Unable to parse seek target as an Integer in key binding:
```
low-batt:
```suggestion
          Unable to parse seek target as a Double in key binding:
```

iina/AppDelegate.swift
```diff
@@ -1269,6 +1269,65 @@ class RemoteCommandController {
...
+              \(key) \(keyBinding.rawAction)
+          """, level: .error)
+      return []
+    }
+    if action.count > 2 {
+      let validFlags: Set<String> = ["exact", "keyframes", "relative"]
```
low-batt: The name `validFlags` is sort of incorrect.. Maybe something like `intervalFlags`? It is not that the other flags are invalid. It is that they change the meaning of the target argument. It is unlikely users will some of the other flags, but IINA has to be ready for them to do so. The most likely one would be this:
```text
seek 10 relative-percent
```

With that command the `10` does not represent 10 seconds. I believe it represents 10% of the total duration. We don't want to try and handle that case. If that argument does not represent a relative number of seconds to seek, we do not want to set `preferredIntervals`.

iina/AppDelegate.swift
```diff
@@ -1269,6 +1269,65 @@ class RemoteCommandController {
...
+      // Multiple flags can be composed using `+`, and each one must be valid
+      let flags = action[2].split(separator: "+")
+
+      guard flags.allSatisfy({ validFlags.contains(String($0)) }) else {
+        log("""
+            Unable to parse seek flag as one of "exact", "keyframes", or "relative":
```
low-batt: Unlike the previous log message when the second argument can't be parsed, this message is about finding a flag that changes the second argument to not represent a relative number of seconds to seek. Need something more like:
```text
Seek command uses a flag other than "exact", "keyframes", or "relative":
```

iina/AppDelegate.swift
```diff
@@ -1269,6 +1269,65 @@ class RemoteCommandController {
...
+                \(key) \(keyBinding.rawAction)
+            """, level: .error)
```
low-batt: Need to remove `, level: .error` to make this just be a debug message. It is not an error for the user to use other flags, it just means IINA decided to not set `preferredIntervals` and this message tells us why.

iina/AppDelegate.swift
```diff
@@ -1351,6 +1444,22 @@ class RemoteCommandController {
...
+      // the FORWARD and REWIND media keys. The preferredIntervals values are set based on the mpv
+      // commands assigned to those keys. When expanded the macOS Now Playing module may display
+      // the interval in the seek backward and seek forward buttons. The user may have changed the
+      // number of seconds to seek causing the value displayed in the Now Playing module buttons to
+      // to be out of date. Merely updating the preferredIntervals values is insufficient to get the
+      // Now Playing module to update its buttons. Support for media keys and remote commands must
+      // be disabled and then re-enabled in a separate task.
+      log("Restarting support for remote commands due to changes to key bindings")
+      disable()
+      DispatchQueue.main.async { self.enable() }
```
low-batt: Replacing with this worked for me:
```suggestion
      // commands assigned to those keys. The Now Playing module may display the interval in the
      // seek backward and seek forward buttons causing the value displayed in the Now Playing
      // module buttons to to be out of date. Merely updating the preferredIntervals values is
      // insufficient to get the Now Playing module to update its buttons. Support for media keys
      // and remote commands must be disabled and then re-enabled.
      log("Restarting support for remote commands due to changes to key bindings")
      disable()
      // Immediately re-enabling media keys and remote commands only partially worked. The Now
      // Playing module would update buttons if the module was expanded, but the buttons shown in
      // the module's small form would still display the old interval. Work around this curious
      // Now Playing behavior by delaying the re-enabling.
      DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + 1) { self.enable() }
```

low-batt (changes requested):
I believe with the changes I'm suggesting this PR will be good to go. At least it passed the testing I did today.

iina/AppDelegate.swift
benniekiss: What about `allowedFlags`? That conveys the purpose

iina/AppDelegate.swift
benniekiss: What are your thoughts on adding `not setting seek interval` to the log message?

`Seek command uses a flag other than "exact", "keyframes", or "relative", not setting seek interval`:

benniekiss:
Other than my last two questions about the `validFlags` var name, and the log message, everything else is addressed. I also tested and can confirm your suggested changes fix the control center widget bug

iina/AppDelegate.swift
low-batt: Perfect

iina/AppDelegate.swift
low-batt: Perfect

iina/AppDelegate.swift
```diff
@@ -1269,6 +1269,65 @@ class RemoteCommandController {
...
+    if action.count > 2 {
+      let allowedFlags: Set<String> = ["exact", "keyframes", "relative"]
+      // Multiple flags can be composed using `+`, and each one must be valid
+      let flags = action[2].split(separator: "+")
+
+      guard flags.allSatisfy({ allowedFlags.contains(String($0)) }) else {
+        log("""
+            Seek flag was not one of \(allowedFlags.map({ "'\($0)'" }).joined(separator: ", ")), not setting seek interval:
+                \(key) \(keyBinding.rawAction)
+            """
```
low-batt: Lost the trailing `)`
```suggestion
            """)
```

low-batt (changes requested):
Just need to fix the compilation error and I think we are good to go.

low-batt:
The new code looks great. Just need to add the missing `)` I noted to fix the build error.

As some users will experience changes in behavior release notes will be needed. See if the text below is coherent.

----

## Improvements

- Support rebindable media keys #1932, #2502, #3512, #4631

  In IINA's settings on the `Key Bindings` tab the key names shown in the table below are tied to  commands in the macOS [Remote Command Center](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter). These commands can be triggered using the keyboard keys and the buttons of the  Now Playing [Control Center](https://support.apple.com/guide/mac-help/quickly-change-settings-mchl50f94f8f/mac) module shown in the table.
  | Key Name | Remote Command | Keyboard | Now Playing |
  | --- | --- | --- | --- |
  | PREV | [Previous track](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/previoustrackcommand) | ⏮ | |
  | PLAYPAUSE | [Toggle play/pause](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/toggleplaypausecommand) | ⏯ | |
  | NEXT | [Next track](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/nexttrackcommand) | ⏭ | |
  | REWIND | [Seek backward](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/seekbackwardcommand) | | ↺ |
  | FORWARD | [Seek forward](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/seekforwardcommand) | | ↻ |
  | PAUSE | [Pause](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/pausecommand)| | ⏸ |
  | PLAY | [Play](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/playcommand) | | ▶ |
  | STOP | [Stop](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/stopcommand) | | |

  **Change in Behavior**

  In previous versions of IINA the mpv commands these keys were bound to were ignored. IINA now uses the mpv commands in the key bindings. Users can now alter the behavior of the media keys as well as other remote commands.
  
  However some of IINA the bindings contained in the `IINA Default` bindings did not match IINA's behavior. To avoid a change in behavior the following keys in the those bindings have been updated:
  
  | Key Name | Previous Binding | Updated Binding |
  | --- | --- | -- |
  | STOP | quit | stop |
  | FORWARD | seek 60 | seek 15 |
  | REWIND | seek -60 | seek -15 |
  
    Users using the `IINA Default` bindings will not experience a change in behavior. But users who duplicated a configuration will experience a change in behavior. To restore the previous behavior users will need to edit their configuration to match the current bindings shown above.
	
	The `mpv Default` and `VLC Default` bindings have not been updated as the new behavior correctly matches the behavior of those players. Therefore the change in behavior is considered a fix. To restore the previous behavior users can duplicate the binding file and then edit the configuration.

iina/AppDelegate.swift
benniekiss: Fixed! I must've removed this just before committing

benniekiss:
Your release note looks great to me! I especially like the keybinding table

iina/AppDelegate.swift
low-batt: Looks good.

low-batt (approved):
I reviewed the changes in detail, pulled changes, built IINA and tested.

Looks good to me!

lhc70000:
Looks great, thanks @benniekiss and @low-batt!

benniekiss:
Wonderful! Pleasure working with you, @low-batt

PR #5867, opened 2025-12-31
Add time-synced lyrics overlay for main window and mini player

ef-molina:
- [x] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)
- [ ] This implements/fixes issue # (no existing issue)

---

## Description

This PR adds **time-synced lyrics (.lrc) support** to IINA with a Spotify-style overlay that works in both the **Main Window** and **Mini Player**.

The feature is fully additive and does not change existing playback, mpv integration, or UI behavior when lyrics are disabled.

---

## Feature Overview

- Displays synced lyrics over:
  - Main window (video / album art)
  - Mini player (music mode)
- Live updates based on playback time
- Shared lyrics visibility toggle across:
  - Main window
  - Mini player
  - Quick Settings
- Automatically hides when:
  - No lyrics are available
  - Lyrics are toggled off
- Restores lyrics immediately when toggled back on
- Adapts visible line count dynamically based on window size

---

## Architecture & Data Flow

**High-level flow:**

PlayerCore (playback time)
- LyricsController.syncTime(_:)
- LyricsEngine (pure logic)
- LyricsOverlayState
- NotificationCenter
- MainWindowController / MiniPlayerWindowController
- LyricsOverlayView (UI only)

### Key design points

- **LyricsEngine**
  - Pure, deterministic logic
  - Binary search for current line
  - Computes surrounding lyric window
  - No UI or PlayerCore dependencies

- **LyricsController**
  - Single source of truth for lyrics state
  - Receives playback time from PlayerCore
  - Emits notification updates only when the current line changes
  - Separates lyrics visibility from lyrics lifecycle (engine is preserved when toggling visibility)

- **LyricsOverlayView**
  - UI-only component
  - No knowledge of playback or notifications
  - Handles layout, font scaling, opacity hierarchy, and animations

- **Window Controllers**
  - Observe lyrics notifications
  - Control visibility
  - Forward state to views
  - Inject dynamic visible-line count via provider closure (no UI ↔ engine coupling)

- **LRCParser**
  - Parses external `.lrc` files only
  - Supports multi-timestamp lines
  - Ignores malformed entries
  - Does not assume lyric source priority

---

## Bug Fixes Included

- Fixed lyrics not reappearing after toggling OFF → ON (engine is no longer destroyed on visibility changes)
- Ensured Mini Player uses the same dynamic visible-line calculation as the Main Window
- Correctly clears lyrics when switching to tracks without `.lrc` files

[...]

PR #5944, opened 2026-03-06
Support External Subtitle Rendering for Audio-Only Files

### LINKED ISSUE #457: 播放mp3时，加载的lrc字幕不显示
**System and IINA version:**

- macOS 10.12.2
- IINA 0.0.6

**Expected behavior:**
播放mp3时，显示lrc字幕

**Actual behavior:**
播放mp3时，加载的lrc字幕不显示

(N/A for feature requests)
播放mp3时，显示lrc字幕

### LINKED ISSUE #972: 播放音频不支持字幕文件
**System and IINA version:**

- macOS 10.12.6
- IINA 0.0.12 build38

已知处理有完整ID3（带封面）的MP4外，其它各种音频格式均不支持字幕文件。已测试有此问题的有flac、ogg、wav等格式，字幕文件为srt、ass

（iina本身不支持lrc

### LINKED ISSUE #5618: 纯音频文件播放时，字幕不显示 | Subtitles for media files without an album cover won't display
### IINA Version

1.4.0-beta2 Build 154

### Release Channel

Nightly

### macOS Version

macOS Sequoia 15.5

### Steps to Reproduce

- Open an .mp3 media file (without an album cover) with its corresponding subtitle file(s) (.lrc, .vtt, etc.) in the same folder. 
- If subtitle not automatically loaded, load it.

### Expected Behavior

Subtitles displays correctly (at least in video mode; would be better if it also works in music mode)

An example of Potplayer (for Windows only), which handles it quite well:

Image

### Actual Behavior

Subtitles won't display in either video mode or music mode:

Image

Image

_No response_

### Logs

_No response_

### Sample Files

https://www.mediafire.com/file/yvfrsttncigxu6j/Sample.zip/file

Every time

### Required Actions

### Optional Actions

yorkyang2333:
- [x] This implements/fixes issue #457, #972, #5618.

---

**Description:**

This PR enables subtitle rendering for all audio-only files (MP3, FLAC, WAV, etc.) and their associated external subtitle files (LRC, SRT, ASS, etc.) by addressing renderer initialization and UI synchronization issues.

### The Problem
When using `vo=libmpv`, the core renderer disables updates if no video track is present. For audio-only media, this prevents subtitle display and stops mpv's internal render clock, which in turn causes IINA's UI synchronization (progress bar and time displays) to freeze during window mode switches.

#### 1. Synthetic Video Track
A persistent synthetic video track is injected via `lavfi-complex` (`color=c=black:s=640x360:r=1[vo]`) when playing audio files with subtitles. This ensures a continuous render clock, preventing UI freezes. The filter is designed to be video-only, preserving original audio tracks and volume controls without any audio hijacking.

#### 2. Native Subtitle Overlay in Music Mode
To provide a clean experience in the mini player without obscuring album art, we implemented a native `NSTextField` overlay:
- The overlay is updated by polling mpv's `sub-text` property during the UI sync loop.
- A 0.1s debounce delay is applied to the hiding logic to prevent flickering during rapid seek operations or buffer flushes.
- The synthetic video track is layered behind the album art to remain invisible.

#### 3. Aspect Ratio and Window Management
Updated `handleVideoSizeChange` to enforce a 1:1 aspect ratio for audio-only files, ensuring the mini player maintains its intended dimensions regardless of the 16:9 synthetic track.

#### 4. Automatic File Matching
Modified `AutoFileMatcher.swift` to include `audio` media types when searching for matching subtitle files, allowing external subtitle files to load automatically for associated audio tracks with matching basenames.

### Verification
- [x] Automatic subtitle loading for various audio/subtitle format pairs (e.g., MP3+LRC, FLAC+SRT).
- [x] Correct rendering in both main window and mini player modes.
- [x] Volume controls and track selection remain functional.
- [x] Responsive and flicker-free subtitle updates during seeking.

### DIFF
--- a/iina/AutoFileMatcher.swift
+++ b/iina/AutoFileMatcher.swift
@@ -26,7 +26,7 @@ class AutoFileMatcher {
   private var subtitles: [FileInfo] = []
   private var subsGroupedBySeries: [String: [FileInfo]] = [:]
   private var unmatchedVideos: [FileInfo] = []
-  
+
   private let subsystem: Logger.Subsystem
 
   private func log(_ message: @autoclosure () -> String, level: Logger.Level = .debug) {
@@ -199,7 +199,8 @@ class AutoFileMatcher {
     let subAutoLoadOption: Preference.IINAAutoLoadAction = Preference.enum(for: .subAutoLoadIINA)
     guard subAutoLoadOption != .disabled else { return }
 
-    for video in filesGroupedByMediaType[.video]! {
+    let currentVideosInfo = filesGroupedByMediaType[.video]! + filesGroupedByMediaType[.audio]!
+    for video in currentVideosInfo {
       var matchedSubs = Set<FileInfo>()
       log("Matching for \(video.filename)")
 
@@ -314,7 +315,7 @@ class AutoFileMatcher {
           if dist < minDistToVideo { minDistToVideo = dist }
         }
         guard minDistToVideo != .max else { continue }
-        sub.minDist = filesGroupedByMediaType[.video]!.filter { sub.dist[$0] == minDistToVideo }
+        sub.minDist = (filesGroupedByMediaType[.video]! + filesGroupedByMediaType[.audio]!).filter { sub.dist[$0] == minDistToVideo }
       }
 
       // match them
@@ -355,7 +356,7 @@ class AutoFileMatcher {
 
       // group video and sub files
       log("Grouping video files...")
-      videosGroupedBySeries = FileGroup.group(files: filesGroupedByMediaType[.video]!).flatten()
+      videosGroupedBySeries = FileGroup.group(files: filesGroupedByMediaType[.video]! + filesGroupedByMediaType[.audio]!).flatten()
       log("Finished with \(videosGroupedBySeries.count) groups")
 
       log("Grouping sub files...")
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -142,6 +142,7 @@ class MPVController: NSObject {
     MPVOption.Subtitles.subPos: MPV_FORMAT_DOUBLE,
     MPVOption.Subtitles.subScale: MPV_FORMAT_DOUBLE,
     MPVOption.Subtitles.subVisibility: MPV_FORMAT_FLAG,
+    MPVProperty.subText: MPV_FORMAT_STRING,
     MPVOption.Equalizer.contrast: MPV_FORMAT_INT64,
     MPVOption.Equalizer.brightness: MPV_FORMAT_INT64,
     MPVOption.Equalizer.gamma: MPV_FORMAT_INT64,
@@ -1402,6 +1403,9 @@ class MPVController: NSObject {
         break
       }
       DispatchQueue.main.async { self.player.subPosChanged(data) }
+
+    case MPVProperty.subText:
+      DispatchQueue.main.async { self.player.subTextChanged() }
 
     case MPVOption.Equalizer.contrast:
       guard let data = UnsafePointer<Int64>(OpaquePointer(property.data))?.pointee else {
--- a/iina/MPVController.swift
+++ b/iina/MPVController.swift
@@ -579,6 +579,10 @@ class MPVController: NSObject {
       }
     }
 
+    // Force mpv window creation. Required to render subtitles/OSD even for audio-only scenarios (e.g. mp3).
+    // Must be set BEFORE mpv_initialize(), since mpv_set_option_string only works pre-init.
+    chkErr(setOptionString(MPVOption.Window.forceWindow, "yes", level: .verbose))
+
     // Set user defined options.
     if Preference.bool(for: .enableAdvancedSettings) {
       if let userOptions = Preference.value(for: .userOptions) as? [[String]] {
--- a/iina/MiniPlayerWindowController.swift
+++ b/iina/MiniPlayerWindowController.swift
@@ -332,12 +332,21 @@ class MiniPlayerWindowController: PlayerWindowController, NSPopoverDelegate {
     guard let window = window else { return }
     let w = player.info.displayWidth, h = player.info.displayHeight
 
-    // For audio-only files (no real video tracks), we want to keep 1:1 aspect
-    // even if a synthetic lavfi-complex track added a video.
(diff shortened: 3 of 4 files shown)


low-batt:
Hello from a _junor_ IINA developer. This is the kind of PR that needs to be reviewed by the senior developers. There is another open PR, #5867, that also adds support for lyrics. It takes a slightly different approach. I can't comment on how the senior developers would like to see this implemented.

Last I heard the next IINA release will be 1.4.2 which is focused on fixing defects. I'm expecting something like this will need to go into IINA 1.5.0 which will include upgrading mpv, plus adding some features.

As the senior developers need to handle this, I will limit my comments, but I will provide a little bit of feedback…

The mpv player supports lyrics. I assume the problem is that support is missing from `vo=libmpv`. Should we be asking mpv for an enhancement?

On this:
> The overlay is updated by polling mpv's sub-text property during the UI sync loop.

I see that code in `PlayerCore.syncUI`. If you right click on the times in the OSC it brings up a context menu that allows you to increase the precision of the displayed times. When you do that the interval between calls to `syncUI` to be reduced. This means the performance of `syncUI` is important and we need to question whenever code is added to it.

IINA already listens for [MPV_EVENT_PROPERTY_CHANGE](https://mpv.io/manual/stable/#command-interface-mpv-event-property-change) events in `MPVController`, usually calling listeners in `PlayerCore`. Couldn't IINA listen for the `sub-text` property changing instead of polling this mpv property?

I did a quick test and ran into a problem.

Try this:
- Play an audio file with lyrics that does not have embedded cover art
- Then open a different audio file with lyrics that has embedded cover art
- The second video plays with lyrics, but without displaying cover art
- mpv emits this warning:

    > 21:04:09.810 [mpv0][w] [cplayer] warn: 
21:04:09.810 [mpv0][w] [cplayer] warn: Audio/Video desynchronisation detected! Possible reasons include too slow
21:04:09.810 [mpv0][w] [cplayer] warn: hardware, temporary CPU spikes, broken drivers, and broken files. Audio
21:04:09.810 [mpv0][w] [cplayer] warn: position will not match to the video (see A-V status field).
21:04:09.810 [mpv0][w] [cplayer] warn: Consider trying `--profile=fast` and/or `--hwdec=auto-safe` as they may help.
21:04:09.810 [mpv0][w] [cplayer] warn: 
- Checking the `Video Track` menu item shows `<None>` selected
- Selecting the thumbnail does not work and mpv emits this error:
    > 21:02:04.850 [mpv0][e] [cplayer] error: Can't switch away from complex filter output.

low-batt:
I pulled the latest commit and repeated the quick test I previously ran. Worked fine. All fixed.

I think we are now waiting for the senior developers. There is usually a flurry of activity once they have time to focus on IINA.

yorkyang2333:
Thanks for the feedback, @low-batt. Regarding the current issue where audio files with "no cover but available subtitles" fail to display lyrics in Video Mode (Main Window), I’d like to provide some context on the methods I attempted and the technical hurdles encountered.

Initially, I tried forcing a dummy black video track via `lavfi-complex` injection, but this consistently triggered severe filter graph locking issues. Specifically, it resulted in the `Can't switch away from complex filter output` error during track transitions and often introduced A/V desynchronization warnings. Even when I tried clearing the filter manually during the `file-ended` event, the race conditions in "Auto-play" scenarios made it impossible to guarantee a smooth transition to subsequent tracks with real covers, sometimes leaving the mpv core in an unstable state.

I also explored overlaying a native `NSTextField` directly on the main video window to handle the text independently. However, this approach significantly complicates the maintenance of the `MainWindowController` view hierarchy. Furthermore, matching the specific rendering characteristics of mpv (such as complex ASS/SSA formatting and dynamic positioning) would require re-implementing significant portions of rendering logic, which felt outside the scope of fixing current regressions.

Given that the priority for this branch is overall stability and a seamless track-switching experience, I decided to pull back on the risky dummy-track logic for now. Subtitle overlays are now functioning reliably in Music Mode, while Video Mode for audio-only files currently adheres to the native mpv constraint requiring a video output for subtitle rendering. If we move to 1.5.0 and gain more flexible filter graph management, we might find a cleaner way to fill this gap.

low-batt:
Thanks very much for that info. I was worried there would be issues with the `lavfi-complex` injection.

The reason for the IINA 1.4.2 release with the old `libmpv` is that it is looking like we will have to drop support for macOS Catalina when we upgrade mpv. Although mpv 0.41.0 does support Catalina, last I heard there were problems building other libraries for Catalina. So we want to fix some serious problems before dropping macOS Catalina.

There are PRs waiting for review for most of the planned work for 1.4.2. I'm still working on a couple of issues and I've left a couple for the senior developers. There is a lot of review work, but I'm hoping that will go quickly.

Upgrading mpv and FFmpeg will fix some serious problems, so I am anxious to move on to IINA 1.5.0. There are some problems in mpv 0.41.0. I am hoping to get patches for from the mpv project.

yorkyang2333:
I've identified and fixed a small layout bug in the Mini Player where the default background placeholder logo would sometimes appear simultaneously with the actual album art. This has been resolved by refining the visibility logic in PlayerCore.swift to ensure the placeholder is hidden whenever an active video track (which includes cover art) is detected.

Screenshot 2026-03-08 at 16 54 03

yorkyang2333:
Merged the fix for UI refresh issues from #5945 into this PR.

This fix addresses a bug where the Mini Player window (music mode) would stop refreshing when a macOS menu bar dropdown was open. By changing the `syncUITimer` in `PlayerCore.swift` to use the `.common` RunLoop mode instead of `.default`, we ensure the UI continues to update smoothly even during menu interaction/event tracking.

yorkyang2333:
PlayerCore: remove unused variable `isAudioOnly`

yorkyang2333:
Superseded by #6030.

I reopened this work on top of the current `develop` branch and replaced the earlier approach with a more stable implementation.

Compared with #5944, the new PR:
- avoids the old `lavfi-complex` / synthetic video direction
- keeps the mini player overlay scoped to music mode only
- keeps the main window path mpv-native instead of using a native AppKit subtitle overlay
- updates subtitle text through property-change events instead of polling from `syncUI`

That rework is the main reason this was resubmitted as a new PR instead of continuing the original branch.

PR #5945, opened 2026-03-07
fix: keep UI timer running during macOS menu tracking

yorkyang2333:

---

**Description:**

This PR fixes an issue where the music mode (MiniPlayer) window stops refreshing when a macOS menu bar dropdown is open.

The root cause was that `syncUITimer` in `PlayerCore.swift` was scheduled only in the `.default` RunLoop mode. Since macOS enters `.eventTracking` mode during menu interaction, the timer would stop firing, leading to a frozen UI.

The fix changes `syncUITimer` to be scheduled in the `.common` RunLoop mode, ensuring it continues to fire and update the UI even when menus are active.


yorkyang2333:
Merged into #5944. Closing this PR.

Issue #5928, opened 2026-02-24
self built IINA w/ embedded libmpv 0.41.0 shows blank video, but standalone mpv 0.41-dev works


comfortablemilk4454:
### IINA Version

1.4.1 Build 160 mpv v0.41.0 FFmpeg 8.0.1  Feb 23, 2026 at 2:38:21 PM develop bd16c84

### Release Channel

Other

### macOS Version

26.4 Beta (25E5218f)

### Steps to Reproduce

play vid

### Expected Behavior

works

### Actual Behavior

audio works but gray screen whole time. only happens when you launch the vid when iina is already running. if you launch iina by clicking on the vid in finder it works. fully works on lastest nightly mpv v0.41.0-dev-gee9a21960 https://nightly.link/mpv-player/mpv/workflows/build/master/mpv-v0.41.0-dev-gee9a21960-22315538242-macos-26-arm.zip, and on latest nightly iina:

1.4.1 Build 160
mpv 0.38.0 FFmpeg 7.0.1

Feb 3, 2026 at 4:11:15 PM
NIGHTLY bd16c84

### Logs

[mpv.log](https://github.com/user-attachments/files/25517810/mpv.log)
[iina.log](https://github.com/user-attachments/files/25517811/iina.log)

### Sample Files

https://gofile.io/d/sPFoDt

### How often can you reproduce the issue

100%

### Required Actions

- [x] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [ ] I tested with the Nightly build and confirmed the issue is not fixed.
- [x] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

low-batt:
Very helpful you reported this as now I know this is reproducible with IINA. I hit it while testing a self build mpv. See mpv issue https://github.com/mpv-player/mpv/issues/17439.

If this is the same problem it only should occur with SDR videos and the `Load ICC profile` setting enabled on the `Video/Audio` tab of IINA's settings.

I suspect the problem was introduced by a change in a 3<sup>rd</sup> party library, possibly the recent [Little CMS 2.18](https://github.com/mm2/Little-CMS/releases) release. Maybe the mpv Nightly build is not using the latest libraries?

Have a look at that mpv issue and see if it matches up with what you are encountering.

comfortablemilk4454:
[...]

PR #5954, opened 2026-03-11
Fix invalid tone-mapping-param value for mpv 0.41.0

shubhamgoel27:
## Summary

- Fix `tone-mapping-param` being set to the string `"default"`, which is invalid for this float-typed mpv option (`OPT_FLOATDEF`). This causes an error with mpv 0.41.0:
  ```
  [cplayer] error: The tone-mapping-param option is out of range: default
  ```
- Use `setDouble` with `Double.nan` instead, which is mpv's actual default for this option (NaN signals "use the tone-mapping curve's built-in default parameter")

This was found while investigating #5928. The blank screen reported there is primarily an upstream issue in Little CMS 2.18 affecting `icc-profile-auto` with `vo=libmpv` (see [mpv#17439](https://github.com/mpv-player/mpv/issues/17439)), but this tone-mapping-param error is a separate IINA-side compatibility issue with mpv 0.41.0 that should be fixed regardless.

## Test plan

- [ ] Build IINA with embedded mpv 0.41.0
- [ ] Play an SDR video with "Load ICC profile" enabled
- [ ] Verify no `tone-mapping-param option is out of range` error in mpv logs
- [ ] Verify tone mapping resets correctly when switching between HDR and SDR content

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/iina/VideoView.swift
+++ b/iina/VideoView.swift
@@ -337,7 +337,7 @@ class VideoView: NSView {
       player.mpv.setString(MPVOption.GPURendererOptions.targetPrim, "auto")
       player.mpv.setString(MPVOption.GPURendererOptions.targetPeak, "auto")
       player.mpv.setString(MPVOption.GPURendererOptions.toneMapping, "auto")
-      player.mpv.setString(MPVOption.GPURendererOptions.toneMappingParam, "default")
+      player.mpv.setDouble(MPVOption.GPURendererOptions.toneMappingParam, Double.nan)
       player.mpv.setFlag(MPVOption.Screenshot.screenshotTagColorspace, false)
     }
   }


low-batt (changes requested):
@shubhamgoel27 Hello from a _junior_ IINA developer.

Thank you very much for posting this fix!

I just entered issue #5956 for this problem. It has been on my todo list to enter an issue specifically for this problem, but it has not been a priority as I am working on IINA 1.4.2, which is a bug fix release before we move on to IINA 1.5.0 which includes upgrading mpv, FFmpeg and other libraries. To be able to track work it is preferable to have one problem per issue.

The root cause of the problem is a regression in mpv. See issue #5956 for references to the mpv issue and the mpv PR that fixed it. Issue #5956 tracks the need for IINA to patch mpv sources with that fix when building `libmpv` 0.41.0 for IINA 1.5.0.

This PR is _very useful_ as it provides a workaround for IINA users that build IINA themselves, but aren't setup to patch and build `libmpv`, which requires you to install a lot of mpv's dependencies to be able to build it. It can be painful to configure a Mac to be able to build `libmpv`. You will see I noted that in issue #5956.

But as this is a mpv regression and there is a fix available, I'm expecting the senior developers will prefer the mpv fix over changing IINA. I'm going to mark this as "Request changes" so the senior developers notice and read this. If there is trouble applying the mpv fix to the mpv 0.41.0 sources, IINA may need this fix.

shubhamgoel27:
> @shubhamgoel27 Hello from a _junior_ IINA developer.
> 
> Thank you very much for posting this fix!
> 
> I just entered issue #5956 for this problem. It has been on my todo list to enter an issue specifically for this problem, but it has not been a priority as I am working on IINA 1.4.2, which is a bug fix release before we move on to IINA 1.5.0 which includes upgrading mpv, FFmpeg and other libraries. To be able to track work it is preferable to have one problem per issue.
> 
> The root cause of the problem is a regression in mpv. See issue #5956 for references to the mpv issue and the mpv PR that fixed it. Issue #5956 tracks the need for IINA to patch mpv sources with that fix when building `libmpv` 0.41.0 for IINA 1.5.0.
> 
> This PR is _very useful_ as it provides a workaround for IINA users that build IINA themselves, but aren't setup to patch and build `libmpv`, which requires you to install a lot of mpv's dependencies to be able to build it. It can be painful to configure a Mac to be able to build `libmpv`. You will see I noted that in issue #5956.
> 
> But as this is a mpv regression and there is a fix available, I'm expecting the senior developers will prefer the mpv fix over changing IINA. I'm going to mark this as "Request changes" so the senior developers notice and read this. If there is trouble applying the mpv fix to the mpv 0.41.0 sources, IINA may need this fix.

@low-batt sounds good! I love IINA as a product, and was just going through the open issues, and thought might take a stab. Feel free to ignore if the mpv fix works!

PR #5959, opened 2026-03-13
Fix: Reduce energy impact from high-frequency timers, GPU compositing, and I/O patterns

rm335:
## Summary

Fixed 9 energy issues across 7 files. These changes reduce unnecessary CPU wakeups, GPU compositing overhead, and disk I/O during playback.

### Timers

- **Fixed `syncUITimer` firing at up to 25 Hz without tolerance** (`PlayerCore.swift`). Added `tolerance = timeInterval * 0.1` after timer creation so macOS can coalesce wakeups with other system timers instead of waking the CPU at exact 40ms intervals.

- **Fixed `ScrollingTextField` animation timer firing at 33 Hz without tolerance** (`ScrollingTextField.swift`). Reduced the interval from 0.03s to 0.05s (20 Hz) which is still visually smooth for text scrolling, and added `tolerance = interval * 0.15`.

- **Fixed Inspector window timer polling mpv while occluded** (`InspectorWindowController.swift`). Added `guard window?.occlusionState.contains(.visible) == true else { return }` at the top of `dynamicUpdate()` so the 1 Hz timer skips mpv property queries when the window is hidden behind other windows. Also added `tolerance = 0.2` to the timer.

- **Fixed plugin `setInterval()` accepting arbitrarily small intervals with no tolerance** (`JavascriptPolyfill.swift`). Repeating timers from JS plugins are now clamped to a minimum of 16ms (~60 Hz) and get `tolerance = timeInterval * 0.1`. One-shot timers (`setTimeout`) are not clamped.

### GPU

- **Fixed 7 `NSVisualEffectView` instances permanently set to `.active`** (`MainWindowController.swift`). Changed to `.followsWindowActiveState` so the GPU stops compositing vibrancy blur when the window is in the background. This is standard macOS behavior — the blur dims when the window loses focus.

- **Fixed `shadowOpacity` not being reset after exiting crop/interactive mode** (`MainWindowController.swift`). Added `self.videoView.layer?.shadowOpacity = 0` in both the animated and immediate `exitInteractiveMode` code paths. Without this, a shadow compositing pass ran over every rendered video frame for the rest of the session after using crop mode once.

### Threading

- **Fixed `Thread.sleep(forTimeInterval: 0.5)` blocking a GCD thread** (`PlayerCore.swift`). Replaced with `DispatchQueue.global(qos: .utility).asyncAfter(deadline: .now() + 0.5)` in `autoSearchOnlineSub()` so the thread is released during the delay instead of being held idle.

### I/O

- **Fixed thumbnail cache writing 300+ individual `file.write()` calls** (`ThumbnailCache.swift`). Coalesced all data (version, file size, modification date, and per-thumbnail blocks) into a single `Data` buffer and writes it once.

### Cleanup

- **Removed unused `var timer: Timer?` property** (`WebSocketServer.swift`). This was declared but never assigned or read anywhere in the codebase.

### Playback timers
1. Open Preferences and enable millisecond precision for the time display
2. Play a local video file
3. Verify the time display still updates smoothly without visible jitter

### Scrolling text
1. Open the mini player (Window > Mini Player)
2. Play a file with a title longer than the mini player width
3. Verify the title scrolls smoothly across the display

### Inspector occlusion
1. Play a video and open the Inspector window (Window > Inspector)
2. Move another window to fully cover the Inspector
3. Open Activity Monitor and verify IINA's CPU usage drops compared to when the Inspector is visible

### Crop mode shadow
1. Play a video
2. Enter crop mode (Video > Crop)
3. Exit crop mode
4. In Xcode debugger: `po self.videoView.layer?.shadowOpacity` should print `0`
5. Verify visually that no shadow artifact remains around the video view

### Visual effect views
1. Play a video with the on-screen controller visible
2. Click on a different app to move IINA to the background
3. Verify the control bar and title bar blur dims slightly (standard macOS behavior)
4. Click back on IINA — blur should return to full vibrancy

### Plugin timers
1. Create a test plugin that calls `setInterval(function() {}, 1)` (1ms interval)
2. Load the plugin and verify in Instruments (Timer Profiler) that the timer fires at ~60 Hz, not 1000 Hz

### Thumbnail cache
1. Play a video that does not yet have a thumbnail cache
2. Wait for thumbnails to generate (visible in the seek bar preview)
3. Close and re-open the same video
4. Verify thumbnails load correctly from cache on the second open

### Network streams
1. Play a network stream (e.g. an HTTP URL)
2. Pause playback
3. Resume playback
4. Verify buffering indicator and time display still work correctly

### DIFF
--- a/iina/InspectorWindowController.swift
+++ b/iina/InspectorWindowController.swift
@@ -121,6 +121,7 @@ class InspectorWindowController: NSWindowController, NSWindowDelegate, NSTableVi
 
     removeTimerAndListeners()
     updateTimer = Timer.scheduledTimer(timeInterval: TimeInterval(1), target: self, selector: #selector(dynamicUpdate), userInfo: nil, repeats: true)
+    updateTimer?.tolerance = 0.2
 
     observers.append(NotificationCenter.default.addObserver(forName: .iinaFileLoaded, object: nil, queue: .main, using: self.fileLoaded))
     observers.append(NotificationCenter.default.addObserver(forName: .iinaMainWindowChanged, object: nil, queue: .main, using: self.fileLoaded))
@@ -301,6 +302,7 @@ class InspectorWindowController: NSWindowController, NSWindowDelegate, NSTableVi
   }
 
   @objc func dynamicUpdate() {
+    guard window?.occlusionState.contains(.visible) == true else { return }
     updateInfo(dynamic: true)
     /// Do not call `reloadData()` (no arg version) because it will clear the selection. Also, because we know the number of rows will not change,
     /// calling `reloadData(forRowIndexes:)` will get the same result but much more efficiently
--- a/iina/JavascriptPolyfill.swift
+++ b/iina/JavascriptPolyfill.swift
@@ -28,7 +28,8 @@ class JavascriptPolyfill {
   }
 
   func createTimer(callback: JSValue, ms: Double, repeats : Bool) -> String {
-    let timeInterval  = ms/1000.0
+    let clampedMs = repeats ? max(ms, 16.0) : ms
+    let timeInterval  = clampedMs/1000.0
     let uuid = NSUUID().uuidString
 
     DispatchQueue.main.async(execute: {
@@ -37,6 +38,7 @@ class JavascriptPolyfill {
                                        selector: #selector(self.callJSCallback),
                                        userInfo: callback,
                                        repeats: repeats)
+      timer.tolerance = timeInterval * 0.1
       self.timers[uuid] = timer
     })
     return uuid
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -616,7 +616,7 @@ class MainWindowController: PlayerWindowController {
     titleBarBottomBorder.fillColor = NSColor(named: .titleBarBorder)!
     cachedScreenCount = NSScreen.screens.count
     [titleBarView, osdVisualEffectView, controlBarBottom, controlBarFloating, sideBarView, osdVisualEffectView, pipOverlayView].forEach {
-      $0?.state = .active
+      $0?.state = .followsWindowActiveState
     }
     // hide other views
     osdVisualEffectView.isHidden = true
@@ -2496,6 +2496,7 @@ class MainWindowController: PlayerWindowController {
       self.sideBarStatus = .hidden
       self.bottomView.subviews.removeAll()
       self.bottomView.isHidden = true
+      self.videoView.layer?.shadowOpacity = 0
       return
     }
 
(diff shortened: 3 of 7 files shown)


low-batt:
Hello from a _junior_ IINA developer. Wow. _Exceptionally_ clear writeup!. With details on testing!

Caught me busy this weekend. I did take a quick peek.

Making use of [tolerance](https://developer.apple.com/documentation/foundation/timer/tolerance) is definitely an important improvement. Makes sense to me.

Quite a while ago I noticed this existing code:
https://github.com/iina/iina/blob/9a5d85fbde95985b7a8cb553ab0385fec07e5fe9/iina/PlayerCore.swift#L2369-L2370

And wanted to delete that call to [sleep](https://developer.apple.com/documentation/foundation/thread/sleep(fortimeinterval:)), but never had time to ask other developers why that is needed or dig into history to see if the commit comment provided a clue. This is the kind of code that really needs a comment as to why it is needed and how `0.5` was picked. I never had time to explore this.

Now that I'm looking at this code, there are some existing problems. TSan was reporting lots of data races related to the values contained in `PlayerCore.info`. Two approaches were taken to fix the data races. For most of the data in `info` there was not a good reason to access the data from threads other than the main thread. For data where it was desirable to allow access from multiple threads, locks were added. The next two lines of the existing code:
https://github.com/iina/iina/blob/9a5d85fbde95985b7a8cb553ab0385fec07e5fe9/iina/PlayerCore.swift#L2371-L2372

Reference `isNetworkResource` and `videoDuration`, which are only supposed to be accessed from the main thread. This is an existing problem, not a problem with this PR.

To fix this the call to `DispatchQueue.main.async` needs to surround all the code, not just the call to `menuFindOnlineSub`. Once that fix is done then there is no need to use the `global` queue. The dispatch to the main thread can use `asyncAfter`.

I will be taking a closer look when I'm more coherent.

rm335:
Great catch, thank you!

Updated to use `DispatchQueue.main.asyncAfter` so `isNetworkResource` and `videoDuration` are accessed on the main thread. This collapses the two nested dispatches (global + main) into a single main thread dispatch, and the inner `DispatchQueue.main.async` for `menuFindOnlineSub` is no longer needed since we're already on main.

Regarding the `0.5` delay,  I kept it as-is since I couldn't find context for why it was chosen either. Happy to add a comment noting it's a heuristic delay to allow subtitle tracks to load before checking whether an online search is needed, if that sounds right.

uiryuu:
This is too messy to review as a single PR. Can you reimplement (or re prompt the AI) this PR with latest develop branch and use GitHub's latest [stacked pull request](https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests) feature? We haven't used this before but I believe it is pretty suitable for changes like this and can make stuff more clear.

PR #5962, opened 2026-03-21
Show detailed online subtitle errors

DavidAGInnovation:
## Summary
- preserve detailed online subtitle errors instead of collapsing them into the generic network error OSD when the server already returned a concrete message
- keep connect and timeout failures mapped to the existing generic network error

## Why
When OpenSubtitles rejects a download with a concrete API message, IINA currently maps that to the generic `Network error` OSD. In the reproduced case, OpenSubtitles returned HTTP 406 with a quota message (`You have downloaded your allowed 5 subtitles for 24h...`), but the UI only showed `Error de red`.

## Change
This updates the native online subtitle workflow to reuse `error.localizedDescription` in the OSD for `CommonError.networkError` and other fallback error cases. That allows server-side messages like quota or authorization failures to reach the user directly.

## Verification
- reproduced an OpenSubtitles download failure where the API returned a concrete quota-exhausted message
- verified the change keeps server-provided error detail instead of converting it to the generic network error path

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -3,6 +3,7 @@ extend-exclude = [
     ".git/",
     "deps/",
     "iina.xcodeproj/",
+    "iina/Assets.xcassets/Symbols/**/*.svg",
     "**/*.strings",
 ]
 ignore-hidden = false
@@ -21,6 +22,7 @@ extend-ignore-re = [
 [default.extend-words]
 "abitrate" = "abitrate"
 "splitted" = "splitted"
+"writeable" = "writeable"
 
 [default.extend-identifiers]
 "caf" = "caf"
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -136,6 +136,9 @@
 "osd.canceled" = "Canceled";
 "osd.cannot_connect" = "Cannot connect";
 "osd.timed_out" = "Timed Out";
+"osd.sub_quota_exceeded" = "Subtitle download limit reached";
+"osd.sub_quota_exceeded.detail" = "Try again after %@";
+"osd.sub_quota_exceeded.detail_unknown" = "Try again later";
 "osd.filter_added" = "Added Filter: %@";
 "osd.filter_removed" = "Removed Filter";
 "osd.file_loop" = "Loop Single File";
--- a/iina/OSDMessage.swift
+++ b/iina/OSDMessage.swift
@@ -84,6 +84,7 @@ enum OSDMessage {
   case canceled
   case cannotConnect
   case timedOut
+  case onlineSubQuotaExceeded(String?)
 
   case fileLoop
   case playlistLoop
@@ -122,6 +123,7 @@ enum OSDMessage {
     case .fileError: fallthrough
     case .foundSub: fallthrough
     case .networkError: fallthrough
+    case .onlineSubQuotaExceeded: fallthrough
     case .savedSub: fallthrough
     case .startFindingSub: fallthrough
     case .timedOut:
@@ -440,6 +442,18 @@ enum OSDMessage {
       return (
         NSLocalizedString("osd.timed_out", comment: "Timed out"),
         .normal
+      )
+
+    case .onlineSubQuotaExceeded(let resetTime):
+      let detail: String
+      if let resetTime, !resetTime.isEmpty {
+        detail = String(format: NSLocalizedString("osd.sub_quota_exceeded.detail", comment: "Try again after %@"), resetTime)
+      } else {
+        detail = NSLocalizedString("osd.sub_quota_exceeded.detail_unknown", comment: "Try again later")
+      }
+      return (
+        NSLocalizedString("osd.sub_quota_exceeded", comment: "Subtitle download limit reached"),
+        .withText(detail)
       )
 
     case .fileLoop:
--- a/iina/OnlineSubtitle.swift
+++ b/iina/OnlineSubtitle.swift
@@ -235,8 +235,8 @@ class OnlineSubtitle: NSObject {
         osdMessage = .cannotConnect
         log("\(prefix)\(cause.localizedDescription)", level: .error)
       case CommonError.networkError(let cause):
-        osdMessage = .networkError
         let error = cause ?? err
+        osdMessage = osdMessageForDetailedError(error, fallback: .networkError, providerName: provider.name)
         log("\(prefix)\(error.localizedDescription)", level: .error)
       case CommonError.timedOut(let cause):
         osdMessage = .timedOut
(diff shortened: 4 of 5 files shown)


low-batt:
Sorry about this big long review. There is some history and details around this area of IINA I thought you should know about…

When I read the title of this PR I smiled. In general I have always found the software industry to be very poor when it comes to handling and reporting errors. Massive amounts of time have been lost tracking down a problem that would have been immediately obvious had the problem been clearly reported. The instance cited here, the reporting of a failure due to the download quota having been exceeded as a `Network Error`, is an atrocious example of poor error reporting. It is especially egregious in that this is a very common error that users will encounter. I was glad to see someone else noticed this bad behavior.

I had plans to address this, but have had to focus on other work. A primary reason this was not a priority for me is because the code that is built into IINA for downloading subtitles from [Open Subtitles](https://www.opensubtitles.com/) has been _deprecated_ in favor of the official [OpenSubtitles plugin](https://github.com/iina/plugin-opensub). One reason for the transition to plugins for downloading subtitles is to make it easier for developers to add support for other subtitle sites independent of an IINA release.

Quite a few of the features that had been requested in this area of IINA, such as being able to manually specify a search string instead of the video title,  have been implemented in the plugin. So the plugin implementation is already superior to the built-in implementation.

For me, the built-in implementation still has one important advantage over the plugin implementation, all communication with [Open Subtitles](https://www.opensubtitles.com/) is fully logged in detail. This is important because of a history of the Open Subtitles servers responding oddly when they are overloaded. Instead of a properly encoded REST API error response the servers would send a HTML page. IINA's client has code to try and deal with these unexpected responses, but as this behavior is intermittent there still may be responses that IINA is not expecting. The plugin does not provide this level of logging.

With the plugin being new the built-in implementation is still useful to compare behaviors. I do not know what the plans are for retaining the built-in implementation. That is a question for @lhc70000.

My thoughts are that as the built-in code is still available for use, it still would be nice to improve the error handling, especially for the common error of a download failing due to the quota being exceeded. So on to the review…

I pulled the PR, built IINA and testing, reproducing the quota exceeded error. On the surface, the changes are good. _However_ there are existing UI design choices that complicate things…

The code uses the on screen display (OSD) to display the message. By default the OSD displays for _one second_. With the changes in this PR the user will have 1 sec to read this:

example2

If the video is wide enough you will see this:

example1

That seems to me to be a bit too much to read in 1 sec.

The error message text Open Subtitles provides is:
> "You have downloaded your allowed 5 subtitles for 24h.Your quota will be renewed in 01 hours and 23 minutes (2026-03-30 23:59:58 UTC) ts=1774910208 "

- The space between the sentences is missing
- There is an extraneous space at the end of the error message (the code in the PR correctly strips it)
- The `ts=1774910208` is just odd as there is a `ts` key with that value in the JSON error response

I never did get around to asking Open Subtitles to fix this error message. But there is a more significant problem related to this:
> to reuse `error.localizedDescription` in the OSD

IINA considers localization a high priority. The built-in code is not passing the [Accept-Language](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Language) header, so I don't think the [Open Subtitles](https://www.opensubtitles.com/) servers have a way of knowing the what language to use in an error message. That header is not being passed because the [Open Subtitles REST API Documentation](https://opensubtitles.stoplight.io/docs/opensubtitles-api/e3750fd63a100-getting-started#important-http-request-headers) did not call for it to be passed. In a quick test I tried passing that header, specifying `fr` for the language. The error message returned was still in English. I'm guessing Open Subtitles does not support localized error messages.

This means not only are we are showing a long, possibly truncated, error message for only one second, it is potentially in the wrong language.

The error response from Open Subtitles can be seen in the [error codes](https://opensubtitles.stoplight.io/docs/opensubtitles-api/12f131ce12132-error-codes) section of the REST API documentation under "406 Not Acceptable". In additon to message strings the response provides:

- `requests`
- `remaining`
- `reset_time_utc`

A side note, in working on this review I found a problem in the client with properly decoding the error response. I have merged a fix for that.

The documentation does not detail these fields, but it looks like `remaining` will be `-1` when the quota is exceeded and `requests` will be one past your quota. This should provide IINA with enough of the dynamic information required for IINA to generate its own localized OSD message, maybe something like this:

proposed


**HOWEVER** this changes the scope of the work. I do not expect there to be objections to us updating deprecated code. But adding new text that needs to be localized means we are asking IINA's translators to provide translations for deprecated code. I am uncertain as to whether there would be objections to that. If we don't want to do work that ends up being rejected, we might want to check in with @lhc70000 before doing additional work.

Thoughts? If you want to continue to work this and need tips on how to deal with localization I will be happy to give you pointers.

But wait, there's more…

I have yet to talk about how the [OpenSubtitles plugin](https://github.com/iina/plugin-opensub) handles this failure. I tested that and there was no feed back at all in the UI. Even clicking on the `Logs` button in the sidebar provided no indication that the request failed. This is a separate issue, but it indicates that error handling and reporting needs to be improved in the plugin implementation as well. So much to do…

DavidAGInnovation:
I updated this to stop feeding the raw OpenSubtitles message directly into the OSD.

For the built-in OpenSubtitles path, 406 quota-exceeded responses are now mapped to a short localized OSD generated from the structured error fields, and the reset time is shown in the user’s locale. I also force a longer OSD timeout for that case, so it’s actually readable.

Other unknown errors now fall back to the existing generic network error instead of showing the raw API text.

lhc70000:
I think it's fine to merge this one. Although please note that we will not actively maintain the built-in OpenSubtitles support.

low-batt:
@DavidAGInnovation As you may have noticed the project is currently reviewing and merging outstanding PRs. Sorry it has taken a while to get back to this one.

As @lhc70000 indicated the built-in code for downloading subtitles from Open Subtitles is under a death sentence due to the switch to the plugin. The plugin is already superior to the built-in code in many ways, however there are a few things that are still missing from the plugin. It is not looking like all of that work will be completed for 1.5.0. So the built-in code has been granted a reprieve on its death sentence, for now.

VERY HELPFUL that you fixed the [typos](https://github.com/crate-ci/typos) configuration. I was just about to investigate the new failures. In this case it does not matter, but always good to put any `typos` related changes in their own PR.

low-batt (approved):
Pulled latest commits, built IINA, uninstalled the OpenSubtitles plug, exceeded the OpenSubtitles download quota to test the new error message. Worked fine.

Changes look good to me!

This is the most important error to communicate properly to the user as especially if they are not using an Open Subtitles account it is very likely they will encounter it as the quota is very low (5). And now with ISPs deploying [CGNAT](https://en.wikipedia.org/wiki/Carrier-grade_NAT) is it even more likely that somebody else has already used up the quota.

This is how IINA now reports the quota exceeded error (when using the deprecated built-in code):
error

PR #5964, opened 2026-03-21
Allow removing downloaded subtitles from the sidebar

DavidAGInnovation:
## Summary
- add a context menu to the primary and secondary subtitle track tables in the quick settings sidebar
- expose a `Remove Downloaded Subtitle` action for downloaded subtitle temp tracks loaded into the current session
- remove the mpv subtitle track and delete the matching temp file so the track disappears from the sidebar immediately

## Why
Downloaded online subtitles are loaded as external temp files and can be saved, but there is currently no matching way to remove them from the subtitles sidebar once they have been loaded.

## Screenshot
subtitles

## Notes
- the action is intentionally limited to downloaded subtitle temp tracks under IINA's temporary directory
- it does not affect embedded subtitle tracks or user-managed external subtitle files outside that temp area
- the action is available from the subtitle track rows via the standard context menu

## Verification
- verified native online subtitle providers store downloaded subtitles under `NSTemporaryDirectory()`
- verified mpv already supports `sub-remove`, which this wires into the sidebar UI


low-batt:
To add to the "why this feature", sometimes you download a subtitle file and you find the timing does not match up with the video you have. Yes you can download another one and switch to that without removing the first one. But it is still nice to have the option to get rid of the subtitle you don't need.

I may have time to review this tomorrow.

low-batt:
I am going to provide some feedback on the code, but do not get too focused on that as I believe design issues need to be addressed first. Mostly I am the wrong person to be reviewing UI changes. I will provide some feedback, but I always defer to the senior developers when it comes to user interface changes. I am going to mention an alternative to a context menu. But that way also would have problems. I'm not sure how this feature should be presented to the user. I want to hear what the senior developers think.

From a "big picture" standpoint we are adding support for the mpv [sub-remove](https://mpv.io/manual/stable/#command-interface-sub-remove) command:
> **sub-remove \[\<id>]**
> Remove the given subtitle track. If the id argument is missing, remove the current track. (Works on external subtitle files only.)

Right away we already have a problem. This command cannot be used to remove a subtitle track that is embedded in the video. This is a problem because the IINA UI does not currently provide any indication as to whether a subtitle is embedded or external. A user would have to open the `Inspector` and look at the `Tracks` tab. This restriction has the potential to cause one of the very frustrating user experiences where an application refuses to perform an operation and it is not apparent to the user as to why the operation is not working.

This PR further restricts this operation:
> -expose a `Remove Downloaded Subtitle` action for downloaded subtitle temp tracks loaded into the current session

Why is IINA imposing additional restrictions on this operation?

On this:
> - add a context menu to the primary and secondary subtitle track tables in the quick settings sidebar

Although context menus are nice, one problem with them is that it is not apparent to a user that a context menu is available. A user would have to discover it by hovering over a downloaded subtitle and right clicking on it.

The first question to ask is whether IINA has an existing UI convention that might apply here. The tables on the `Key Bindings` tab of IINA's settings have `+` and `-` buttons. Same for the `Additional mpv options` table on the `Advanced` tab. If those buttons were added here the `+` could bring up the same dialog used for the `Load External Subtitle…` menu option. The `-` button would have to differ in that when an internal subtitle was selected the button would have to be disabled and dimmed out. A tooltip could explain why the button is dimmed (only external subtitles can be removed). _However_ this table differs from the ones in settings in that selecting an entry has a side effect. It selects the subtitle to be displayed. A context menu does have the advantage of being able to delete an entry without selecting it.

Note that this selection behavior does provide an odd user experience in one case. Clicking on a subtitle will highlight it and then go back to the one that was previously selected if that subtitle is already being displayed. I did not find this restriction in the mpv documentation, but mpv will not allow a subtitle to be displayed as both the primary and secondary subtitles. IINA needs to dim and disallow entries and provide a tooltip explaining that a subtitle can't be selected as it is already being displayed. This is an existing problem that needs to be addressed at some point (not in this PR).

This PR does not just add support for the mpv `sub-remove` command, it adds in an additional action:
> and delete the matching temp file

If this subtitle file was downloaded by an IINA plugin, that plugin may not be expecting that file to disappear. A plugin could remember all the files it downloaded and provide an interface to allow you to save the file even after it is no longer being used. I'm thinking it is the responsibility of the code that created the temporary file and added it to the track list to listen for the event indicating the track list to changed, look to see if the track they added is no longer in the track list, and if so then that code is responsible for removing the temporary file.

One of the reasons why that approach has an advantage is because IINA does not have full control over mpv. The plugin API provides low level access to the mpv client API. IINA key bindings allow a user to directly use mpv commands. In the case of key bindings IINA does intercept certain mpv commands, but that is not reliable. Users can completely by-pass IINA by directly using the mpv [JSON IPC](https://mpv.io/manual/stable/#json-ipc). Users can also use [Conditional auto profiles](https://mpv.io/manual/stable/#conditional-auto-profiles) to dynamically change mpv configuration. Keying off events handles these other ways of altering `libmpv` behavior.

A side note on this:
> - verified native online subtitle providers store downloaded subtitles under NSTemporaryDirectory()

I believe IINA is using [NSTemporaryDirectory()](https://developer.apple.com/documentation/foundation/nstemporarydirectory%28%29) because we only recently were forced to drop support for macOS 10.11 thru 10.14 when mpv dropped support for macOS releases before 10.15, Catalina. Not now, but the next time we are working on a release that includes refactoring/cleanup this code can be updated to use [FileManager.temporaryDirectory](https://developer.apple.com/documentation/foundation/filemanager/temporarydirectory) (introduced in macOS 10.12).

iina/es.lproj/Localizable.strings
```diff
@@ -52,6 +52,7 @@
 "quicksetting.hdr" = "HDR";
 "quicksetting.hide" = "Ocultar";
 "quicksetting.reset_speed" = "Restablecer velocidad a 1x";
+"quicksetting.sub_remove_downloaded" = "Eliminar subtítulo descargado";
```
low-batt: How localization is handled in IINA is special due to our integration with [Crowdin](https://crowdin.com/project/iina). The requirements can be found in [Contributing](https://github.com/iina/iina?tab=contributing-ov-file) under [Localizations](https://github.com/iina/iina?tab=contributing-ov-file#localizations):
> We use [Crowdin](https://crowdin.com/project/iina) for localization. When there is a UI text change, only add/update the strings in `Base.lproj` and `en.lproj`. Please do not include any localization changes for other languages in the pull request. New strings on the develop branch from the `en.lproj` folder will be synced automatically to Crowdin, and we will fetch the latest translations from Crowdin before releasing the next version.

Thus for the synchronization with Crowdin to work, the changes to `es.lproj/Localizable.strings` must be removed from this PR. The Spanish translation for the new string will need to be added using Crowdin once this PR has been merged. And the changes made to `en.lproj/Localizable.strings` will need to be duplicated in `Base.lproj/Localizable.strings`.

A while back there was an effort to switch to [string catalogs](https://developer.apple.com/documentation/xcode/localizing-and-varying-text-with-a-string-catalog) but there was difficulty getting synchronization with Crowdin to work at that time.

iina/QuickSettingViewController.swift
```diff
@@ -658,6 +666,39 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
...
+      return subTableView
+    }
+    if menu === secondarySubtitleContextMenu {
+      return secSubTableView
+    }
+    return nil
+  }
+
+  private func temporarySubtitleDirectoryPath() -> String {
+    Utility.tempDirURL.path.hasSuffix("/") ? Utility.tempDirURL.path : Utility.tempDirURL.path + "/"
```
low-batt: Might want to consider using a computed property, lazy variable or constant instead of a function for something this simple that doesn't take parameters.

iina/QuickSettingViewController.swift
```diff
@@ -658,6 +666,39 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
...
+  }
+
+  private func isRemovableDownloadedSubtitle(_ track: MPVTrack) -> Bool {
+    guard track.type == .sub, track.isExternal, let filename = track.externalFilename else { return false }
```
low-batt: Adding this code makes IINA inconsistent with how it identifies downloaded subtitles. The other code that does this is in `MainMenuActions.saveDownloadedSub`
https://github.com/iina/iina/blob/18d4cb807abaeb8711281f4682f8b7ecf61bfa87/iina/MainMenuActions.swift#L432

I'm thinking the new code in this PR is better than the existing code. No changes needed to the PR. Just pointing out the existing code should be improved at some point.

iina/QuickSettingViewController.swift
```diff
@@ -920,6 +961,32 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
     }
   }
 
+  @objc func removeDownloadedSubtitleFromMenu(_ sender: NSMenuItem) {
+    guard let track = sender.representedObject as? MPVTrack,
+          isRemovableDownloadedSubtitle(track),
+          let filename = track.externalFilename else { return }
+
+    player.mpv.command(.subRemove, args: [String(track.id)], checkError: false, level: .verbose) { code in
```
low-batt: My own opinion is that IINA code is way too [tightly coupled](https://en.wikipedia.org/wiki/Coupling_%28computer_programming%29). I've not had a chance to discuss this with the very busy senior developers. But I believe there is some recognition of this. The [Contributing](https://github.com/iina/iina?tab=contributing-ov-file) guidelines under [Current Structure](https://github.com/iina/iina?tab=contributing-ov-file#current-structure) says:
> - `PlayerCore` encapsulates general playback functions.
> 	- Setting options/properties directly through `MPVController` is discouraged.

My experience has been that "discouraged" is never enough. The build or a unit test has to fail to ensure such rules are followed. Therefore `PlayerCore.mpv` _must be_ `private`. But it is already too late. There is a lot of code that by-passes `PlayerCore` and accesses it's `mpv` property directly. To address this requires a lot of refactoring, but currently other work takes priority. All we can do is to try and keep new code from adding to the problem.

So although there is a lot of IINA code that references the `MPVController` directly through the `PlayerCore` as is done here, it is preferable to add a new method to `PlayerCore` in the `MPV Commands` section:
```swift
  func subRemove(id: Int? = nil)
```

And have the `QuickSettingViewController` call that.

iina/QuickSettingViewController.swift
```diff
@@ -920,6 +961,32 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
...
+        self.player.log("Failed removing subtitle track #\(track.id): error code \(code)", level: .error)
+        return
+      }
+
+      do {
+        try FileManager.default.removeItem(atPath: filename)
+      } catch CocoaError.fileNoSuchFile {
+        // Ignore if the temp subtitle file was already cleaned up.
+      } catch {
+        self.player.log("Failed removing downloaded subtitle file \(filename): \(error.localizedDescription)", level: .warning)
```
low-batt: As `PlayerCore` is not involved in downloading subtitle files I think it is better to log this as a window operation using `self.mainWindow.log`.

low-batt (changes requested):
Localization has to be corrected. I think a few other changes should be made. Wait and see what the other reviewers say.

iina/es.lproj/Localizable.strings
DavidAGInnovation: Fixed. I removed the `es.lproj` change, kept the new string in `en.lproj`, and duplicated it in `Base.lproj` so it follows IINA’s Crowdin workflow.

iina/QuickSettingViewController.swift
DavidAGInnovation: Good suggestion. I changed `temporarySubtitleDirectoryPath()` to a computed property.

iina/QuickSettingViewController.swift
DavidAGInnovation: Agreed. I added a `PlayerCore.subRemove(id:completion:)` wrapper and switched `QuickSettingViewController` to call that instead of reaching into `player.mpv` directly.

iina/QuickSettingViewController.swift
DavidAGInnovation: Makes sense. I left the new check as-is for this PR, but I agree the existing `saveDownloadedSub` path should be cleaned up separately so both places use the same identification logic.

iina/QuickSettingViewController.swift
DavidAGInnovation: Adjusted. The subtitle-file removal logging now uses `self.mainWindow.log`.

DavidAGInnovation:
I'll make this pass.

iina/PlayerCore.swift
```diff
@@ -1385,6 +1385,11 @@ class PlayerCore: NSObject {
     }
   }
 
+  func subRemove(id: Int? = nil, completion: @escaping (Int32) -> Void = { _ in }) {
+    let args = id.map { [String($0)] } ?? []
+    mpv.command(.subRemove, args: args, checkError: false, level: .verbose, completion)
+  }
```
low-batt: I should have mentioned the error logging should also be moved to `PlayerCore`. Sorry! Looking through the  `PlayerCore` methods there aren't any that return mpv error codes. Something like this would be closer to how some of the other `PlayerCore` methods handle errors:
```Suggestion
  func subRemove(id: Int? = nil) -> Bool {
    let args = id.map { [String($0)] } ?? []
    var result = true
    mpv.command(.subRemove, args: args, checkError: false, level: .verbose) { code in
      if code < 0 {
        result = false
        self.log("Failed removing subtitle track #\(id ?? 0): error code \(code)", level: .error)
      }
    }
    return result
  }
```

That allows the call in `removeDownloadedSubtitleFromMenu` to simply be:
```swift
guard player.subRemove(id: track.id) else { return }
```

This will also correct the compilation error that caused the build to fail.

low-batt:
See my comment above. That will also fix the build failure.

I will try and get a senior developer to look at this PR and comment. They are pretty busy reviewing the older PRs at the moment.

iina/PlayerCore.swift
DavidAGInnovation: Good catch. I updated it so the mpv error handling now lives in `PlayerCore`, and `subRemove(id:)` returns `Bool` instead of exposing the raw callback in `QuickSettingViewController`.

DavidAGInnovation:
I addressed the concrete code-review items in the latest commits, including the `PlayerCore.subRemove(id:)` wrapper, moving the mpv error logging into `PlayerCore`, and the build fix.

low-batt (commented):
We are now at the point where we need to hear from the senior developers as to whether they want to go with the context menu.

uiryuu (approved):
LGTM

DavidAGInnovation:
Superseded by a clean replacement pull request with a single commit and no merge/force-push history.

PR #5965, opened 2026-03-21
Add local development testing assets and commands

anduimagui:
## Summary
- add a bundled sample video at `example/video/big_buck_bunny.mp4` for repeatable manual playback testing
- add `Taskfile.yml` tasks for downloading local dependencies, building the app, launching it, and launching it with the sample video
- make local development workflows easier to run with short commands like `task dev` and `task dev-example`


low-batt:
Ignore the Spelling failure. A new release of [typos](https://github.com/crate-ci/typos) is detecting an existing miss-spelling, unrelated to this PR. I've pushed a commit to [develop](https://github.com/iina/iina/commits/develop/) that fixes this.

Taskfile.yml
```diff
@@ -0,0 +1,35 @@
...
+vars:
+  DERIVED_DATA: DerivedData
+  APP_PATH: '{{.DERIVED_DATA}}/Build/Products/Debug/IINA.app'
+  EXAMPLE_VIDEO: example/video/big_buck_bunny.mp4
+
+tasks:
+  deps:
+    desc: Download prebuilt dylibs and youtube-dl
+    cmds:
+      - mkdir -p deps/lib deps/executable
```
low-batt, line 12: I'm not understanding why this is not just executing  `./other/download_libs.sh` as discussed in the [Using the pre-compiled libraries](https://github.com/iina/iina?tab=readme-ov-file#using-the-pre-compiled-libraries) section of the README? That script is the official way to populate the pre-built portions of `deps`.

low-batt:
Hello from a _junior_ IINA developer. This is the kind of PR that needs to be reviewed by the senior developers. I will give you some feedback, but really the other developers need to comment on this.

I'm having trouble seeing much of a benefit. Currently IINA is expecting developers to use Xcode. This PR introduces the need for a new tool, [Task](https://taskfile.dev/). It adds always downloading IINA's dependencies. We could of added that to the Xcode build. But that would add additional load to IINA's servers and of course the extra build time. And that would raise the question as to whether there is a way to speed that up, such as coming up with a way to check to see if anything has changed, rather than always doing a full download. IINA's dependent libraries rarely change.

I prefer to manually update the dependencies. I want to be aware that changes to the dependencies could be responsible for a failure and not the changes I'm working on.

At least that is my first reaction. It is the opinion of the senior developers that matters…

uiryuu:
Sorry but I believe this is not how the devs do building/testing with IINA. You can keep that locally if you find it handy

Issue #5811, opened 2025-11-16
Secondary subtitle position setting is not saved after application restart


Joe00011:
### IINA Version

14.1

### Release Channel

Release

### macOS Version

Sequoia 15.7.2

### Steps to Reproduce

1. Open a video file with both main and secondary subtitles. 
2. Go to Subtitle menu → Secondary Subtitle → Position and change the position (e.g., move it up or down). 
3. Confirm the secondary subtitle position has changed in the player. 
4. Close IINA. 
5. Open IINA again and play the same video.

### Expected Behavior

The secondary subtitle should remain at the position I set before closing the application.

### Actual Behavior

The secondary subtitle position is reset to the default when IINA is reopened. Only the secondary subtitle position is affected; the position for the main subtitle is saved correctly.

### How often can you reproduce the issue

Always

### Required Actions

- [x] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [ ] I tested with the Nightly build and confirmed the issue is not fixed.
- [ ] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

low-batt:
Did not reproduce for me.

For IINA to remember settings for a file the IINA `Resume last playback position` must be enabled. This activates mpv's [Watch Later](https://mpv.io/manual/stable/#watch-later) feature. You also have to quit IINA before you get to the end of the video.

What settings are saved by [Watch Later](https://mpv.io/manual/stable/#watch-later) is controlled by the mpv [watch-later-options](https://mpv.io/manual/stable/#options-watch-later-options) option. IINA adjusts the value of this option because the default value is missing some options for the secondary subtitles. You can see this happening in the IINA log file:
```text
23:20:37.289 [mpv0][d] Adding secondary-sid to watch-later-options
23:20:37.289 [mpv0][d] Adding secondary-sub-delay to watch-later-options
[...]

PR #5974, opened 2026-03-28
Added secondary subtitle features and search

Cleedek:

---
🚀 **QUICK SUMMARY**
This PR introduces three user-facing enhancements and two technical fixes to improve subtitle management:

🎛️ Secondary Subtitle Control: Adds a toggle in Preferences > Subtitles to set a permanent vertical position for secondary subtitles (matching the primary subtitle settings).

💾 Session Persistence: Ensures that secondary subtitle positions are correctly saved and restored when resuming a video (fixes a known watch-later limitation).

🔍 Subtitle Search Filter: Adds a live-search field to the Online Subtitle Downloader to quickly filter results by name, language, or date.

🔧 Compatibility Fix: Replaces a macOS 13+ specific property with a backwards-compatible constant to prevent build failures on older macOS versions.

____________________________________

**MORE DETAILED**

**Feature Changes Overview
This submission adds three user-facing improvements to subtitle handling, one session-persistence fix, and one build compatibility fix.

1. 🎛️ Secondary Subtitle Vertical Position in Preferences
Files changed: iina/Preference.swift, iina/MPVController.swift, iina/PrefSubViewController.swift, iina/Base.lproj/PrefSubViewController.xib

What changed:

The Preferences → Subtitle → Position section previously only allowed configuring the vertical position (sub-pos) for primary subtitles. Secondary subtitles had no equivalent static preference — their position could only be adjusted during playback via the Quick Settings panel, and the value was lost on restart.

A "Primary Subtitles / Secondary Subtitles" segmented control has been added above the Vertical position field, following the same established pattern already used by the ASS Subtitles → Override Level section.

Details:
* Added secondarySubPos preference key (Preference.Key) with a default value of 100.
* Registered secondarySubPos as an mpv user option bound to secondary-sub-pos, so the saved value is applied at startup automatically — identical to how subPos works for primary subtitles.
* The segmented control dynamically rebinds the text field between values.subPos and values.secondarySubPos using bind(_:to:withKeyPath:options:), the same mechanism subOverrideLevelSegmentedControlAction uses.

2. 💾 Secondary Subtitle Position Persisted in Watch-Later Files
File changed: iina/MPVController.swift

What changed:
IINA uses mpv's watch-later feature to restore playback state (including subtitle position) when re-opening a video. A known mpv issue (#14417) means that secondary-sub-pos is not included in the list of saved options by default, even though the equivalent sub-pos for primary subtitles is.
IINA already contained a workaround for similar missing secondary subtitle options (secondary-sid, secondary-sub-delay). This extends that workaround to also include secondary-sub-pos.

Details:
* secondary-sub-pos is now automatically appended to watch-later-options if sub-pos is present but secondary-sub-pos is not. This mirrors what was already done for secondary-sid and secondary-sub-delay.

3. 🔍 Search/Filter Field in Online Subtitle Downloader
Files changed: iina/SubChooseViewController.swift, iina/Base.lproj/SubChooseViewController.xib

What changed:
When searching for online subtitles and clicking "Search Online", IINA presents a sheet with a list of results. Previously there was no way to filter this list — users had to scroll through all results manually.

A live search field has been added above the results table. As the user types, the list is filtered in real time across subtitle name, language, and date columns.

Details:
* NSSearchField is added programmatically in viewDidLoad and positioned between the instructional label and the results table with proper Auto Layout constraints.
* A filteredSubtitles array drives the table view. It mirrors the full subtitles array by default and is updated on each keystroke via NSSearchFieldDelegate.controlTextDidChange.
* Filtering is case-insensitive and matches across all three description fields returned by getDescription().
* The XIB's scroll view top constraint is updated to accommodate the new search field's height (16 pt gap above, 8 pt gap below).

4. 🔧 Build Fix: NSTextAttachment.character Compatibility
File changed: iina/LanguageTokenField.swift

What changed:
NSTextAttachment.character is only available on macOS 13+. The original code used this property unconditionally, causing a build failure on earlier deployment targets.

Details:
* Replaced NSTextAttachment.character with the backwards-compatible NSAttachmentCharacter constant (Unicode U+FFFC), which works on all supported macOS versions.
**

### DIFF
--- a/iina/Base.lproj/PrefSubViewController.xib
+++ b/iina/Base.lproj/PrefSubViewController.xib
@@ -24,6 +24,8 @@
                 <outlet property="subOverrideLevelSegmentedControl" destination="mqJ-QH-Pfa" id="hcK-8B-EzR"/>
                 <outlet property="subOverrideLevelSlider" destination="7UX-IL-hdk" id="fem-nv-O1m"/>
                 <outlet property="subOverrideLevelText" destination="FR3-Ar-5Dw" id="BSP-zH-sos"/>
+                <outlet property="subPosSegmentedControl" destination="pos-Sg-Ctl" id="psc-1a-abc"/>
+                <outlet property="subPosTextField" destination="HKD-K0-lsV" id="ptf-2b-bcd"/>
                 <outlet property="subShadowColorWell" destination="Gtc-ii-hE0" id="hh5-ON-mLf"/>
                 <outlet property="subSourcePopUpButton" destination="UHS-ce-8t5" id="zmW-Jx-JMg"/>
                 <outlet property="subSourceStackView" destination="bSz-xd-6o2" id="UOh-XZ-MhE"/>
@@ -1164,13 +1166,6 @@
                         <color key="textColor" name="controlTextColor" catalog="System" colorSpace="catalog"/>
                         <color key="backgroundColor" name="textBackgroundColor" catalog="System" colorSpace="catalog"/>
                     </textFieldCell>
-                    <connections>
-                        <binding destination="5Up-Ab-aAm" name="value" keyPath="values.subPos" id="BIO-yY-56q">
-                            <dictionary key="options">
-                                <bool key="NSContinuouslyUpdatesValue" value="YES"/>
-                            </dictionary>
-                        </binding>
-                    </connections>
                 </textField>
                 <textField identifier="AccessoryLabelVerticalPos" focusRingType="none" horizontalHuggingPriority="251" verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="9sd-YV-QMQ">
                     <rect key="frame" x="316" y="64" width="16" height="16"/>
@@ -1180,6 +1175,22 @@
                         <color key="backgroundColor" name="controlColor" catalog="System" colorSpace="catalog"/>
                     </textFieldCell>
                 </textField>
+                <segmentedControl verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="pos-Sg-Ctl">
+                    <rect key="frame" x="120" y="178" width="304" height="21"/>
+                    <constraints>
+                        <constraint firstAttribute="width" relation="greaterThanOrEqual" constant="300" id="psc-w-cst"/>
+                    </constraints>
+                    <segmentedCell key="cell" controlSize="small" borderStyle="border" alignment="left" segmentDistribution="fillEqually" style="rounded" trackingMode="selectOne" id="psc-cell">
+                        <font key="font" metaFont="smallSystem"/>
+                        <segments>
+                            <segment label="Primary Subtitles" selected="YES"/>
+                            <segment label="Secondary Subtitles" tag="1"/>
+                        </segments>
+                    </segmentedCell>
+                    <connections>
+                        <action selector="subPosSegmentedControlAction:" target="-2" id="psc-act"/>
+                    </connections>
+                </segmentedControl>
                 <button translatesAutoresizingMaskIntoConstraints="NO" id="pyx-K4-cgm">
                     <rect key="frame" x="118" y="31" width="327" height="18"/>
                     <buttonCell key="cell" type="check" title="Display subtitles in letterboxes while in full screen" bezelStyle="regularSquare" imagePosition="left" state="on" inset="2" id="2XY-Un-0Fg">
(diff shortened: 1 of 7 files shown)


low-batt:

I'm especially excited to see the added setting for the secondary subtitle position. Support for this has been requested in issue #5811. I've wanted to see that implemented. Definitely needed. UI changes need to be reviewed by the senior developers. But I can point out what I expect they might object to.

In [Contribution workflow](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md#contribution-workflow) there is this guidance:
> - Please submit separate pull requests for different features; i.e. try not to shove multiple, unrelated changes into one pull request.

I'm not understanding why the enhancements to secondary subtitle settings needs to be combined with adding a search field to the subtitle downloader. The senior developers sometimes allow this, but it is definitely better to break changes into separate smaller PRs as much as possible. Anyway, I believe these changes are going to run afoul of other IINA plans that you could not be aware of…

For IINA 1.4.0 IINA has started to work on addressing [technical debt](https://en.wikipedia.org/wiki/Technical_debt) and problems that were considered risky to fix. What went out as IINA 1.4.0 was supposed to be 1.4.0-beta2, but macOS Tahoe broke IINA so badly that fixes for some of the macOS 26 problems were quickly added and IINA 1.4.0 was rushed out the door. IINA 1.4.1 was a quick follow up. There are still quite a few regressions that need fixing.

Originally the plan was to start developing IINA 1.5.0 which includes upgrading the libraries IINA is dependent upon including mpv and FFmpeg. That effort ran into problems building for macOS Catalina 10.15. It is likely IINA will need to move to macOS Big Sur 11.0 as the oldest supported macOS version. To leave users trapped on macOS Catalina a better release IINA is currently working on an IINA 1.4.2 bug fix release. As IINA 1.4.2 is intended to be a stabilization release there is resistance to adding changes that do not address defects. With IINA 1.4.2 likely being the last release for macOS Catalina it is important not to introduce any regressions.

Why this matters has to do with the enhancements to secondary subtitle settings. We are continuing to address technical debt. One part of that is rewriting the settings window in SwiftUI. That work is going on in the [new-pref-ui](https://github.com/iina/iina/tree/new-pref-ui) branch and the plan is to merge it for IINA 1.5.0, replacing the current XIB based code. That means the portion of this PR that modifies `PrefSubViewController.xib` can only be used for IINA 1.4.2. And if it is merged for 1.4.2 then the same enhancement must be added to the new SwiftUI code.

I'm explaining this because I believe more changes to `PrefSubViewController.xib` are needed. That raises the question as to whether you want to put more work into this PR when this code's lifetime will end with IINA 1.5.0 and the swift to SwiftUI, assuming the senior developers allow it into 1.4.2.

When enhancing a UI it is important to look over the existing UI for a pattern that should be followed when adding similar controls. There are two places where something similar exists. The first place is on the `SUBTITLES` tab of the side panel:

existing2

The second place is in the `ASS Subtitles` section on the `Subtitle` tab of IINA's settings:

existing1

Notice in both cases the `Primary Subtitles` and `Secondary Subtitles` buttons are nested within a box along with the settings for which IINA supports separate values for primary and secondary subtitles.

This is what the UI added by this PR looks like:

new

The `Primary Subtitles` and `Secondary Subtitles` buttons are not nested in a box. How does the user know that these buttons only control the `Vertical position` setting?

I'm thinking the `Vertical position` setting needs to be moved up so that it is right after the `Primary Subtitles` and `Secondary Subtitles` buttons and then these controls need to be nested in a box, and aligned similar the controls seen in the `ASS Subtitles` section.

Thoughts? Do you want to make these changes and see if the senior developers will work this into IINA 1.4.2?

Another reason why it is good to use separate PRs for unrelated changes is that reviews become long when changes are combined into one PR…

About the change to add a subtitle search filter:

filtered

The problem the senior developers might have with this change is that the built-in code for downloading subtitles has been deprecated in favor of plugin implementations. One reason for the transition to plugins for downloading subtitles is to make it easier for developers to add support for other subtitle sites independent of an IINA release. Quite a few of the features that had been requested in this area of IINA, such as being able to manually specify a search string instead of the video title,  have been implemented in the plugin. So the plugin implementation is already superior to the built-in implementation. The [OpenSubtitles plugin](https://github.com/iina/plugin-opensub) has its own UI:

plugin

Adding filtering means the old the deprecated code has a feature the new plugin doesn't.

On this:
> 🔧 Compatibility Fix: Replaces a macOS 13+ specific property with a backwards-compatible constant to prevent build failures on older macOS versions.

Did an AI propose this change?

The [Building](https://github.com/iina/iina?tab=readme-ov-file#building) section of IINA's README contains this note:
> 2. Open iina.xcodeproj in the latest public version of Xcode. _IINA may not build if you use any other version._

I'm testing this with Xcode 26.3 running under macOS 15.7.4. Right clicking on `NSTextAttachment.character` (in the `develop` branch) shows me this source that indicates `character` is available from macOS 10 onward:
```swift
extension NSTextAttachment {

    @available(macOS 10.0, *)
    public class var character: Int { get }
}
```

Building the PR with Xcode 26.3 failed for me with this error:
```text
/Users/low-batt/Documents/builds/pr-tests/iina/iina/LanguageTokenField.swift:131:50: error: 'NSAttachmentCharacter' has been renamed to 'NSTextAttachment.character'
    let attachmentChar = Character(UnicodeScalar(NSAttachmentCharacter)!)
                                                 ^~~~~~~~~~~~~~~~~~~~~
                                                 NSTextAttachment.character
AppKit.NSAttachmentCharacter:3:12: note: 'NSAttachmentCharacter' was obsoleted in Swift 4.2
public var NSAttachmentCharacter: Int { get }
```

Support for Swift 4.2 was added in Xcode 10.

What version of Xcode are you using that required switching from `NSTextAttachment.character` to `NSAttachmentCharacter`?

low-batt:
The Spelling failure is due to a new release of [typos](https://github.com/crate-ci/typos) that is detecting an existing miss-spelling, unrelated to this PR. I've pushed a commit to [develop](https://github.com/iina/iina/commits/develop/) that fixes this. Rebasing this PR will correct this failure.

The build failure is due to the use of `NSAttachmentCharacter` as I discussed above.

getong:
Currently, iina does not recognize local file with the same name vtt subtitle, but vlc works.
I change to convert vtt to srt , but it still not work.
srt subtile file, It works sometimes, but sometimes it does not work.
But vlc always work.
Hope this pr works.

I finally find the bug.
when use iina open a single video file, it will find the vtt subtitle, it works.
But if iina open at first, add many video into playlist, iina does not search for local vtt subtitle files.
Please consider this bug, and fix it.

Issue #5993, opened 2026-04-13
Feature Request: Native Apple Glass (Liquid Glass / Frosted UI) Theme for macOS


Mikasa-san:
**What you want IINA to do:**

I would like IINA to add an optional Apple Liquid Glass style theme for macOS, with a more modern transparent layered look for the player window and controls.

**What IINA does currently:**

Currently, IINA has a clean design, but it does not offer a Liquid Glass style appearance.

**Why you think this should be added:**

This would make IINA look more modern and more consistent with Apple’s latest design language on macOS. It could also make the player feel more polished and visually appealing while remaining optional for users who prefer the current style.

**Examples of other projects that have something similar:**

Apple’s recent macOS interface elements that use a Liquid Glass style appearance.

LH-and-FPGA:
<img width="656" height="407" alt="Image" src="https://github.com/user-attachments/assets/80f09cc2-d78a-4190-becf-d8b7056b7433" />

I’m not part of the IINA team, but I’m also interested in this direction. I’ve been prototyping a Liquid Glass-style appearance in my own fork, and it’s still under development. The idea is to make it an optional theme, so it wouldn’t affect users who prefer the current design. If I can get it into a polished enough state, I’d like to submit a PR and let the maintainers evaluate whether it makes sense for the project.

LH-and-FPGA:
<img width="467" height="114" alt="Image" src="https://github.com/user-attachments/assets/59dcdb3e-9dc0-4327-8795-d014e596f611" />

frieza213:
@LH-and-FPGA thats so cool, also any chances the video progress bar will also look the the volume bar with the circular type of progress indicator?

<img width="1304" height="256" alt="Image" src="https://github.com/user-attachments/assets/84108b9a-5ced-4b6e-8bd7-1ef49d6aa52b" />

ydzhou:
BTW, any plan to revamp the bottom-style control? I like how Apple implements in their TV App.
<img width="1712" height="1012" alt="Image" src="https://github.com/user-attachments/assets/b700ebd0-62ef-446c-b966-fd4995146e9f" />

uiryuu:
Fixed in #6132.

uiryuu:
[...]

PR #6001, opened 2026-04-15
Adopt Liquid Glass design for macOS 26

LH-and-FPGA:
- I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)
- This implementsissue #5993.

---

**Description:**
# Summary

Adopt Apple's Liquid Glass design language (introduced in macOS 26) across IINA's UI. All changes are gated behind `#available(macOS 26, *)`, older systems remain unaffected.

### Initial Window

- Replace `NSVisualEffectView` + gradient overlay with a full-size `NSGlassEffectView`, making the welcome window transparent with a glass material backdrop
- Hide the legacy blur layer, gradient layer, and separator on macOS 26
- Window is set to `isOpaque = false` with clear background to allow desktop show-through

### Floating Control Bar (OSC)

- Introduce `DraggableGlassEffectView` (`NSGlassEffectView` subclass) to replace the `NSVisualEffectView`-based control bar
- Migrate all subviews (play/pause, volume, timeline, toolbar buttons) and their Auto Layout constraints into the glass container
- Preserve drag-to-reposition functionality by forwarding mouse events from the glass view back to `ControlBarView`
- Fix show/hide fade animations (`fadeableViews`) to target the glass view instead of the hidden VEV

### Sliders

- `VolumeSliderCell`: Skip custom `drawBar` override on macOS 26 to allow system Liquid Glass track rendering
- `PlaySliderCell`: Skip custom `drawKnob` override on macOS 26 to allow system Liquid Glass knob animation
- `VolumeSlider`: Replace custom cell with standard `NSSliderCell` on macOS 26 for full press-to-expand glass knob effect

### Other UI Elements

- Add `applyLiquidGlass(cornerRadius:)` extension on `NSVisualEffectView` for reusable VEV → Glass migration (hides VEV, reparents subviews, migrates constraints)
- Apply Liquid Glass to:
  - OSD overlay
  - sidebar
  - bottom control bar
  - title bar
  - PiP overlay
  - buffer indicator
  - additional info view
- `MiniPlayer`: Apply Liquid Glass to background, playlist wrapper, and close button

## Files Modified

- `ControlBarView.swift` — Glass reparenting + `DraggableGlassEffectView`
- `MainWindowController.swift` — Conditional Liquid Glass setup, fade animation fix
- `InitialWindowController.swift` — Transparent window + full-size glass
- `Extensions.swift` — `applyLiquidGlass()` helper
- `MiniPlayerWindowController.swift` — MiniPlayer glass effects
- `VolumeSlider.swift` / `VolumeSliderCell.swift` — Standard cell swap
- `PlaySliderCell.swift` — System knob rendering

## Compatibility

- All changes gated behind `#available(macOS 26, *)`
- Requires Xcode 26+ to compile
- macOS 25 and earlier: zero behavior change


uiryuu:
First of all, thanks for contributing to IINA! Adopting the liquid glass is something that we definitely want to do. However please note that because we also have internal plans to update/revamp UI in the following releases, we might ask you to change (or even revert back) some of the changes.

Some issues I've noticed so far:
- When the OSC layout is not floating, a dangling `NSGlassEffectView` is still in the window
- The OSD doesn't fade; sometimes artifacts appear on the corners of the OSD
- As the dev team as discussed before, we must have an option to let the user to opt out the Liquid Glass. Such that people don't like the new glass effect can still use the old apperances.

LH-and-FPGA:
Thanks for taking a look, @uiryuu, and for the heads-up about the planned UI revamp.

Before I sink more time into polishing this, would it help to clarify scope? A couple of questions:

- Is there a rough timeline or design direction for the internal revamp you can share? If Liquid Glass adoption is likely to land as part of that work, I'd rather keep this PR minimal.
- Are you open to landing this in stages, or would you prefer one PR that covers everything including the opt-out?

On the concrete issues:

1. Dangling NSGlassEffectView in non-floating OSC layouts: I'll guard the glass setup behind the floating-layout check.
2. OSD fade + corner artifacts: I'll fix the fade target and add a corner mask on the glass view.
3. Opt-out preference: fully agree this needs to exist; happy to add it.

Happy to go in whichever direction works best for the team.

LH-and-FPGA:
Add checkbox to toggle Liquid Glass effect on UI panel.

LH-and-FPGA:
Fix OSC and OSD as well. But the OSD is a little bit dark rn. Maybe I need to adjust the effect?

uiryuu:
> Is there a rough timeline or design direction for the internal revamp you can share? If Liquid Glass adoption is likely to land as part of that work, I'd rather keep this PR minimal.

There's a revamp of the settings pages on going (https://github.com/iina/iina/tree/new-pref-ui), and we are trying to land it in 1.5.0. In this version we want to give IINA a refreshed look, including the settings, Liquid Glass among others. We've discussed about adopting Liquid Glass quite a while ago and I had some prototype to see how does it look like, but nothing serious was implemented. But either ways we want to keep the PR simple and we can always improve it later together.

> Are you open to landing this in stages, or would you prefer one PR that covers everything including the opt-out?

As I said above, I'd like to keep this PR simple. But I'd like to have the opt-out in preference before merging. Of cause we have to listen to other devs' input on this as well.

I'll check other fixes later, thanks.

low-batt:
This has been fixed.

~With the latest commits this is what IINA looks like for me running under macOS 26.4. This is with the OSC layout configured to be `Bottom`. Looks like part of the floating OSC is visible?:~
bottom-osc

low-batt:
Update: Setting name has been changed to `Liquid Glass`, although there are other concerns about this setting. See comments below.

For the name of the [Liquid Glass](https://developer.apple.com/documentation/technologyoverviews/liquid-glass) setting, I'm thinking `Use Liquid Glass effect` could be shortened to `Use Liquid Glass` or even just `Liquid Glass`. See what the other reviewers think.

The floating OSC with `Use Liquid Glass effect` enabled:

transparent

With `Use Liquid Glass effect` enabled and the macOS `Reduce transparency` system setting enabled (found under `Display` on the `Accessibility` tab):

reduced

With `Use Liquid Glass effect` disabled:

disabled

With `Use Liquid Glass effect` disabled and `Reduce transparency` enabled:

disabled-reduced

lhc70000:
Thanks for the PR. I don’t have time to review the code right now, but I’ll do it in a few days. Here’s a general direction:

- The first step should be to implement the effect itself without changing the settings UI. As other maintainers mentioned, we’re rewriting the settings window.
- I suggest that Liquid Glass should become an option alongside the existing Dark and Light choices. This is because it doesn’t provide control over the material or translucency. There’s no Dark+Glass or Light+Glass combination.
- The code needs to handle switching between Liquid Glass and NSVisualEffectView on the fly. It might be better to encapsulate all these logic into a class.
- We need to consider which windows and views can be glassified: the Welcome window, OSC, OSD, and sidebars. Can the user customize this behavior?

LH-and-FPGA:
Thank you for the feedback.

I’ll fix the OSC issue first and adjust the naming of the Liquid Glass option. Maybe we should discuss a little bit about how the switch works? Does that mean I need to add a separate Liquid Glass control on "Theme"?

As for the scope, the current changes are limited to the OSC, OSD, and Welcome window. I was also considering applying the effect to the inspector for a more consistent experience, but I’m happy to follow the maintainers’ preferred direction here since I’d like to keep this PR as simple as possible.

I’ll also look into supporting switching between Liquid Glass and NSVisualEffectView on the fly. That said, I have not yet found many existing projects that handle this kind of runtime switching, so I may need to explore the best way to structure it.

LH-and-FPGA:
The OSC is fixed. I also adjusted the alpha value to make the OSC a little bit more transparent.

LH-and-FPGA:
image

low-batt:
@lhc70000 I'm totally confused by this:
> There’s no Dark+Glass or Light+Glass combination.

The OSC screenshots in this PR show Dark+Glass and Light+Glass (plus a few other combinations)?

low-batt:
After watching the WWDC 2025 video [Meet Liquid Glass](https://developer.apple.com/videos/play/wwdc2025/219/) I now understand the comment from @lhc70000 about a Liquid Glass setting should really be a new value for the `Theme` setting, as in Liquid Glass is not light or dark, it is adaptive.

However the IINA `Theme` setting has a `Match System Appearance`. That means this IINA setting is tied to the macOS setting. In macOS Tahoe the macOS system setting `Appearance` still has the same choices, `Auto`, `Light` and `Dark`. As Apple did not add Liquid Glass to that setting, I don't think IINA should change its `Theme` setting.

At the moment, I'm thinking the current separate `Liquid Glass` setting is appropriate.

uiryuu:
Currently, I don't see the glass is adaptive to the background and change its appearance. I highly suspect that the current implementation is wrong. 

As per apple's documentation:
> `NSGlassEffectView` only guarantees the `contentView` will be placed inside the glass effect; arbitrary subviews aren’t guaranteed specific behavior with regard to z-order in relation to the content view or glass effect.

For example, for the floating OSC, we should set the OSC view to the `contentView` of the glass effect view, instead of using the glass as the OSC.

@LH-and-FPGA Can you try to implement the glass in this way?

> We need to consider which windows and views can be glassified: the Welcome window, OSC, OSD, and sidebars. Can the user customize this behavior?

I don't feel like we give the user the ability to choose which views to become glass. That's too complicated to me. If we want to give the user some flexibility, we might provide the user a option to either use less glass or use more glass. And we define the 2 sets of glassifiable views.

(edited) For now, I think the the initial window shouldn't be a whole glass. The only fully glass window I can find throughout the system is the Spotlight, and you can argue that it isn't a window. Maybe we can try to make the buttons in the initial window to be glassified to make it pretty.

low-batt:
The OSC should probably be closer to the transparency seen in the screenshots provided by @miloshimself in discussion #5988. The Apple [Meet Liquid Glass](https://developer.apple.com/videos/play/wwdc2025/219/) video provides some useful advice on how Liquid Glass should be used.

uiryuu:
Just for reference, this is QuickTime's OSC. They probably used `NSGlassEffectView.Style.regular` here:

image

LH-and-FPGA:
Thanks for the suggestions.

On the adaptive background feature: I’m not sure we should implement it. Apple does this on the OSC in iPhone/iPad, but if you play content where the background changes rapidly, the transitions aren’t that smooth. I’d rather keep the OSC transparent but consistent. That said, if you’d like to try it, I can add a build flag so we can test both implementations side by side.

On the welcome window: I could add a build flag so the maintainer can choose the appearance. Personally, I like having some unique touches that set this app apart from other players as long as readability isn’t compromised.

On transparency: I’ll drop the alpha value to 0.1.
I’ll also extend Liquid Glass to the title bar and the top/bottom OSC.

LH-and-FPGA:
I changed it to NSGlassEffectView.contentView. You can switch between the two implementations by changing kUseGlassContentViewAPI in ControlBarView. Once we decide which version to keep, I’ll remove the other one. I tested it with a few videos, but I couldn’t see any difference on my side no matter what kind of video frame on the back.

uiryuu:
Thanks for the update. I've found some buggy behavior when `kUseGlassContentViewAPI = false`, so I think we would definitely use the updated version.

| `kUseGlassContentViewAPI = false`                            | `kUseGlassContentViewAPI = true`|
| ----------------------------------- | ----------------------------------- |
| original | fixed |

However, I can only observe the OSD being adaptive, the OSC is always dark or light according to the overall appearance. While I agree with @LH-and-FPGA that maybe we shouldn't make the OSD change color so much, but I just want to know what's the difference between the OSD and OSC in the current implementation? Can we control the glass appearance?

LH-and-FPGA:
I updated my code using the newly recommended Liquid Glass methods. I still need some time to figure out why the Liquid Glass effect in OSC remains unchanged.

lhc70000:
This implementation still requires further debugging and refinement. I will merge your work into a new branch `liquid-glass` for easier collaborative development. Please feel free to add any additional fixes by submitting PRs targeting the new branch.

LH-and-FPGA:
Understood, thank you very much. I’m currently occupied with another task, but I plan to begin debugging and further development in early May.

LH-and-FPGA:
@lhc70000 @low-batt Hi! I'm back and would like to ask whether this PR is still being considered. Should I open further fixes and adjustments against the liquid-glass branch going forward? Thanks!

low-batt:
@LH-and-FPGA Sorry I'm slow to respond. I've not been following this PR closely as I defer to the other developers on UI issues. There is a lot of work going on right now addressing different issues, so developers might be slow to respond.

I see the [liquid-glass](https://github.com/iina/iina/tree/liquid-glass) branch has been created and a commit from you has been merged. When I am unable to get timely advice I just plow ahead. The more fully complete and working a PR is, the greater chance of it being merged. I accept that by plowing ahead without guidance sometimes my PR might be rejected and my changes end up in the trash.

In this case providing support for liquid glass is a part of the next IINA release. So PRs will get attention. Up to you, but I would move head with a new PR against the `liquid-glass` branch.

PR #6008, opened 2026-04-16
PiP: add close-window entry behavior

DavidAGInnovation:
## Summary
- add a `Close` option for the existing "When entering Picture-in-Picture" window behavior setting
- keep the player running when PiP closes the source window, then recreate the main window state when PiP exits
- leave the existing `Do nothing`, `Hide`, and `Minimize` behaviors unchanged

## Why
Some PiP workflows are cleaner if entering PiP removes the source player window entirely instead of leaving it visible, hidden, or minimized.

## Validation
- `xcodebuild -project iina.xcodeproj -scheme iina -configuration Debug -destination platform=macOS CODE_SIGNING_ALLOWED=NO build`
- `git diff --check`


uiryuu:
What's the difference between hidden and closing from a user's perspective? Can you give an example of such "PiP workflow"?

DavidAGInnovation:
From the user side, `Hide` keeps the player window alive and simply orders it out, so exiting PiP brings that same window back in its previous state. `Close` removes the source window entirely while PiP is active, which can feel cleaner if the user treats PiP as the only remaining surface and does not want an associated hidden window hanging around.

One example is using PiP as a temporary detached viewer while working across many windows and Spaces. In that case, a user may want to enter PiP to leave only the PiP window on screen, without also keeping a hidden player window that reappears later unless PiP is explicitly exited.

DavidAGInnovation:
For context, the motivation is to support the same user-facing style as on Apple TV on macOS: entering PiP collapses playback into the floating PiP viewer, rather than leaving the regular player window visible/minimized.

DavidAGInnovation:
- Rebased pip-close-window onto current upstream/develop.
- Dropped the duplicate CI/download script commit that is already upstream.
- Added the missing new-settings UI localization for the fourth PiP behavior: Close.

DavidAGInnovation:
@low-batt

DavidAGInnovation:
image

As a reminder, this is so that only the window from the right appears when PiP is used.

Can this be merged? @uiryuu @low-batt

uiryuu:
Can you rebase this branch? Note that we are not maintaining the old preferences, so please discard all changes in the old preference, and implement them in the new settings, thanks.

DavidAGInnovation:
Closing this in favor of a smaller follow-up that makes Hide the default Picture-in-Picture window behavior, matching the native Apple TV-style collapse without adding a separate Close-source-window mode.

Issue #5972, opened 2026-03-28
Lack of confirmation before deleting entire key configuration file


Openstreetmapler:
### IINA Version

1.4.1 Build 160

### Release Channel

Release

### macOS Version

12.7.5 (Monterey)

### Steps to Reproduce

1. Open Settings
2. Go to Key Binds settings page
3. If you don’t have any custom key configuration, create them
4. Press the lower-left minus button while the custom key configuration entry is selected

### Expected Behavior

There’s a confirmation message: “Are you sure you want to delete the key configuration ‘LOREM IPSUM’?”

### Actual Behavior

No confirmation message, key configuration is deleted without warning.

The issue with this is that I keep confusing the two minus buttons with each other: I often click on the minus button for deleting the whole key configuration when I only want to remove a single entry on the right pane (which has its own minus button)!

### Crash Report or Process Sample

Not applicable

### How often can you reproduce the issue

Always

### Required Actions

- [x] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [ ] I tested with the Nightly build and confirmed the issue is not fixed.
- [ ] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

Openstreetmapler:
By the way, I need some help on how to create a key command “Resize window so that its height (width for vertical videos) is 720 px” – my attempts to do so (`set autofit-larger 720`, `autofit-larger=720`, `set geometry 720x720`) have been fruitless.

low-batt:
Instead of IINA asking for confirmation it might be possible for IINA to move the key binding configuration to the macOS trash and then allow it to be "put back"  from the trash. The UI would have to make it clear that is what is happening.

I'm a _junior_ IINA developer and defer to the senior developers on the UI design issues. I'll leave it to them to decide how to address this problem.

As there is no recovery from accidentally deleting a key binding configuration, I'm going to label this as a bug, rather than an enhancement request.

On the key binding question…

We use issues to track work. That works better if separate problems have their own issue. I've created issue #5975 for that.

Openstreetmapler:
@low-batt Some extra info for you:

> As there is no recovery from accidentally deleting a key binding configuration, I'm going to label this as a bug, rather than an enhancement request.

For your interest: I’ve recovered my configuration from my Time Machine backups.

> On the key binding question…
> 
[...]

PR #5572, opened 2025-06-30
Close add setting to prevent accidental scrolling, #5451

low-batt:
This commit will:
- Add settings `disablePlaySliderScrolling` and `disableVolumeSliderScrolling` to `Preferences`
- Add a `scrollWheel` method to `PlaySlider` that ignores scroll wheel events if `disablePlaySliderScrolling` is enabled
- Add a `VolumeSlider` class with a similar `scrollWheel` method and change `MainWindowController.xib` to use this class for the volume slider in the OSC
- Add a `MiniPlaySlider` class with a similar `scrollWheel` method and change `MiniPlayerWindowController.xib` to use this class for the play slider in the mini player

This allows users to disable scrolling within the playback and volume sliders to prevent accidentally changing a slider when using input devices where it is easy to unintentionally generate a scroll event, such as a mouse with a scroll wheel that spins freely.

- [x] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)
- [x] This implements/fixes issue #5451.

---

**Description:**

PR #6016, opened 2026-04-20
Merge the new Settings Window

lhc70000:
Merging the new Settings Window since 1.4.2 is released, and we are working towards 1.5.0.

It is not finished yet, but we need the infrastructure to incorporate it with other PRs.

The new Settings Window drops all XIB files and utilizes Swift's result builder syntax to achieve a declarative UI creation for most views.

**Remaining tasks** in the new Settings Window:

- Localization: need a script to migrate old translations to the new strings file
- Searching
- Online plugin list
- Preference key (the old preference window is expected to be removed after 1.5.0)
- Issue https://github.com/iina/iina/issues/5972 add a confirmation before deleting a configuration file
- Better key binding display

**New project-wise utilities**

The interfaces are subject to changes. Feel free to discuss.

- A new `Dialogs` class for alerts, prompts, open/save panels, using the `async/await` API for sheet windows. New code should use or extend this class.
- A new `ConstraintsSugar` file with several extensions adding Auto Layout Constraints, e.g `view.padding(.top(8))`.


lhc70000:
When merging, please don't squash. We keep the commit history for large features.

uiryuu (approved):
LGTM. Need this to be merged first to avoid further conflicts

low-batt:
Before merging the build has to pass!

There are compilation errors due to missing availability tests for methods not available in Catalina like this one in `SettingsPageSubtitles`:
```text
/Users/runner/work/iina/iina/iina/Pages/SettingsPageSubtitles.swift:280:37: error: 'init(systemSymbolName:accessibilityDescription:)' is only available in macOS 11.0 or newer
    let boldButton = SButton(image: NSImage(systemSymbolName: "bold", accessibilityDescription: nil)!)
```

I will look closer at this PR tomorrow.

iina/Pages/SettingsPageVideoAudio.swift
```diff
@@ -0,0 +1,259 @@
...
+
+    audioDevicePopUp.target = self
+    audioDevicePopUp.action = #selector(audioDeviceAction)
+
+    view.addSubview(audioDevicePopUp)
+    audioDevicePopUp.padding(.top(-4), .bottom(8), .leading(SettingsSubListView.padding), .trailing(8))
+  }
+
+  @objc func audioDeviceAction(_ sender: Any) {
+    let device = audioDevicePopUp.selectedItem!.representedObject as! [String: String]
```
low-batt: On a Mac with multiple audio output devices:
- Open the new settings window
- Click on `Audio`
- Attempt to change the `Preferred audio device` setting
- IINA crashes on this line

The crash is likely because `PlayerCore.getAudioDevices` was changed from returning `[[String: String]]` to returning `[MPVAudioDevice]` in PR #5616. This change was required because mpv creates two audio devices for each device, one for use with Core Audio, and one for use with AVFoundation. When displaying audio devices IINA must only show the ones that are usable with the currently selected audio driver.

iina/Pages/SettingsPageGeneral.swift
```diff
@@ -0,0 +1,195 @@
...
+              .bindTo(.pauseWhenLeavingFullScreen)
+            SettingsItem.Switch()
+              .bindTo(.pauseWhenGoesToSleep)
+          }
+      }
+
+      SettingsListView {
+        SettingsItem.Switch()
+          .image(name: "macwindow.badge.plus")
+          .bindTo(.alwaysOpenInNewWindow)
```
low-batt, line 67: `Always open media in new window` was replaced by `Allow multiple player windows` with subordinate settings `Group media in playlist when opening multiple files at once` and `Allow opening the same file in multiple windows` in PR #5868.

iina/Pages/SettingsPageUI.swift
```diff
@@ -0,0 +1,440 @@
...
+      SettingsListView {
+        SettingsItem.Switch()
+          .image(name: "arrow.right.and.line.vertical.and.arrow.left")
+          .bindTo(.controlBarStickToCenter)
+        SettingsItem.Switch()
+          .image(name: "ruler")
+          .bindTo(.showChapterPos)
+        SettingsItem.Switch()
+          .image(name: "slider.horizontal.below.square.and.square.filled")
+          .bindTo(.showRemainingTime)
```
low-batt, line 118: The `Show remaining time instead of total duration` has a subordinate setting `Base remaining time on playback speed` added in PR #3820.

iina/Pages/SettingsPageUI.swift
```diff
@@ -0,0 +1,440 @@
...
+      }
```
low-batt, line 127: Two new settings are missing, `Prevent scroll wheel from changing playback position` and `Prevent scroll wheel from changing volume`. These were added in PR #5572.

iina/Pages/SettingsPageUI.swift
```diff
@@ -0,0 +1,440 @@
...
+    }
+  }
+
+  private func sectionThumbnail() -> [NSView] {
+    return section {
+      SettingsListView(title: .text_ThumbnailPreview) {
+        SettingsItem.Switch()
+          .image(name: "photo")
+          .bindTo(.enableThumbnailPreview)
```
low-batt, line 171: Missing the new subordinate setting `Include files on a mounted remote drive` added in PR #5571.

iina/Pages/SettingsPageUI.swift
```diff
@@ -0,0 +1,440 @@
...
+          .image(name: "pip.enter")
+          .withDetailView {
+            SettingsItem.PopupButton()
+              .bindTo(.windowBehaviorWhenPip, ofType: Preference.WindowBehaviorWhenPip.self)
+            SettingsItem.Switch()
+              .bindTo(.pauseWhenPip)
+          }
+        SettingsItem.Switch()
+          .image(name: "pip.swap")
+          .bindTo(.togglePipByMinimizingWindow)
```
low-batt, line 196: Missing the new subordinate setting `Only for video files` added in PR #5068.

iina/Pages/SettingsPageVideoAudio.swift
```diff
@@ -0,0 +1,259 @@
...
+          .image(name: "character.book.closed")
+          .withDetailView(
+            SettingsAccessory.LanguageSelector()
+              .bind(to: .audioLanguage)
+          )
+      }
+    }
+  }
+
+  private func sectionReplayGain() -> [NSView] {
```
low-batt, line 143: As `Replay Gain` is really an audio thing it is slightly odd to have its own section like this. The section was intended to be a quick and dirty addition instead of doing a full refactoring and splitting audio and video into their own tabs.

Should `Video/Audio` be split into separate tabs for the new settings?

iina/Pages/SettingsPageSubtitles.swift
```diff
@@ -0,0 +1,520 @@
...
+  private func sectionAutoLoad() -> [NSView] {
+    return section {
+      SettingsListView(title: .text_AutoLoad) {
+        SettingsItem.PopupButton()
+          .image(name: ["bolt.badge.automatic", "bolt.badge.a"])
+          .bindTo(.subAutoLoadIINA, ofType: Preference.IINAAutoLoadAction.self)
+        SettingsItem.General(title: .text_Advanced)
+          .withExpandingDetailView {
+            SettingsItem.LongInput()
+              .bindTo(.subAutoLoadPriorityString)
```
low-batt, line 55: Missing `Please enter a comma-separated list of strings.` help text below setting.

iina/Pages/SettingsPageSubtitles.swift
```diff
@@ -0,0 +1,520 @@
...
+              .controlSize(.small)
+            SettingsItem.LongInput()
+              .bindTo(.subAutoLoadSearchPath)
```
low-batt, line 59: Missing `Directories are separated by colons (;). Relative paths and wildcards (*) at the end are allowed.` help text below setting.

iina/Pages/SettingsPageSubtitles.swift
```diff
@@ -0,0 +1,520 @@
...
+
+fileprivate class SubtitlesColorView: SettingsAccessory.Base {
+  override init(l10n: SettingsLocalization.Context) {
+    super.init(l10n: l10n)
+
+    let colorLabel = makeLabel(.text_Color)
+    let colorWell = makeColorWell()
+
+    let backgroundLabel = makeLabel(.text_Background)
+    let backgroundWell = makeColorWell()
```
low-batt: I'm not finding the code where the colorwell is tied into settings. Something is not right as the colors all show as white in the new settings, but not in the old settings.

iina/Pages/SettingsPageKeyBindings.swift
```diff
@@ -0,0 +1,530 @@
...
+          .image(name: ["playpause.circle", "playpause"])
+          .withHelpLink(AppData.wikiLink.appending("/Manage-Key-Bindings"))
+      }
+    }
+  }
+  
+  private func sectionConfiguration() -> [NSView] {
+    return section {
+      SettingsListView {
+        SettingsItem.General(title: .text_KeyBindingSet)
```
low-batt, line 46: Missing the help text to let the user know they can double click on a binding to edit it that the old settings have.

iina/Pages/SettingsPagePlugin.swift
```diff
@@ -0,0 +1,919 @@
...
+        descLabel.textColor = .secondaryLabelColor
+        
+        let stackView = NSStackView(views: [nameStackView, descLabel])
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        stackView.orientation = .vertical
+        stackView.alignment = .leading
+        
+        self.aboutBtn = NSButton()
+        aboutBtn.size(width: 18, height: 18)
+        aboutBtn.image = .findSFSymbol(["info.circle"])
```
low-batt, line 339: I was confused by the use of the info button. I expecting that would just describe the plugin. I didn't expect plugin settings to be associated with that button.

low-batt:
I could not figure out how to easily see the full option name:
mpv-options

iina/Pages/SettingsPageKeyBindings.swift
```diff
@@ -0,0 +1,530 @@
...
+
+  @objc func showConfFileAction(_ sender: AnyObject) {
+    let url = URL(fileURLWithPath: currentConfFilePath)
+    NSWorkspace.shared.activateFileViewerSelecting([url])
+  }
+
+  @objc func deleteConfFileAction(_ sender: AnyObject) {
+    guard isCurrentConfigEditable() else { return }
+    do {
+      try FileManager.default.removeItem(atPath: currentConfFilePath)
```
low-batt, line 357: Issue #5972 is about the lack of a confirmation before deleting a configuration file. Because the file is not moved to the trash, there is no recovery.

iina/Pages/SettingsPageKeyBindings.swift
```diff
@@ -0,0 +1,530 @@
...
+    searchField.bind(
+      .predicate, to: mappingController, withKeyPath: "filterPredicate",
+      options: [
+        .predicateFormat: "(readableAction contains[cd] $value) OR (prettyCommand contains[cd] $value)",
+      ]
+    )
+
+    editorView.addSubview(editorStackView)
+    editorStackView.padding(.leading(16), .trailing(0), .top(0), .bottom(12))
+
```
low-batt, line 205: I do not see `Display raw values`, like the old settings has? Very useful to be able to see the raw values.

The nice cleaner easier to read presentation really emphasizes the need to display the keys in some sort of order other than "random mess".

iina/SettingsWindow.swift
```diff
@@ -0,0 +1,409 @@
...
+
+  func tableViewSelectionDidChange(_ notification: Notification) {
+    let tableView = notification.object as! NSTableView
+
+    if let prevIndex = (notification.userInfo?["NSTableViewPreviousRowSelectionUserInfoKey"] as? NSIndexSet),
+       prevIndex.count > 0 {
+      let prevRow = prevIndex.firstIndex
+      loadPage(at: tableView.selectedRow > prevRow ? tableView.selectedRow - 1 : tableView.selectedRow)
+      tableView.removeRows(at: IndexSet(integer: prevRow + 1), withAnimation: [.effectFade, .slideUp])
+      tableView.insertRows(at: IndexSet(integer: tableView.selectedRow + 1), withAnimation: [.effectFade, .slideDown])
```
low-batt:
```suggestion
      let options: NSTableView.AnimationOptions = Preference.bool(for: PK.disableAnimations) ?
        [] : [.effectFade, .slideDown]
      tableView.removeRows(at: IndexSet(integer: prevRow + 1), withAnimation: options)
      tableView.insertRows(at: IndexSet(integer: tableView.selectedRow + 1), withAnimation: options)
```

iina/SettingsWindow.swift
```diff
@@ -0,0 +1,409 @@
...
+    }
+    if newOrigin.y > rect.origin.y {
+      newOrigin.y = rect.origin.y
+    }
+    if rect.origin.y > newOrigin.y + clipView.bounds.height - rect.height {
+      newOrigin.y = rect.origin.y - clipView.bounds.height + rect.height
+    }
+    
+    NSAnimationContext.runAnimationGroup({ context in
+      context.duration = 0.25
```
low-batt:
```suggestion
      context.duration = AccessibilityPreferences.adjustedDuration(0.25)
```

iina/SettingsWindow.swift
```diff
@@ -0,0 +1,409 @@
...
+      }
+    }
+
+    guard let firstVisibleTitle = firstVisibleTitle,
+          let sectionNameStackView = sectionNameStackView else { return }
+
+    let titleIndex = sectionNames.firstIndex(of: firstVisibleTitle) ?? 0
+
+    NSAnimationContext.runAnimationGroup({ context in
+      context.duration = 0.1
```
low-batt:
```suggestion
      context.duration = AccessibilityPreferences.adjustedDuration(0.25)
```

low-batt (commented):
A few suggested changes to support the `Disable animations` setting.

lhc70000:
I will add the missing options. For any new features that are not in the current Preferences window, I will add them later.

> I could not figure out how to easily see the full option name:

There is no easy way to do so. NSTableView needs to be embedded in a scroll view to show table headers. I updated the default column sizes for now.

iina/Pages/SettingsPagePlugin.swift
lhc70000: I think it's commonly used in iOS (e.g. WiFi list, Bluetooth devices list). However, if you can find a better SFSymbol, feel free to suggest one.

iina/Pages/SettingsPageKeyBindings.swift
lhc70000: This requires a more intuitive design. I will improve it later.

iina/Pages/SettingsPageVideoAudio.swift
lhc70000: Indeed it's slightly odd, but since it has many sub-settings, I think it's OK to have its own section. We can also rename the sub-settings to like "Replay gain adjustment", "Replay gain fallback".

> Should Video/Audio be split into separate tabs for the new settings?

Currently it's fine, since we don't have much settings in this page.

iina/AppDelegate.swift
```diff
@@ -11,6 +11,7 @@ import MediaPlayer
 import Sparkle
 
 let IINA_ENABLE_PLUGIN_SYSTEM = true
+let IINA_ENABLE_NEW_SETTINGS = false
```
uiryuu, line 14:
```suggestion
let IINA_ENABLE_NEW_SETTINGS = (InfoDictionary.shared.buildType == .debug)
```

Maybe we enable the new settings automatically for debug builds, so that the nightly will still use the old preference panel.

iina/Base.lproj/Localizable.strings
```diff
@@ -145,8 +145,8 @@
 
 "preference.title" = "Settings";
 "preference.general" = "General";
-"preference.ui" = "UI";
-"preference.video_audio" = "Video/Audio";
+"preference.ui" = "Interface";
+"preference.video_audio" = "Video & Audio";
```
uiryuu, line 149: These changes have been reverted before; maybe a misoperation?

uiryuu (changes requested):

iina/Base.lproj/Localizable.strings
lhc70000: No, we want this change.

iina/AppDelegate.swift
lhc70000: It still has many features missing. We need to at least allow people to revert to the legacy Preferences window so they are not stuck. So let's do it later.

lhc70000:
Missing options added.

iina/Pages/SettingsPageVideoAudio.swift
low-batt: Sounds good.

iina/Pages/SettingsPagePlugin.swift
low-batt: If Apple is doing this, then it is fine with me.

iina/AppDelegate.swift
low-batt: I too think we should do this after addressing the currently known issues. When users first encounter the new settings window, that is when you get your best review as they look at it closely in detail. We don't want them wasting time reporting missing settings that we already know about. I think we should bring it up to the level where we don't expect them to report anything, before enabling it.

iina/SettingsWindow.swift
```diff
@@ -19,6 +19,7 @@ fileprivate extension NSView {
   }
 }
 
+@available(macOS 11.0, *)
```
low-batt, line 22: This has caused 4 new warnings in this class as it contains statements like this that are no longer needed because of the class level `@available`:
```swift
if #available(macOS 11.0, *) {
```

low-batt (approved):
As the build is working we can merge this disabled and continue enhancing.

Would be nice to cleanup those 4 compiler warnings before merging.

iina/SettingsWindow.swift
lhc70000: Removed.

PR #6017, opened 2026-04-20
Add script for generating DMG file

low-batt:
This commit will:
- Add `generate_dmg.sh` to the `other` directory
- Add `dmg_background.png` to the `other` directory

Running the `generate_dmg.sh` script generates an IINA DMG file in Xcode's build directory.

This script requires that `create-dmg` has been installed.

This script copies the new script conventions introduced in the recent changes to `download_libs.sh` by DavidAGInnovation.

---

**Description:**


low-batt:
When mounted under macOS Sequoia and earlier the volume shows IINA's icon. This does not happen on my Mac running macOS Tahoe 26.4. Posts on the net accuse Apple of breaking this feature and so far not fixing it.

Edit: Indeed the above behavior was another indication of the quality of macOS Tahoe. After upgrading to macOS 26.4.1 the icons of mounted volumes are back to being displayed.

@DavidAGInnovation Your great enhancements to `download_libs.sh` inspired me to create this script. My bash skills are rusty. If you notice something that could be done better do not hesitate to point it out.

DavidAGInnovation:
I think the `create-dmg` call is using the wrong source path. Its second positional arg is a source folder whose contents are copied into the image, so passing `IINA.app` will copy the bundle internals rather than the `IINA.app` bundle itself. That would also make `--icon "IINA.app"` miss the expected item. We probably need a staging directory that contains `IINA.app`, then pass that directory to `create-dmg`.

`-v/--verbose` looks broken at the moment: it’s shown in `printUsageHelp()` and handled in the `case`, but the `getopt` spec only allows `-h` and doesn’t include `verbose`. So `-v` should fail during option parsing before we reach the `case` block. Also, the long-option list still includes `arch`, `yt-dlp-src`, and `parallel`, which seem copied over from `download_libs.sh`.

low-batt:
@DavidAGInnovation Thanks for taking a look and reporting problems!

I moved this to a draft as the option parsing is definitely not right. I did copy that from `download_libs.sh` and missed updating the call to `getopt`. But that code isn't working for me with `download_libs.sh` either as invalid options are not rejected:
```text
low-batt@urp iina (develop $=)$ other/download_libs.sh --badoption
Starting downloads in parallel...
Downloading libarchive.13.dylib...
Downloading libass.9.dylib...
```

I'm confused about the `getopt` available on the Mac. The man page does not mention parsing long options.

I found the [create-dmg](https://github.com/create-dmg/create-dmg) `source_folder` argument confusing. From its description:
> All contents of source_folder will be copied into the disk image.

From that description you would expect the top level of the DMG volume to be populated with the contents of the `IINA.app` directory. But the generated DMG has the app and the shortcut to `/Applications`. Maybe it recognizes it is a `.app` file?

I'll check the `create-dmg` sources and look for an explanation.

low-batt:
On the `getopt` problem…

The use of `getopt` was added 2 years ago in commit 83b6725 by @semty. This is what the code in `download_libs.sh` looks like now:

```bash
# Reset in case getopts has been used previously in the shell.
if ! OPTS=$(getopt -o "h": --long "arch:,yt-dlp-src:,parallel:,help": -n 'parse-options' -- "$@"); then
  echo -e "${RED}Failed parsing options.${NC}" >&2
  exit 1
fi
```

It is not detecting invalid options:
```bash
low-batt@urp iina (develop $=)$ other/download_libs.sh --badoption
Starting downloads in parallel...
Downloading libarchive.13.dylib...
```

The man page indicates the `getopt` supplied by macOS originates from BSD and is simplistic, with no concept of long form options: 
<details>
<summary><b>man getopt:</b></summary>

```text
GETOPT(1)                                                                            General Commands Manual                                                                            GETOPT(1)

NAME
     getopt – parse command options

SYNOPSIS
     args=`getopt optstring $*` ; errcode=$?; set -- $args

DESCRIPTION
     The getopt utility is used to break up options in command lines for easy parsing by shell procedures, and to check for legal options.  Optstring is a string of recognized option letters
     (see getopt(3)); if a letter is followed by a colon, the option is expected to have an argument which may or may not be separated from it by white space.  The special option ‘--’ is used
     to delimit the end of the options.  The getopt utility will place ‘--’ in the arguments at the end of the options, or recognize it if used explicitly.  The shell arguments ($1 $2 ...) are
     reset so that each option is preceded by a ‘-’ and in its own shell argument; each option argument is also in its own shell argument.

EXIT STATUS
     The getopt utility prints an error message on the standard error output and exits with status > 0 when it encounters an option letter not included in optstring.

EXAMPLES
     The following code fragment shows how one might process the arguments for a command that can take the options -a and -b, and the option -o, which requires an argument.

           args=`getopt abo: $*`
           # you should not use `getopt abo: "$@"` since that would parse
           # the arguments differently from what the set command below does.
           if [ $? -ne 0 ]; then
                   echo 'Usage: ...'
                   exit 2
           fi
           set -- $args
           # You cannot use the set command with a backquoted getopt directly,
           # since the exit code from getopt would be shadowed by those of set,
           # which is zero by definition.
           while :; do
                   case "$1" in
                   -a|-b)
                           echo "flag $1 set"; sflags="${1#-}$sflags"
                           shift
                           ;;
                   -o)
                           echo "oarg is '$2'"; oarg="$2"
                           shift; shift
                           ;;
                   --)
                           shift; break
                           ;;
                   esac
           done
           echo "single-char flags: '$sflags'"
           echo "oarg is '$oarg'"

     This code will accept any of the following as equivalent:

           cmd -aoarg file1 file2
           cmd -a -o arg file1 file2
           cmd -oarg -a file1 file2
           cmd -a -oarg -- file1 file2

SEE ALSO
     getopts(1), sh(1), getopt(3)

HISTORY
     Written by Henry Spencer, working from a Bell Labs manual page.  Behavior believed identical to the Bell version.  Example changed in FreeBSD version 3.2 and 4.0.

BUGS
     Whatever getopt(3) has.

     Arguments containing white space or embedded shell metacharacters generally will not survive intact; this looks easy to fix but is not.  People trying to fix getopt or the example in this
     manpage should check the history of this file in FreeBSD.

     The error message for an invalid option is identified as coming from getopt rather than from the shell procedure containing the invocation of getopt; this again is hard to fix.

     The precise best way to use the set command to set the arguments without disrupting the value(s) of shell options varies from one shell version to another.

     Each shellscript has to carry complex code to parse arguments halfway correctly (like the example presented here).  A better getopt-like tool would move much of the complexity into the
     tool and keep the client shell scripts simpler.

macOS 26.4                                                                                August 1, 2015                                                                               macOS 26.4
~
(END)
```
</details>

@semty, @DavidAGInnovation Do you have a better version of `getopt` installed? Maybe the one provided by [brew](https://formulae.brew.sh/formula/gnu-getopt)?

From [getopt-example](https://github.com/util-linux/util-linux/blob/master/misc-utils/getopt-example.bash) (what brew installs):
```text
TEMP=$(getopt -o 'ab:c::' --long 'a-long,b-long:,c-long::' -n 'example.bash' -- "$@")
```

I believe I have the `getopt` macOS provides:
```bash
low-batt@urp ~$ which getopt
/usr/bin/getopt
low-batt@urp ~$ ls -la /usr/bin/getopt
-rwxr-xr-x  1 root  wheel  100912 Mar 20 01:25 /usr/bin/getopt
low-batt@urp ~$ 
```

For the script in this PR I have simplified it to use single character options and follow the usage example in the man page. At some point in the future the option parsing in the `download_libs.sh` script will need to be fixed. Other work takes priority at the moment.

@DavidAGInnovation THANKS for catching this problem!


On the [create-dmg](https://github.com/create-dmg/create-dmg#usage) `<source_folder>` argument…

I investigated the behavior of this argument:
```text
create-dmg [options ...] <output_name.dmg> <source_folder>
```

That argument is used as the value for the `-srcfolder` option when calling `hdiutil create`. From the man page for `hdiutil`:

```text
-srcfolder source
                           copies file-by-file the contents of source into image, creating a fresh (theoretically defragmented) filesystem on the destination.  The resulting image is thus
                           recommended for use with asr(8) since it will have a minimal amount of unused space.  Its size will be that of the source data plus some padding for filesystem
                           overhead.  The filesystem type of the image volume will match that of the source as closely as possible unless overridden with -fs.  Other size specifiers, such as
                           -size, will override the default size calculation based on the source content, allowing for more or less free space in the resulting filesystem.  -srcfolder can be
                           specified more than once, in which case the image volume will be populated at the top level with a copy of each specified filesystem object.  -srcdir is a synonym.
```

To me that still sounds like the contents of the `IINA.app` directory will be copied into the top level of the disk image, exactly the problem we are worried about.

I had to run tests to figure this out. My conclusion is this is yet another case of the industry producing software whose behavior does not match the documentation along with poor deceptive names for options . Based on the name of the option you would expect `hdiutil` to _fail_ if the value of `-srcfolder` is not a folder. It does not. If you specify a text file `hdiutil` copies the file into the top level of the disk image without a complaint. I also confirmed that if the value of  the `-srcfolder` option is an application, it does not treat the application as a directory. It copies the application to the top level of the disk image.

The `hdiutil` man page should cover this unexpected behavior of the `-srcfolder` option. Of course if it did that then it would become obvious that the option name is inappropriate and really should just be `-src`.

The script is working for me as seen here:

working


I am marking this ready for review.

DavidAGInnovation:
I checked my local environment and confirmed I’m using macOS `/usr/bin/getopt`, not GNU `getopt`.

I opened a separate PR to fix `download_libs.sh` option parsing without relying on GNU-style `getopt` syntax: https://github.com/iina/iina/pull/6024

That PR preserves the existing long options, including `--skip-plugins`, and now rejects unknown options like `--badoption`.

lhc70000:
Looks good to me. Since you have already included some checks, here is another check that I always run, and it would be great to also include it:

```
check_plugin_works() {
    pluginkit -a "IINA.app/Contents/PlugIns/OpenInIINA.appex"
    
    if ! pluginkit -mAvvv -p com.apple.Safari.extension | grep -q "e/IINA.app"; then
        echo "Plugin not found in Safari extension list" >&2
        exit 1
    fi

    pluginkit -r "IINA.app/Contents/PlugIns/OpenInIINA.appex"
}
```

low-batt:
I've added a modified version of that check. I made the grep check for the path to the extension. I have lots of copies of the extension installed and Safari won't let me remove them without uninstalling IINA.

PR #6019, opened 2026-04-22
chore: remove code to support macOS Catalina

uiryuu:

---

**Description:**
For the 1.5.0 release and the new preference ui, we are planning to drop support for macOS Catalina (10.15). Now minimum supporting OS is macOS 11 for Intel Macs and macOS 12 for Apple Silicon Macs.

Global search `#available(macOS 11` shows nothing; global search `#unavailable(macOS 12` shows 1 occurrence and that is necessary for identifying macOS 11.


iina/PrefPluginViewController.swift
```diff
@@ -294,6 +294,75 @@ class PrefPluginViewController: PreferenceViewController, PreferenceWindowEmbedd
     }
   }
 
+  private func handleInstallationError(_ error: Error) {
```
low-batt: What does this have to do with Catalina?

low-batt (approved):
I don't understand why the new code for plugin errors is in this PR.

Otherwise all looks good.

iina/PrefPluginViewController.swift
uiryuu: This must be a problem with git rebase, checking...

PR #6021, opened 2026-04-22
Suppress the UTI warning

uiryuu:

---

**Description:**
This commit tries to suppress the UTI warning. The new API should perform like before (filter using file extension)


low-batt (approved):
Reviewed change in detail. Pulled PR, tested. Looks good to me.

HOWEVER, in testing I noticed that IINA's open panel for File/Open allows you to select subtitle files, which results in a "Nothing to open" Alert. Seems like something to improve. This is existing behavior unrelated to this PR.

PR #6024, opened 2026-04-23
Fix download_libs option parsing

DavidAGInnovation:
## Summary
- replace the GNU-style getopt invocation with explicit Bash option parsing
- reject unknown options, missing option values, empty option values, and unexpected positional arguments
- preserve existing long options including --skip-plugins

## Tests
- bash -n other/download_libs.sh
- ./other/download_libs.sh --help
- ./other/download_libs.sh --badoption
- ./other/download_libs.sh -x
- ./other/download_libs.sh --arch
- ./other/download_libs.sh --arch=arm64 --yt-dlp-src=iina --parallel=1 --skip-plugins --help
- ./other/download_libs.sh -- --extra
- git diff --check

### DIFF
--- a/other/download_libs.sh
+++ b/other/download_libs.sh
@@ -19,12 +19,6 @@ YELLOW='\033[1;33m'
 BLUE='\033[0;34m'
 NC='\033[0m' # No Color
 
-# Reset in case getopts has been used previously in the shell.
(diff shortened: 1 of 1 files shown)


low-batt (approved):
Reviewed changes. Pulled PR and tested error reporting as well downloading. All worked.

Changes look good to me.

low-batt:
@DavidAGInnovation Thanks so much for the fix!

PR #6027, opened 2026-04-24
thumbnail: add thumbnail width option in the settings window

uiryuu:
- [x] This implements/fixes issue #885.

---

**Description:**
The preference key for the thumbnail width has been in the codebase for quite a while, so this commit just add the option to the UI, and set the width of the thumbnail view accordingly.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -2569,11 +2569,12 @@ class MainWindowController: PlayerWindowController {
         let (videoWidth, videoHeight) = player.videoSizeForDisplay
         let displayAspectRatio = CGFloat(videoWidth) / CGFloat(videoHeight)
 
-        let height = round(120 / displayAspectRatio)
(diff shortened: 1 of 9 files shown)


iina/MainWindowController.swift
```diff
@@ -2569,11 +2569,12 @@ class MainWindowController: PlayerWindowController {
         let (videoWidth, videoHeight) = player.videoSizeForDisplay
         let displayAspectRatio = CGFloat(videoWidth) / CGFloat(videoHeight)
 
-        let height = round(120 / displayAspectRatio)
```
lhc70000, line 2572: Should use `thumbnailWidth / 2`. The default value is 240px image = 120pt on screen

lhc70000 (changes requested):
Another important problem:

- When the user changes the width, the existing thumbnail cache should be cleared. Otherwise the user might see a blurred image from the cache.
- Consider adding a description label alerting the user.

PR #6028, opened 2026-04-24
New Settings Window update

lhc70000:
**Description:**

- Updated key binding page layout with the "show raw values" option
- Add confirmation before deleting a key binding set (#5972)
- Updated icons using custom SFSymbols
- Added the preference key for the new settings window (haven't added to the UI)
- Fixed several layout problems on older systems


uiryuu (approved):
LGTM

iina/Pages/SettingsPageKeyBindings.swift
```diff
@@ -47,6 +47,8 @@ class SettingsPageKeyBindings: SettingsPage {
           .withValueView(configEditor.chooserView)
         SettingsItem.Custom()
           .view(configEditor.editorView)
+        SettingsItem.Switch()
+          .bindTo(.displayKeyBindingRawValues)
```
low-batt, line 51: Glad to see this implemented. From a UI standpoint the old settings were a bit better in that this setting was not hidden. In the new settings you won't know this setting exists until you scroll all the way to the end of the table.

iina/Assets.xcassets/Symbols/custom.lock.badge.questionmark.symbolset/Contents.json
```diff
@@ -0,0 +1,12 @@
+{
+  "info" : {
+    "author" : "xcode",
+    "version" : 1
+  },
+  "symbols" : [
+    {
+      "filename" : "custom.lock.badge.questionmark.svg",
```
low-batt, line 8: The new lock badge looks nice!

iina/Pages/SettingsPageKeyBindings.swift
```diff
@@ -366,6 +376,10 @@ fileprivate class ConfigEditor: SettingsAccessory.Base {
   }
 
   @objc func addKeyMappingAction(_ sender: AnyObject) {
+    guard isCurrentConfigEditable() else {
+      Utility.showAlert("duplicate_config", sheetWindow: view.window)
```
low-batt: The table still allows selecting a key binding in a default key binding configuration. Correctly the buttons for deleting and editing are not shown. But there is no indication on the individual entry as to why that is. Maybe show the locked button on each entry and also show this alert when that button is clicked? And maybe not allow selection? Showing the locked button on each entry would catch the user's attention more than the single small lock button next to the search. Makes it clearer to new users what is going on.

low-batt (approved):
Changes look good.

A couple of suggestions on the Key Bindings tab.

low-batt:
We will need to figure out how to make `typos` happy again.

iina/Pages/SettingsPageKeyBindings.swift
lhc70000: Yes, I was also thinking about this. I tried showing the locked button in each row and it's not bad. I will update it.

iina/Pages/SettingsPageKeyBindings.swift
lhc70000: It's a trade-off. I tried several different layouts but seems that there's no good place for this option. Considering the users who want this are advanced users, and it will stay on once enabled, I think the current position is acceptable. Users who want this option can find it, just with a little bit extra hassle.

PR #6029, opened 2026-04-24
Fix auto-play when Bluetooth device connects, #6025

### LINKED ISSUE #6025: Unintended auto-play when connecting or switching Bluetooth devices
I’ve noticed that whenever I connect a Bluetooth headset to my Mac, IINA automatically resumes video playback, even if the video was previously paused. This also occurs when I switch between different Bluetooth audio devices.
This behavior started appearing about a week ago. Previously, the player would correctly remain paused until I manually resumed it.
I would appreciate it if you could look into this issue. Thank you for developing such a great media player!

low-batt:
This commit will:
- Add `PLAYONLY` and `PAUSEONLY` keys to all default key bindings files except input.conf which already contained these keys
- Change the `RemoteCommandController.enable` method to use these new keys instead of the `PLAY` and `PAUSE` keys

This commit applies the changes made by mpv in mpv PR https://github.com/mpv-player/mpv/pull/7380 to IINA. The `PLAY` and `PAUSE` keys were configured to the mpv `cycle pause` command. That definitely does not match the expected behavior of the `MPRemoteCommandCenter` [playCommand](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/playcommand) and [pauseCommand](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/pausecommand). Over backward compatibility concerns the mpv project added the new keys `PLAYONLY` and `PAUSEONLY` with the commands `set pause no` and `set pause yes` and changed their `MPRemoteCommandCenter` implementation to use the new keys. This commit will apply similar changes to IINA's `MPRemoteCommandCenter` implementation.

- [x] This implements/fixes issue #6025.

---

**Description:**

### DIFF
--- a/iina/AppDelegate.swift
+++ b/iina/AppDelegate.swift
@@ -1382,7 +1382,7 @@ class RemoteCommandController {
 
     // For each command, apply a configured keybinding or fallback to default values.
     remoteCommand.playCommand.addTarget { _ in
-      if let action = PlayerCore.keyBindings["PLAY"] {
+      if let action = PlayerCore.keyBindings["PLAYONLY"] {
         PlayerCore.lastActive.mainWindow.handleKeyBinding(action)
       } else {
         PlayerCore.lastActive.resume()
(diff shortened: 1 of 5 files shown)


lhc70000:
I wonder whether we have backward compatibility concerns since we are macOS only. What would be the down side if we simply change the action to `set pause yes/no`?

low-batt:
Two reasons I did it this way:

- This way automatically fixes the problem for users who already have their own custom key bindings as they would be missing these new keys and IINA would use its default behavior
- This keeps IINA somewhat aligned with mpv, not critical, but desirable as it reduces the work when a developer needs to compare IINA to mpv and look for changes in key bindings we might want to consider importing

PR #6030, opened 2026-04-25
Support external subtitles for audio-only media

yorkyang2333:
- [x] This implements/fixes issue #457, #972, #5618.

---

**Description:**

This PR supersedes #5944.

It reintroduces audio-only external subtitle support on top of the current `develop` branch, but narrows the mini-player path on purpose instead of continuing to push for richer subtitle fidelity there.

## Problem

IINA could auto-load external subtitle files for videos, but audio-only media still had multiple gaps. Audio files were not part of the existing external subtitle auto-match flow, and audio-only playback in the main window still needed a stable mpv-native subtitle path under `vo=libmpv`. At the same time, the open design question for music mode was not simply whether subtitles could be rendered there, but what kind of subtitle behavior actually belonged in the mini player.

## What Changed

- Added audio files to IINA's existing external subtitle matching pipeline so audio-only media can reuse the same external subtitle auto-load behavior as video files.
- Kept the main window as the primary audio-only subtitle rendering target on the existing mpv-native path.
- Refined the audio-only redraw/layout handling so plain-text subtitles do not sit awkwardly in the middle of the window.
- Narrowed the mini player to a basic `sub-text` display path only.
- Removed the richer ASS/SSA-specific observation, parsing, and higher-fidelity mini-player rendering paths.
- Corrected the music mode enter/exit toggle behavior.
- Moved `force-window=yes` later so Xcode launch no longer hangs before media load, while still switching to `force-window=immediate` after render initialization.

## Why Keep A Minimal Mini Player Text Path

The reason I did not remove mini-player text entirely is that I do not think the maintainer's reference to `#5867` and the use case for this PR are exactly the same problem. `#5867` is a lyric-oriented `.lrc` direction, where the expected experience is lyric presentation and, eventually, lyric-specific behavior. This PR is still about treating subtitle files as subtitles for general audio-only media.

That distinction matters because music is not the whole category here. Audio-only playback can also mean recordings, interviews, broadcasts, lectures, commentary tracks, and other material where the subtitle file is not functioning as stylized song lyrics but simply as timed text. For those cases, removing the richer mini-player fidelity work makes sense, but removing every mini-player text path entirely felt like an unnecessary loss in usability. The current revision therefore keeps only the weakest possible mini-player behavior: plain text from `sub-text`, with none of the ASS/SSA-specific fidelity work that previously made the mini-player path much broader.

More broadly, I do not think the lyric-style `.lrc` path and the minimal subtitle-style text path have to be mutually exclusive. My preference would be to let both forms coexist and leave the final choice to the user, because they serve different audio-only scenarios instead of one single presentation model.

## Relationship to #5867

I agree that `#5867` is the better fit for `.lrc` lyric-style presentation. That is exactly why this revision stops trying to make the mini player behave like a richer subtitle renderer. Instead, it leaves lyric-focused presentation to the lyric-specific direction, while keeping this PR centered on general subtitle-file support for audio-only media. In other words, I think these two directions are closer to complementary user-facing choices than to competing implementations of the same feature.

## Why This Differs From #5944

This PR is still a clean reimplementation on top of the current `develop` branch rather than a carry-forward of the older branch. It does not use `lavfi-complex`, a synthetic track, or a dummy black video layer, and it does not use a native `NSTextField` overlay in the main window. The main window stays on mpv-native subtitle rendering, and the mini player is now intentionally reduced to a minimal plain-text path. It also continues to rely on `sub-text` property-change events instead of polling subtitle text from `syncUI`, and it does not repeat the earlier `syncUITimer` common-mode fix because that follow-up already landed in `develop` separately.

## Validation

Validation on this revision focused on the narrowed audio-only behavior rather than richer mini-player subtitle fidelity. The checks covered external subtitle auto-match for audio files, main-window audio-only plain-text subtitle positioning, plain-text mini-player subtitle display, music mode enter/exit after the toggle fix, and Xcode launch no longer hanging before media load. The build command used for this revision remained `xcodebuild -project iina.xcodeproj -scheme iina -configuration Debug -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO build`.

## Risks / Non-goals

This PR still does not introduce online lyrics, embedded lyric parsing, or new subtitle preferences. It also intentionally does not keep full mini-player ASS/SSA fidelity or lyric-style scrolling presentation. The main window remains the primary rendering target for audio-only subtitles, while the mini player keeps only a minimal plain-text path.


iina/PlayerCore.swift
```diff
@@ -3063,13 +3115,96 @@ class PlayerCore: NSObject {
...
+  }
+
+  func musicModeVideoSizeForDisplay() -> (Int, Int) {
+    guard currentMediaIsAudio == .isAudio else {
+      return videoSizeForDisplay
+    }
+    return info.musicModeArtworkSize ?? (1, 1)
+  }
+
+  private func refreshMiniPlayerArtwork() {
```
uiryuu: Move this function to `MiniPlayerWindowController`, unless some reason that this function is better placed here?

iina/PlayerCore.swift
```diff
@@ -233,6 +233,10 @@ class PlayerCore: NSObject {
   /// functionality for the duration of this player even if the preference is `true`. But if they manually change the
   /// "music mode" status again, change this to `false` so that the preference is honored again.
   var overrideAutoSwitchToMusicMode = false
+  private var cachedMiniPlayerSubScale: Double?
+  private var isMiniPlayerSubtitleRenderSuppressed = false
+  private var isInternalMiniPlayerSubScaleChange = false
```
uiryuu: Seems like these variables are all related to the mini player UI logic, so consider moving them to `MiniPlayerWindowController`. Besides, add more comments for these variables, and if the comments are clear enough, consider shorten the variable names.

iina/PlayerCore.swift
```diff
@@ -3063,13 +3115,96 @@ class PlayerCore: NSObject {
...
+    }
+
+    setMiniPlayerSubtitleRenderingSuppressed(true)
+    let subtitleText = mpv.getString(MPVProperty.subText) ?? ""
+    if subtitleText.isEmpty {
+      miniPlayer.clearSubtitleLabel()
+    } else {
+      miniPlayer.updateSubtitleLabel(with: subtitleText)
+    }
+  }
```
uiryuu: Ditto

uiryuu (changes requested):
Pulled and tested. The subtitles worked fine in the main window; the mini player cannot handle some complex subtitles. But I guess we can do little on this since we only get the subtitle text from mpv and render ourselves.

I think the functions that only related to mini player's UI logic should be moved to `musicModeVideoSizeForDisplay`. See below for details.

iina/PlayerCore.swift
yorkyang2333: Also moved the audio-only main-window redraw workaround out of PlayerCore into MainWindowController in e62f8e9e, so the remaining core-side code is just dispatching UI updates. I also re-ran a local Debug macOS build after the refactor.

iina/PlayerCore.swift
yorkyang2333: Addressed in e62f8e9e. The mini-player-only subtitle state moved to MiniPlayerWindowController, and I renamed/commented that state there as cachedSubScale, isSubRenderSuppressed, and isManagedSubScaleChange.

iina/PlayerCore.swift
yorkyang2333: Moved this out of PlayerCore in e62f8e9e. The music mode display sizing and subtitle/artwork overlay helpers now live in MiniPlayerWindowController, and PlayerCore just forwards the relevant state changes.

yorkyang2333:
Summary of the follow-up changes in `e62f8e9e`:

- Moved the music-mode-only UI state and helpers out of `PlayerCore`.
- `MiniPlayerWindowController` now owns music mode display sizing, album-art visibility, subtitle overlay rendering, and subtitle-scale suppression bookkeeping.
- Renamed/commented the mini-player subtitle state there to make the ownership clearer.
- Moved the audio-only main-window redraw workaround into `MainWindowController`.
- `PlayerCore` now only forwards the relevant playback / track / subtitle state changes to the window controllers.

Validation re-run after the refactor:

- `xcodebuild -project iina.xcodeproj -scheme iina -configuration Debug -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO build`

The requested changes should all be addressed now, and I have re-requested review.

yorkyang2333:
Follow-up update for the latest mini player subtitle work.

This commit is broader than a pure visual tweak. It expands the mini player subtitle path while keeping the main window on the existing mpv-native rendering path.

What changed in this round:

- added additional mpv subtitle observers for the mini player (`sub-text`, `sub-text/ass`, `sub-text-ass`, `sub-text/ass-full`) so subtitle updates can drive a richer native overlay path
- replaced the previous single plain-text label in the mini player with an overlay container that can host multiple subtitle cue views
- kept the plain text fallback for non-ASS subtitles
- added best-effort ASS parsing/rendering in the mini player, including ASS script/style parsing, active event parsing, and styled cue rendering
- for external `.ass` / `.ssa` subtitles, the mini player can also parse the subtitle file directly and render cues from the current playback time instead of relying only on flattened subtitle text
- kept the mini-player-specific subtitle suppression / restore flow so mpv subtitle rendering is still hidden only while the native mini player overlay is active
- tuned the subtitle presentation afterward by removing the heavy outline look, reducing the default sizing, and adjusting text/shadow handling so the mini player output reads closer to the original style
- relaxed the centered / positioned cue layout width so mini player overlay clipping is less aggressive

Validation:
- `xcodebuild -project iina.xcodeproj -scheme iina -configuration Debug -destination "platform=macOS" CODE_SIGNING_ALLOWED=NO build`
- manual checks on mini player plain text rendering, ASS styling fallback, centered / positioned cue behavior, and switching back to the main window subtitle path

This is still best-effort mini player rendering rather than full libass parity, but it should cover substantially more subtitle cases than the previous single-label path.

yorkyang2333:
Rebased the branch on top of the latest upstream changes.

uiryuu:
Thanks for updating the PR.

I tested the miniplayer again, I still found the quality of the subtitle is not as good as that in the main window. Also I think when people use the music mode they will probably not paying much attention to complex subtitles and expect them to be rendered on the small miniplayer screen. Even if they want the lyrics to be rendered, most people would use a `.lrc` file, in which case we are in favor with the solution in #5867.

Therefore, I want to change the goal of this PR: we only render the subtitle in the main window, and don't touch the mini player at all. So please revert back all changes for the mini player. Let's do code-level review after this.

For the subtitles rendered in the main window, I found that the srt subtitles are always rendered in the middle of the screen, which looks pretty weird. Can we do something about that?

yorkyang2333:
Thanks for the direction update. I narrowed this revision substantially, but I want to be explicit about the one point where I intentionally did not go all the way to “main window only”.

I removed the richer mini-player subtitle work entirely. There is no longer any ASS/SSA-specific observation, no direct ASS/SSA parsing, no multi-cue overlay, and no style-restoration path. The mini player now keeps only a minimal plain-text `sub-text` display path.

The reason I kept that minimal path is that I do not think the use case for this PR is identical to the lyric-oriented direction in `#5867`. I agree that `#5867` is the better fit for `.lrc` lyric-style presentation. This PR, however, is still treating subtitle files as subtitles for general audio-only media, not only for song lyrics. Audio-only playback here can also mean recordings, interviews, broadcasts, lectures, and similar material where timed text is still useful even if richer lyric-style presentation is not the goal.

Because of that, this revision removes the broader fidelity work but keeps only the weakest possible mini-player behavior: plain text from `sub-text`. My preference would be for these two forms to coexist and be left as a user choice, because the lyric-style `.lrc` path and the minimal subtitle-style text path solve different audio-only scenarios rather than one single presentation model.

In the main window, I also adjusted the audio-only plain-text subtitle positioning so it no longer sits awkwardly in the middle of the screen. I fixed the music mode enter/exit toggle path as well, and moved `force-window=yes` later so Xcode launch no longer hangs before media load.

If you want the mini player to show nothing at all in this PR, I can narrow it one step further. Otherwise, if the remaining minimal plain-text path is acceptable, I think this version is ready for code-level review.

uiryuu:
I still don't like the idea to use the miniplayer to render subtitles. My reasons:
1. For other audio-only media the users should still use the main window to properly view the subtitles.
1. The main target for the mini player is still playing music, therefore adding lyrics display is reasonable.
2. The code to messy to render subtitles for the miniplayer. We want to keep it relatively simple for maintainability.

Let's see what other devs' opinion in this...

yorkyang2333:
> I still don't like the idea to use the miniplayer to render subtitles. My reasons:
> 
> 1. For other audio-only media the users should still use the main window to properly view the subtitles.
> 2. The main target for the mini player is still playing music, therefore adding lyrics display is reasonable.
> 3. The code to messy to render subtitles for the miniplayer. We want to keep it relatively simple for maintainability.
> 
> Let's see what other devs' opinion in this...

I see... I'll rebase this branch on a regular basis to keep updated and get prepared for further review. Thanks a lot anyway!

lhc70000:
Thanks for your PR and quick updates. You said it provided better support for recordings, interviews, broadcasts, lectures, commentary track, and I agree with it. Therefore, the scope of this PR should be limited to these use cases, which separates itself from #5867.

1. We cannot have two separate mechanisms for the main window and miniplayer. My suggestion is: don't add any overlay, let mpv draw the subtitles in both modes. Maintain a reasonable window size when drawing in the main window. Display the video view instead of the default album art in music mode if drawing subtitles.
2. Could you explain why you have an audio-only subtitle layout? I think this will confuse the user (changing mpv parameters without notice/synchronization), and the apply/restoration logic is complicated.
3. Please explain your logic in `requestAudioOnlyRedraw` and `forceAudioOnlyRedraw`, including why it's necessary to have a `pendingAudioOnlyRedraw`, why choose the `0.08` and `0.2` constants, why call `displayIdle` if it's paused, etc.

uiryuu:
I also agree if we can use the same rendering path of subtitles in both main window and mini player through mpv. I just don't want custom rendering on the mini player.

yorkyang2333:
@uiryuu @lhc70000 Thanks for the detailed review and guidance! 

After reconsidering the implementation and reviewing your feedback, I completely agree that the previous approach of manually intercepting subtitle rendering and maintaining a custom overlay for the Mini Player was overly complex and error-prone. The hacky workarounds (like intercepting `sub-scale`, manually adjusting `videoAlignY`, and forcing redraws) indeed introduced unnecessary maintenance burdens and messed up the synchronization of the UI states.

I have just pushed a new commit that completely reverts those custom overlays and layout hacks. The new implementation falls back entirely to `mpv`'s native rendering path for audio-only subtitles, as you suggested.

Here is a summary of the latest changes:
1. **Removed Custom Rendering logic:** Completely removed `MiniPlayerSubtitleOverlayView`, plain text subtitle labels, and all related methods (`requestAudioOnlyRedraw`, `refreshAudioOnlySubtitleLayout`, etc.) from both `MainWindowController` and `MiniPlayerWindowController`.
2. **Removed Subtitle Property Hooks:** Restored `subScaleChanged` to its original state and removed the `MPVProperty.subText` observer, as we no longer need to extract plain text manually.
3. **Mini Player Artwork Visibility Fix:** In `MiniPlayerWindowController.swift`, I updated the `refreshArtworkVisibility()` logic. Now, when the player is in music mode, it checks if there is an active subtitle track (`isSubVisible` && `sid != 0`). If subtitles are present, it simply hides the default album art, revealing the underlying `videoView` so that `mpv` can natively draw the subtitles there.
4. **Window Sizing:** Since IINA's `videoSizeForDisplay` already handles fallback dimensions when no video track is present (`AppData.widthWhenNoVideo`), we don't need to manually force window resizing. `mpv` handles the rendering bounds correctly on its own.

This refactor keeps the codebase clean and relies on `mpv` doing what it does best. Please let me know if there are any further adjustments needed or if you'd like me to test any specific edge cases. Thanks again for steering this in the right direction!

Present   |  Original
:-------------------------:|:-------------------------:
   |

yorkyang2333:
Fixed an issue where manually triggering the display, hiding, or addition of subtitles in Mini Player (audio-only mode) does not refresh the interface, causing the default artwork to remain visible and obscure the video view where subtitles are rendered.

**Root Cause:**
1. **Missing UI Refresh Notification:** In `PlayerCore.swift`, when manual subtitle visibility toggles (`subVisibilityChanged`, `secondSubVisibilityChanged`) or track changes (`sidChanged`, `secondarySidChanged`) occurred, the system only sent OSD alerts and internal notifications. It failed to instruct the Mini Player to refresh its view.
2. **Incomplete Artwork Hiding Logic:** In `MiniPlayerWindowController.swift`, the `refreshArtworkVisibility` method only checked the primary subtitle state (`isSubVisible && sid != 0`). If a user only enabled the secondary subtitle, the artwork would erroneously fail to hide.

**Changes:**
* Added `if isInMiniPlayer { miniPlayer.refreshArtworkVisibility() }` to the four subtitle event handlers in `PlayerCore.swift` (`sidChanged`, `subVisibilityChanged`, `secondarySidChanged`, `secondSubVisibilityChanged`) to ensure the Mini Player correctly updates on state changes.
* Updated `refreshArtworkVisibility` in `MiniPlayerWindowController.swift` to consider both primary and secondary subtitle states:
  ```swift
  let hasSubtitles = (player.info.isSubVisible && player.info.sid != 0) || (player.info.isSecondSubVisible && player.info.secondSid != 0)
  ```

iina/MainWindowMenuActions.swift
```diff
@@ -87,7 +87,11 @@ extension MainWindowController {
   }
 
   @objc func menuSwitchToMiniPlayer(_ sender: NSMenuItem) {
-    player.switchToMiniPlayer()
+    if player.isInMiniPlayer {
+      player.switchBackFromMiniPlayer()
+    } else {
+      player.switchToMiniPlayer()
+    }
```
uiryuu: I believe this change is for "Corrected the music mode enter/exit toggle behavior". I haven't found any bug regarding "music mode enter/exit toggle behavior". Please revert all changes regarding this issue. If you found a concrete bug, please make another PR with steps to reproduce the bug, thanks.

iina/MainWindowController.swift
```diff
@@ -3270,7 +3269,11 @@ class MainWindowController: PlayerWindowController {
     case .fullScreen:
       toggleWindowFullScreen()
     case .musicMode:
-      player.switchToMiniPlayer()
+      if player.isInMiniPlayer {
+        player.switchBackFromMiniPlayer()
+      } else {
+        player.switchToMiniPlayer()
+      }
```
uiryuu: Revert

iina/MenuController.swift
```diff
@@ -517,6 +517,7 @@ class MenuController: NSObject, NSMenuDelegate {
     fullScreen.title = isInFullScreen ? Constants.String.exitFullScreen : Constants.String.fullScreen
     pictureInPicture?.title = isInPIP ? Constants.String.exitPIP : Constants.String.pip
     miniPlayer.title = player.isInMiniPlayer ? Constants.String.exitMiniPlayer : Constants.String.miniPlayer
+    miniPlayer.target = player.isInMiniPlayer ? player.miniPlayer : player.mainWindow
```
uiryuu: Revert

iina/PlayerCore.swift
```diff
@@ -535,6 +535,10 @@ class PlayerCore: NSObject {
     // Better to directly reset icc-profile-auto. See issue #5727 for details.
     mpv.setFlag(MPVOption.GPURendererOptions.iccProfileAuto, false)
 
+    // Delay force-window until an actual file load to avoid Xcode-launched app startup hanging
+    // while mpv tries to create a VO before IINA has entered its normal media-open path.
+    mpv.setString(MPVOption.Window.forceWindow, "yes", level: .verbose)
```
uiryuu, line 540: What is this hang, can you provide steps to reproduce?

iina/PlayerCore.swift
```diff
@@ -632,6 +636,10 @@ class PlayerCore: NSObject {
   func initVideo() {
     // init mpv render context.
     mpv.mpvInitRendering()
+    // `force-window=immediate` makes audio-only subtitle rendering work with `vo=libmpv`,
+    // but setting it before render initialization can race the VO thread against IINA's
+    // render context setup. Switch to `immediate` only after the render context exists.
+    mpv.setString(MPVOption.Window.forceWindow, "immediate", level: .verbose)
```
uiryuu, line 642: Ditto

iina/PlayerCore.swift
```diff
@@ -910,6 +913,7 @@ class PlayerCore: NSObject {
   ///     running when the mpv core is shutdown it may call into mpv triggering a crash.
   func stop() {
     guard info.state != .shutDown else { return }
+
```
uiryuu: Why empty lines are added here?

iina/PlayerCore.swift
```diff
@@ -2329,6 +2334,7 @@ class PlayerCore: NSObject {
     // restart even while paused. See issue #5337.
     syncUI(.time)
     reloadSavedIINAfilters()
+
```
uiryuu: Ditto

iina/PlayerWindowController.swift
```diff
@@ -698,6 +698,12 @@ class PlayerWindowController: NSWindowController, NSWindowDelegate {
       AppDelegate.shared.openFile(self)
     case .openURL:
       AppDelegate.shared.openURL(self)
+    case .toggleMusicMode:
+      if player.isInMiniPlayer {
+        player.switchBackFromMiniPlayer()
+      } else {
+        player.switchToMiniPlayer()
+      }
```
uiryuu: Revert

uiryuu (changes requested):
See the inline comments

yorkyang2333:
@uiryuu 
I've reverted the changes related to the music mode toggle behavior and cleaned up the extraneous empty lines.

Regarding the `force-window` configuration in `PlayerCore.swift`:
These lines are actually the core logic needed to render subtitles in audio-only mode with `vo=libmpv` (without `force-window=immediate`, mpv will not create a video output layer for audio files, so subtitles have nowhere to render).

As for the hang mentioned: Setting `force-window` at the very beginning of mpv initialization causes IINA to hang when launched from Xcode, because mpv's VO thread races against IINA's render context setup. To fix this, I delayed setting `force-window=yes` until an actual file is loaded, and then switched it to `immediate` safely after `mpvInitRendering()` is called. These changes are strictly necessary for the audio-only subtitles feature to work.

Please let me know if there's anything else!

iina/MainWindowMenuActions.swift
yorkyang2333: Reverted in the latest commit.

iina/MainWindowController.swift
yorkyang2333: Reverted.

iina/MenuController.swift
yorkyang2333: Reverted.

iina/PlayerWindowController.swift
yorkyang2333: Reverted.

iina/PlayerCore.swift
yorkyang2333: Removed in the latest commit.

iina/PlayerCore.swift
yorkyang2333: Removed.

iina/PlayerCore.swift
yorkyang2333: Setting `force-window` at the very beginning of mpv initialization causes IINA to hang when launched from Xcode, because mpv's VO thread races against IINA's render context setup. To fix this, I delayed setting `force-window=yes` until an actual file is loaded.

iina/PlayerCore.swift
yorkyang2333: Switched to `immediate` safely after `mpvInitRendering()` is called. These `force-window` changes are strictly necessary for the audio-only subtitles feature to work, as without `force-window`, `vo=libmpv` will not create a video output layer for audio files, so subtitles have nowhere to render.

yorkyang2333:
Reviewed against CONTRIBUTING.md — no violations found. Changes are well-scoped to the audio-only subtitle feature.

Minor cleanup applied:
- Removed the empty `chore: trigger CI` commit
- Fixed two extra blank lines in `MiniPlayerWindowController.swift` (one before closing brace, one at EOF)

iina/PlaybackInfo.swift
```diff
@@ -118,15 +118,33 @@ class PlaybackInfo {
...
+    guard let albumArtTrack,
+          let width = albumArtTrack.demuxW,
+          let height = albumArtTrack.demuxH,
+          width > 0,
+          height > 0 else {
+      return nil
+    }
+    return (width, height)
+  }
+
```
lhc70000: Each of these properties is used only in a single place. They should be `let` variables inside the functions where they are used. The possibility of reusing them in the future is thin.

iina/PlayerCore.swift
```diff
@@ -3089,17 +3124,17 @@ class PlayerCore: NSObject {
...
-    let noVideoTrack = info.videoTracks.isEmpty
-    let noAudioTrack = info.audioTracks.isEmpty
-    if noVideoTrack && noAudioTrack {
+    switch info.isAudio {
+    case .unknown:
       return .unknown
+    case .isAudio:
+      return .isAudio
+    case .notAudio:
+      return .notAudio
```
lhc70000, line 3133: @uiryuu We need to do something about this. 
@yorkyang2333: you don't need to update this part, we will handle it afterwards.

iina/PlayerWindowController.swift
```diff
@@ -698,6 +698,8 @@ class PlayerWindowController: NSWindowController, NSWindowDelegate {
       AppDelegate.shared.openFile(self)
     case .openURL:
       AppDelegate.shared.openURL(self)
+    case .toggleMusicMode:
+      player.switchToMiniPlayer()
```
lhc70000: Please remove this line as it's out of scope.

lhc70000 (changes requested):

uiryuu (approved):

yorkyang2333:
Addressed @lhc70000's latest review:

1. **Removed single-use properties from `PlaybackInfo`** — `hasRealVideoTrack`, `albumArtTrack`, and `musicModeArtworkSize` have been removed as computed properties. Each is now inlined as a local `let` inside the single function that uses it:
   - `hasRealVideoTrack` → local `let` inside `isAudio`
   - `albumArtTrack` → local `let` inside `videoSizeForDisplayInMusicMode()` and `refreshArtworkVisibility()`
   - `musicModeArtworkSize` → logic expanded inline inside `videoSizeForDisplayInMusicMode()`

2. **Removed out-of-scope `toggleMusicMode` case** from `PlayerWindowController.handleIINACommand(_:)`.

(The `PlayerCore.swift` item at line 3133 was noted as something the team will handle — left untouched.)

iina/PlaybackInfo.swift
yorkyang2333: fixed

iina/PlayerWindowController.swift
yorkyang2333: fixed

PR #6035, opened 2026-04-26
sub: reload quicksetting for sub page when sid changes

uiryuu:
---

**Description:**
When `sid` is changed from mpv (e.g. drag a sub file to window, the new sub file is enabled automatically), the sub page is not reloaded.

Step to reproduce:
- Open a video with ass subtitle enabled
- Open the quick settings subtitle panel
- Notice that the controls for sub text color are disabled
- Keep the panel opened
- Drag a srt subtitle file to the window, the srt subtitle should be enabled
- The controls should be enabled, but they are not

### DIFF
--- a/iina/QuickSettingViewController.swift
+++ b/iina/QuickSettingViewController.swift
@@ -317,11 +317,7 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
     }
     observe(.iinaVIDChanged) { [unowned self] _ in self.videoTableView.reloadData() }
     observe(.iinaAIDChanged) { [unowned self] _ in self.audioTableView.reloadData() }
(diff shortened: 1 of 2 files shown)


iina/QuickSettingViewController.swift
```diff
@@ -320,6 +320,7 @@ class QuickSettingViewController: NSViewController, NSTableViewDataSource, NSTab
     observe(.iinaSIDChanged) { [unowned self] _ in
       self.subTableView.reloadData()
       self.secSubTableView.reloadData()
+      reload()
```
lhc70000: `reload()` should already contain both `self.subTableView.reloadData()` and `self.secSubTableView.reloadData()`, right?

lhc70000 (changes requested):

PR #6036, opened 2026-04-26
Enhancements to the utilities settings page

Uranosphaerite:

---

**Fixes**
* Fixed thumbnail cache size not showing until it's cleared

**Additions**
* Chrome and Firefox extensions open in their respective apps when possible
* A button to get to the Safari extension's settings
   * This currently shows as a second Chrome button because I don't want to mess up the localization files
   * This [does not work](https://stackoverflow.com/questions/47456460/safari-app-extension-not-detected-in-showpreferencesforextension) when debugging, the app will have to be archived and exported to test it
   * The extension identifier is hardcoded which might be a bad idea

I was trying to make my own version of this page so I thought I'd bring over a couple improvements from there. Loving all the work that's been done over the past few weeks

### DIFF
--- a/iina/Pages/SettingsPageUtilities.swift
+++ b/iina/Pages/SettingsPageUtilities.swift
@@ -7,6 +7,7 @@
 //
 
 import UniformTypeIdentifiers
+import SafariServices.SFSafariApplication
 
 class SettingsPageUtilities: SettingsPage {
   override var title: String {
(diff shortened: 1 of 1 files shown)


uiryuu (changes requested):
Thanks for contributing to IINA. I really like the changes. I've tested uninstalling Chrome and click on the Chrome button, the link is opened normally in my default browser.

I wanted to add the localization for "Safari", but I can't push to this branch, so please apply this patch below:

[l10n.patch](https://github.com/user-attachments/files/27108085/l10n.patch)

uiryuu:
Thanks for your contribution. I'll help you to add the translation locally

PR #6040, opened 2026-04-28
Monitor file changes to auto-reload subtitles if they are modified (#3724)

svobs:
- [x] This implements/fixes issue #3724.

---

**Description:**
Back-porting this code from my IINA Advance project - might be advantageous to get this code into upstream.

This uses code from an earlier version of [vChewing_FolderMonitor](https://github.com/vChewing/vChewing-macOS/tree/main/Packages/vChewing_FolderMonitor) (the link on the bug report page is broken due to rename).

In `PlayerCore`, the complementary methods `startWatchingSubFile` and `stopWatchingSubFile` start and stop the file monitor, respectively - the former first calls the latter to stop any previous monitoring, so it can be called repeatedly without error.

The method `startWatchingSubFile` is called whenever the sub track changes, but the monitor is only set up if the current sub track is determined to be external. 
The method `stopWatchingSubFile` is called both in `PlayerCore`'s `fileStarted()` and `stop()`.

When the file changes, it triggers the following callback which reloads the sub track.
``` Swift
fileMonitor.fileDidChange = { [self] in
  mpv.command(.subReload, args: [String(currentSubTrack.id)], checkError: false)
}
```

### DIFF
--- a/iina/FileMonitor.swift
+++ b/iina/FileMonitor.swift
@@ -1,0 +1,59 @@
+/// (c) 2018 Daniel Galasko
+/// Ref: https://medium.com/over-engineering/monitoring-a-folder-for-changes-in-ios-dc3f8614f902
+/// Modified from FolderMonitor for use with a single file.
+import Foundation
+
+public class FileMonitor {
+  // MARK: Properties
+
+  /// A file descriptor for the monitored file.
+  private var monitoredFileFD: CInt = -1
+  /// A dispatch queue used for sending file changes in the file.
+  private let fileMonitorQueue = DispatchQueue(label: "com.iina.FileMonitorQueue", attributes: .concurrent)
+  /// A dispatch source to monitor a file descriptor created from the file.
+  private var monitorSource: DispatchSourceFileSystemObject?
+  /// URL for the file being monitored.
+  public let url: URL
+
+  public var fileDidChange: (() -> Void)?
+
+  // MARK: Initializers
+
+  public init(url: URL) {
+    self.url = url
+  }
+
+  // MARK: Monitoring
+
+  /// Listen for changes to the file (if we are not already).
+  public func startMonitoring() {
+    guard monitorSource == nil, monitoredFileFD == -1 else {
+      return
+    }
+    // Open the file referenced by URL for monitoring only.
+    monitoredFileFD = open(url.path, O_EVTONLY)
+    // Define a dispatch source monitoring the file for additions, deletions, and renamings.
+    monitorSource = DispatchSource.makeFileSystemObjectSource(
+      fileDescriptor: monitoredFileFD, eventMask: .all, queue: fileMonitorQueue
+    )
+    // Define the block to call when a file change is detected.
+    monitorSource?.setEventHandler { [weak self] in
+      guard let self = self else { return }
+      self.fileDidChange?()
+    }
+    // Define a cancel handler to ensure the file is closed when the source is cancelled.
+    monitorSource?.setCancelHandler { [weak self] in
+      guard let self = self else { return }
+      close(self.monitoredFileFD)
+      self.monitoredFileFD = -1
+      self.monitorSource = nil
+    }
+    // Start monitoring the file via the source.
+    monitorSource?.resume()
+  }
+
+  /// Stop listening for changes to the file, if the source has been created.
+  public func stopMonitoring() {
+    monitorSource?.cancel()
+  }
+}
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -222,6 +222,8 @@ class PlayerCore: NSObject {
     return controller
   }()
 
+  private var subFileMonitor: FileMonitor? = nil
+
   lazy var info: PlaybackInfo = PlaybackInfo(self)
 
   var syncUITimer: Timer?
@@ -911,6 +913,7 @@ class PlayerCore: NSObject {
   func stop() {
     guard info.state != .shutDown else { return }
     savePlaybackPosition()
+    stopWatchingSubFile()
 
     // The player may already be stopped in which case the state must not be set to stopping.
     if info.state != .idle {
@@ -1441,6 +1444,38 @@ class PlayerCore: NSObject {
     }
     mainWindow?.quickSettingView.reload()
   }
+
+  private func startWatchingSubFile() {
+    guard let currentSubTrack = info.currentTrack(.sub) else { return }
+    guard let externalFilename = currentSubTrack.externalFilename else {
+      log("Sub track \(currentSubTrack.id) is not an external file", level: .verbose)
+      return
+    }
+
+    // Stop previous watch (if any)
+    stopWatchingSubFile()
+
+    let subURL = URL(fileURLWithPath: externalFilename)
+    let fileMonitor = FileMonitor(url: subURL)
+    fileMonitor.fileDidChange = { [self] in
+      mpv.command(.subReload, args: [String(currentSubTrack.id)], checkError: false) { [self] code in
+        if code >= 0 { return }
+        log("Failed reloading sub track \(currentSubTrack.id): error code \(code)", level: .error)
+      }
+    }
+    subFileMonitor = fileMonitor
+    log("Starting FS watch of sub file \(subURL.path.quoted)", level: .verbose)
+    fileMonitor.startMonitoring()
+  }
+
+  private func stopWatchingSubFile() {
+    guard let subFileMonitor else { return }
+
+    log("Stopping FS watch of sub file \(subFileMonitor.url.path.quoted)", level: .verbose)
+    subFileMonitor.stopMonitoring()
+    self.subFileMonitor = nil
+  }
+
 
   func setAudioDelay(_ delay: Double) {
     mpv.setDouble(MPVOption.Audio.audioDelay, delay)
@@ -2027,6 +2062,9 @@ class PlayerCore: NSObject {
       }
     }
 
+    // Stop watchers from prev media (if any)
+    stopWatchingSubFile()
+
     NowPlayingInfoManager.shared.updateInfo(withTitle: true)
 
     // Auto load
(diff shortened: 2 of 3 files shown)


low-batt:
Tuesday is my busy day. I'll take a closer look at this tomorrow…

Should we be embedding this in the player, or making an official IINA plugin for this?

uiryuu:
During test, I found it unstable and often fail reloading with error code -12:
```
18:42:18.884 [player0][e] Failed reloading sub track 11: error code -12
```
Have you encountered this before?

An OSD to display the error in such case is desired.

svobs:
> During test, I found it unstable and often fail reloading with error code -12
>
> Have you encountered this before?
>
> An OSD to display the error in such case is desired.

What did you do to produce that error? I was only able to produce the error by moving or deleting the currently selected sub file. In that case, it seems valid for loading to fail. I can add an OSD message to handle it. But if you are getting an this error from just changing the file's contents there must be something else going wrong...

> Should we be embedding this in the player, or making an official IINA plugin for this?

I don't see why it would need to be optional. Various other apps automatically detect changes to the files being viewed. For example, I tried overwriting a PDF that was open in Preview.app, and the document window refreshed automatically with the changes.

uiryuu:
> What did you do to produce that error? I was only able to produce the error by moving or deleting the currently selected sub file. In that case, it seems valid for loading to fail. I can add an OSD message to handle it. But if you are getting an this error from just changing the file's contents there must be something else going wrong...

I don't think I've done anything weird. I just edit the ass file with TextEdit.app while IINA window is open with the ass file loaded. When the file is saved, the error occurs. It's also not 100%, it initially worked a few times. I also checked the mpv command call and compare to `reloadAllSubs`, nothing strange, but `reloadAddSubs` works every time without error.

low-batt:
Interesting. GitHub is having serious database problems with pull requests. Seems to extend to posting comments to the wrong pull request. Browser was showing 6041, but my comment ended up here.

iina/PlayerCore.swift
```diff
@@ -1442,6 +1445,38 @@ class PlayerCore: NSObject {
     mainWindow?.quickSettingView.reload()
   }
 
+  private func startWatchingSubFile() {
```
low-batt, line 1448: These are in the "MPV Commands" section, which are public methods for other classes to use. This probably should be in the "Utils" section.

iina/PlayerCore.swift
```diff
@@ -1442,6 +1445,38 @@ class PlayerCore: NSObject {
...
+    let subURL = URL(fileURLWithPath: externalFilename)
+    let fileMonitor = FileMonitor(url: subURL)
+    fileMonitor.fileDidChange = { [self] in
+      mpv.command(.subReload, args: [String(currentSubTrack.id)], checkError: false) { [self] code in
+        if code >= 0 { return }
+        log("Failed reloading sub track \(currentSubTrack.id): error code \(code)", level: .error)
+      }
+    }
+    subFileMonitor = fileMonitor
+    log("Starting FS watch of sub file \(subURL.path.quoted)", level: .verbose)
```
low-batt, line 1467: IINA is not using `path.quoted` in any other log messages. In a lot of messages the convention is "bla bla file: path". Some log messages are missing the colon. Would be good to be consistent.

As this is a new feature and this method should not be called at a high rate, I think this should be a normal debug message. Once the feature is proven we could make it a verbose message.

iina/PlayerCore.swift
```diff
@@ -1442,6 +1445,38 @@ class PlayerCore: NSObject {
...
+      log("Sub track \(currentSubTrack.id) is not an external file", level: .verbose)
+      return
+    }
+
+    // Stop previous watch (if any)
+    stopWatchingSubFile()
+
+    let subURL = URL(fileURLWithPath: externalFilename)
+    let fileMonitor = FileMonitor(url: subURL)
+    fileMonitor.fileDidChange = { [self] in
```
low-batt, line 1460: Need a debug level log message indicating IINA is reloading the file. The message should contain the file path. Something like: "Reloading subtitle file: file-path"

iina/PlayerCore.swift
```diff
@@ -1442,6 +1445,38 @@ class PlayerCore: NSObject {
...
+      mpv.command(.subReload, args: [String(currentSubTrack.id)], checkError: false) { [self] code in
```
low-batt, line 1461: How does this code know _for certain_ that `currentSubTrack` is still set to the sub ID that was being monitored? What is enforcing the order of event processing? Could this be processing a file change event that is stale?

iina/PlayerCore.swift
```diff
@@ -1442,6 +1445,38 @@ class PlayerCore: NSObject {
...
+        if code >= 0 { return }
```
low-batt, line 1462: Right now the user gets no feedback that IINA noticed the changed subtitle file and reloaded it. Upon a successful return an OSD message should be shown. Doesn't have to contain the filename. Maybe `Reloaded Subtitles` and `Reloaded Secondary Subtitles`. We aren't just monitoring the primary subtitles, right?

iina/PlayerCore.swift
```diff
@@ -1442,6 +1445,38 @@ class PlayerCore: NSObject {
...
+    }
+    subFileMonitor = fileMonitor
+    log("Starting FS watch of sub file \(subURL.path.quoted)", level: .verbose)
+    fileMonitor.startMonitoring()
+  }
+
+  private func stopWatchingSubFile() {
+    guard let subFileMonitor else { return }
+
+    log("Stopping FS watch of sub file \(subFileMonitor.url.path.quoted)", level: .verbose)
```
low-batt, line 1474: As this is a new feature and this method should not be called at a high rate, I think this should be a normal debug message. Once the feature is proven we could make it a verbose message.

iina/FileMonitor.swift
```diff
@@ -0,0 +1,59 @@
...
+
+  // MARK: Monitoring
+
+  /// Listen for changes to the file (if we are not already).
+  public func startMonitoring() {
+    guard monitorSource == nil, monitoredFileFD == -1 else {
+      return
+    }
+    // Open the file referenced by URL for monitoring only.
+    monitoredFileFD = open(url.path, O_EVTONLY)
```
low-batt, line 34: Possibly this could be more swifty and use [FileDescriptor.open](https://developer.apple.com/documentation/system/filedescriptor/open(_:_:options:permissions:retryoninterrupt:)-4ql4b) and [eventOnly](https://developer.apple.com/documentation/system/filedescriptor/openoptions/eventonly).

iina/FileMonitor.swift
```diff
@@ -0,0 +1,59 @@
...
+/// Modified from FolderMonitor for use with a single file.
+import Foundation
+
+public class FileMonitor {
+  // MARK: Properties
+
+  /// A file descriptor for the monitored file.
+  private var monitoredFileFD: CInt = -1
+  /// A dispatch queue used for sending file changes in the file.
+  private let fileMonitorQueue = DispatchQueue(label: "com.iina.FileMonitorQueue", attributes: .concurrent)
```
low-batt, line 12: Why a dedicated queue? As much as possible we are trying to restrict access to `PlayerCore.info` to the main thread to avoid the need for locking to coordinate access.

iina/FileMonitor.swift
```diff
@@ -0,0 +1,59 @@
...
+    }
+    // Define a cancel handler to ensure the file is closed when the source is cancelled.
+    monitorSource?.setCancelHandler { [weak self] in
+      guard let self = self else { return }
+      close(self.monitoredFileFD)
+      self.monitoredFileFD = -1
+      self.monitorSource = nil
+    }
+    // Start monitoring the file via the source.
+    monitorSource?.resume()
```
low-batt, line 52: This is not what the Apple documentation for [makeFileSystemObjectSource](https://developer.apple.com/documentation/dispatch/dispatchsource/makefilesystemobjectsource(filedescriptor:eventmask:queue:)#Discussion) tell you to do. It says:
> The returned dispatch source is in the inactive state initially. When you are ready to begin processing events, call its activate() method.

Probably should be following the Apple doc.

low-batt (changes requested):
The BIG issue is that the code does not work. Try changing a `.srt` file using TextEdit multiple times. The subtitles are initially reloaded. Then the changes go unnoticed. 

I believe this is because some editors save the changes to a new file and then unlink the existing file and link the new file in its place. I believe this leaves IINA monitoring the unlinked file. Possibly something like that is happening.

The secondary subtitle is not monitored. There will probably be one user out there that is working with multiple subtitles. Seems inconsistent to me that the secondary subtitle behaves differently.

svobs:
Once again, what I thought was a relatively straightforward change turns out to be a lot more involved, haha. Guess I brought this on myself.

> The BIG issue is that the code does not work. Try changing a `.srt` file using TextEdit multiple times. The subtitles are initially reloaded. Then the changes go unnoticed.
> 
> I believe this is because some editors save the changes to a new file and then unlink the existing file and link the new file in its place. I believe this leaves IINA monitoring the unlinked file. Possibly something like that is happening.

In my testing I hadn't seen this. I had not considered the possibility that some editors may be replacing the file rather than modifying it. Probably would need to cancel the monitoring of the file on any change and set it up again based on file path, to ensure a new file descriptor gets picked up. But in a situation with many successive replacement operations a race condition could result which may cause update detection to stop working...

Maybe it's due to quirks like this that Apple discourages the use of this API. I hadn't noticed until filing this that the source code I based this on has since been completely rewritten using `NSFilePresenter`, which claims to transparently handle concurrent access, as well as files served via iCloud.

Seems like a complete rewrite using vChewing's latest version (using `NSFilePresenter`) might be the best way to proceed. I'll work on this and submit a new PR.

low-batt:
> Once again, what I thought was a relatively straightforward change turns out to be a lot more involved

Happens to me a lot too!

The other developers ok'ed adding this, assuming it can be made to work. I looked at examples of doing this and they all seemed too simplistic. I didn't find any sophisticated examples that might handle the file being replaced.

PR #6042, opened 2026-04-29
chore: suppress Xcode warnings

uiryuu:

---

**Description:**
- update Xcode settings
  - "Enable Generated Asset Symbol Extensions" is enabled, so the NSColor names for custom colors are automatically generated. Existing NSColor extensions are no longer needed thus removed.
  - `ENABLE_USER_SCRIPT_SANDBOXING` is enabled, which requires `write_info_plist_header.sh` to be added to `inputFileListPaths`. Otherwise the shell script doesn't have the permission to run.
- Changes the type of `VideoView` to `NSOpenGLView` to suppress the warning to use `NSOpneGLView` for `wantsExtendedDynamicRangeOpenGLSurface` and `wantsBestResolutionOpenGLSurface`.
- Use `NSKeyedUnarchiver.unarchivedObject(ofClasses:from:)` instead of `NSUnarchiver.unarchiveObject(with:)`.

App behaviors should not change.

### DIFF
--- a/iina/LegacyMigration.swift
+++ b/iina/LegacyMigration.swift
@@ -61,7 +61,7 @@ class LegacyMigration {
 
       // Deserialize & convert legacy pref value to modern string format:
       guard let legacyData = Preference.data(for: legacyKey) else { continue }
-      guard let color = NSUnarchiver.unarchiveObject(with: legacyData) as? NSColor,
+      guard let color = try? NSKeyedUnarchiver.unarchivedObject(ofClass: NSColor.self, from: legacyData),
             let mpvColorString = color.usingColorSpace(.deviceRGB)?.mpvColorString else {
         Logger.log("Failed to convert color value from legacy pref \(legacyKey.rawValue)", level: .error)
         continue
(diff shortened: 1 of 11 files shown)


iina/LegacyMigration.swift
```diff
@@ -61,7 +61,7 @@ class LegacyMigration {
 
       // Deserialize & convert legacy pref value to modern string format:
       guard let legacyData = Preference.data(for: legacyKey) else { continue }
-      guard let color = NSUnarchiver.unarchiveObject(with: legacyData) as? NSColor,
+      guard let color = try? NSKeyedUnarchiver.unarchivedObject(ofClass: NSColor.self, from: legacyData),
```
low-batt: This **must not** be changed. This is the code that reads the **old non-keyed** archive and then writes in a new format. This is about migrating existing color prefs created by previous IINA versions and stored in the old archive format.

This is my example to the Swift developers of why they need to give us a way to suppress compiler warnings, which last I heard they stubbornly refused to do.

low-batt (changes requested):
I'm still reviewing, but the change to `LegacyMigration` must be removed.

low-batt:
I'm testing under macOS Sequoia and it is crashing for me when I try and play a video:
crash

PR #6045, opened 2026-04-29
Mention linking yt-dlp to deps/executable/youtube-dl 

rockfordroeNG:

---

**Description:**

When following the manual build instructions, xcode fails to build the app bundle due to a missing youtube-dl dependency, this should be mentioned in the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,15 +98,22 @@ IINA uses mpv for media playback. To build IINA, you can either fetch copies of 
 	port contents mpv | grep '\.dylib$' | xargs other/change_lib_dependencies.rb /opt/local
 	```
 
-5. Open `iina.xcodeproj` in the [latest public version of Xcode](https://apps.apple.com/app/xcode/id497799835). *IINA may not build if you use any other version.*
+5. Link the *yt-dlp* dependency to deps/executable 
+ 
+   ```console
+   mkdir -p deps/executable
+   ln -s $(which yt-dlp) deps/executable/youtube-dl
+   ```
+   
+6. Open `iina.xcodeproj` in the [latest public version of Xcode](https://apps.apple.com/app/xcode/id497799835). *IINA may not build if you use any other version.*
 
-6. Remove all references to `.dylib` files from the Frameworks group in the sidebar and add all the `.dylib` files in `deps/lib` to that group by clicking  "Add Files to iina..." in the context menu.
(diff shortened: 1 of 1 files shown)


low-batt:
The README does need to be updated. Sorry about that.

However it is not that it is missing instructions to link `yt-dlp` to `youtube-dl`…

What is missing is that the [Using the pre-compiled libraries](https://github.com/iina/iina#using-the-pre-compiled-libraries) section does not mention that `other/download_libs.sh` does more than download the precompiled libraries. This is what you should see at the end after it has downloaded the libraries:

```text
Downloaded libzstd.1.dylib
Downloaded libzmq.5.dylib
Downloading yt-dlp...
yt-dlp downloaded
Downloading latest plugin release for iina/plugin-online-media...
Downloaded iina-plugin-ytdl-0.9.10.iinaplgz
Downloading latest plugin release for iina/plugin-userscript...
Downloaded iina-plugin-userscript-0.9.3.iinaplgz
Downloading latest plugin release for iina/plugin-opensub...
Downloaded iina-plugin-opensub-0.9.6.iinaplgz
All downloads completed.
```

It is populating `deps/executable` with `yt-dlp` renamed to `youtube-dl` and also populating `deps/plugins` with the official IINA plugins.

We need to figure out why that did not happen in your case.

GitHub has rolled out what they call [Rate limits](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2026-03-10). I have not seen them interfere with downloading `yt-dlp`, but we definitely have had trouble with them blocking downloads of IINA plugins.

rockfordroeNG:
I understand that the general workflow is to use the script to download and rename the binaries. However, I am choosing not to do so as the README has instructions for if I want to compile IINA with a locally compiled mpv. The GPLv3 also extends to the scripts and data needed to compile the executable as per the Corresponding Source clause. Even without factoring in the GPLv3, I would like this to be merged as there is no documentation on how to proceed properly with the manual compilation if the script happens to fail or the server storing the binaries is down for whatever reason.

low-batt:
Ah, I see what you mean. Sorry, busy day. I will check with the other developers on this. I think they will approve.

uiryuu:
Very minor nitpick, maybe we can use
```markdown
1. test
1. test
```
which will yield
1. test
1. test

So that the number of changed lines will be reduced...

PR #6050, opened 2026-05-01
Don't render OSD when in miniplayer

uiryuu:

---

**Description:**
During testing #5750, I found that even the current player is miniplayer when the main window isn't shown, the OSD is still rendered if `sendOSD()` is called, which is a waste of resource. This commit adds a check in `sendOSD()` and directly return if the current player is mini player. I've checked that currently there's no directly call to `displayOSD()` other than through `sendOSD()`.


low-batt (approved):
Pulled PR, built IINA, tested and confirmed change is working.

PR #6059, opened 2026-05-07
ui: add a toolbar to the inspector window

uiryuu:

---

**Description:**
New           |  Old
:-------------------------:|:-------------------------:
  |  

This commit is a relatively small change:
- The segmented control is moved from content view to the title bar.
  - This looks more like the segmented control on top of the Apple's Calendar image

  - I've also tried to use the `NSTabView` as the content view of the window, and set the style of the tab view to `NSTabViewController.TabStyle.toolbar` (according to the HIG), but it looks awful.
- Changed label color of the field names from `Label Color` to `Secondary Label Color`
  - QuickTime's inspector also does something like this. 
  - When the color below the inspector window is not white, these labels will have less contrast and more vibrancy.

We can also consider to make the titlebar transparent. The current implementation is not transparent to preserve the old look of our inspector window. If we want a transparent titlebar, I'll update this PR.

Normal Titlebar           |  Transparent Titlebar
:-------------------------:|:-------------------------:
  |


low-batt (approved):
Pulled PR, built IINA, tested inspector window.

Changes look good to me.

low-batt:
I really like the normal title bar. It causes the bar to be clearly separate from the content. The buttons stand out more. I definitely like it better than the transparent bar. But that might just be me…

PR #6062, opened 2026-05-08
Highlight OSC toolbar button when applicable

uiryuu:

---

**Description:**
Highlight OSC toolbar buttons when the corresponding view is enabled/shown. Idea from #5867, but implemented a universal solution for all buttons on the OSC toolbar.

Example (pip, playlist, fullscreen enabled):
telegram-cloud-photo-size-5-6334452269520196098-y

### DIFF
--- a/iina/Preference.swift
+++ b/iina/Preference.swift
@@ -859,7 +859,7 @@ struct Preference {
     func alternateImage() -> NSImage? {
       switch self {
       case .settings: return makeSymbol(["gearshape.fill"], NSImage.actionTemplateName)
-      case .playlist: return makeSymbol(["list.bullet.rectangle.fill", "list.bullet.fill"], "playlist")
+      case .playlist: return makeSymbol(["list.bullet.rectangle.fill", "list.bullet"], "playlist")
       case .pip: return makeSymbol(["pip.exit"], "pip")
       case .fullScreen: return makeSymbol(["arrow.down.forward.and.arrow.up.backward.rectangle", "arrow.down.right.and.arrow.up.left"], "fullscreen")
       case .plugins: return makeSymbol(["puzzlepiece.extension.fill"], "plugin")
--- a/iina/Preference.swift
+++ b/iina/Preference.swift
@@ -837,21 +837,33 @@ struct Preference {
       }
     }
 
-    func image() -> NSImage {
-      func makeSymbol(_ names: [String], _ fallbackImage: NSImage.Name) -> NSImage {
+    private func makeSymbol(_ names: [String], _ fallbackImage: NSImage.Name) -> NSImage {
         guard #available(macOS 14.0, *) else { return NSImage(named: fallbackImage)! }
         let configuration = NSImage.SymbolConfiguration(pointSize: 14, weight: .medium)
         return NSImage.findSFSymbol(names, withConfiguration: configuration)
       }
+
+    func image() -> NSImage {
       switch self {
       case .settings: return makeSymbol(["gearshape"], NSImage.actionTemplateName)
       case .playlist: return makeSymbol(["list.bullet.rectangle", "list.bullet"], "playlist")
-      case .pip: return makeSymbol(["pip.swap"], "pip")
+      case .pip: return makeSymbol(["pip.enter"], "pip")
       case .fullScreen: return makeSymbol(["arrow.up.backward.and.arrow.down.forward.rectangle", "arrow.up.left.and.arrow.down.right"], "fullscreen")
-      case .musicMode: return makeSymbol(["music.microphone", "music.mic"], "toggle-album-art")
+      case .musicMode: return makeSymbol(["music.microphone"], "toggle-album-art")
       case .subTrack: return makeSymbol(["captions.bubble.fill"], "sub-track")
       case .screenshot: return makeSymbol(["camera.shutter.button"], "screenshot")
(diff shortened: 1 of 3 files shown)


low-batt:
I've always disliked that everything in the OSC is grayscale except for the volume slider:

OSC

Never had time to ask about changing that to match up with the progress slider that we changed to be greyscale.

Should highlighting of the buttons use the default control accent color, or should we keep the OSC grayscale?

This makes the main window OSC inconsistent with the mini-player OSC. The situation is a bit different with the mini-player in that normally when the main window OSC buttons are active it is usually for a short period of time, but it is normal to keep the album cover visible. So if we made this same change in the mini player I would have to stare at a blue button all the time. I'd find that a bit irritating. Greyscale would not be so prominent.

Thoughts?

Otherwise, the changes look good to me.

uiryuu:
> This makes the main window OSC inconsistent with the mini-player OSC.

Actually I was going to add this button highlighting to the mini player as well...

> So if we made this same change in the mini player I would have to stare at a blue button all the time. I'd find that a bit irritating. Greyscale would not be so prominent.

I wanted to say that because we are using the `controlAccentColor` as the highlight color, and you can control this color in the system settings: 
image

If you find the blue color irritating, you can choose the Graphite color as the tint color, so that the highlight will not be that jarring. However, when I change the color to Graphite to test, the highlight color is actually lighter than the normal control color:

image

So maybe we can try to add `.bold()` to the symbol instead of using colors?

uiryuu:
I found a better solution: we manually pick the alt image for the button, and use the alt image appropriately. This way, we can even handle enter/exit fullscreen correctly.

Example (in fullscreen, in pip, w/ quicksettings panel open)
image

iina/Preference.swift
```diff
@@ -837,24 +837,36 @@ struct Preference {
...
       case .plugins: return makeSymbol(["puzzlepiece.extension"], "plugin")
       }
     }
 
+    func alternateImage() -> NSImage? {
+      switch self {
+      case .settings: return makeSymbol(["gearshape.fill"], NSImage.actionTemplateName)
+      case .playlist: return makeSymbol(["list.bullet.rectangle.fill", "list.bullet.fill"], "playlist")
+      case .pip: return makeSymbol(["pip.exit"], "pip")
+      case .fullScreen: return makeSymbol(["arrow.down.forward.and.arrow.up.backward.rectangle", "arrow.down.right.and.arrow.up.left"], "fullscreen")
```
low-batt, line 864:
```suggestion
      case .fullScreen: return makeSymbol(["arrow.down.forward.and.arrow.up.backward.rectangle.fill", "arrow.down.right.and.arrow.up.left.fill"], "fullscreen")
```

iina/Preference.swift
```diff
@@ -837,24 +837,36 @@ struct Preference {
...
       case .screenshot: return makeSymbol(["camera.shutter.button"], "screenshot")
       case .plugins: return makeSymbol(["puzzlepiece.extension"], "plugin")
       }
     }
 
+    func alternateImage() -> NSImage? {
+      switch self {
+      case .settings: return makeSymbol(["gearshape.fill"], NSImage.actionTemplateName)
+      case .playlist: return makeSymbol(["list.bullet.rectangle.fill", "list.bullet.fill"], "playlist")
+      case .pip: return makeSymbol(["pip.exit"], "pip")
```
low-batt, line 863: Since there isn't a `pip.exit.fill` I'm not sure what we can do about this one. As the main window is often hidden when PiP is being used I don't think highlighting it is critical.

low-batt (commented):
Missing `.fill` on the symbol for exiting fullscreen.

low-batt:
I like the look of the `.fill` symbols.

I did not consider changing the system `controlAccentColor` as it is only on the OSC that I find objectionable. As merely moving the mouse shows the OSC, I desire the OSC to be subtle/greyscale. I always thought the intent for the OSC was for it to be greyscale, since nearly all of it already is. I figured nobody so far had the time to customize the volume slider like was done for the progress slider. Some apps such as video and still editing software take this to the extreem where all controls are greyscale. Such apps usually only provide a dark theme. I see no reason to do that. It is the OSC and the titlebar that needs to not be distracting.

uiryuu:
@low-batt We are not using `.fill` for all symbols. The pip symbol is "pip.enter" and "pip.exit"; and the full screen is "arrow.up.backward.and.arrow.down.forward.rectangle" and "arrow.down.forward.and.arrow.up.backward.rectangle".

uiryuu:
Even though the blue color of the volume slider didn't bother me before, I kinda agree with you now. However I failed to find a good way to set the color of the track's fill color. I tried to set the `NSSlider.trackFillColor` to `NSColor.mainSliderBarLeft`, which is the color we used to fill the left side of the main slider, the left side of the volume slider becomes brighter when the window is focused, unlike the main slider. I think we can deal with this in another PR.

low-batt:
Yes, no need to deal with the volume slider in this PR.

As my eyes are bad, it is easier for me to notice the full screen button is enabled when `.fill` is used than to see the change in the little arrows in the symbol. But it makes sense to have a convention where we only do that when the symbol doesn't already change to indicate the state.

low-batt (approved):
Good enhancement.

lhc70000:
It might also be useful to emit the sidebar change notifications to plugins.

PR #6071, opened 2026-05-16
Change volume slider to be greyscale

low-batt:
This commit will:
- Change `VolumeSliderCell.drawBar` to draw the full slider bar itself using greyscale colors
- Change `drawBar` to round the ends of the slider bar (matching the playback position slider)
- Change `drawBar` to correctly draw the gap in the bar representing 100% volume

These changes will cause the volume slider to match up with the greyscale color scheme of the reset of the on screen controller. They will also fix a problem with failing to draw the small gap that marks the position representing 100% volume. This indicator is present when the IINA `Maximum volume` setting is used to allow the volume to exceed 100%.

---

**Description:**

### DIFF
--- a/iina/Assets.xcassets/Colors/VolumeSliderBarLeft.colorset/Contents.json
+++ b/iina/Assets.xcassets/Colors/VolumeSliderBarLeft.colorset/Contents.json
@@ -1,0 +1,38 @@
+{
+  "colors" : [
+    {
+      "color" : {
+        "color-space" : "srgb",
+        "components" : {
+          "alpha" : "0.400",
+          "blue" : "0.000",
+          "green" : "0.000",
+          "red" : "0.000"
+        }
+      },
+      "idiom" : "universal"
+    },
+    {
+      "appearances" : [
+        {
+          "appearance" : "luminosity",
+          "value" : "dark"
+        }
+      ],
+      "color" : {
+        "color-space" : "srgb",
+        "components" : {
+          "alpha" : "0.300",
+          "blue" : "1.000",
+          "green" : "1.000",
+          "red" : "1.000"
+        }
+      },
+      "idiom" : "universal"
+    }
+  ],
+  "info" : {
+    "author" : "xcode",
+    "version" : 1
+  }
+}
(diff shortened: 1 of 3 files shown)


low-batt:
I'm not good at selecting colors, so some tweaking may be needed.

The following screenshots were taken with the IINA `Maximum volume` setting set to 160.

Under macOS Sequoia 15.7.4…

IINA 1.4.2, you can see the 100% indicator is missing:

1 4 2-light-sequoia

1 4 2-dark-sequoia

This PR:

pr-light-sequoia

pr-dark-sequoia

mini-player

Under macOS Tahoe 26.4.1 and this PR:

pr-light-tahoe

pr-dark-tahoe

uiryuu (approved):

low-batt:
For displaying the 100% volume position to the user the current `VolumeSliderCell.drawBar` code uses the same method as is used in the `PlaySliderCell.drawBar` code for displaying chapter positions, namely a small gap in the slider bar. When it comes to why this is not currently working in the volume slider, my guess is that in some AppKit release the [NSSliderCell.drawBar](https://developer.apple.com/documentation/appkit/nsslidercell/drawbar(inside:flipped:)) code was changed such that the clipping path established by calling [NSBezierPath.setClip](https://developer.apple.com/documentation/appkit/nsbezierpath/setclip()) was overridden. As `PlaySliderCell.drawBar` does not make use of  `super.drawBar` the technique is still working for the chapter indicators.

There are issues regarding how hard the chapter indicators are to see. In some cases the user thought the `Show chapter position in progress bar` setting was completely broken because they did not see the chapter markers.

At some point I believe we need to revisit the UI for indicating chapter positions and the 100% volume position. I'm thinking these should be very tiny lines, something similar to tick marks, that when you hover over them they display a help tag with either the chapter name or number and in the case of the volume slider just a `100%` help tag.

An enhancement for a future PR perhaps?

uiryuu:
Since `VolumeSliderCell` and `PlaySliderCell` use similar `drawBar`, can we provide a common extension function to reduce code duplication and also makes it easier to change their behavior later?

low-batt:
Initially I had thought about creating a base class. I didn't end up doing that as the amount of code that is sharable with `PlaySliderCell` is pretty tiny. The `PlaySliderCell.drawBar` deviates from what the volume slider has to deal with due to the additional customizations in the play slider, such as the custom knob.

PR #6074, opened 2026-05-16
Searching in the new Settings window (with refactor)

lhc70000:
Needed to refactor the whole framework to support searching.

- The builder only creates a tree of lightweight representations (Page, Section, List, Item).
- All strings for searching can be gathered through `registerSearchEntry()` from this tree without creating the `NSView`s. Each component can implement custom logic determining what content is searchable.
- When displaying the page, the actual `NSView` hierarchy is built by `makeView()`.
- Each component has an `itemID` corresponding to the materialized `NSView`'s tag.

Known issues / Todo:

- `SettingsItem.Custom()` currently doesn't automatically inherit the above behavior.


low-batt:
Just getting started looking at this. When there is only one result this truncation looks odd:
truncated

uiryuu:
Found several bugs.

1. Search some thing, click on the first entire, then quickly click on the second entry before the first animation is finished. Then the animation state is broken, all animation are very short afterwards.
2. Sometimes the search result window is not triggered. Try this: input "pause" (the window displays now), <C-a> then retype "pause".

Some suggestions:
1. I think multi-line labels are better than truncating the strings. Some of our settings are pretty long (maybe longer in other languages), truncating them make it harder to understand the difference between results.
2. Nipick
image
image
Sometimes it's on top of the highlighted box, sometimes it's on the bottom.

3. When the result popover is open, switching to another app dismisses it. When we focus on the settings window again, we should bring up the popover. Or consider putting the search results in the sidebar (like macOS' settings) so that we don't have to manage the state of the popover.
4. Make the popover longer to hold more entires/make it draggable to configure the length.

lhc70000:
- Made popover's size change with its content. Increased maximum height.
- Removed top/bottom padding for first/last items. Removed floating table group headers.
- Fixed popover getting dismissed when retyping the same search string.
- Fixed disrupted animation when rapidly clicking results.

We can think about how to display the results better to avoid truncating later, but perhaps not in this PR.

PR #6084, opened 2026-05-19
Sync playlist with external file deletions

hussainarslan:
## Summary
- Add a small player-owned `PlaylistFileMonitor` that watches parent folders for local playlist files and debounces filesystem events.
- Refresh the playlist from mpv before updating monitor subscriptions or pruning missing entries, keeping mpv as the source of truth.
- Remove missing local playlist items through the existing `PlayerCore.playlistRemove(_:)` path while leaving network/URL entries untouched.

## Why
Deleting or renaming a playlist file in Finder currently leaves a stale row in IINA's playlist. Trying to play that row fails because the file no longer exists. This keeps the playlist aligned with local filesystem changes without adding preference UI or changing playback APIs.

## Contribution notes
- Checked `CONTRIBUTING.md`; this PR targets `iina/iina`'s `develop` branch.
- Re-ran issue search for an exact duplicate and did not find one. Closest adjacent issue found: #4548.
- No localization changes, no generated mpv source edits, and the project file change is only to add the new Swift file.

## Validation
- `git diff --check`
- `plutil -lint iina.xcodeproj/project.pbxproj`
- `xcodebuild -project iina.xcodeproj -scheme iina build`
- `xcodebuild -project iina.xcodeproj -scheme iina test` cannot run because the `iina` scheme is not configured for the test action.

### DIFF
--- a/iina.xcodeproj/project.pbxproj
+++ b/iina.xcodeproj/project.pbxproj
@@ -10,6 +10,7 @@
 		1326717E20852D0D000FA7E2 /* SubChooseViewController.xib in Resources */ = {isa = PBXBuildFile; fileRef = 1326718020852D0D000FA7E2 /* SubChooseViewController.xib */; };
 		3412DB8E2C2FC64C00BBC142 /* Equalizations.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3412DB8D2C2FC64800BBC142 /* Equalizations.swift */; };
 		348F71962DE2A819006C587D /* MPVCommandWrappers.swift in Sources */ = {isa = PBXBuildFile; fileRef = 348F71952DE2A814006C587D /* MPVCommandWrappers.swift */; };
+		BD69B6CBE0BF44C68B1A703A /* PlaylistFileMonitor.swift in Sources */ = {isa = PBXBuildFile; fileRef = C55FCB1A2ADE4A1FBFF29EFF /* PlaylistFileMonitor.swift */; };
 		34ACAB722C142A0500F871A0 /* libintl.8.dylib in Frameworks */ = {isa = PBXBuildFile; fileRef = 34ACAB2D2C142A0300F871A0 /* libintl.8.dylib */; };
 		34ACAB732C142A0500F871A0 /* libgraphite2.3.dylib in Frameworks */ = {isa = PBXBuildFile; fileRef = 34ACAB2E2C142A0300F871A0 /* libgraphite2.3.dylib */; };
 		34ACAB742C142A0500F871A0 /* liblcms2.2.dylib in Frameworks */ = {isa = PBXBuildFile; fileRef = 34ACAB2F2C142A0300F871A0 /* liblcms2.2.dylib */; };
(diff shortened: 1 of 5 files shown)


uiryuu:
Tried this PR and it works. I think we can have this. I'll take a closer look later.

hussainarslan:
@uiryuu let me know if theres something needed from my side to get this merged

hussainarslan:
@lhc70000  @low-batt Hey guys, just following up on this pr. Let me know if there are any comments here.

low-batt:
Hello from a _junior_ IINA developer!

Sorry we have been slow to respond. Everyone is overloaded right now.

As I am a _junior_ IINA developer my opinions are not what matter. It is the senior developers who determine what gets merged. But here are some quick thoughts from me…

Usually IINA will merge PRs that are outside of the goals of the release that is being developed. I have been advocating that for the release that is currently under development we need to be choosey and be more focused on the goals of the release to avoid development taking as long as it did for 1.4.0.

Users are impatiently waiting for IINA to upgrade mpv, FFmpeg and other libraries IINA is dependent upon. That is one of the primary goals of the next release. IINA has been working on some significant UI changes in a branch. That was resulting in duplicate work as changes to the existing UI code had to then be recoded in the code being developed. So the other important goal is to finish these UI changes and get them merged to eliminate the need for duplicate work. These two goals represent a significant amount of work. This PR does not align with these goals.

Now that alone does not rule out merging a PR. Fine to merge simple safe PRs outside the goals of the release. But this PR definitely does not contain simple safe changes. It introduces a new `DispatchQueue`, but the real risk is use of file monitoring. Another PR that proposed use of file monitoring was just closed. See [this comment](https://github.com/iina/iina/pull/6040#pullrequestreview-4200043618) from me as to the concern that caused that PR to be abandoned. Certain types of file operations that we think of as modifying the file actually unlink the current file from the directory and replace it with the new contents. I'm expecting there is at least a rare theoretical race condition that needs to be examined in this code. Anyway, code changes like this need to be looked at closely.

In general developers should always consider where the appropriate place to add changes is. In the case of IINA the question is whether a new feature makes sense for the [mpv player](https://mpv.io/). If it does then ideally that feature would be added to `libmpv` and IINA would only need to be changed to make use of the new feature. The concept of synchronizing the playlist with the file system is applicable to mpv as well. Ideally this would be implemented by the mpv project. IINA will add changes that really should be implemented in `libmpv`. But this is another consideration to keep in mind.

Another consideration that is a factor is how much users are requesting a feature. The feature the closed PR #6040 was trying to add had been requested by some users. I could have missed it, but I do not remember any user requesting synchronizing playlist deletions.

None of the concerns I just listed prevents this PR from being merged. It is all about where reviewing this PR ranks given all the other work that is going on.

Remember, I'm a _junior_ IINA developer. Need to see what the senior developers think. They may decide they want to include this PR in the current release. At the moment they are overloaded, so it may be a while before they respond.

One thing that is missing is logging. With something like this we really need to be able to look at a log file and see what is going on. Also possibly there should be an OSD message when the playlist is altered. Wait on that to see what the other reviewers think.

PR #6088, opened 2026-05-22
Stabilizing the AutoLayout constraints helper API

lhc70000:
I added these extensions when designing the new settings page for common layout tasks.

Considering that we are gradually ditching xibs, we expect future code will use these extensions frequently. So it's important to make sure they are developer-friendly and grammatically sound.

Proposed API:

```swift
view.padding(.all)
view.padding(.horizontal(8))
view.padding(.top, .leading(4), from: aView)
view.spacing(.trailing(8), from: aView)
view.center(x: true, with: aView)
view.alignBaseLine(with: aView)
```

Please let me know if you have any suggestions.


iina/ConstraintsSugar.swift
uiryuu:
```suggestion
    else { fatalError("A constraint must have a value") }
```

iina/ConstraintsSugar.swift
```diff
@@ -163,7 +163,7 @@ extension NSView {
   }
 
   @discardableResult
-  func spacing(to aView: NSView? = nil, _ constraintList: ALConstraint...) -> Self {
+  func spacing(_ constraintList: ALConstraint..., from aView: NSView? = nil) -> Self {
```
uiryuu: Current `spacing(_,_)` cannot handle `.all()`. We should implement direction decomposition like that in `padding(_,_)`.

iina/ConstraintsSugar.swift
```diff
@@ -126,7 +126,7 @@ extension NSView {
   }
 
   @discardableResult
-  func center(with aView: NSView? = nil, x: Bool? = nil, y: Bool? = nil) -> Self {
+  func center(x: Bool? = nil, y: Bool? = nil, with aView: NSView? = nil) -> Self {
```
uiryuu: If `aView` is not in the superview, the constraint might not be correctly added. Consider
1. Putting the constraint in the common ancestor of `self` and `aView`
2. Or restricting `aView` to be in the superview in the doc

iina/ConstraintsSugar.swift
uiryuu: Spacing should only be applied in the sibling views; the constraints between self and superview should be set by `padding(_,_)`. Therefore we should make `aView` non-nil and never fallback to superview.

I checked that currently all `spacing(_,_)` calls don't pass nil as `aView`, so it's safe to change the signature.

uiryuu: "baseline" should be 1 word

iina/ConstraintsSugar.swift
uiryuu: `spacing(_, to aView)` would be better.

iina/ConstraintsSugar.swift
uiryuu: Maybe use an OptionSet (`[.x, .y]`) instead of `(x = true, y = true)`.

iina/ConstraintsSugar.swift
```diff
@@ -82,7 +82,7 @@ struct ALConstraint {
 
 extension NSView {
   @discardableResult
-  func padding(to aView: NSView? = nil, _ constraintList: ALConstraint...) -> Self {
+  func padding(_ constraintList: ALConstraint..., from aView: NSView? = nil) -> Self {
     let constraints = constraintList.flatMap { c in
       ALConstraint.Direction.list.compactMap {
         c.direction.contains($0) ? ALConstraint(direction: $0, relation: c.relation, constant: c.constant)  : nil
```
uiryuu: Double spaces

uiryuu (changes requested):
Would be nice if a small cheatsheet is provided at the beginning of this file.

iina/ConstraintsSugar.swift
lhc70000: Different from padding, we can't have `.all`, `.horizontal` or `.vertical` spacing, so we don't need decomposition.

iina/ConstraintsSugar.swift
low-batt: Something along these lines?
```swift
  struct OrientationOptions: OptionSet {
    let rawValue: Int
    static let x = OrientationOptions(rawValue: 1 << 0)
    static let y = OrientationOptions(rawValue: 1 << 1)
  }
  func center(_ align: OrientationOptions) {}
  func example() {
    center(.x)
    center([.x, .y])
  }
```
I would think that would be easier to read, especially since using both should be the rare case.

low-batt:
This PR got me to stress test layout. Here a two glitches. I confirmed they are not related to this PR. I was able to trigger them with `develop` branch.

In this case the `sans-serif` button probably could shrink instead of squishing the `Font` text so much:

font

No idea what is triggering the duplicate tabs. I think I've seen this before:

tabs

lhc70000:
Updated.

Issue #6089, opened 2026-05-22
PluginStorePanel fails to compile for Intel


low-batt:
### IINA Version

1.4.3

### Release Channel

Other

### macOS Version

macOS 26.5

### Steps to Reproduce

Build the current `develop` branch for Intel.

### Expected Behavior

Build succeeds.

### Actual Behavior

Build fails compiling `PluginStorePanel.swift`.

Commit 4762762 added use of the SwiftUI view method [onSubmit(of:_:)](https://developer.apple.com/documentation/swiftui/view/onsubmit(of:_:)) as well as other methods that require macOS 12+, conflicting with the Intel build's requirement to support macOS 11.

### How often can you reproduce the issue

Every time.

### Required Actions

- [x] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [x] I tested with the Nightly build and confirmed the issue is not fixed.
- [ ] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

lhc70000:
SwiftUI is not stable on macOS 11 and it's very difficult to test as I don't have an Intel Mac. My proposed solution is to make the plugin store macOS 12+ and only show this basic dialog on macOS 11:

<img width="814" height="714" alt="Image" src="https://github.com/user-attachments/assets/9d424fa9-97a1-4b9b-b82f-cf4650c8b1a1" />

I will push this after #6088 is merged.

We can also announce in the 1.5.0 release note that we are considering dropping macOS 11 support. If any user has a specific use case to stick on macOS 11, they can let us know, otherwise we will drop it perhaps in 1.6.0.

low-batt:
Dropping support will affect users trapped on macOS Big Sur because support for their old Macs were dropped by macOS Monterey. Information about that can be found in [macOS 12 Monterey Requirements Full List of Compatible Mac Computers](https://mrmacintosh.com/macos-12-requirements-full-list-of-compatible-mac-computers/) on [Mr. Macintosh](https://mrmacintosh.com/). The Macs dropped were in the 2013 - early 2015 range.

The other reason why people are trapped on old macOS versions is due to abandonware. Especially when it is device drivers for expensive hardware. Usually no possibility of switching to some other software in that case.

[...]

PR #6095, opened 2026-05-23
Reorder fields on the file tab of the inspector

low-batt:
This commit will move the new title and comment fields on the file tab of the inspector window so that they are right after the file path field.

---

**Description:**


low-batt:
This is what the `File` tab looks like now:
inspector

low-batt:
The CI build failure is due to issue #6089.

uiryuu:
Sorry to let you change this again, but I think the comment is more ﻿﻿natural to be at the bottom.

low-batt:
I moved the comment up next to the title after finding examples where it is very relevant and related to the title, like in the above example.

uiryuu (approved):

Issue #5645, opened 2025-08-24
Side panel causes aspect ratio distortion


low-batt:
### IINA Version

IINA 1.4.0-beta1

### Release Channel

Beta

### macOS Version

macOS 15.6

### Steps to Reproduce

- Download the `Frame rate bug test file.zip` file attached to [this comment](https://github.com/iina/iina/issues/5365#issuecomment-3217458296)
- Unpack the zip file
- Start playing the video `THE NEW 2025 BMW M5 TOURING #bmw #shorts.mp4`
- Pause playback
- Click on `Show Video Panel` under the `Video` menu
- Notice window expands and picture is distorted
- Click on `Hide Video Panel` under the `Video` menu
- Notice hiding the side panel did not restore the window size, the picture remains distorted
- Only if you attempt to resize the window will it snap back to the correct aspect ratio

### Expected Behavior

Desirable for IINA to never distort the video. Acceptable for IINA to distort the video while the side panel is shown, but IINA must not continue to show a distorted picture after the side panel is dismissed.

### Actual Behavior

The video remains distorted after the side panel is dismissed.

With the current development version of IINA the behavior has changed. The window does not enlarge, so the picture is not distorted. _However_ all is not well. Many constraint errors are reported to the Xcode console when the side panel is shown:
<details>
<summary><b>Constraint Errors:</b></summary>

```text
Conflicting constraints detected: (
    "<NSLayoutConstraint:0x600002079cc0 customAspectRatio.trailing == NSView:0x1440f6b20.trailing   (active, names: customAspectRatio:0x1422103d0 )>",
    "<NSLayoutConstraint:0x60000205fbb0 NSSegmentedControl:0x1440f9060.leading == NSView:0x1440f6b20.leading   (active)>",
    "<NSAutoresizingMaskLayoutConstraint:0x600002059590 h=-&- v=-&- NSClipView:0x1440f5bd0.minX == 0   (active, names: '|':NSScrollView:0x144956400 )>",
    "<NSAutoresizingMaskLayoutConstraint:0x600002059680 h=-&- v=-&- H:[NSClipView:0x1440f5bd0]-(15)-|   (active, names: '|':NSScrollView:0x144956400 )>",
    "<NSAutoresizingMaskLayoutConstraint:0x600002059db0 h=-&- v=-&- NSScrollView:0x144956400.minX == 0   (active, names: '|':NSTabView:0x132686ca0 )>",
[...]

PR #6098, opened 2026-05-25
Grow window to prevent the quicksettings from distortion

uiryuu:
- [x] This implements/fixes issue #5645.

---

**Description:**
Currently, the minimal window size is too small to fit the quick settings panel + 20% margin, and the quick settings panel will shrink, which breaks the quick settings view. This commit first fixes the quick settings panel's width, then grow the window if needed. The quick settings are no longer distorted.

Before: see #5645 
Now:

https://github.com/user-attachments/assets/d6f0830b-9422-43b3-913b-336213b972c1


low-batt (approved):
Pulled PR, built IINA, tested.

Changes look good to me.

PR #6099, opened 2026-05-25
New Settings Window update

lhc70000:
1. Made the plugin store macOS 12+ only (#6089). I removed the wording asking the user to submit an issue. It only says the feature is not available on macOS 11.
2. Mitigated the label width issue mentioned by @low-batt in #6088. (Couldn't reproduce the duplicate sidebar item problem; need further investigation)
3. Implemented initial window size/position bindings. There might still be some missing features in the new Settings window; need a double-check.


uiryuu:
LGTM. In testing the prompt for macOS 11, I noticed a lot of issues under macOS 11. Let me address those in another PR.

Update: fixes for macOS 11 are already pushed to this PR

low-batt:
I'm looking at this in detail.

The typos warnings are:
```text
Warning: "swich" should be "switch" or "swish".
Warning: "programatically" should be "programmatically".
Warning: "programatically" should be "programmatically".
```

low-batt:
I confirmed the layout issue with `Font` has been addressed.

I tried to reproduce the problem with duplicate tabs and wasn't able to trigger it. I will test some more.

`Auto hide after` can not be disabled.

`Base remaining time on playback speed` should be a subordinate setting of `Show remaining time instead of total duration`

Should `Thumbnail width` be a subordinate setting to `Enable thubnail preview` like the others? I think yes.

`Load ICC profile` is in a separate section from `Enable HDR mode` and `Enable tone mapping`. I'm guessing this represents the current IINA code that only applies tone mapping when HDR is active. But I believe that is a defect we should address in 1.5.0. You want to be able to apply tone mapping to HDR videos when displaying them on SDR displays. So I'm thinking tone mapping should be separate from HDR mode.

The `Gapless audio` setting is missing from the `Video & Audio` tab.

On the `Subtitles` tab the help comment for `Preferred language` is centered instead of left justified like others.

On the `Network` tab we now call out  `yt-dlp` but still to `youtube-dl`. Is that due to mpv still referring to it?

On `Control` tab `Single click to` list missing text. Same for others except `Middle click to`.

text


On the `Advanced` tab nothing happens when I click `Open Log Directory`.

Would be nice of the link in this help was clickable:

clickable


On `Plugins` I could not find `Permissions`. Or figure out how to uninstall a plugin.

Starting to look into the code changes…

low-batt:
I'm testing on my Intel Mac, so the build problems are fixed. Typos need to be fixed, other stuff I pointed out can be addressed in another PR as it will be good to get the Intel build working.

I guess localization has not been implemented yet?:

language

Some minor stuff:

Several of the settings related swift files contain mulitiple classes. Might want ot make use of `// Mark: -` to make it easier to spot the different classes in Xcode.

Documentation comments are supported on class properties, so in In SettingsKeyBindings, the comment shown below can be changed to a `///` comment. That will cause Xcode to include the comment in `Quick Help` when you select the `isLoadingConfig` property:
```swift
  // This variable is to prevent `NSTableView.reloadData()` in the `loadConfigFile` to trigger `loadConfigFile` again thus forming an infinite loop
  var isLoadingConfig = false
```

lhc70000:
Most of the comments addressed. Below are the remaining ones:

> Auto hide after can not be disabled.

Do we really need this ?

> On the Network tab we now call out  yt-dlp but still to youtube-dl. Is that due to mpv still referring to it?

Yes, just for legacy reasons. We might need to rename it.

> Or figure out how to uninstall a plugin.

It's inside the three-dots menu after the plugin name. Maybe we need to use a more prominent icon?

low-batt:
On the question of whether we need support for disabling `Auto hide after`…

My answer is absolutely yes because it is a currently supported feature. Not supporting it in the new settings would be considered a regression by some users. We don't want there to be any reasons for users to object to the new settings.

On the `youtube-dl` issue…

I'm good with whatever you decide to do on that. Good that we are now mentioning `yt-dlp` as users have been confused about what IINA is using.

On uninstalling plugins…

I totally missed the `⋯` was a button that triggered a context menu. Because there is an "Install" button visible I was expecting a corresponding "Uninstall" button.

On permissions…

I never would have thought to look for that on the `About` tab. As it is about security I prefer it to be more prominent and have its own tab like in the current settings.

Issue #3815, opened 2022-06-11
Support Live Text for video on macOS Ventura


arguiot:
**What you want IINA to do:**
Be able to support Live Text for video. There should be a button when you pause a video on any frame that contains text to be able to copy, translate, look up, share, etc.

**What IINA does currently:**
Doesn't support this feature.

**Why you think this should be added:**
Because it's available in QuickTime Player on macOS Ventura.

SENTINELITE:
Bumping this one. Having to open QT for videos is a huge time killer, for me.

Would love to see this added!

maqboolkhan:
bumping as well

HaileyesusZ:
would be great if we have this.

sudokar:
Can we have this, please?

uiryuu:
Fixed in #6101. Will be available in 1.5.0.

Issue #5805, opened 2025-11-11
Clickable subtitle


Ali72:
First of all, thank you for this great app!
I especially appreciate the dual subtitle feature — it’s extremely useful and well implemented.

What I want IINA to do:
I’d love to have clickable subtitles, so that when I click on a subtitle line, I can either copy it or trigger a custom action (for example, sending the text to a translation tool).

What IINA does currently:
Currently, subtitles are displayed on screen but are not interactive — you can’t select, copy, or interact with them.

Why I think this should be added:
This feature would make IINA even more powerful for language learners or people who watch foreign-language content. Being able to copy or interact with subtitle text (similar to what PotPlayer offers on Windows) would make studying and translating much easier.

Examples of other projects that have something similar:

PotPlayer (Windows): Allows clicking subtitles to copy or translate lines.

Adifagos:
And what happens to the remaining 99% of the populace that just wishes to watch a movie but, assuming this went through, everytime they'd click for a pause, they'd then have to watch _where_ they clicked? Because you literally just made it "interactive"?
Or wonder _what_ it is they clicked on and why this happens? I mean this isn't exactly media player 'default' behaviour, is it.
Maybe there should also be a toggle for that in settings? Even more work?
How about the floating UI for those that use it? What if one wanted to click on subtitles and simultaneously have it floating? Yet more work?
No matter the abundance of issues already in a backlog?

English isn't my native language, it isn't even my _second_ language. Kindly take a guess as to whether i ever relied on movies to master it.

Another user's thoughts, by all means ignore freely.

low-batt:
This sounds like a request to support Apple's [Live Text](https://support.apple.com/en-us/120004) feature. I looked into that quite a while ago when Apple first introduced that feature. It is supported on macOS, but most of the documentation is focused on the iPhone and iPad. I was unable to get Apple's sample program to work and turned my attention to other matters. Later I did hear from another IINA developer who was working on it. I've not checked with them. I've been assuming they got busy and turned their attention to something else as well.

Other users have also requested this feature. As @Adifagos noted this would require a setting to enable or disable this feature. And as @Adifagos also noted there are some UI design problems with exactly how such a feature would work that need to be solved.

kerim:
This is precisely what I want as well. I'm wondering if it is possible to add this feature via IINA's plugin system?

kerim:
Turns out it is! I've vibe-coded a solution here. Works on my MacBook Air.

[...]

Issue #4961, opened 2024-05-29
Some feature requests


giovariot:
I'm trying to use the program to summarize some video lessons, so using it for this task I've thought about 2 useful features that would be useful for that task as well as some other thoughts about the player in general.

I'd love to be able to select text using the live text feature of macOS while the video is paused. Do you think it would be possible to implement such feature using ImageAnalysisInteraction?

Also I tried editing the Key Bindings configuration in order to use the ⏪️⏩️ keyboard keys (corresponding to the F7 and F9) while writing text on another program to make them seek -5 and +5 seconds instead of playing next or previous items. Still, even if I edited the NEXT and PREV keys into the mpv commands "seek -5" and "seek 5", pressing ⏪️⏩️ it's still starting to play other items.

I've also noticed the player customizable toolbar is missing a simple audio language selection button similar to the one used for subtitles, which is a control present in the standard mpv version. I'd love to add such a button to the custom toolbar as it'd be very comfy.

I've also noticed the app is still using mpv 0.35 instead of the newer 0.38 and ffmpeg 6.0 instead of the newer 7.0. Compared to the standalone mpv player playback and jumps from small window to full screen iina feels less snappy, could this usage of older version be the reason why?

Also a nice ease-of-use function would be a double-click-to-reset feature on the sliding bar for controlling video equalization (such a double click to bring back the brightness control to zero and so on). Now double clicking anywhere on the slider it just sets the value to the clicked area.

Thanks a lot to the developers, this is a really great project.

low-batt:
Caught us very busy working on a new release. So for now, just a couple of comments…

> Compared to the standalone mpv player playback and jumps from small window to full screen iina feels less snappy, could this usage of older version be the reason why?

By default `mpv` will use a zooming animated transition effect when transitioning from a small window into full screen mode. I would not term that snappy. Are you setting the `mpv` [native-fs](https://mpv.io/manual/stable/#options-native-fs) option to `no`? With that setting `mpv` will directly transition to full screen mode without any transition. That could be the snappy behavior you are referring to.

The behavior of IINA when transitioning from a small window into full screen mode depends upon two settings:
- macOS [Reduce motion](https://support.apple.com/guide/mac-help/stop-or-reduce-onscreen-motion-mchlc03f57a1/mac) 
- IINA `Use legacy full screen` 

For the fastest transition from a small window into full screen mode enable both of these options. The IINA `Use legacy full screen` setting can found in the `Behavior` section on the `General` tab.

[...]

Issue #4168, opened 2023-01-09
Selectable subs


matsumurae:
**What you want IINA to do:**

I would like to have selectable subs like [Voracious](https://voracious.app/) does. Having a dict is not required (for me, but maybe it's interesting), I just want to select the word that I don't know how to write.

**What IINA does currently:**

I can tweak the subs but they're not selectable T-T

**Why you think this should be added:**

It could be useful for lang learners to select words or even full sentence of subs.

If selectable it's not possible, maybe something like [AnimeBook](https://animebook.github.io/) does with the subs: having them somewhere so you can follow. Pretty similar to what [Animelon](https://animelon.com/) does.

**Examples of other projects that have something similar:**

- [Voracious](https://voracious.app/)
- [Memento](https://ripose-jp.github.io/Memento/).
- [AnimeBook](https://animebook.github.io/)
- [Animelon](https://animelon.com/)

low-batt:
Really excellent description of the requested functionality. The references to other projects are very helpful.

A few quick thoughts…

IINA uses a library from the [mpv project](https://github.com/mpv-player/mpv) for audio/video support. With respects to subtitles, certain subtitle related features such as downloading from subtitle sites is provided by IINA code. However the majority of subtitle processing is performed by mpv. It is mpv that renders subtitles into the video frame. IINA merely displays the frame constructed by the mpv library.

For IINA to be able to implement this it would need information from mpv about the subtitle it rendered into the video frame. Of interest is the mpv property [sub-text](https://mpv.io/manual/stable/#command-interface-sub-text). From the mpv manual:
> The current subtitle text regardless of sub visibility. Formatting is stripped.
> If the subtitle is not text-based (i.e. DVD/BD subtitles), an empty string is returned.
>
>    This property is experimental and might be removed in the future.

That last statement is concerning. IINA would not want to construct a new feature based on a mpv property that is experimental and might be removed. Before attempting to implement such a subtitle feature IINA would need to check in with the [mpv project](https://github.com/mpv-player/mpv) on the current status of this property.

The reference to DVD/BD subtitles has to do with subtitles that are image based. There are many different subtitle formats. Some of these formats are not text based. They are small "pictures" to embed in the video. [Optical character recognition](https://en.wikipedia.org/wiki/Optical_character_recognition) would be required to obtain text from such subtitles.

matsumurae:
Glad to know that my explanations are clear. I'm used to write issues as I'm a web dev haha

[...]

PR #6101, opened 2026-05-26
Add live text support

uiryuu:
- [x] This implements/fixes issue #3815, #4168, #5805, #4961.

---

**Description:**
This PR added live text support using Apple's [VisionKit](https://developer.apple.com/documentation/visionkit). This feature is only supported on macOS 13+, so the settings and OSC toolbar button will only appear on applicable systems.

Upon pausing the video, a screenshot command is sent to mpv and subsequently, the image analysis is triggered via `ImageAnalyzer` of `VisionKit`. Then the results will be shown on `ImageAnalysisOverlayView`. This overlay will blocks mouse interaction to enable text selection and the menus.

In the current implementation, 2 preference keys control the live text behaviors:
- `enableLiveText`: The live text is only enabled if this key is true. Setting this key to true will trigger image analysis for every video pause (sub change, file change, playback position change, etc).
- `liveTextOverlay`: Since the overlay view blocks user interactions with the video view, we use this key to control whether or not to show the overlay. This key can be controlled by:
  - A OSC Toolbar button
  - Settings page

When the user want to use live text, they will first enable the live text feature in the settings, then enable the overlay as well. If the user wants to dynamically control the overlay, they can add the OSC toolbar button, so they can show the overlay on-the-fly.

When the user interact with they overlay, the fadeable UI elements are temporarily hidden, until the user interaction is finished. This includes:
- clicking on the live text button (on the bottom-right corner, and btw the inset of this button is dynamically changed based on the current OSC position, to ensure the OSC will not block this button)
- selecting the text in the image
- interacting with the menu

After the user finished inspecting the analyzed image, they can resume the playback, which will invalidate the previous analysis and remove the overlay from the view hierarchy.

**Examples:**
Text (for embedded subtitles, external subtitles, and video content) OCR:
image

Translation:
image

Object extraction:
image


low-batt:
I think we need a menu items for live text and for the overlay, similar to how IINA usually has many ways to control a feature. Probably should also have controls on the `Video` panel.

The overlay messes me up because I use right click to pause and resume playback. Once I pause playback, right clicking the mouse no longer restarts playback.

I didn't find any calls to [ImageAnalyzer.isSupported](https://developer.apple.com/documentation/visionkit/imageanalyzer/issupported). I _think_ that is ok as I _think_ that all Macs capable of running macOS Ventura support this. But I wasn't able to find Apple documentation indicating this is true.

I'm seeing a lot of these warnings:
```text
15:32:58.839 [window0][w] No data for tracking area
```

IINA is not taking into account that Live Text may be disabled in System Settings as shown here:

settings

I think that might be why I see this in the log file?:
```text
15:28:22.215 [window0][v] Recreating tracking area
Visual isTranslatable: NO; reason: observation failure: noObservations
```

As this is a new features it would be good to have debug level messages indicating when the overlay has been added and removed.

Try this:
- Pause playback
- Enter full screen mode 
- Exit full screen mode

For me that causes this to happen:

constraints


The log looks like this:
```
15:28:22.105 [window0][d] Exiting full screen mode
15:28:22.136 [window0][d] Restoring dock settings
15:28:22.142 [window0][d] Forcing drawing, window resized during animated enter or exit full screen
15:28:22.143 [mpv0][v] Set property: window-scale=0.6666666666666666
Conflicting constraints detected: (
    "<NSAutoresizingMaskLayoutConstraint:0x6000035b13b0 h=-&- v=-&- VKFlippedGlassContainerView:0x7f8443b1c290.minY == 1360   (active, names: '|':VKCActionInfoView:0x7f843afb5e80 )>",
    "<NSAutoresizingMaskLayoutConstraint:0x6000035b1400 h=-&- v=-&- V:[VKFlippedGlassContainerView:0x7f8443b1c290]-(8)-|   (active, names: '|':VKCActionInfoView:0x7f843afb5e80 )>",
    "<NSAutoresizingMaskLayoutConstraint:0x6000035b1540 h=--& v=--& VKCActionInfoView:0x7f843afb5e80.height == 1320   (active)>"
).

Will attempt to recover by breaking <decode: bad range for [%@] got [offs:848 len:174 within:0]>.
Unable to simultaneously satisfy constraints:
(
    "<NSAutoresizingMaskLayoutConstraint:0x6000035b13b0 h=-&- v=-&- VKFlippedGlassContainerView:0x7f8443b1c290.minY == 1360   (active, names: '|':VKCActionInfoView:0x7f843afb5e80 )>",
    "<NSAutoresizingMaskLayoutConstraint:0x6000035b1400 h=-&- v=-&- V:[VKFlippedGlassContainerView:0x7f8443b1c290]-(8)-|   (active, names: '|':VKCActionInfoView:0x7f843afb5e80 )>",
    "<NSAutoresizingMaskLayoutConstraint:0x6000035b1540 h=--& v=--& VKCActionInfoView:0x7f843afb5e80.height == 1320   (active)>"
)

Will attempt to recover by breaking constraint 
<NSAutoresizingMaskLayoutConstraint:0x6000035b1400 h=-&- v=-&- V:[VKFlippedGlassContainerView:0x7f8443b1c290]-(8)-|   (active, names: '|':VKCActionInfoView:0x7f843afb5e80 )>

Set the NSUserDefault NSConstraintBasedLayoutVisualizeMutuallyExclusiveConstraints to YES to have -[NSWindow visualizeConstraints:] automatically called when this happens.  And/or, set a symbolic breakpoint on LAYOUT_CONSTRAINTS_NOT_SATISFIABLE to catch this in the debugger.
15:28:22.167 [window0][d] Forcing drawing, window resized during animated enter or exit full screen
15:28:22.169 [mpv0][v] Set property: window-scale=0.6666666666666666
15:28:22.188 [window0][d] Forcing drawing, window resized during animated enter or exit full screen
15:28:22.189 [window0][d] Exited full screen mode
```

uiryuu:
> I think we need a menu items for live text and for the overlay, similar to how IINA usually has many ways to control a feature. Probably should also have controls on the Video panel.

Added a menu item to toggle the overlay. I don't want to add it to the quicksettings, because it's pretty awkward to control the overlay in the quicksettings, you have to close the quicksettings to see the whole video; and secondly, we @lhc70000 is working on the quicksettings rn. So even if we want to add the toggle it has to be after that.

> The overlay messes me up because I use right click to pause and resume playback. Once I pause playback, right clicking the mouse no longer restarts playback.

That's the reason why I make a toggle on the OSC toolbar, so that the user can choose whether or not to use the overlay. When there's no overlay, the input is not intercepted. And this live text overlay heavily use right click menu for user interactions. The user can right click to extract object from the view. Therefore I don't think it's possible to circumvent the right click menu of the overlay completely.

> I'm seeing a lot of these warnings:

I failed to figure out what is causing this. Maybe the overlay is using its own tracking mechanism, but somehow the events are forwarded to our main window's content view. Can we remove this warning log and just ignore events without userInfo?

> IINA is not taking into account that Live Text may be disabled in System Settings as shown here:

Added `AppDelegate.liveTextAvailable` to check wehther or not live text is enabled in the system settings.

> As this is a new features it would be good to have debug level messages indicating when the overlay has been added and removed.

Added logs.

> For me that causes this to happen:

Fixed by remove the overlay before entering/exiting full screen. Also fixed bugs when the window aspect ratio is unlocked. Basically we have to re-analysis the image after the window now aspect ratio is changed.

low-batt:
> Fixed by remove the overlay before entering/exiting full screen.

NOT fixed for me. Entering and exiting full screen mode still triggers the constraint errors for me.

> Added a menu item to toggle the overlay.

Sorry I didn't bring this up before. As we are redoing the settings window, now would be a good time to cleanup some wording. The use of the word `Enable` has always bothered me. Look through `System Settings`. There is no use of `Enable` in the names of settings.  The state of the button indicates whether the setting is enabled or disabled. So rather than add `Enable Live Text` for the setting, I'm thinking it should just be  `Live Text`. Of course we'd want to rename other settings to make things consistent. With the big settings change this is a good time to do that.
 
I mentioned this feature to someone and they were confused. They did not understand what "live text" was referring to. I think we are stuck with that term because that is Apple's official name. The description of the `Enable Live Text` setting is clear and should be understandable by users. But I'm thinking the setting should also have a help link, probably to [Use Live Text to interact with text in a photo on Mac](https://support.apple.com/guide/photos/live-text-interact-a-photo-pht6bc6bd5f5/mac). 

The `Show Live Text overlay` setting and the new `Show Live Text Overlay` menu item are a problem for multiple reasons. The word `Show` seems very inappropriate to me. Show usually means displaying something. The user would expect clicking on that to "show" something. Worse is "overlay". This is an example of implementation developer speak sneaking into the UI. The description refers to "overlay" and "video view". That is weird developer speak to a lot of users.

Effectively the feature has 3 states?:
- Disabled `Enable Live Text` disabled
- Standby `Enable Live Text` enabled `Show Live Text overlay` disabled
- Active `Enable Live Text` & `Show Live Text overlay` enabled

Do we need that? Is there a big advantage to that over just `Live Text` enabled/disabled?

The nested `Show Live Text overlay` is also the only nested setting to use a symbol. I think it is preferable to not use symbols on subordinate settings. Easier to see the indentation without the symbol.

As a user likely will have to toggle this, a new key binding is desirable.

>  I don't want to add it to the quicksettings, because it's pretty awkward to control the overlay in the quicksettings

Sounds fine. 

> Therefore I don't think it's possible to circumvent the right click menu of the overlay completely.

Also sounds fine.

> Added `AppDelegate.liveTextAvailable` to check wehther or not live text is enabled in the system settings.

Apparently that needs to be checked in more places? I'm seeing logs indicating IINA is attempting to analyze an image and failing when I disable the macOS setting.

Would be good to be able to clearly tell from a log file if Live Text is not available due to running on macOS 12 or 11. That one would be more for users as we already can see the macOS version in the log. But it would be good to have a log for us to make it clear that Live Text is not available because the user disabled it in `System Settings`. I noticed users in the Apple forum confused as to why Live Text was not working, only to eventually track it down to the setting in `System Settings` being turned off.

uiryuu:
> NOT fixed for me. Entering and exiting full screen mode still triggers the constraint errors for me.

Not sure what caused this. I'm still invesgating. Also I can only trigger this when *exiting* fullscreen, not entering fullscreen. Don't know why either.

> Do we need that? Is there a big advantage to that over just Live Text enabled/disabled?

Having `Enable Live Text` let the user to save energy when they don't want to run image analysis in the background every time they pause the video. You can argue that QuickTime doesn't have this option, but I bet you will prioritize energy saves over this.

Having `Show Live Text overlay` let the user to have the ability to remove the overlay so that the interactions with the video view can be used normally (right click, etc).

> Apparently that needs to be checked in more places? I'm seeing logs indicating IINA is attempting to analyze an image and failing when I disable the macOS setting.

Currently I'm not updating this variable until the user restart IINA. Let's first resolve other important issues, I'll add support for this later.

low-batt:
The constraint error occurs for me only when exiting full screen mode as well.

The `Enable Live Text` setting makes sense. What I don't understand is why the `Show Live Text overlay` is needed. Why can't you just toggle off `Enable Live Text` when you want to interact normally with the view?

uiryuu:
> The constraint error occurs for me only when exiting full screen mode as well.

Will push a fix soon.

> The Enable Live Text setting makes sense. What I don't understand is why the Show Live Text overlay is needed. Why can't you just toggle off Enable Live Text when you want to interact normally with the view?

My original thought was that, because the analysis takes time, if you just have 1 toggle to control the live text, then when you enable that with the video paused, you'll have to wait for a bit until the analysis finishes. However, now I think about it, it doesn't really matter. The analysis is pretty fast (at least on my M3 Max machine). I'll simplify the logic so that only 1 toggle controls live text.

low-batt:
I definitely agree with the switch to just a single setting. That will be a lot easier for users to understand.

Several problems in the screenshot below taken under macOS 15.7.4. Duplicate `Enable Live Text` menu items. And should just be `Live Text`. It is the checkmark that indicates whether this feature is enabled or not and what clicking on the menu item does (enables or disables).

Then there is the odd lock symbol for `Lock Window Aspect Ratio`. Looks really out of place.

menu

I assume the constraint error has not be addressed, as I can still reproduce it. It also occurs if you resize the window, or switch to music mode.

uiryuu:
> Several problems in the screenshot below taken under macOS 15.7.4. Duplicate Enable Live Text menu items. And should just be Live Text. It is the checkmark that indicates whether this feature is enabled or not and what clicking on the menu item does (enables or disables).

Renamed.

> Then there is the odd lock symbol for Lock Window Aspect Ratio. Looks really out of place.

The symbol was introduced by #6100. Fixed in this revision. The wrong labels were due to my mistake when duplicating the localization strings. I forgot to change to the correct keys.

> I assume the constraint error has not be addressed, as I can still reproduce it. It also occurs if you resize the window, or switch to music mode.

I think this issue is finally fixed by this commit. Now the overlay view is recreated every time without reusing the previous one. Please try it out.

During debugging, I also notice a small defect. When the window is unlocked, with media paused, and you try to change the window size, there will be a short image retention of the paused frame. This is caused by the overlay, even though the current implement already tried to remove the overlay at the earliest possible time (in the `windowWillStartLiveResize`). Hopefully that will not be that interfering...

low-batt:
Really looking good. But I still triggered a constraint error:
- Start playing a video and pause
- Double click on the title
- Double click again on the title
- Constraint error

I'm also able the trigger this using the green button, as I have the button configured to not go into full screen mode.

I tried the CI build again and this is how it failed:
```test
fatal: unable to access 'https://github.com/groue/GRMustacheSpec.git/': Could not resolve host: github.com
```

I'm thinking GitHub is currently having some trouble…

uiryuu:
I pushed another commit to clear the analysis and re-request the analysis without considering whether or not the aspect ratio of the window is locked. This will ensure the overlay is removed during any window resize operation. Hopefully the constraint error will not occur anymore.

low-batt:
I'm still getting a constraint error following the instructions I posted earlier about double clicking on the title bar twice.

low-batt:
Constraint error still there with the latest commit. It is happening because this is not a "live resize", so the clearing of the analysis that is done is `windowWillStartLiveResize` is never called. Only the `windowWillResize` method is being called. I tracked the reason why this is not being treated as a `live resize` to my having `Reduce motion` enabled in `System Settings`. Try enabling that and the constraint error should reproduce for you.

uiryuu:
@low-batt Please test again!

iina/ViewLayer.swift
```diff
@@ -215,6 +215,35 @@ class ViewLayer: CAOpenGLLayer {
...
+      if let handler {
+        let image = framebufferSnapshot(width: Int(dims[2]), height: Int(dims[3]))
+        handler(image)
+      }
+    }
+  }
+
+  // MARK: - Snapshot
+
+  private let snapshotLock = NSLock()
```
low-batt: We have our own `Lock` class in `Utils`.

iina/ViewLayer.swift
```diff
@@ -215,6 +215,35 @@ class ViewLayer: CAOpenGLLayer {
...
+  /// Capture the next rendered frame as an `NSImage`. Forces a redraw so this works while paused.
+  func captureSnapshot() async -> NSImage? {
+    await withCheckedContinuation { continuation in
+      snapshotLock.lock()
+      // Drop any previous in-flight request — only the latest caller wins.
+      let previous = pendingSnapshotHandler
+      pendingSnapshotHandler = { image in
+        continuation.resume(returning: image)
+      }
+      snapshotLock.unlock()
```
low-batt: Hard to do it here with the way the code is structured, but a nice pattern to use with locks is to have a `defer` statement right after the call to `lock()` that does the unlocking.

iina/LiveTextSupport.swift
```diff
@@ -0,0 +1,112 @@
...
+//
+//  Created by Yuze Jiang on 5/26/25.
+//  Copyright © 2025 lhc. All rights reserved.
+//
+
+import Cocoa
+import VisionKit
+
+fileprivate let subsystem = Logger.makeSubsystem("livetext", ["text.viewfinder"])
+fileprivate func liveTextLog(_ str: String, level: Logger.Level = .debug) {
```
low-batt: Log helper methods like this now use `@autoclosure () -> String` as the type of the message string. See one of the similar methods. If the string being passed contains expressions that need to be evaluated, that won't happen if the logger determines logging is not enabled or this message will not be logged due to the level.

low-batt (commented):
Constraint error all fixed. Just a couple of trivial issues.

low-batt (approved):
Looks ready to be merged to me!

uiryuu:
Rebased to latest dev and removed unnecessary commits from other branches.

uiryuu:
Merged locally

Issue #6096, opened 2026-05-24
Holding option key when opening a new file from menu bar does nothing


psmpm:
### IINA Version

1.4.3

### Release Channel

Release

### macOS Version

Sequoia 15.7.4

### Steps to Reproduce

1. Open a media file.
2. Observe that the first option in the "File" menu bar item is either "Open..." or "Open in New Window..." depending on whether the "Allow multiple player windows" setting is on.
3. Hold the option key and observe that the first option in the "File" menu bar item has switched to the opposite text.
4. While holding the option key, press the first item and open another media file.

### Expected Behavior

According to the text of the menu item, the opposite to the usual behavior should be carried out when it is selected while holding down the option key.

### Actual Behavior

The same behavior occurs as if the option key was never held. Technically, according to the name of the "Allow multiple player windows" setting, this is correct when that setting is disabled, because holding the option key should open a new window and that does not happen, i.e. it is not allowed. However, the behavior is still incorrect when that setting is on, because the item no longer says that it will open a file in a new window and yet it does anyway. It seems this change was made (along with the change to rename "Always open media in new window" to "Allow multiple player windows") in #5868. Therefore, it seems like it was intended to remove this inversion behavior entirely, but the author of that PR (cc @Cykelero) did not realize to remove the text changing when holding the option key. The argument for removing this was that it was hard to discover and that it would be somewhat confusing with the parallel change to opening the same media file twice. I disagree with both of those points, since holding the option key for variant choices is a very common pattern in MacOS, and you could always just make the alternate option do what it says it is going to do. I also think it's sad to lose functionality here, since I would prefer to open files in new windows by default but have the temporary option to open one in the same window.

### How often can you reproduce the issue

Always

### Required Actions

- [x] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [x] I tested with the Nightly build and confirmed the issue is not fixed.
- [ ] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

psmpm:
TL;DR Either the inversion option suggested by the text changing, removed by #5868, should be reimplemented, or the text changing should be removed as to not lie to the user. I would prefer the former as this is functionality that I used.

low-batt:
Reproduced for me.

I've not looked into this yet, but this seems like a defect to me. IINA's setting establish default behavior. But you can choose to override many of those settings. For example, if you enable hardware decoding in IINA's settings you can use the video panel to turn it off.

[...]

PR #6109, opened 2026-05-29
Fix: issue #6096 Holding option key when opening new file from menu bar corrected

leifgrove:
- [ x] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)
- [ x] This implements/fixes issue #6096 

---

Reversed functionality of option key when choosing file open from menu bar, without removing changes made in #5868, expected behavior: holding option while choosing open file from menu bar results in opening file in same window if the option to "allow multiple player windows" is checked. if the option is not checked, the option key changes the text to "open in new window"

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -82,7 +82,7 @@ class PlayerCore: NSObject {
   }
 
   static func activeOrNewForMenuAction(isAlternative: Bool) -> PlayerCore {
-    let useNew = Preference.bool(for: .alwaysOpenInNewWindow) != isAlternative
+    let useNew = Preference.bool(for: .alwaysOpenInNewWindow) == isAlternative
     return useNew ? newPlayerCore : active
   }
   


low-batt:
Thank you for your PR! As it is a regression, fixing issue #6096 is a priority.

Always good to do a deep analysis of what exactly went wrong with the PR that introduced the regression. That might be needed in this case as I don't think the proposed fix is working?

Follow these instructions:
- Start IINA
- The settings panel appears
- In the `Behavior` section disable `Allow multiple player windows`
- Open a video
- Open the `File` menu
- Press ⌥
- The `Open…` menu item changes to `Open in New Window…`
- Click on `Open in New Window…`
- In the open panel select a different video and click `Open`
- IINA starts playing the new file
- Open the `Window` menu
- Notice there is only _one_ window open
- Notice the window has the title of the _previous_ video

PR #6115, opened 2026-05-31
Added section headers to playlist context menu (macOS 14+)

uiryuu:

---

**Description:**
Added section headers to the playlist context menu for macOS 14+ to increase its readability.

Before:
image

Now:
image


low-batt:
Looks good!

I'm thinking it would be more consistent if instead of taking up two lines the `Subtitles` section title had a suffix provding the status, like `Subtitles (36 matched)` instead of a separate line.

subtitles

low-batt:
Thinking some more…

Isn't "matched" is the wrong term? It assumes the subtitles were added by `Auto Load` matching the file. That is the usual case. But the subtitles could have been added using the `Load External Subtitle…` menu. The menu only knows that X number of subtitles have been loaded. So it probably should be `Subtitles (36 loaded)`.

PR #6116, opened 2026-06-02
Fix watch later MD5 sum calculation

low-batt:
This commit will:
- Change the `Utility.mpvWatchLaterMd5` method to match the sum calculated by mpv
- Add a new `ignorePath` parameter to `mpvWatchLaterMd5` in support of the mpv `ignore-path-in-watch-later-config` option
- Add a new `ignorePathInWatchLaterConfig` property to `PlayerCore` that gets and caches the value of the mpv `ignore-path-in-watch-later-config` option
- Change `PlayerCore.refreshCachedVideoInfo` to use the `ignorePathInWatchLaterConfig` property when calling `mpvWatchLaterMd5`
- Remove the call to `mpvWatchLaterMd5` from `PlaybackInfo`
- Change `PlayerCore.openMainWindow` to call `mpvWatchLaterMd5` and set `PlaybackInfo.mpvMd5`
- Change `PlayerCore.fileStarted` to call `mpvWatchLaterMd5` and set `PlaybackInfo.mpvMd5`
- Add a `mpvMd5` parameter to the `PlaybackHistory` initializer and remove the call to `mpvWatchLaterMd5`
- Add an `ignorePath` parameter to `HistoryController.add`
- Change `HistoryController.add` to call `mpvWatchLaterMd5` and pass the sum when constructing the `PlaybackHistory` object
- Change `PlayerCore.fileLoaded` to pass `ignorePathInWatchLaterConfig` when calling `HistoryController.add`
- Add a `cat_watch_later_file.sh` script to `other` that makes it easy for developer to see the contents of a watch later file

This commit will correct the MD5 sum calculation for YouTube videos to match the sum calculated by mpv. The current code is calculating the sum based on the URL path, which for a YouTube video is just `/watch`. This prevents IINA from finding the watch later file and also causes the history file to have only one entry for all YouTube videos.

---

**Description:**

### DIFF
--- a/iina/HistoryController.swift
+++ b/iina/HistoryController.swift
@@ -91,16 +91,18 @@ class HistoryController: NSObject {
   ///   - url: URL of the media being played.
   ///   - duration: Total duration of the media.
   ///   - title: Title of the media (if available).
-  func add(_ url: URL, duration: Double, title: String?) {
+  ///   - ignorePath: When `true`, only the URL's filename will be used for the sum if the URL does not contain a scheme.
+  func add(_ url: URL, duration: Double, title: String?, _ ignorePath: Bool) {
     guard Preference.bool(for: .recordPlaybackHistory) else { return }
     $tasksOutstanding.withLock { $0 += 1 }
     queue.async { [self] in
+      let mpvMd5 = Utility.mpvWatchLaterMd5(url, ignorePath)
       $history.withLock { history in
-        if let existingItem = history.first(where: { $0.mpvMd5 == url.path.md5 }),
+        if let existingItem = history.first(where: { $0.mpvMd5 == mpvMd5 }),
            let index = history.firstIndex(of: existingItem) {
           history.remove(at: index)
         }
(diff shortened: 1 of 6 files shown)


low-batt:
As seen here, with this commit the `Playback History` window will now correctly show the history of YouTube videos played and correctly show playback progress for videos that can be resumed:
history

uiryuu (approved):
The changes look good to me. Just some suggestions further improving the history window to be considered...

We've saved the title string in the history db, which is far more useful than the URL in the history window.

Screenshot 2026-06-04 at 14 18 46

For online resources, the icon is just empty file icon, which makes no sense. Maybe we can set it to 🌐 to indicate its not a local file.

The column name "file" is also weird to me. Also when grouped by "folder", all URLs are grouped in the "/" group:

image

We make these improvements in followup PRs.

uiryuu:
Or maybe I can propose a new playback history window in swiftUI/pure code appkit to get rid of the xib to include these improvements.

PR #6117, opened 2026-06-04
Rewrite HistoryWindowController to ditch xib

uiryuu:

---

**Description:**
This commit rewrites the xib using pure swift code (AppKit). Visually almost 1:1 recreation from the old xib, with bug fixes and additions:
- Added a delete file menu item
- Added keyboard shortcuts for all menu actions
- Added menu item action images
- Use toolbar
- Validate all actions before executing

Localization is yet to be added...


iina/HistoryWindowController.swift
```diff
@@ -44,35 +45,77 @@ class HistoryWindowController: NSWindowController, NSOutlineViewDelegate, NSOutl
...
   ]
 
-  override var windowNibName: NSNib.Name {
-    return NSNib.Name("HistoryWindowController")
-  }
-
-  @IBOutlet weak var outlineView: NSOutlineView!
-  @IBOutlet weak var historySearchField: NSSearchField!
+  let scrollView = NSScrollView()
+  let outlineView = NSOutlineView()
```
low-batt: The XiB used the IINA `OutlineView` class to support the ability to disable the sliding animation when the disclosure triangles are collapsed and expanded.
```suggestion
  let outlineView = OutlineView()
```

low-batt:
Yeah! I was hoping to see the history window changed to have the controls in the title like the log window.

typos caught the miss-spelling of "History" in this code:
```swift
    let toolbar = NSToolbar(identifier: "HitoryWindowToolbar")
```

First the comments from PR #6116…

----
On this from PR #6116:
> We've saved the title string in the history db, which is far more useful than the URL in the history window.

Yes, BUT this is a source of complaints about the playlist. Users have complained about not being able to see the URL in the playlist when we just show the title. Maybe when there is a title we show that and then in dim smaller text we show the filename / URL under the title?

On this from PR #6116:
> For online resources, the icon is just empty file icon, which makes no sense. Maybe we can set it to 🌐 to indicate its not a local file.

I am going to question why we are showing an icon at all. Unlike the Finder, clicking any of these files will open it in IINA. The icon is really tiny. If we make the above change then you will be able to see the URL, so no need to have an icon to indicate it is not local.

On this from PR #6116:
> The column name "file" is also weird to me.

This is a problem in many places in IINA and mpv. Maybe change it to be `Media`? Ha! Just noticed you have already made that change.

On this from PR #6116:
> Also when grouped by "folder", all URLs are grouped in the "/" group:

I propose changing the button to be `Folder / Domain`. When that button is selected streaming entries would be grouped by domain. Sorting would be by reversed domain name. So all `.com` entries would be separate from `.net` entries. Secondary sorting would be on date like is done for folders.

----

Feedback on the PR behavior…

This fixing one of my complains. In the current window, if you search and then click on an item the history window is updated to reflect that entry is now on today's date, but in doing that the old implementation did not reapply the filter. The filter is correctly reapplied in this PR.

As seen here when testing under macOS 15.7.4 the position slider is broken:

history

The values in the `Played at` column look at bit messy. Maybe remove the `,`, left justify the date and right justify the time. That would align the AM / PM. Maybe something like that?

Also the scroll bar is missing.

When you click on column titles they react and briefly highlight the column title acknowledging the click. Users might do this thinking it could change the sort order, like it does in Finder. As we don't support that the column titles should not react to clicks.

By default the disclosure triangles are all expanded. I believe that is the correct choice for the default behavior.  When displaying by `Folder / Domain` a user currently will need to scroll down to find the folder or site they are interested in. That would be easier to do if all of the disclosures were collapsed. Should we have a button to do that? I didn't find any guidance in the [HIG](https://developer.apple.com/design/human-interface-guidelines/disclosure-controls#Disclosure-triangles) about this.

On this:
> Added a delete file menu item

menu

Users seem to want this. I really like read only programs where I know I can't accidentally change something, including especially deleting a file. I prefer a true pure read only player. I'm constantly annoyed when I accidentally trigger Preview editing features.

I vaguely remember some sort of issue or discussion a while back accusing IINA deleting a file. There are a number of issues related to deletion. I usually don't comment on them. Anyway…

This seems extra wrong to me in that this is not the Apple way. This should not be `Delete file…`. It should be `Move to Trash`. That would also better match up with the trashcan symbol.

The other thing that stands out is the lack of a symbol for `Show in Finder`. There is a SF Symbol for Finder. It requires macOS 26+. Maybe we can use that and create our own similar symbol for older OS?

uiryuu:
> I am going to question why we are showing an icon at all. Unlike the Finder, clicking any of these files will open it in IINA. The icon is really tiny. If we make the above change then you will be able to see the URL, so no need to have an icon to indicate it is not local.

So you want to remove the icon all together? I guess the reason why we have the icon was because at first we assume all media are local files, therefore adding their preview (icon) is natural.

> I propose changing the button to be Folder / Domain. When that button is selected streaming entries would be grouped by domain. Sorting would be by reversed domain name. So all .com entries would be separate from .net entries. Secondary sorting would be on date like is done for folders.

"Folder / Domain" is good to me. But I'd like to keep 2 levels of domain names, so that video from youtube.com and x.com can be categorized separately. Will implement in this PR.

> As seen here when testing under macOS 15.7.4 the position slider is broken:

I need to test it on macOS 11 later to fix these issues.

> Also the scroll bar is missing.

Added vertical scroll bar (with audo hide enabled).

> When you click on column titles they react and briefly highlight the column title acknowledging the click. Users might do this thinking it could change the sort order, like it does in Finder. As we don't support that the column titles should not react to clicks.

I believe there's not a standard way to do it other than subclass the NSTableHeaderView and override the mouseDown method. The behavior is the same as in the old xib implementation.

> This seems extra wrong to me in that this is not the Apple way. This should not be Delete file…. It should be Move to Trash. That would also better match up with the trashcan symbol.

Will change it to "Move to Trash…".

> The other thing that stands out is the lack of a symbol for Show in Finder. There is a SF Symbol for Finder. It requires macOS 26+. Maybe we can use that and create our own similar symbol for older OS?

I should have only add these action images for macOS 26. I'll add a helper extension to make sure the action images are only added on macOS 26+ for menu items.

uiryuu:
Update:
image


This is the current look of the playback history window. I've tried to resolve some of the issues raised in the comments. Besides, in order to disable the animation from the NSProgressIndicator, I used NSLevelIndicator here. There's no way to make it shorter other than subclassing it. So let me know how you guys think about it

Besides, a network symbol is added for online media; the online media is grouped by their domain names. A subtitle is added to show their filename. The text field on the first line is now media's title.

Also the segmented control toolbar item now uses symbols instead of text labels, which make it looks more like Finder.

low-batt:
Sorry I've been slow to respond…

This is looking good!

On the two line form, I was suggesting that should only be used when there is a title. When a title is not available, just display the pathname or URL as the primary. No need to display "No title".

Maybe help tags for the two sort buttons?

Did you decide against a button to collapse all disclosure triangles?

When I collapse them there is quite a gap between entries as seen in the screenshot below.

Alignment has improved, but the time is not right aligned.

align

When I change the sort order the first entry shows `_` as the name of the disclosure triangle. That seems a bit odd:

underscore


This is what one of those entries looks like when I enable logging of the history file:

```text
15:57:07.990 [history][v] History[230] added: 11/16/2025 14:07:29 duration: 00:00
  archive:///Users/low-batt/Documents/builds/iina-official/issue-5805/clickable-subtitles-v0.0.17.iinaplugin%7C/clickable-subtitles.iinaplugin/node_modules/iina-plugin-definition/pages/media/plugin-structure.png
  MD5: c3c07a85040d7f853678e33ad5dfaf14
```

I think the other odd entry is from testing the `--stdin` IINA CLI option? I think that is this entry:

```text
15:57:07.976 [history][v] History[97] added: 03/08/2026 21:38:53 duration: 10:35
  - -- file:///Users/low-batt/
  MD5: 118e7a1fbfc38872143fd291c23eb6ba
```

Attempting to scroll to the bottom of the window is crashing for me:

crash

I forgot that if I play something with the released version of IINA all the titles in my history file will be erased. I will have to generate some entries with titles again.

These improvements are great!

low-batt:
The main thread stack trace from a crash:

<details>
<summary>Main thread crash:</summary>

```text
CALayer position contains NaN: [nan 10]. Layer: <CALayer:0x6000004eafc0; position = CGPoint (0 0); bounds = CGRect (0 0; 0 0); backgroundColor = <CGColor 0x6000020e7300> [<CGColorSpace 0x6000020e42a0> (kCGColorSpaceICCBased; kCGColorSpaceModelRGB; sRGB IEC61966-2.1)] ( 0 0.478431 1 1 )>
(
	0   CoreFoundation                      0x000000019b2c7a30 __exceptionPreprocess + 176
	1   libobjc.A.dylib                     0x000000019ad8ab90 objc_exception_throw + 88
	2   CoreFoundation                      0x000000019b2c7920 +[NSException exceptionWithName:reason:userInfo:] + 0
	3   QuartzCore                          0x00000001a42bcca4 _ZN2CA5Layer12set_positionERKNS_4Vec2IdEEb + 172
	4   QuartzCore                          0x00000001a42bcbd0 -[CALayer setPosition:] + 52
	5   QuartzCore                          0x00000001a42c1668 -[CALayer setFrame:] + 388
	6   AppKit                              0x000000019f7aaa70 -[_NSLevelIndicatorContinuousCapacityCore layoutSublayersOfLayer:forState:] + 488
	7   AppKit                              0x000000019f7aa840 -[_NSLevelIndicatorContinuousCapacityCore updateLayer:forState:] + 928
	8   AppKit                              0x000000019f96e67c -[NSLevelIndicatorCell updateLayerWithFrame:inView:] + 288
	9   AppKit                              0x000000019f21c558 _NSViewUpdateLayer + 104
	10  AppKit                              0x000000019f1655d8 +[NSAppearance _performWithCurrentAppearance:usingBlock:] + 72
	11  AppKit                              0x000000019f945984 __29-[NSViewBackingLayer display]_block_invoke + 148
	12  AppKit                              0x000000019f19873c -[NSFocusStack performWithFocusView:inWindow:usingBlock:] + 96
	13  AppKit                              0x000000019f9453c8 -[NSViewBackingLayer display] + 280
	14  QuartzCore                          0x00000001a42d20b4 _ZN2CA5Layer17display_if_neededEPNS_11TransactionE + 748
	15  QuartzCore                          0x00000001a4466474 _ZN2CA7Context18commit_transactionEPNS_11TransactionEdPd + 492
	16  QuartzCore                          0x00000001a42b33a8 _ZN2CA11Transaction6commitEv + 644
	17  AppKit                              0x000000019f297fb4 __62+[CATransaction(NSCATransaction) NS_setFlushesWithDisplayLink]_block_invoke + 272
	18  AppKit                              0x000000019fce13fc ___NSRunLoopObserverCreateWithHandler_block_invoke + 64
	19  CoreFoundation                      0x000000019b255e28 __CFRUNLOOP_IS_CALLING_OUT_TO_AN_OBSERVER_CALLBACK_FUNCTION__ + 36
	20  CoreFoundation                      0x000000019b255d10 __CFRunLoopDoObservers + 536
	21  CoreFoundation                      0x000000019b255380 __CFRunLoopRun + 784
	22  CoreFoundation                      0x000000019b2549e8 CFRunLoopRunSpecific + 572
	23  HIToolbox                           0x00000001a6cf627c RunCurrentEventLoopInMode + 324
	24  HIToolbox                           0x00000001a6cf931c ReceiveNextEventCommon + 216
	25  HIToolbox                           0x00000001a6e84484 _BlockUntilNextEventMatchingListInModeWithFilter + 76
	26  AppKit                              0x000000019f175a34 _DPSNextEvent + 684
	27  AppKit                              0x000000019fb145cc -[NSApplication(NSEventRouting) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 688
	28  AppKit                              0x000000019faaf0c8 -[NSScroller trackKnob:] + 1016
	29  AppKit                              0x000000019faafd1c -[NSScroller mouseDown:] + 632
	30  AppKit                              0x000000019f30fbe4 -[NSWindow(NSEventRouting) _handleMouseDownEvent:isDelayedEvent:] + 3664
	31  AppKit                              0x000000019f29bc54 -[NSWindow(NSEventRouting) _reallySendEvent:isDelayedEvent:] + 492
	32  AppKit                              0x000000019f29b8a0 -[NSWindow(NSEventRouting) sendEvent:] + 288
	33  AppKit                              0x000000019fb1334c -[NSApplication(NSEventRouting) sendEvent:] + 1504
	34  AppKit                              0x000000019f71242c -[NSApplication _handleEvent:] + 60
	35  AppKit                              0x000000019f168c0c -[NSApplication run] + 520
	36  AppKit                              0x000000019f13f2dc NSApplicationMain + 880
	37  IINA.debug.dylib                    0x0000000106e0a30c __debug_main_executable_dylib_entry_point + 12
	38  dyld                                0x000000019adcab98 start + 6076
)
```
</details>

uiryuu:
> On the two line form, I was suggesting that should only be used when there is a title. When a title is not available, just display the pathname or URL as the primary. No need to display "No title".

Fixed. Now the 2 line form is used when title is available. Otherwise only a centered line with filename is displayed.

> Maybe help tags for the two sort buttons?

Added tooltips for the segments.

> Did you decide against a button to collapse all disclosure triangles?

~~I forgot this after finding a fit symbol for this for a long time, sorry. Will add a button to do this soon.~~
Done.

> When I collapse them there is quite a gap between entries as seen in the screenshot below.

Changed the style to `.plain` to remove the spacing.

> Alignment has improved, but the time is not right aligned.

Change the alignment to `.right`

> When I change the sort order the first entry shows _ as the name of the disclosure triangle. That seems a bit odd:

The "_" is only a fallback mechanism when the media is not file and `url.host` is nil. I guess `archive://` makes the system thinks this is not a file, and the `url.host` surely cannot get the host name from this. What is `archive://`?

> Attempting to scroll to the bottom of the window is crashing for me:

The problem might caused by an infinite value being pass to the `NSLevelIndicator`. Added additional checks to make sure the ratio is a finite value, otherwise hide the view.

low-batt:
Scrolling working fine now.

That the second entry in the screenshot below has a title that matches the filename is what PR #6124 fixes.

There are a couple of problems with the menu. First it doesn't work. If I click `Play` nothing happens for a URL. I'm thinking the menu should match the style of the File / Open menu. So it should switch to `Play URL` when the history entry isn't a file. The File / Open menu won't show `in New Window` when IINA doesn't have anything open, which was the case when I took this screenshot:

play-youtube

Notice some of the dates use a small font. I'm not understanding why:

small-font

This shows what I found odd about the file icon. In this screenshot the icon indicates the file normally opens in VLC:

vlc

I've not been able to get IINA to use an `archive:` URL scheme. Expanding the history window you can see that somehow I got IINA to show `.png` files within an IINA plugin file:

archive

uiryuu:
> That the second entry in the screenshot below has a title that matches the filename is what PR https://github.com/iina/iina/pull/6124 fixes.

I've merged #6124 into this branch, and it works fine.

> There are a couple of problems with the menu. First it doesn't work. If I click Play nothing happens for a URL. I'm thinking the menu should match the style of the File / Open menu. So it should switch to Play URL when the history entry isn't a file. The File / Open menu won't show in New Window when IINA doesn't have anything open, which was the case when I took this screenshot:

Fixed online media URL is not handled correctly. And the "Open new" menu item is hidden if there's no window.

> Notice some of the dates use a small font. I'm not understanding why:

Not sure why, but I've tried to set the font explicitly. Normally it should be handled by the system. Please try it and see whether the issue still persists.

> This shows what I found odd about the file icon. In this screenshot the icon indicates the file normally opens in VLC:

This icon is get directly from `NSWorkspace`, it is the icon that Finder shows. So maybe on your system the file type is opened by VLC by default, or some sort of cache problem. I don't think we can solve that kind of issue.

low-batt:
> This icon is get directly from NSWorkspace, it is the icon that Finder shows. So maybe on your system the file type is opened by VLC by default, or some sort of cache problem. I don't think we can solve that kind of issue.

Yes, that kind of file is assigned to VLC on my system. I merely showed this as an example of why I thought it was a bit odd to be showing the icon used by Finder.

The incorrect fonts problem is still there. It is somehow tied to scrolling. The fonts of a particular entry will change as I scroll around. Maybe related to issue #6126 that I just entered? This is what I'm seeing with the latest commit:

bold


Otherwise, there was an odd entry that I think came from an iina-cli test of streaming from stdin. I believe that is a backend problem. We probably shouldn't be adding a history entry for streaming from stdin.

So I think it is only the scrolling problem that is left.

uiryuu:
I make the group header rows not reusable. Please test again. I think https://github.com/iina/iina/issues/6126 is not related to this issue.

low-batt:
The scrolling problem is fixed!

Sorry I hadn't noticed this before. The filename is correctly dimmed indicating the file does not exist. But the progress slider is sill shown (likely the watch later file is still around). And the `Play` menu item is not disabled:

deleted

low-batt:
As you can see below CI `typos` check found some misspellings that need to be corrected.

I'm still bothered by the dates being misaligned. I find that harder to read and it looks messy. I searched the net for advice for handling dates in columns. All kinds of differing bad advice out there. Nothing useful.

This shows the alignment that bothers me:

dates


The code is doing this:
```swift
      return DateFormatter.localizedString(from: entry.addedDate, dateStyle: .short, timeStyle: .short)
```

Can we do something like this, or is there some time localization that would make the asumption about the size being 7 or 8 characters invalid?:
```swift
      let date = DateFormatter.localizedString(from: entry.addedDate, dateStyle: .short, timeStyle: .none)
      var time = DateFormatter.localizedString(from: entry.addedDate, dateStyle: .none, timeStyle: .short)
      // If needed, pad the time so that the end of the dates line up.
      if time.count < 8 { time = " " + time }
      return date + ", " + time
```

This is what that would look like:

aligned-date

low-batt:
In answer to my own question, the code shown does not work for RTL languages. Investigating…

low-batt:
I think these changes will properly align the dates.

Add a new property:
```swift
  /// The number of characters required to display a localized time when each component requires two digits.
  private let timeWidth: Int = {
    var components = DateComponents()
    components.hour = 12
    components.minute = 59
    components.second = 59
    let date = Calendar.current.date(from: components)!
    let time = DateFormatter.localizedString(from: date, dateStyle: .none, timeStyle: .short)
    return time.count
  }()
```

Change the alignment assignment in `makeTimeCellView` to this:
```swift
    textField.alignment = outlineView.userInterfaceLayoutDirection == .leftToRight ? .right : .left
```

And change `getTimeString` to be:
```swift
  private func getTimeString(from entry: PlaybackHistory) -> String {
    if groupBy == .lastPlayed {
      return DateFormatter.localizedString(from: entry.addedDate, dateStyle: .none, timeStyle: .short)
    } else {
      let date = DateFormatter.localizedString(from: entry.addedDate, dateStyle: .short, timeStyle: .none)
      var time = DateFormatter.localizedString(from: entry.addedDate, dateStyle: .none, timeStyle: .short)
      // If needed, pad the time so that the end of the dates line up.
      let padNeeded = timeWidth - time.count
      if padNeeded > 0 { time = String(repeating: " ", count: padNeeded) + time }
      return date + ", " + time
    }
  }
```

This is what it looks like with a RTL language:

rtl

uiryuu:
I found a easier way to setup the equally-lengthed date/time string. Take a look and see if you are ok with it.

low-batt:
Not sure why, but it didn't work with RTL. See screenshot below.

I'm not excited about the use of 24 hour times. I thought the 12 hour time with AM / PM was easier to understand.

bad

uiryuu:
Now the time respect the system settings regarding to 12/24h. The RTL should also be fixed since every time string should be the same length now.

low-batt (approved):
Tested latest commit. All looks good. Fine to merge as is.

My last suggestion to think about whether the buttons to expand and collapse the disclosure triangles should be combined into a single toggle button whose icon and help tag change based on the state.

uiryuu:
> My last suggestion to think about whether the buttons to expand and collapse the disclosure triangles should be combined into a single toggle button whose icon and help tag change based on the state.

I didn't do that was because if some groups are expanded and some are collapsed, both "expand all" and "collapse all" are valid actions, so we need 2 buttons anyways.

PR #6118, opened 2026-06-07
ci: authenticate GitHub release API requests

yorkyang2333:

---

**Description:**

CI sometimes fails while installing dependencies because the workflow downloads the latest official plugin releases through unauthenticated GitHub API requests. When GitHub rate-limits the runner IP, the dependency install step fails with HTTP 403 before the build can start.

This change passes the workflow `GITHUB_TOKEN` into the dependency install step and uses it when querying GitHub's release API. Authenticated requests have a higher rate limit, which should make plugin downloads more reliable in CI.

Testing performed:

- Ran `bash -n other/download_libs.sh`
- Ran `git diff --check`
- Verified `.github/workflows/ci.yml` can be parsed as YAML

### DIFF
--- a/other/download_libs.sh
+++ b/other/download_libs.sh
@@ -214,9 +214,14 @@ fetch_latest_plugin_asset() {
   local repo="$1"
   local response_file
   local status_code
+  local curl_args=()
+
+  if [[ -n "${GITHUB_TOKEN:-}" ]]; then
+    curl_args=(-H "Authorization: Bearer ${GITHUB_TOKEN}")
+  fi
 
   response_file=$(mktemp) || return 1
-  status_code=$(curl -s -L -o "$response_file" -w "%{http_code}" "https://api.github.com/repos/${repo}/releases/latest") || {
+  status_code=$(curl -s -L "${curl_args[@]}" -o "$response_file" -w "%{http_code}" "https://api.github.com/repos/${repo}/releases/latest") || {
     echo -e "${RED}Failed to contact GitHub for ${repo}.${NC}" >&2
     rm -f "$response_file"
     return 1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,8 @@
 name: CI
 
+permissions:
+  contents: read
+
 on:
   push:
       branches: [develop]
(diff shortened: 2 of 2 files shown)


other/download_libs.sh
```diff
@@ -214,9 +214,14 @@ fetch_latest_plugin_asset() {
   local repo="$1"
   local response_file
   local status_code
+  local curl_args=()
+
+  if [[ -n "${GITHUB_TOKEN:-}" ]]; then
```
low-batt, line 219: Excellent timing for this PR!! We were just discussing that Nightly builds were failing due to this problem.

Passing the token in a `GITHUB_TOKEN` environmental variable works, but it deviates from this script's conventions which is to control it through arguments/options.  Running the script with `--help` there is no mention of the token feature:

```text
low-batt@gag iina (ci/github-token-auth $)$ other/download_libs.sh --help

Usage:
    other/download_libs.sh [-h|--help]:           Displays this help message
    other/download_libs.sh [--arch] <ARCH>:       Architecture to download dylibs for: universal | arm64 | x86_64
    other/download_libs.sh [--yt-dlp-src] <SRC>:  Source to download youtube-dl from: github | iina
    other/download_libs.sh [--parallel] <N>:      Number of parallel downloads (default: 5)
    other/download_libs.sh [--skip-plugins]:      Skip downloading official plugins

low-batt@gag iina (ci/github-token-auth $)$ 
```

If we are going with the environmental variable I think it is desirable to add some comments explaining what this is all about. Something like this:
```text
  # GitHub imposes rate limits on their REST API. See:
  # https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api
  # Using a GITHUB_TOKEN avoids the low rate limit for unauthenticated users. See:
  # https://docs.github.com/en/actions/concepts/security/github_token
  # IINA's CI builds will usually fail attempting to down plugins unless a GITHUB_TOKEN
  # is used, so ci.yml obtains the token and sets the environment variable.
  if [[ -n "${GITHUB_TOKEN:-}" ]]; then
```

See what the other reviewers think.

uiryuu: What about not including any plugins for nightly builds?

low-batt: For the CI build triggered by pull requests we are skipping including plugins. I think that is acceptable for PR builds. For the Nightly build it is desirable to not deviate too much from what would be generated for a release. So I think it is undesirable to skip plugins for the Nightly. The changes in this PR should allow the Nightly CI build to be able to successfully obtain the plugins.

yorkyang2333:
@low-batt If there's anything needed from my side, please let me know.

low-batt (approved):
Changes look good.

Proceeding with merging.

low-batt:
@yorkyang2333 Thanks again for this PR! Sorry for the slow review. Lots going on right now.

PR #6124, opened 2026-06-11
Fix filename being stored as title in history

low-batt:
This commit will change `PlayerCore.fileLoaded` to check if the mpv property `media-title` value matches the filename of the media and if so, do not store it as the title in the history entry.

---

**Description:**


uiryuu:
Merged to `noxib-history-window`, tested, works well.

PR #6129, opened 2026-06-13
fix: use pushed_at instead of updated_at for plugin update time in the new settings page

yorkyang2333:

---

**Description:**

This PR fixes a misleading "Updated X time ago" string for Community Plugins in the Plugin Store Panel of the new settings page.

Previously, `PluginStorePanel.swift` used the `updated_at` field from the GitHub repository API to display the relative update time. However, `updated_at` reflects the repository's last activity (which can include events like receiving a star or modifying an issue label), which does not reliably correspond to the last code push. This discrepancy often caused confusion for users and plugin authors, as the time shown inside IINA did not match the "Updated on..." time typically displayed on the GitHub website (which uses `pushed_at`).

To resolve this, we now attempt to decode the `pushed_at` field from the GitHub API payload. The UI uses `pushed_at` as the primary timestamp for relative date calculation, falling back to `updated_at` if `pushed_at` is unavailable. This fallback design ensures robustness: in rare edge cases (e.g., completely empty repositories), the GitHub API might not return a `pushed_at` field. By making `pushedAt` optional and using the nil-coalescing operator (`repo.pushedAt ?? repo.updatedAt`), we guarantee that the UI accurately displays the push time under normal circumstances, while gracefully degrading to `updated_at` when necessary, preventing crashes or missing timestamps.

### Examples

GitHub Website Time (`pushed_at`):
Screenshot 2026-06-13 at 20 52 15

Comparison of how it looks in IINA:

| Before (`updated_at`) | After (`pushed_at`) |
| --- | --- |
|  |  |

### DIFF
--- a/iina/PluginStorePanel.swift
+++ b/iina/PluginStorePanel.swift
@@ -323,6 +323,7 @@ struct GitHubRepo: Codable {
   let language: String?
   let htmlUrl: URL
   let updatedAt: Date
+  let pushedAt: Date?
   let owner: Owner
 
   struct Owner: Codable {
@@ -340,6 +341,7 @@ struct GitHubRepo: Codable {
     case language
     case htmlUrl = "html_url"
     case updatedAt = "updated_at"
+    case pushedAt = "pushed_at"
     case owner
   }
 }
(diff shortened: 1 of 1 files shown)


yorkyang2333:
I've added an additional refinement to the date formatting logic: when a repository hasn't been updated for more than 30 days, the native `RelativeDateTimeFormatter` suffers from strict rounding/truncation issues. For example, it truncates 55+ days down to '1 month ago', which is misleading.

To perfectly match the official GitHub UI behavior, the time is now formatted as an absolute date (e.g., 'Updated on Apr 19') if the last push was over 30 days ago, while keeping the relative format for recent updates.

uiryuu:
Thanks!

PR #6138, opened 2026-06-20
chore: add bilingual audio to plugin list

glechic:

---

**Description:**
Add bilingual audio to plugin list
A plugin that lets two people watch the same movie together, each hearing it in their own language — one track plays through the left speaker, the other through the right.


low-batt (approved):
I verified URL and other fields are correct.

PR #6139, opened 2026-06-20
Add AirPlay video casting

dobreo:
Casts the current file (including MKV) to an AirPlay 2 receiver — tested on an LG webOS TV.

## Approach
mpv renders through a CAOpenGL layer the system can't AirPlay, so this remuxes the current file in-process to fMP4 HLS using **libavformat** (the `libav*` libraries IINA already links — **no bundled ffmpeg binary**), serves it over a small `NWListener` LAN HTTP server, and streams it to the receiver with an `AVPlayer` (`allowsExternalPlayback`) engaged via an `AVRoutePickerView` in the OSC. While casting, mpv pauses behind a "casting" overlay and AVPlayer is the transport master (OSC position, play/pause, seek, track selection).

## Features
- **Audio** — the selected audio track is honored. AAC/AC-3 are stream-copied; other codecs (E-AC3, DTS, …) are transcoded to AAC with a standard channel layout (Apple's decoder rejects PCE-based layouts), and the track's container A/V offset is preserved as leading silence so audio stays in sync.
- **Subtitles** — the selected text subtitle is transcoded to a WebVTT HLS rendition.
- **Starts from the beginning** — the remux is paced to ~real time so the receiver doesn't ride a live edge that a full-speed stream-copy would race to the end; pacing is released once playback is established to fill the buffer.
- **Seek anywhere** — an instant `AVPlayer` seek within the remuxed range; beyond it, the file is re-remuxed from the target and the cast restarts there.

## Adds
`HLSRemuxer` (libavformat remux), `HLSFileServer` (LAN server), `HLSPackager`, `AirPlayCoordinator`, an `.airplay` OSC toolbar button, and `NSAllowsLocalNetworking`.

Note: casting bypasses mpv for rendering, so mpv-side ASS styling / speed / audio-delay controls don't apply to the cast.

### DIFF
--- a/iina.xcodeproj/project.pbxproj
+++ b/iina.xcodeproj/project.pbxproj
@@ -231,6 +231,9 @@
 		845E70B2210CB59D00D06C42 /* ScrollingTextField.swift in Sources */ = {isa = PBXBuildFile; fileRef = 845E70B1210CB59D00D06C42 /* ScrollingTextField.swift */; };
 		8460FBA91D6497490081841B /* PlaylistViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8460FBA71D6497490081841B /* PlaylistViewController.swift */; };
 		846121BD1F35FCA500ABB39C /* DraggingDetect.swift in Sources */ = {isa = PBXBuildFile; fileRef = 846121BC1F35FCA500ABB39C /* DraggingDetect.swift */; };
+		FFFFFF010000000000FF0B02 /* HLSPackager.swift in Sources */ = {isa = PBXBuildFile; fileRef = FFFFFF010000000000FF0B01 /* HLSPackager.swift */; };
+		FFFFFF010000000000FF0B04 /* HLSFileServer.swift in Sources */ = {isa = PBXBuildFile; fileRef = FFFFFF010000000000FF0B03 /* HLSFileServer.swift */; };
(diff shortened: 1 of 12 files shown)


lhc70000:
Thanks for the PR. I’ve considered this approach before, and it’s impressive that you’ve developed a working implementation.

The main concern that prevented me from working on this before was that it bypasses mpv entirely and almost looks like a standalone application. Features that most users want in IINA, such as ASS subtitles rendering and delay/speed control, are not available.

I also don’t think that the sole advantage of Airplay outweighs the significant increase in app size from bundling a ffmpeg binary. IINA already contains most ffmpeg libraries, which makes it seem redundant; but fftools is not available, and it’s hard to include.

My current opinion is that it can be implemented as a plugin. IINA can provide additional API to facilitate the plugin’s integration into its interface: adding an OSC button, providing more playback control callbacks, and downloading ffmpeg.

dobreo:
Reworked to address the earlier feedback:

- **No bundled ffmpeg** anymore — the remux now runs in-process via `libavformat`/`libavcodec` (the libraries IINA already links), so there's no added binary-size cost.
- The full feature set now works end-to-end: MKV, the selected audio track (incl. E-AC3 → AAC transcode with A/V-offset correction), WebVTT subtitles, play/pause, audio/subtitle track switching, and seek to any position.

On the plugin suggestion: the AVKit/AirPlay handoff (`AVPlayer` + `allowsExternalPlayback` + `AVRoutePickerView`) can't be driven from the current JS plugin API. Would you be open to IINA exposing an `AVPlayer`/AirPlay capability that a plugin could use? I'm happy to split it that way if there's an API surface you'd prefer.

PR #6147, opened 2026-06-22
Add custom floating video window mode

NoneNano:
## Summary
- add a custom floating video window mode that moves the existing video view into an IINA-managed floating NSWindow
- expose the mode through a new `toggle-custom-floating-window` IINA command and key binding entry
- keep the existing system Picture-in-Picture path unchanged

## Notes
This is intended as a first-pass alternative to macOS system PiP for users who want a freely positionable floating video window. The first version uses a standard floating window shell; a custom borderless/hover-controls UI can be layered on in follow-up work.

## Validation
- `git diff --check`
- `plutil -lint iina.xcodeproj/project.pbxproj`

`xcodebuild` was not run locally because the current machine only has Command Line Tools selected, not a full Xcode installation (`xcodebuild requires Xcode`).

### DIFF
--- a/.github/workflows/build-iina.yml
+++ b/.github/workflows/build-iina.yml
@@ -1,0 +1,32 @@
+name: Build IINA
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          ref: codex/custom-floating-window
+
+      - name: Install dependencies
+        run: ./other/download_libs.sh --skip-plugins
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build
+        run: xcodebuild -project iina.xcodeproj ONLY_ACTIVE_ARCH=NO -scheme iina -configuration Nightly
+
+      - name: Archive
+        run: |
+          APP=$(find /Users/runner/Library/Developer/Xcode/DerivedData -name "IINA.app" -type d -maxdepth 5 | head -1)
+          echo "Found: $APP"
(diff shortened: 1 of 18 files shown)


lhc70000:
What is the benefit of this vs the "float on top" option in IINA?

NoneNano:
Thanks for asking. The main benefit I’m trying to achieve is different from the existing "float on top" option.

My original goal is to have a floating video window that can be placed anywhere above any other window, including fullscreen Spaces, without losing its topmost behavior and without being forced to snap to the screen edge.

The current pain point is that IINA’s existing Picture-in-Picture behavior uses macOS’s system PiP component. That window is controlled by macOS, so IINA cannot really customize its behavior, and the user is forced into the system edge-snapping behavior.

So this PR is an attempt to create an IINA-managed floating video window instead: freely movable, not edge-snapped, and able to stay above fullscreen windows. It is closer to the behavior of browser-style floating video windows than macOS system PiP.

To be transparent, I built this with vibe coding as an experimental prototype, and I have already run into several bugs while testing it. I would really appreciate your help either fixing the current implementation, or suggesting/writing a cleaner implementation that supports the behavior above. Thank you very much.

NoneNano:
I also want to add two concrete bugs I’m seeing in the current prototype, so the problem is clearer:

1. The new custom floating window sometimes unexpectedly loses its topmost behavior. In particular, it no longer stays above fullscreen apps/Spaces, and I’m not sure what causes this regression.

2. When IINA’s guide/open-file window is not open, the floating window cannot be moved. If that window is shown again, the floating window becomes movable again.

These may be related to how the custom `NSPanel` is configured, activated, or attached to IINA’s window/session state, but I’m not confident about the correct AppKit-side fix.

Issue #6075, opened 2026-05-16
Icon Composer App Icons


kra-mo:
Hi,

I created new icons for the app in the new .icon format. I made sure that it is aligned to the grid, and used colors from the official palette.

| Regular | Beta | Debug | Nightly |
|--------|--------|--------|--------|
| <img width="1024" height="1024" alt="Image" src="https://github.com/user-attachments/assets/e83d858d-1294-400c-b45b-4513aa79a305" /> | <img width="1024" height="1024" alt="Image" src="https://github.com/user-attachments/assets/49c699a4-6ed6-4c15-8412-00dee3e1d78a" /> | <img width="1024" height="1024" alt="Image" src="https://github.com/user-attachments/assets/8f097402-cfa2-4877-bd76-83c4c25b8cf3" /> | <img width="1024" height="1024" alt="Image" src="https://github.com/user-attachments/assets/a3679b00-4e94-4091-b778-bc05d2739c21" /> | 

Do they look to your liking? :)

Here are the files: [Icons.zip](https://github.com/user-attachments/files/27858930/Icons.zip)

To make it official, I license them under GPL-3.0 so they can be included in the app.

pparanoiidd:
I do think new branding should be seriously considered for the Liquid Glass update, if it hasn’t been already. The current icon feels a bit outdated, especially for an app that aims to blend seamlessly into Apple’s design language.

metalnakls:
<img width="1472" height="966" alt="Image" src="https://github.com/user-attachments/assets/de2729b8-cfa2-4f85-9733-591bff7f5c7e" />
<img width="1472" height="966" alt="Image" src="https://github.com/user-attachments/assets/2b88ea8c-5f75-4d64-8368-a7df13047408" />

more colors

[IINA Color + Light .zip](https://github.com/user-attachments/files/28165132/IINA.Color.%2B.Light.zip)

rslnrdnk:
Looks nice!

kra-mo:
Updated for OS 27 refractions :)

<img width="1024" height="1024" alt="Image" src="https://github.com/user-attachments/assets/e9424936-4866-4a4c-ac04-d2d01a8ec497" />

[IINA.icon.zip](https://github.com/user-attachments/files/29216911/IINA.icon.zip)

metalnakls:
<img width="1472" height="966" alt="Image" src="https://github.com/user-attachments/assets/80f55ba6-746f-4027-abf5-c15e0f8b360f" />

[...]

PR #6148, opened 2026-06-22
Update app icons using Icon Composer

kra-mo:
- [x] This implements/fixes issue #6075.

---

**Description:**

Updated the app icons to use the modern format and design.
More in the linked issue.

I found Xcode and its build system quite confusing, so please forgive me if I did it wrong and please tell me if something else also needs to be changed.

Preview:

Image


metalnakls:
a PR for a dynamic version that uses the [proper color / dark mode](https://developer.apple.com/documentation/Xcode/configuring-your-app-icon) can be found here: https://github.com/iina/iina/pull/6185

Image

Image

shoutout @kra-mo for the badges and the initial commit though.

kra-mo:
I mean I wouldn’t say “proper”, it is a stylistic preference, I followed the app icon’s current styling. I think it’s part of the app’s brand, but let’s see what the maintainers think.

Please also keep in mind that just translating the icon to a color variant is not that straightforward. The smaller triangles in the dark version are darker than the big one, making it still look like a play button from afar as the smaller ones kind of blend into the background. This is not the case with the proposed colorful version, it looks much more like a diamond shape from afar.

metalnakls:
youre right. the play button should be more apparent. I force pushed an [updated commit](https://github.com/iina/iina/commit/b6ac7487e4a41e95f523915a3c86deab8b72fc33) for that PR.

for color mode: I specifically used _proper_ as a dynamic dark theme icon is what the HIG expects from a modern app. thats why I linked the doc.
its what most of the apps in the 26 releases do — turn dark mode on when its dark.

Uranosphaerite:
Plenty of Apple apps and third-party apps have dark-colored icons for light mode: compass, magnifier, wallet, passwords, watch, and Fusion all do. 

The background does change in the icon posted originally; the light-themed icon has a more grey background. I think that it would be better to keep the dark icon even in light mode because that's just IINA's branding - even the welcome window is dark in light mode.

metalnakls:
well these always-dark icons are an exception, not a rule.
I honestly think the triangles themselves are the branding. the colors are just there to match with the rest of the system.

IINA has been sitting in my Utilities folder for damn near six years now, probably a blink of an eye for some of the og maintainers. it always worked like a solid ass program where you can throw anything at it and it does just work. its a solid tool that never prioritized . even Liquid Glass is still not out for the general public a whole year after the first 26 betas dropped. I do want a pretty app icon and I know many other people do.

ss

hope you can keep an open mind. maybe we could even experiment with a light mode icon thats more than just a gradient.

PR #6150, opened 2026-06-23
ui: optimize usage of symbols

uiryuu:

---

**Description:**
This commit:
- Removed unused icon files
- Use `repeat` and `speaker.wave` instead of old "loop" and "volume" icons
  - The "loop" icons was acutally made from sfsymbols before
- Use `plus`, `minus`, and `trash` from sfsymbols
- Change the repeat button type from toggle to momentary push in

Visual changes:
- The volume symbol is now always centered (same as Music.app)
- The `minus` is shaper on lower resolution displays
- Fix the repeat button changes state when pressed (before releasing)

See #6135 for visual changes.


lhc70000:
Merged in 8edf399.

PR #6155, opened 2026-06-26
feat: automatically show subtitles for audio files

yorkyang2333:

---

**Description:**
This is a small improvement to PR #6030. It adds `sub-visibility=yes` and `subs-with-matching-audio=yes` to the default `userOptions` in `Preference.swift`. This ensures that subtitles are automatically loaded and visible by default when playing audio-only files, so users no longer have to manually toggle them on.

### DIFF
--- a/iina/Preference.swift
+++ b/iina/Preference.swift
@@ -1182,7 +1182,7 @@ struct Preference {
     .enableLogging: false,
     .logLevel: Logger.Level.debug.rawValue,
     .displayKeyBindingRawValues: false,
-    .userOptions: [[String]](),
+    .userOptions: [["sub-visibility", "yes"], ["subs-with-matching-audio", "yes"]],
     .useUserDefinedConfDir: false,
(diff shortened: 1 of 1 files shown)


uiryuu:
In my case, when I opened an audio file with an ass file with same filename in the same folder, the sub file is already loaded and visible. I also don't think `sub-visibility` is set to false by default so there's no need to set it explicitly imo.

Besides, we are not changing `userOptions`. Those options will appear here:

image

and the user can remove them if they don't want. We ship our default mpv options during startup at `MPVController.mpvInit()`. You can see a bunch of `setUserOption()` (settings that we have a GUI to configure) and `chkErr(setOptionString())` (settings that we don't have to sync with the UI) there.

yorkyang2333:
Closing this PR as the reviewer pointed out the default behavior is already correct and modifying userOptions is not the right approach for internal defaults. Thanks for the review!

PR #6157, opened 2026-06-26
UI: Hide HDR setting when unsupported

yorkyang2333:

---

**Description:**

In unsupported HDR videos, showing a disabled (grayed-out) HDR toggle switch in the quick settings pane is confusing for users. 

This PR completely hides the HDR setting row when HDR is not available for the current video.

**Visual Changes:**

| Before | After |
| :---: | :---: |
|  |  |

**Technical Details:**
- Modified `createSwitchesView()` inside `SidebarVideoPane.swift`.
- The HDR row's visibility is dynamically toggled via `isHidden` based on `player.info.hdrAvailable`.

### DIFF
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -54,6 +54,7 @@
 "quicksetting.hwdec" = "Hardware Decoding";
 "quicksetting.deinterlace" = "Deinterlace";
 "quicksetting.hdr" = "HDR";
+"quicksetting.unsupported" = "Unsupported";
 "quicksetting.hide" = "Hide";
 "quicksetting.reset_speed" = "Reset speed to 1x";
 
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -54,7 +54,6 @@
 "quicksetting.hwdec" = "Hardware Decoding";
 "quicksetting.deinterlace" = "Deinterlace";
 "quicksetting.hdr" = "HDR";
-"quicksetting.unsupported" = "Unsupported";
 "quicksetting.hide" = "Hide";
 "quicksetting.reset_speed" = "Reset speed to 1x";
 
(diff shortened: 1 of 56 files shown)


uiryuu:
1. What about "Not supported" instead of "Unsupported"? Not a native speaker, just my 2¢
2. I think it's too much to make a dedicate class for this specifically. Would be nicer if you can replace the `NSSwitch` with the text field if HDR is not supported.
3. Please revert all the localizations and other than base and English (see https://github.com/iina/iina/blob/develop/CONTRIBUTING.md#localizations). I know it's a little bit lame especially in AI era, we might change this policy in the future (after migrating to StringCatalog after 1.5.0 release), but please follow it for now.

yorkyang2333:
Thanks for the review! I've addressed all the feedback in the latest commit.

> 1. What about "Not supported" instead of "Unsupported"? Not a native speaker, just my 2¢

Done! I have changed the English string from "Unsupported" to "Not supported".

> 2. I think it's too much to make a dedicate class for this specifically. Would be nicer if you can replace the `NSSwitch` with the text field if HDR is not supported.

I completely agree! I have removed the dedicated `HDRControlView` class. The view is now created inline inside `createSwitchesView()`, and `player.observe(.iinaHDRChanged)` dynamically toggles the visibility between the switch and the text field. This is much cleaner.

> 3. Please revert all the localizations and other than base and English (see https://github.com/iina/iina/blob/develop/CONTRIBUTING.md#localizations). I know it's a little bit lame especially in AI era, we might change this policy in the future (after migrating to StringCatalog after 1.5.0 release), but please follow it for now.

Understood. I have properly reverted the changes to all other language files and only kept the `quicksetting.unsupported` key in `en.lproj` and `Base.lproj`.

lhc70000:
Is it better to just hide the whole setting?

yorkyang2333:
Good point. I've updated the PR to completely hide the entire HDR setting row if HDR is unsupported instead of displaying the "Not supported" text. I also completely removed the `quicksetting.unsupported` localization key since it's no longer needed.

PR #6160, opened 2026-06-26
Fix plugin update checks failing

yorkyang2333:

---

**Description:**
Fixes the issue where the Plugins settings page displays "Error checking for updates" globally due to individual plugins lacking an `Info.json` file in their `main` or `master` branch.
- Modifies `JavascriptPlugin.swift` to check both `main` and `master` branches for `Info.json` to handle plugins using `main` as their default branch.
- Modifies `checkNewVersion` to safely return `nil` instead of throwing an error when `Info.json` is not found on either branch (treating a missing `Info.json` as having no updates).
- Updates `SettingsPagePlugin.swift` to catch and isolate individual plugin update check failures, preventing a single failure from aborting the entire check process for all plugins.

### DIFF
--- a/iina/JavascriptPlugin.swift
+++ b/iina/JavascriptPlugin.swift
@@ -521,7 +521,8 @@ class JavascriptPlugin: NSObject {
       if let error = lastError {
         throw error
       } else {
-        throw PluginError.cannotDownload("Info.json not found on main or master", "")
+        // Both branches returned 404, so we assume no update is available.
+        return nil
       }
     }
 
--- a/iina/JavascriptPlugin.swift
+++ b/iina/JavascriptPlugin.swift
@@ -489,25 +489,51 @@ class JavascriptPlugin: NSObject {
   }
 
   func checkNewVersion() async throws -> String? {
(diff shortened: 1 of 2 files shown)


lhc70000:
The first two changes don't fix anything; they modify the behavior:

- Developers are expected to use a `master` branch for update checks. Should we support `main` or even fetch the repo's default branch before checking updates? We will need to update the documentation if we make the move.
- If Info.json is missing, this seems to be an error to me: the developer didn't properly set up the repo, and it needs to be flagged rather than failing silently.

> Updates SettingsPagePlugin.swift to catch and isolate individual plugin update check failures, preventing a single failure from aborting the entire check process for all plugins.

This is worth fixing, but there are multiple problems in the code.

I'm closing the PR for now, and will soon push a fix for the last problem with your credit. For the other ones, please open an issue first.

yorkyang2333:
> I'm closing the PR for now, and will soon push a fix for the last problem with your credit. For the other ones, please open an issue first.

Thanks! And please refer to issue #6166.

PR #6161, opened 2026-06-27
Resizable sidebar

uiryuu:
- [?] This implements/fixes issue #. I'm not aware of any, maybe there is I couldn't find.

---

**Description:**
This PR makes all sidebars resizable (in main window). Because the quick settings were designed without considering resizable sidebar, the majority changes of this PR is to tweak the autolayout constraints in the quick settings.

Major changes:
- Put `widthPrefKey` and `widthRange` to the subclasses of SidebarViewController.
- Made the `widthRange` uniform (300...500). This is an RFC. If we agree these values (or some other uniform values), I'll change the default implementation of `widthRange`. However, keeping the widthRange uniform across all sidebars are beneficial, including
  - We can save left sidebar width, right sidebar width in the user defaults, instead of saving width for each sidebar type
  - By doing so, we can eliminate the animation when switch between different sidebar types that are on the same side
- Made the factory label in `UIHelper` has content compression resistance priority of `.defaultLow`, and truncating by tail to prevent it compress other UI elements when the width of superview is not enough
- Made most of the `NSSegmentedControl` has segments with equal widths (`segmentDistribution = .fillEqually`)
- Call `updateMask()` for the `HorizontalScrollViewWithIndicator` during `layout()` to make it reactive to sidebar width changes

It's been a breeze to tweak UI without having to messing up with xib files 😆

### DIFF
--- a/iina/PlaylistViewController.swift
+++ b/iina/PlaylistViewController.swift
@@ -23,12 +23,6 @@ class PlaylistViewController: SidebarViewController {
   override var leadingPrefKey: Preference.Key {
     .sidebarPlaylistDisplayAtLeading
   }
-  override var widthPrefKey: Preference.Key {
-    .playlistWidth
-  }
-  override var widthRange: ClosedRange<CGFloat> {
-    300...500
-  }
 
   override var allTabs: [SidebarViewController.TabType] {
     [.playlist, .chapters]
--- a/iina/PlaylistViewController.swift
+++ b/iina/PlaylistViewController.swift
@@ -23,6 +23,12 @@ class PlaylistViewController: SidebarViewController {
   override var leadingPrefKey: Preference.Key {
     .sidebarPlaylistDisplayAtLeading
   }
+  override var widthPrefKey: Preference.Key {
+    .playlistWidth
+  }
+  override var widthRange: ClosedRange<CGFloat> {
+    300...500
+  }
 
   override var allTabs: [SidebarViewController.TabType] {
     [.playlist, .chapters]
--- a/iina/SidebarSliderView.swift
+++ b/iina/SidebarSliderView.swift
@@ -74,14 +74,10 @@ class SidebarSliderView: NSView {
       tickMarkView.addSubview(label)
       label.padding(.vertical)
       if i == 0 {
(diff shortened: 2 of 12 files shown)


iina/PlaylistViewController.swift
```diff
@@ -23,6 +23,12 @@ class PlaylistViewController: SidebarViewController {
   override var leadingPrefKey: Preference.Key {
     .sidebarPlaylistDisplayAtLeading
   }
+  override var widthPrefKey: Preference.Key {
+    .playlistWidth
+  }
+  override var widthRange: ClosedRange<CGFloat> {
+    300...500
```
lhc70000: Why the width range is different from previous settings (240-800)?

iina/UIHelper.swift
```diff
@@ -55,6 +55,9 @@ class UIHelper {
     if let font {
       textField.font = font
     }
+    textField.setContentCompressionResistancePriority(.defaultLow, for: .horizontal)
+    textField.allowsDefaultTighteningForTruncation = true
+    textField.lineBreakMode = .byTruncatingTail
```
lhc70000: - Are you sure this will not affect other `ui.label`s?
- Better to extract this into a parameter like `canCompress`, so you don't need to manually set the priority to high for some labels

iina/SidebarSliderView.swift
```diff
@@ -63,21 +64,28 @@ class SidebarSliderView: NSView {
...
+      if i == 0 {
+        // pin the leading edge for the first label
+        label.leadingAnchor.constraint(equalTo: tickMarkView.leadingAnchor).isActive = true
+      } else if i == tickMarkLabels.count - 1 {
+        // pin the trailing edge for the last label
+        label.trailingAnchor.constraint(equalTo: tickMarkView.trailingAnchor).isActive = true
+      } else {
+        // center the middle labels. Only these are tracked for repositioning in layout(); the
+        // array index is offset by one from the tick index (the first label is excluded).
+        let constraint = label.centerXAnchor.constraint(equalTo: tickMarkView.leadingAnchor, constant: 0)
```
lhc70000, line 81: Unnecessary comments

PR #6162, opened 2026-06-27
Fix crashes that occur during "Reload All Plugins" and initial window load

LeeMoon-glow:
Summary

Fix four crashes that occur during plugin reload and initial window load.

Background

During development of a plugin that uses setInterval to call sidebar.postMessage every 3 seconds, I encountered consistent crashes when performing "Reload All Plugins". Investigation revealed multiple interrelated bugs in IINA's JavaScript plugin infrastructure.

Changes

1. JavascriptPolyfill.swift — Fix timer leak and stale callbacks after plugin deinit

    Problem: createTimer dispatches Timer.scheduledTimer to DispatchQueue.main.async. If deinit fires before the async block executes, the timers dictionary is empty so removeAllTimers() does nothing. The async block then creates a Timer whose target retains JavascriptPolyfill, keeping it alive and firing JS callbacks in the old (still alive) JSContext indefinitely — a memory leak that also triggers crash \# 2.

    Fix: 
    - Add an isInvalidated flag set by removeAllTimers(); the async block checks it and skips timer creation if the polyfill has been invalidated.
    - In callJSCallback, guard plugin != nil and self-invalidate the timer if the plugin instance is gone, breaking the retain cycle.

2. JavascriptAPISidebarView.swift — Guard nil pluginInstance in API methods

    Problem: pluginInstance is declared weak var pluginInstance: JavascriptPluginInstance! (implicitly unwrapped optional). After plugin reload, old timers (see \# 1) continue firing and call sidebar.postMessage. By then the old JavascriptPluginInstance is deallocated, pluginInstance is nil, and accessing pluginInstance! triggers an 
    EXC_BREAKPOINT.
    
    Fix: Guard loadFile, show, and postMessage with guard let instance = pluginInstance else { return }. Capture the local strong reference in closures instead of accessing pluginInstance directly.

3. JavascriptAPISidebarView.swift — Remove removePluginTab from cleanUp

    Problem: During deinit, cleanUp calls removePluginTab → updatePluginTabs → getAllTabs which reads PlayerCore.plugins. This happens while clearPlugins is mid-mutation of pluginMap on the same call stack, triggering a Swift exclusivity violation (SIGABRT in swift_beginAccess).
    
    Fix: Remove the removePluginTab call from cleanUp. reloadPlugin (called after clearPlugins) already calls updatePluginTabs(), which rebuilds the complete tab list from scratch, automatically replacing stale tabs.

4. MainWindowController.swift — Guard nil oscFloatingView in windowDidResize

    Problem: When opening a video file, PlayerCore.openMainWindow triggers NIB loading which may fire NSWindowDidResizeNotification before windowDidLoad line 490 assigns oscFloatingView. Since oscFloatingView is OSCFloatingView! (IUO), accessing it in windowDidResize causes a nil force-unwrap crash.
    
    Fix: Use optional binding: if oscPosition == .floating, let view = oscFloatingView.

Testing

  Setup: Build IINA with these changes, then install the [plugin-mumm-playlist-tree](https://github.com/LeeMoon-glow/plugin-mumm-playlist-tree), which uses setInterval + sidebar.postMessage.
  
  Steps:
  1. Launch IINA and open a video file → verify no crash during window load.
  2. Navigate to Plugins > Reload All Plugins → verify no crash.
  3. Repeat step 2 multiple times → verify no crash.
  4. Verify plugin sidebar tab still works correctly after reload.
  
  Results: All four crashes are resolved. The plugin survives repeated reloads and continues to function normally.

### DIFF
--- a/iina/JavascriptAPISidebarView.swift
+++ b/iina/JavascriptAPISidebarView.swift
@@ -23,21 +23,18 @@ class JavascriptAPISidebarView: JavascriptAPI, JavascriptAPISidebarViewExportabl
   private lazy var messageHub = JavascriptMessageHub(reference: self)
 
   override func cleanUp(_ instance: JavascriptPluginInstance) {
-    // removePluginTab is not called here to avoid exclusivity conflicts during deinit.
-    // reloadPlugin (called after clearPlugins) will trigger updatePluginTabs to refresh
-    // the tab list, automatically removing stale tabs.
+    player!.mainWindow.sidebars.pluginView.removePluginTab(withIdentifier: instance.plugin.identifier)
   }
 
   func loadFile(_ path: String) {
-    guard let instance = pluginInstance else { return }
     guard player!.mainWindow.loaded else {
       throwError(withMessage: "sidebar.loadFile called when window is not available. Please call it after receiving the \"iina.window-loaded\" event.")
       return
     }
-    let rootURL = instance.plugin.root
+    let rootURL = pluginInstance.plugin.root
     let url = rootURL.appendingPathComponent(path)
     Utility.executeOnMainThread {
-      let nav = instance.sidebarTabView.load(URLRequest(url: url))
+      let nav = pluginInstance.sidebarTabView.load(URLRequest(url: url))
       if nav == nil {
         throwError(withMessage: "Failed to load ")
       }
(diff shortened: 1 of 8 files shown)


lhc70000:
Nice catch. However, only problem 3 can be reproduced with your plugin on my machine, and the fix is incorrect.

- Problem 1: I cannot reproduce. Your AI said
  > If deinit fires before the async block executes, the timers dictionary is empty so removeAllTimers() does nothing

  This is possible, but seems highly unlikely and requires millisecond-level timing (plugin adds a timer immediately after user clicking reload plugin in the UI). I doubt it's constantly reproducible (let me know if you have steps to reproduce). The fix is fine, but I wonder whether the `plugin != nil` guard is needed.

- Problem 2: should not crash if problem 1 is fixed. But again, problem 1 seems unlikely to me.

- Problem 3: need to be fixed, but removing the `removePluginTab()` line is incorrect. It's required when you disable a plugin. The new sidebar is still under development, so I prefer to fix it later by us.

- Problem 4: out of scope. (This is a potential crash, and exists not only in windowDidResize. Essentially we need to `guard loaded` in addition to `guard let window` everywhere.)

In short, we can merge this PR if only the problem 1 fix is included, and you should either justify the `plugin != nil` guard or remove it. I will fix problem 3 and 4 later.

lhc70000:
Other fixes pushed to develop (4a2ba4d, b2f88fa).

Issue #718, opened 2017-06-07
Shuffle playlist button


anonmonogatari:
**System and IINA version:**

- macOS 10.12.5
- IINA 0.0.11

**Expected behavior:**
Shuffle button should be toggleable. It should shuffle playlist when toggled on and revert changes when toggled off.

**Actual behavior:**
Shuffle button shuffles playlist when pressed. Can't revert changes once you pressed it.

**Steps to reproduce:**
Press shuffle button.

danleyzberg:
Following this feature request. Love the app!

devros42:
This would be great.  Similar to how VLC handles it where it doesn't actually shuffle around the list in the playlist, but jumps around randomly and then when shuffle is pressed again it simply stops shuffling and would then just play the next track from the original playlist order

devros42:
mentioned here:

https://github.com/mpv-player/mpv/issues/6250

qianbinbin:
I think it's a bug rather than enhancement, the list would not recover once the shuffle button clicked, which is apparently not the user want.
Apple Music supports Playlist and Playing Next: https://support.apple.com/guide/music/use-the-playing-next-queue-musb1e6d1c76/mac

sajjadalis:
> This would be great. Similar to how VLC handles it where it doesn't actually shuffle around the list in the playlist, but jumps around randomly and then when shuffle is pressed again it simply stops shuffling and would then just play the next track from the original playlist order

I second this. The shuffle button should play a random file similar to VLC. Shuffling the playlist items makes problems if you want to listen to them in order.

GsHeri:
8 years later, shuffle button still destroys the playlist

uiryuu:
Fixed in #6172. Will be included in the upcoming 1.5.0 release.

PR #6171, opened 2026-07-02
fix: make playlist shuffle reversible (#718)

tathagat22:
- [x] This implements/fixes issue #718.

---

**Description:**

Right now the shuffle button in the playlist just calls `playlist-shuffle`
every time it's pressed. Once you click it there's no way back to the
original order short of reloading everything, which is what #718 has been
asking about for a while.

This makes the button a toggle:

- First press shuffles the playlist (`playlist-shuffle`).
- Second press restores the previous order (`playlist-unshuffle`).

mpv already gives us `playlist-unshuffle`, which puts the list back exactly
the way it was before the last shuffle, so I lean on that instead of trying
to snapshot and rebuild the order by hand. To make the state visible, the
button now tints while a shuffle is active, the same way the loop button
next to it reflects its state.

A few details:

- The shuffled state is a single `isShuffled` flag on `PlaybackInfo`.
- It's reset to `false` whenever the playlist changes for another reason
  (new media opened, items added / removed / moved), since an unshuffle
  wouldn't match the list after that. The next press then just does a fresh
  shuffle.
- Advancing to the next track during normal playback doesn't touch the flag,
  so a shuffled playlist stays shuffled (and the button stays lit) as it
  plays through.
- The `--shuffle` command-line path also sets the flag, so a playlist that
  was shuffled at launch shows the button as active.

Tested by opening a folder and toggling shuffle on/off (order restores
correctly), and by editing the playlist after shuffling to confirm the
button clears and the next press re-shuffles rather than doing a stale
unshuffle.

### DIFF
--- a/iina/PlaybackInfo.swift
+++ b/iina/PlaybackInfo.swift
@@ -235,6 +235,14 @@ class PlaybackInfo {
   ///     To avoid the need to lock multiple locks the cache properties are always accessed while holding the playlist lock. The cache
   ///     properties are private to force all access to be through class methods that properly coordinate thread access.
   @Atomic var playlist: [MPVPlaylistItem] = []
+
+  /// Whether the playlist is currently shuffled via the shuffle button.
+  ///
+  /// Tracks whether a `playlist-shuffle` is in effect so it can be reverted with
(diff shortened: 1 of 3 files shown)


uiryuu:
Thanks for your contribution! I've edited your changes in #6172, which will be reviewed soon.

PR #6178, opened 2026-07-05
Fix Live Text UI issues: OSC autohide glitch on shortcuts & uncoordinated menu icon

yorkyang2333:
- [x] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)
- [ ] Related issue: #
- [ ] Related Design Proposal: Not applicable
---
- [x] AI was used to write the code in this pull request.
  - [ ] The code is fully written by AI / I am an AI agent.
---

**Description:**

This PR resolves two UI/UX issues regarding On-Screen Controller (OSC) auto-hiding behavior and menu bar icon styling consistency:

### 1. Fix OSC staying stuck when pausing or seeking via shortcuts
- **Problem**: When triggering playback actions such as play/pause or seeking (fast-forward/rewind) via keyboard shortcuts, the bottom On-Screen Controller (OSC) would automatically show up and stay visible indefinitely without auto-hiding. Furthermore, simple shortcut presses should not needlessly evoke the bottom OSC when OSD feedback is already cleanly displayed at the top-left.
- **Fix**: 
  - Added Live Text analysis state checks (`!liveTextSupport.isAnalyzing`) before triggering OSC auto-show in `updateTimeLabel()`, preventing unwanted OSC wake-ups during playback shortcut operations.
  - Ensured `inWindow.updateTimer()` is explicitly called when video playback is paused/resumed, properly initiating the auto-hide timer.
  - Reset all interaction state flags (`isSelected`, `isMenuOpen`, `isHighlighted`) during Live Text analysis clearing to ensure clean state transitions.

### 2. Remove uncoordinated icon from "Live Text" menu item
- **Problem**: The "Live Text" item in the main menu bar and context menu had a `secondaryImage="text.viewfinder"` SF Symbol attribute set in `MainMenu.xib`. It was the only menu item with an icon among standard text-only menu items (such as `Lock Window Aspect Ratio`), creating an inconsistent and uncoordinated visual appearance.
- **Fix**: Removed the `secondaryImage="text.viewfinder"` attribute from `MainMenu.xib`, bringing it into full visual consistency with native macOS menu styling.

---

### Comparison

| Feature / Group | Before | After |
| :---: | :---: | :---: |
| **1. OSC Auto-hide on Shortcuts** |  |  |
| **2. Menu Bar Icon Consistency** |  |  |

PR #6179, opened 2026-07-05
refactor: update timer in `showUI()` by default

uiryuu:
- [ ] Related issue: #
---
---

**Description:**
Currently, most showUI() calls are followed by updateTimer(), with some exceptions. This commit calls updateTimer() inside showUI(), and use a parameter to not call updateTimer(). This elimiates the need of exposing updateTimer() to public (introduced in #6178).

Also fixes updateTimer() not being called in
`windowDidFailToExitFullScreen` and `windowDidFailToEnterFullScreen`.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -934,7 +934,7 @@ class MainWindowController: PlayerWindowController {
     }
 
     fadeableViews.update()
-    showUI()
+    showUI(startTimer: false)
 
     if isFloating {
       fragControlViewMiddleButtons1Constraint.constant = 24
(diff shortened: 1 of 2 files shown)


iina/MainWindowController.swift
```diff
@@ -1991,6 +1989,10 @@ class MainWindowController: PlayerWindowController {
     // The OSC may not have been updated while it was hidden to avoid wasting energy. Make sure it
     // is up to date.
     player.refreshSyncUITimer()
+    if startTimer {
+      self.updateTimer()
```
low-batt, line 1993: The timer is not just for hiding UI, it also hides the cursor. By adding the call to `updateTimer` in `showUI` at this point means it will not be called if the checks at the start of `showUI` cause a premature exit. Won't that cause the cursor to not be hidden in some cases where it used to be?

uiryuu: You are very correct. I'm closing this for now as its not very necessary either

PR #6189, opened 2026-07-08
Fix misleading Open URL load failure message

### LINKED ISSUE #6134: Open URL window claims the URL is "not valid" just because a video fails to open
### IINA Version

1.4.3

### Release Channel

Release

### macOS Version

15.7.1

### Steps to Reproduce

- Open a URL via e.g. the Open in IINA Safari plugin
- The URL fails to load after an apparent attempt to open it, i.e. the URL is _not invalid or the Open button wouldn't even have been enabled_
- The window blames the actual URL with red text, even though the problem is not with the URL

### Expected Behavior

A non-misleading error message

### Actual Behavior

Blames a completely normal and valid youtube URL (`https://www.youtube.com/watch?v=_<alphanumeric string>`) for being invalid. Well, either that or a Swedish translation error.

_No response_

### Logs

_No response_

### Sample Files

_No response_

Every time

### Required Actions

### Optional Actions

jjoanna2-debug:
- [x] Related issue: #6134
---
---

**Description:**

Fixes #6134.

## Summary
- Keep the existing invalid-URL message for malformed input.
- Show the existing open-failure message when a validated URL fails to load.
- Reset the URL field color when the sheet state is cleared.

## Verification
- `git diff --check`
- `xcodebuild -list -project iina.xcodeproj`
- `./other/download_libs.sh`
- `xcodebuild -project iina.xcodeproj -scheme iina -configuration Debug -destination 'platform=macOS,arch=arm64' CODE_SIGNING_ALLOWED=NO build`

The first build attempt reached Swift compilation but failed at link/copy because `deps/lib` was missing. After running the documented dependency download, the same Debug build succeeded.

## Behavior Notes
This does not change URL parsing, scheme handling, credential storage, or the string passed to mpv. The change is limited to UI feedback after a previously accepted URL fails to load.

## AI Usage
I used AI assistance to inspect the issue path, draft the patch, and run validation. I reviewed the diff and verification output before opening this PR.


low-batt:
Hello from a _junior_ IINA developer. We will have to see what the other developers think. But I will share some thoughts…

When you have to fix something it is always good to pause, step back and take a real good look at the problem and question everything, rather than going for a quick fix.

However… There is a lot going on right now. Wait and see what they other developers think. They may decide a quick partial fix is fine for now. 

The problem at hand is that the error reporting is not properly communicating the true nature of the failure. It still isn't. This works fine:
```text
/Applications/IINA.app/Contents/MacOS/iina-cli file:///Users/low-batt/Movies/big_buck_bunny.mp4
```

When running under this PR and using `Open URL…`:

invalid

That URL is not invalid from a user's point of view. Users expect such an error message to mean that the supplied string cannot be parsed as a valid URL. But that is a valid URL and accepted by the IINA CLI. That is not why IINA is rejecting this URL. IINA is demanding the URL passed contain a host.

If instead I enter this in `Open URL…`:
```text
file://example.com/Users/low-batt/Movies/big_buck_bunny.mp4
```

IINA ignores the host and successfully opens the file. So we have an inconsistency with the IINA CLI.

The error message now shown for a failure while _loading_ is:
```text
Cannot open file or stream!
```

That is both a good and bad choice. It is good in that this is the same message IINA uses in other cases. But it is bad in that it is inconsistent with what IINA told the user it was doing which was:
```text
Loading Media…
```

To be consistent you would expect:
```text
Cannot load file or stream!
```

Since IINA said it was "loading". But rather than change the error message, I'm thinking "Loading Media…" should be changed to "Opening Media…" to be consistent with the term "Open" being used in most other contexts.

Again, before making any changes, see what the other developers say.

jjoanna2-debug:
Thanks for testing this. I agree the `file://` case exposes a real existing inconsistency, but it is a different path from #6134: it is rejected before opening by the existing `url.host` validation in `controlTextDidChange`. This PR does not change URL acceptance rules.

#6134 concerns a network URL that passes validation, reaches `PlayerCore.openURL`, and then fails during playback loading. In that path, displaying `alert.invalid_url` is specifically wrong; this PR changes only that post-validation failure message to the existing localized `alert.error_open` message.

I would keep the file-URL acceptance inconsistency and the "Loading"/"Opening" terminology out of this patch so the bug fix remains scoped. They can be addressed separately if maintainers want those behavior or wording changes. I will wait for maintainer direction before expanding the PR.

low-batt:
Agree. Wait for the other IINA developers to chime in. My guess is that it is likely they will decide to deal with the problem I noted as a separate issue.

I always end up diving right into the work and forgetting to thank people for their submissions…

Thank you very much for investigating the problem and proposing a fix!

jjoanna2-debug:
Following up for maintainer direction on the narrow scope here. The PR remains a one-file fix for #6134, is mergeable, and has green macOS 26.5 build and spelling checks. It changes only the post-validation failure message; the separate `file://` acceptance inconsistency discussed above remains intentionally out of scope.

@uiryuu, when you have a moment, could you confirm whether this scope is acceptable or identify any change needed before review?

low-batt:
Last I heard from the senior developers they were very focused on some major changes. I expect once they have finished that work there will be a flurry of reviews and merging.

uiryuu:
I agree this is a partial fix for https://github.com/iina/iina/issues/6134, but is this good enough to close that issue?

jjoanna2-debug:
@uiryuu Yes—the merged change is sufficient to close #6134 as reported. That issue covers a valid network URL that passes field validation, reaches loading, and is then misleadingly labeled invalid when opening fails. The merged code now distinguishes that post-validation failure and shows the existing `alert.error_open` message.

The `file://` host requirement discussed above happens earlier, during field validation. It is a separate URL-acceptance inconsistency and should not keep #6134 open.

PR #6190, opened 2026-07-08
Save playlist entries relative to playlist file

### LINKED ISSUE #6182: Saving a playlist to the same folder as the files doesn't use relative paths to the files
### IINA Version

1.4.4

### Release Channel

Release

### macOS Version

15.7

### Steps to Reproduce

- Save the current queue of files that are all residing in the same folder as a playlist to that same folder
- The playlist references the files with absolute paths that will break as soon as the folder is moved

### Expected Behavior

A relative path to the files should be used when possible. E.g. the greatest common denominator file path prefix (including the future path of the playlist) can be found in order to generate relative file paths, which covers the common case where all files are in the same folder.

### Actual Behavior

Absolute paths are used in the generated playlist file

_No response_

### Logs

_No response_

### Sample Files

_No response_

always

### Required Actions

### Optional Actions

jjoanna2-debug:
- [x] Related issue: #6182
---

**Description:**

Fixes #6182.

## Summary
- Write local file entries relative to the saved playlist file.
- Preserve network URLs and non-file playlist entries unchanged.
- Keep the existing save panel flow and `.m3u8` output format.

## Verification
- `git diff --check`
- `swift - <<'EOF' ... EOF` behavior probe covering same-folder, child-folder, sibling-folder, network URL, and pre-relative playlist entries
- `xcodebuild -project iina.xcodeproj -scheme iina -configuration Debug -destination 'platform=macOS,arch=arm64' CODE_SIGNING_ALLOWED=NO build`

The [CI](https://github.com/iina/iina/actions/runs/32006506679) and [Spelling](https://github.com/iina/iina/actions/runs/32006506703) workflows for this revision currently report `action_required`; repository approval is required before their jobs can run.

## Behavior Notes
This does not read, probe, or modify playlist item files. It only rewrites local file path strings before writing the user-selected playlist file.


jjoanna2-debug:
Following up for review on this focused fix for #6182. The PR remains mergeable and changes only `MainMenuActions.swift`: local paths are written relative to the saved playlist, while network URLs and existing non-file entries remain unchanged. The behavior probe covers same-folder, child, sibling, network, and pre-relative entries, and the macOS Debug build passed.

@uiryuu, could you confirm whether this behavior is acceptable or identify any required changes?

uiryuu:
I think we should provide an option to save the playlist with relative path or full path. Otherwise I'm afraid that we will break some users' normal workflow. I want to know your opinions @anohren @low-batt

anohren:
> Otherwise I'm afraid that we will break some users' normal workflow.

Sounds like a reasonable possibility, but personally I know nothing about this. 

Spontaneously I'd be tempted to at least suggest that if the playlist file is saved into the same folder that also contains every single referenced file, or is saved into the parent folder of that one single folder, that the playlist file and its referenced files are meant to be moved as a single unit and that relative paths is a sensible automatic decision.

A checkbox might still be the least surprising behavior compared to switching the behavior silently.

jjoanna2-debug:
@uiryuu I do not think we need to add a choice to the save panel for this fix. The current patch is broader than #6182 because it converts every local entry to a relative path, including paths that may climb through several parent directories.

I would narrow the rule instead: write a relative entry only when the media file is in the playlist directory or one of its descendants, and preserve the absolute path for local files outside that tree. Network URLs and existing non-file entries would remain unchanged. That directly fixes the portable same-folder case without changing workflows that deliberately reference media elsewhere.

This is close to @anohren’s directory-based suggestion, but applied deterministically per entry. It avoids adding a storage-format decision to every save operation and avoids choosing a new global default. If users later need to override that bounded behavior, an explicit option can be considered separately.

Issue #6192, opened 2026-07-12
RTL language window controls overlap sidebar


low-batt:
### IINA Version

develop @ commit c5dccbc

### Release Channel

Other

### macOS Version

macOS 26.5.1

### Steps to Reproduce

- Click on `System Settings…` under the  menu
- Navigate to `Languages & Region`
- Set  `Preferred Languages` to a [left-to-right](https://en.wikipedia.org/wiki/Left-to-right_language#Directionality_and_orientation) language
- In the `Applications` section add IINA and set it to a right-to-left language
- Open IINA
- Start playing a video
- Click on `Show Video Panel` under the `Video` menu
- The side panel appears
- The window controls overlap the top of the sidebar

<img width="1445" height="1418" alt="Image" src="https://github.com/user-attachments/assets/281273b5-5557-4a89-8b43-f38c0c78fa95" />

The windows controls on top of the sidebar:
<img width="3840" height="489" alt="Image" src="https://github.com/user-attachments/assets/34878eaf-d728-49b8-9065-d64526b9da32" />

### Expected Behavior

For a right-to-left language the window controls are on the right side of the title bar and do not overlap the sidebar.

### Actual Behavior

When a right-to-left language is configured at the system level the  window controls are on the right side of the title bar and do not overlap the sidebar. However when a left-to-right language is configured at the system level and IINA is configured with a right-to-left language the window controls appear on the left side of the title bar and interfere with the sidebar. When the sidebar is not shown, the window controls will interfere with the float on top status indicator (if enabled).

### How often can you reproduce the issue

Every time.

### Required Actions

- [x] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [x] I tested with the Nightly build and confirmed the issue is not fixed.
- [ ] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

low-batt:
## Analysis

From the Apple documentation for the [NSWindow.windowTitlebarLayoutDirection](https://developer.apple.com/documentation/appkit/nswindow/windowtitlebarlayoutdirection) property:
> ## windowTitlebarLayoutDirection
> **The direction the window’s title bar lays text out, either left to right or right to left.**
> 
> macOS 10.12+
> ```text
> var windowTitlebarLayoutDirection: > NSUserInterfaceLayoutDirection { get }
> ```
> 
> ## Discussion
>
[...]

PR #6193, opened 2026-07-12
Fix RTL language window controls overlap sidebar, #6192

low-batt:
This commit will change `MainWindow` to override the `windowTitlebarLayoutDirection` property with an implementation that returns the layout direction of the view containing the window close button, instead of the implementation provided by `NSWindow` that uses the primary system language.

- [x] Related issue: #6192
---
---

**Description:**

### DIFF
--- a/iina/MainWindow.swift
+++ b/iina/MainWindow.swift
@@ -11,6 +11,20 @@ import Cocoa
 class MainWindow: NSWindow {
   var forceKeyAndMain = false
 
+  /// The direction the window’s title bar lays text out, either left to right or right to left.
+  ///
+  /// As discussed in the [windowTitlebarLayoutDirection](https://developer.apple.com/documentation/appkit/nswindow/windowtitlebarlayoutdirection)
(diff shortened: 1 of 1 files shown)


low-batt:
See issue #6192 for a full discussion of the root cause of the problem.

Currently AppKits behavior means there are 3 directionalities:
- left-to-right
- right-to-left
- hybrid, with the directionality of the window controls not matching the rest of the app

This PR only fixes the main window problem shown in #6192. That means the window control buttons will still be on the left side of other IINA windows. So one question for the reviewers is whether all windows should be fixed. Maybe putting the fix in a new class, something like `WindowRTLWorkaround` (not sure what to call it) that `MainWindow` would extend and would be used directly by other IINA windows so that all windows appear as they would when the system language is a RTL language?

low-batt:
I've converted this PR to a draft as PR #6233 provides a comprehensive fix. I have kept this PR open just in case reviewers prefer the limited fix provided by this PR.

low-batt:
Closing this PR as the comprehensive fix in PR #6233 has been merged.

PR #6195, opened 2026-07-12
Show OSD even when applied volume is constrained

mooons:
- [ ] Related issue: n/a
---
---

**Description:**

When the volume is already at 0 or 100, further volume adjustments are clamped to the same value. mpv does not emit a property-change event when the volume isn't changed, so IINA does not display the volume OSD in such cases.

Proposed change: sendOSD directly when the requested volume is constrained and the applied volume is already current. This matches the behavior of other media players like Movist or PotPlayer.

Note that normal volume changes continue to use mpv's property-change event. This won't duplicate OSD notifications.

Testing done:

- Built the Release configuration for arm64 and x86_64.
- At volume 100, pressing volume-up again would now display the volume OSD at 100.
- At volume 0, pressing volume-down again would now display the volume OSD at 0.

(I made the initial change locally back in May 2022. I found this behavior is still unchanged as of today, so I rebased the change and figured I should contribute this back.)

Pls kindly review :)

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -1276,9 +1276,15 @@ class PlayerCore: NSObject {
     let maxVolume = Preference.integer(for: .maxVolume)
     let constrainedVolume = volume.clamped(to: 0...Double(maxVolume))
     let appliedVolume = constrain ? constrainedVolume : volume
+    let shouldSendConstrainedVolumeOSD =
+      constrain && volume != constrainedVolume && info.volume == appliedVolume
     info.volume = appliedVolume
     mpv.setDouble(MPVOption.Audio.volume, appliedVolume, level: .verbose)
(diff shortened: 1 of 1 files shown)


low-batt:

Thank you for submitting this PR!

Currently if a user presses a key to raise the volume there is no feedback. The user might wonder if they are pressing the wrong key. With this change the user can tell that IINA recognized the request to increase the volume. From the slider in the OSD they can tell the volume did not change because it is at the max.

For UI changes I defer to the senior developers, but this change certainly makes sense to me.

low-batt (approved):
Pulled PR, built IINA and tested the enhancement.

Changes look good to me.

uiryuu:
Thanks!

mooons:
Thanks @low-batt and @uiryuu for reviewing this!

PR #6197, opened 2026-07-13
Remove deprecated OpenGL surface properties to suppress Xcode compiler warnings

yorkyang2333:
- [ ] Related issue: #
---
---

**Description:**

Fixes compiler deprecation warnings in `VideoView.swift` as `VideoView` hosts a `CAOpenGLLayer` (`ViewLayer`), and Retina high-resolution scaling and HDR / EDR content dynamic range are managed directly via `CAOpenGLLayer` properties (`contentsScale` and `wantsExtendedDynamicRangeContent`):
- `'wantsBestResolutionOpenGLSurface' was deprecated in macOS 10.14: Use NSOpenGLView instead.`
- `'wantsExtendedDynamicRangeOpenGLSurface' was deprecated in macOS 10.14: Use NSOpenGLView instead.`

### DIFF
--- a/iina/VideoView.swift
+++ b/iina/VideoView.swift
@@ -70,8 +70,6 @@ class VideoView: NSView {
 
     // other settings
     autoresizingMask = [.width, .height]
-    wantsBestResolutionOpenGLSurface = true
-    wantsExtendedDynamicRangeOpenGLSurface = true
 
     // dragging init
     registerForDraggedTypes([.nsFilenames, .nsURL, .string])
--- a/iina/VideoView.swift
+++ b/iina/VideoView.swift
@@ -70,6 +70,11 @@ class VideoView: NSView {
 
     // other settings
     autoresizingMask = [.width, .height]
+    // Unlike upstream mpv (video/out/mac/view.swift), IINA intentionally omits setting
+    // `wantsBestResolutionOpenGLSurface` and `wantsExtendedDynamicRangeOpenGLSurface` on NSView.
+    // They were deprecated in macOS 10.14 and are redundant here because VideoView hosts a
+    // custom CAOpenGLLayer (`ViewLayer`), where high-resolution Retina scaling is managed via
+    // `contentsScale` and EDR mode is configured via `wantsExtendedDynamicRangeContent`.
 
     // dragging init
     registerForDraggedTypes([.nsFilenames, .nsURL, .string])


low-batt:
Hello from a _junior_ IINA developer. I'm always happy when I see other developers care about reducing compiler warnings.

I could be remembering wrong, but I have a vague memory of @uiryuu investigating removing this code and finding that caused some kind of problem.

@uiryuu Please have a look at this change.

Some background…

It is critical that this particular section of IINA uses `libmpv` in the way that the [mpv project](https://github.com/mpv-player/mpv) intends for it to be used. For that reason it is desirable that as much as possible IINA code be aligned with the similar code in the mpv player that uses `libmpv`.

Over time the equivalent mpv code has changed. IINA code has not kept up with those changes, some of which correct problems. For IINA 1.4.0 an effort was made to bring [ViewLayer](https://github.com/iina/iina/blob/develop/iina/ViewLayer.swift) closer in alignment with the equivalent mpv player code. This effort was limited by time constraints so there is still work to be done to incorporate all of the changes mpv has made.

Recently the mpv player has diverged from IINA with a move from [OpenGL](https://www.opengl.org/) to [Vulkan](https://www.vulkan.org/) / [MoltenVK](https://moltengl.com/moltenvk/) / [Metal](https://developer.apple.com/metal/) by default. Better performance allowed mpv to switch from using a background thread for rendering to always using the main thread for rendering. The mpv project decided the performance problems that change introduced when using [OpenGL](https://www.opengl.org/) were acceptable. 

It is very desirable to switch to only using the main thread for rendering. That would correct some pauses that occur during animations and other UI glitches. However this was tested and the UI performance of IINA unacceptable unless the background thread is used. This will have to be reevaluated once IINA has moved away from [OpenGL](https://www.opengl.org/). This is documented at the start of the `ViewLayer` class:
https://github.com/iina/iina/blob/c5dccbc8554aafa01efda91cd3f28ae2c3c51a98/iina/ViewLayer.swift#L61-L67

What all this means is that one thing we want to know is whether the equivalent mpv code is still setting these properties. I checked and the answer is yes it does in [view.swift](https://github.com/mpv-player/mpv/blob/master/video/out/mac/view.swift).

Now this does not mean IINA has to set these. But I would like to see some sort of comment that IINA is intentionally deviating from mpv so that the next effort to align IINA code with mpv can easily tell these setting have been intentionally removed.

yorkyang2333:
@low-batt Thanks for the detailed background and review!

> But I would like to see some sort of comment that IINA is intentionally deviating from mpv so that the next effort to align IINA code with mpv can easily tell these setting have been intentionally removed.

I have added a comment in `VideoView.swift` documenting this intentional deviation from upstream `view.swift` so future alignment efforts clearly see why these properties are omitted here.

low-batt:
@yorkyang2333 Thanks for that update!

We are now waiting on @uiryuu to review. The senior developers are very busy right now, so it could be a while before they get to this. If I get their attention I will mention this PR.

uiryuu:
Thanks for the PR, but I think at this stage we are not touching legacy GL-related code unless there's bug or other justifications. I think our code using OpenGL contains a lot of in imperfections and suboptimal decisions, but since we are moving away from GL, I don't want to make modifications without good reasons.

low-batt:
@uiryuu On this:
> I think at this stage we are not touching legacy GL-related code unless there's bug or other justifications

I've been too busy to get to it, but I was thinking of making a tweak that might mitigate issues #6213 and  #6212, which I recently encountered. I do not remember any reports of crashes like this before. Maybe a macOS regression? It appears to be a race condition. It is very hard to reproduce so it is very hard to diagnose.  

IINA fell way behind mpv and was missing equivalent fixes in this area that had been made upstream. For 1.4.0 I added a lot of them, but ran out of time before completing that work. I had plans to revisit that once `develop` was upgraded to mpv 0.41.0.

At one point I encountered a crash that I traced to this statement in the `ViewLayer` shadow copy initializer:
https://github.com/iina/iina/blob/aab1b341b41d72b2357c658f4ade37effc8c2b92/iina/ViewLayer.swift#L151

An open PR from me removes that statement, but that PR was not merged due to other issues with it. I have been working on updating that PR, but keep getting interrupted.

On this PR, I vaguely remember something about you trying to make this same change to eliminate the compiler warnings and encountering a problem? Or is my memory faulty? I tested this PR and didn't encounter a problem. 

Upstream, mpv 0.41.0 no longer uses  OpenGL by default, but it still can be used. I just searched the mpv source code for `wantsBestResolutionOpenGLSurface` and `wantsExtendedDynamicRangeOpenGLSurface` and mpv is still setting them in [view.swift](https://github.com/mpv-player/mpv/blob/master/video/out/mac/view.swift). However from looking through the output when doing a build, eliminating compiler warnings is not a priority for the mpv project. Retaining these statements does keep IINA aligned with the mpv code.

Should I hold off making any changes in this area?

uiryuu:
I too remembered something bad happened when removing these 2 lines, but I cannot recall what exactly was that.

I think we only fix bugs (like crashes) in this area. Otherwise we don't put too much time and effort on this.

PR #6200, opened 2026-07-15
build: support macOS 27 in Availability.xcconfig

yorkyang2333:
- [ ] Related issue: #
---
---

**Description:**

This pull request adds support for building IINA using Xcode 27 by defining `AVAILABLE_270000` in `Configs/Availability.xcconfig` in order to resolve the following compiler error:
> `Conditional compilation flags must be valid Swift identifiers (rather than '$AVAILABLE_270000')`

---

**Question regarding `MACOSX_DEPLOYMENT_TARGET`**

I also noticed that Xcode 27 dropped support for `MACOSX_DEPLOYMENT_TARGET = 11`, throwing the following error while building:
> `The macOS deployment target 'MACOSX_DEPLOYMENT_TARGET' is set to 11/11.0, but the range of supported deployment target versions is 12.0 to 27.0.x.`

Since IINA still has support for `MACOSX_DEPLOYMENT_TARGET = 11` for x86_64, what's the recommended workflow for contributors building locally with newer Xcode versions? Should we just override the target locally to `12.0` (e.g., via `skip-worktree` or `-xcconfig`), or would the team consider adding something like `#include? "Local.xcconfig"` to `Shared.xcconfig` to make local developer overrides cleaner? Or is there any plan to bump the minimum deployment target to `12.0` across the board in the near future? (Currently I just set `MACOSX_DEPLOYMENT_TARGET = 12` locally with `skip-worktree` btw)

If any configuration adjustments are desired by the maintainer team, I would be very happy to prepare a follow-up PR. Thanks anyway!


low-batt (approved):
Changes look good to me.

@yorkyang2333 Thank you very much for posting this PR!

PR #6203, opened 2026-07-17
Settings: split pasted mpv options into key and value

yorkyang2333:
- [ ] Related issue: #
---
---

**Description:**

Automatically split pasted mpv options into the Key and Value columns in the new settings page.

For example, pasting `hr-seek-framedrop=yes` into either cell now sets:
- Key: `hr-seek-framedrop`
- Value: `yes`

### DIFF
--- a/iina/SettingsPageAdvanced.swift
+++ b/iina/SettingsPageAdvanced.swift
@@ -115,6 +115,7 @@ class SettingsPageAdvanced: SettingsPage {
       SettingsList {
         SettingsItem.General(title: .text_AdditionalMpvOptions)
           .image(name: ["document.badge.gearshape", "doc.badge.gearshape"])
+          .hasDescription(content: .text_AdditionalMpvOptions_desc)
           .extraViews(mpvOptionsEditor.delBtn, mpvOptionsEditor.addBtn)
         SettingsItem.Custom()
           .view(mpvOptionsEditor.view)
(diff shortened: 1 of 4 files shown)


iina/SettingsPageAdvanced.swift
```diff
@@ -164,6 +201,30 @@ fileprivate class MPVOptionsEditor: SettingsAccessory.Base, NSTableViewDelegate,
     UserDefaults.standard.synchronize()
   }
 
+  private static func parsePastedOption(_ string: String) -> [String]? {
+    let parts = string.split(separator: "=", maxSplits: 1, omittingEmptySubsequences: false)
+    guard parts.count == 2 else { return nil }
```
uiryuu: Try a=b=c, I think in this case we should still consider this as invalid. So worth considering remove the maxSplits param in split.

uiryuu (changes requested):

iina/SettingsPageAdvanced.swift
yorkyang2333: Removed

iina/SettingsPageAdvanced.swift
```diff
@@ -164,6 +201,30 @@ fileprivate class MPVOptionsEditor: SettingsAccessory.Base, NSTableViewDelegate,
     UserDefaults.standard.synchronize()
   }
 
+  private static func parsePastedOption(_ string: String) -> [String]? {
+    let parts = string.split(separator: "=", omittingEmptySubsequences: false)
+    guard parts.count == 2 else { return nil }
```
lhc70000: Some options can have multiple `=`s, e.g. `vf=foo:option1=value1:option2:option3=value3,bar`

lhc70000 (commented):
Please remove your latest commit because the value can indeed contain multiple `=`s

iina/SettingsPageAdvanced.swift
yorkyang2333: Reverted and modified.

PR #6206, opened 2026-07-19
fix: prevent stale Live Text overlay during playback

yorkyang2333:
- [x] Related issue: #6101 
---
---

**Description:**

Rapidly toggling pause and playback could leave the Live Text overlay visible while the video was playing.

When playback resumed before image analysis finished, `clearAnalysis()` returned early because no overlay existed yet. The pending analysis could then complete and insert a stale overlay.

This change cancels pending analysis even when its overlay has not been created. It also rechecks cancellation, playback state, and the Live Text setting before displaying analysis results.

### DIFF
--- a/iina/LiveTextSupport.swift
+++ b/iina/LiveTextSupport.swift
@@ -69,7 +69,7 @@ class LiveTextController: NSObject {
   }
 
   func clearAnalysis() {
-    guard #available(macOS 13.0, *), isShown else { return }
+    guard #available(macOS 13.0, *), analysisTask != nil || isShown else { return }
     clearAnalysisImpl()
   }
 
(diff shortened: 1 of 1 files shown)


iina/LiveTextSupport.swift
```diff
@@ -119,6 +119,11 @@ extension LiveTextController: ImageAnalysisOverlayViewDelegate {
         }
         try Task.checkCancellation()
         let analysis = try await ImageAnalyzer().analyze(image, orientation: .up, configuration: .init([.text]))
+        try Task.checkCancellation()
+        guard mainWindow.player.info.state == .paused, Preference.isLiveTextEnabled else {
+          liveTextLog("Discarding image analysis results because Live Text is no longer active")
```
uiryuu: Other changes look good to me, but what's the reason of the log? Under what circumstance can this log be triggered?

uiryuu (changes requested):

iina/LiveTextSupport.swift
yorkyang2333: Removed

uiryuu:
Merged to #6262. If you have any other suggestions, feel free to raise on that PR.

PR #6210, opened 2026-07-20
Fix HDR tone mapping not working on SDR display, #5119

### LINKED ISSUE #5119: HDR tonemapping not working on external display
**System and IINA version:**

- macOS 14.6 (23G80)
- IINA 1.3.5 Build 141 (mpv 0.35.0-419-gf79458476b-dirty, FFmpeg 6.0)

**Expected behavior:**
IINA applies tone mapping to my HDR video so I can watch it on my external display without HDR support

**Actual behavior:**
tone mapping doesn't seem to do anything at all (setting target peak to multiple options between 100-320)

See difference in behavior between IINA, mpv, and Infuse (look at the eyes of the left character):
Screenshot 2024-08-08 at 15 29 42
IINA is darkest, mpv is only slightly brighter, Infuse is brightest

Now when we're trying to tone map:
Screenshot 2024-08-08 at 16 31 55

Now the strange thing is that when I drag the window to the Macbook Pro Display, the IINA log says:
```
16:38:05.535 [iina][d] Falling back to nominal display refresh rate: 120.0 from 0.0
16:38:05.536 [mpv0][d] Set property: override-display-fps=120.0
16:38:05.540 [iina][d] Refreshing HDR for player0 @ display1: "Built-in Retina Display" visible frame (-1684.0, 0.0, 1684.0, 1079.0) EDR: {supports=true maxPotential=15.989999771118164 maxCurrent=1.0}
16:38:05.540 [hdr][d] HDR gamma=pq, primaries=bt.2020, sig_peak=5.133005142211914
16:38:05.540 [hdr][d] Will activate HDR color space instead of using ICC profile
16:38:05.540 [mpv0][d] Set property: icc-profile=
16:38:05.540 [mpv0][d] Set property: target-prim=bt.2020
16:38:05.540 [mpv0][d] Set property: target-trc=pq
16:38:05.540 [mpv0][d] Set property: screenshot-tag-colorspace=true
16:38:05.541 [hdr][d] Will enable tone mapping target-peak=100 algorithm=auto
16:38:05.541 [mpv0][d] Set property: target-peak=100
16:38:05.541 [mpv0][d] Set property: tone-mapping=auto
```

But then when I draw it back to the external display it says:
```
16:40:03.187 [iina][d] Falling back to nominal display refresh rate: 49.98677433262449 from 0.0
16:40:03.187 [mpv0][d] Set property: override-display-fps=49.98677433262449
16:40:03.188 [iina][d] Refreshing HDR for player0 @ display2: "34UM95" visible frame (0.0, 0.0, 3440.0, 1415.0) EDR: {supports=false maxPotential=1.0 maxCurrent=1.0}
16:40:03.188 [hdr][d] HDR gamma=pq, primaries=bt.2020, sig_peak=5.133005142211914
16:40:03.188 [hdr][d] HDR video was found but the display does not support EDR mode
16:40:03.188 [hdr][d] Loading ICC profile
16:40:03.192 [mpv0][d] Set property: icc-profile=/Library/ColorSync/Profiles/Displays/34UM95-0FE29938-AFCB-4FB0-8DFB-F94D1D12DAFE.icc
16:40:03.192 [mpv0][d] Set property: target-trc=auto
16:40:03.192 [mpv0][d] Set property: target-prim=auto
16:40:03.192 [mpv0][d] Set property: target-peak=auto
16:40:03.192 [mpv0][d] Set property: tone-mapping=auto
16:40:03.192 [mpv0][d] Set property: tone-mapping-param=default
16:40:03.192 [mpv0][d] Set property: screenshot-tag-colorspace=false
16:40:03.877 [mpv0][d] Set property: window-scale=0.44765625

```

**seemingly ignoring the tone mapping settings...**
I've also attached the mpv log below

[mpv.log](https://github.com/user-attachments/files/16549890/mpv.log)

**Steps to reproduce:**

Always

low-batt:
This commit will:
- Remove setting of mpv tone mapping options from `VideoView.requestEdrMode`
- Add a `isHDRVideo` method to `VideoView`
- Add `setToneMappingForHDR` and `setToneMappingForSDR` methods to `VideoView`
- Change` VideoView.refreshEdrMode` to set the tone mapping options based on whether the video is HDR or SDR

This will change IINA to apply tone mapping when a HDR video is played on a SDR display. Currently tone mapping is only applied when playing a HDR video on a HDR display.

- [x] Related issue: #5119
---
---

**Description:**

### DIFF
--- a/iina/VideoView.swift
+++ b/iina/VideoView.swift
@@ -334,9 +334,6 @@ class VideoView: NSView {
       videoLayer.wantsExtendedDynamicRangeContent = false
       player.mpv.setString(MPVOption.GPURendererOptions.targetTrc, "auto")
       player.mpv.setString(MPVOption.GPURendererOptions.targetPrim, "auto")
-      player.mpv.setString(MPVOption.GPURendererOptions.targetPeak, "auto")
-      player.mpv.setString(MPVOption.GPURendererOptions.toneMapping, "auto")
-      player.mpv.setString(MPVOption.GPURendererOptions.toneMappingParam, "default")
       player.mpv.setFlag(MPVOption.Screenshot.screenshotTagColorspace, false)
     }
   }
@@ -421,7 +418,14 @@ extension VideoView {
       NSScreen.logEDR("Refreshing HDR for \(player.subsystem.rawValue) on display\(displayId)",
                       screen, subsystem: hdrSubsystem)
     }
-    let edrEnabled = requestEdrMode()
+    let edrEnabled: Bool?
+    if isHDRVideo() {
+      edrEnabled = requestEdrMode()
+      setToneMappingForHDR()
+    } else {
+      edrEnabled = false
+      setToneMappingForSDR()
+    }
     let edrAvailable = edrEnabled != false
     if player.info.hdrAvailable != edrAvailable {
       player.info.hdrAvailable = edrAvailable
@@ -430,27 +434,24 @@ extension VideoView {
     if edrEnabled != true { setICCProfile() }
   }
 
-  func requestEdrMode() -> Bool? {
+  /// Returns `true` if the video being played is a HDR video.
+  /// - Returns: `true` if the video is known to be a HDR video, `false` if the video is SDR or the required information is not
+  ///     available.
+  private func isHDRVideo() -> Bool {
     guard let mpv = player.mpv else { return false }
-
     guard let primaries = mpv.getString(MPVProperty.videoParamsPrimaries), let gamma = mpv.getString(MPVProperty.videoParamsGamma) else {
       logHDR("Video gamma and primaries not available")
       return false
     }
-  
     let peak = mpv.getDouble(MPVProperty.videoParamsSigPeak)
     logHDR("Video gamma=\(gamma), primaries=\(primaries), sig_peak=\(peak)")
 
     // HDR videos use a Hybrid Log Gamma (HLG) or a Perceptual Quantization (PQ) transfer function.
     guard gamma == "hlg" || gamma == "pq" else { return false }
 
-    var name: CFString? = nil
     switch primaries {
-    case "display-p3":
-      name = CGColorSpace.displayP3_PQ
-
-    case "bt.2020":
-      name = CGColorSpace.itur_2100_PQ
+    case "bt.2020", "display-p3":
+      return true
 
     case "bt.709":
       return false // SDR
@@ -459,6 +460,10 @@ extension VideoView {
       logHDR("Unsupported color space: gamma=\(gamma) primaries=\(primaries)", level: .warning)
       return false
     }
+  }
+
+  func requestEdrMode() -> Bool? {
+    guard let mpv = player.mpv else { return false }
 
     guard (window?.screen?.maximumPotentialExtendedDynamicRangeColorComponentValue ?? 1.0) > 1.0 else {
       logHDR("HDR video was found but the display does not support EDR mode")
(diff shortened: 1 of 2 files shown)


low-batt:
As discussed in issue #5119 the initial HDR implementation took a conservative approach. At that time there were many serious HDR related problems in macOS. An important priority was to be able to disable IINA's HDR support.

Time to take a close look at this code and update as needed. This PR is just focused on the problem reported in #5119, namely tone mapping is not being applied to a HDR video when played on a SDR display. First a look at how IINA currently behaves…

### Current Behavior

With tone mapping enabled, `Target peak` set to `0`, `Algorithm` set to `mobius` while playing a SDR video:
```text
13:15:28.726 [mpv0][d] Set property: target-peak=auto
13:15:28.726 [mpv0][d] Set property: tone-mapping=auto
13:15:28.726 [mpv0][d] Set property: tone-mapping-param=default
```

The same behavior occurs while playing a HDR video on a display that does not support HDR. The same behavior also occurs if tone mapping is disabled.

With the same settings playing a HDR video on a XDR display:

```text
13:26:59.758 [hdr0][d] Successfully obtained information about the display
13:26:59.758 [hdr0][d] Found NonReferencePeakHDRLuminance: 1600
13:26:59.758 [hdr0][d] Will enable tone mapping: target-peak=1600 algorithm=mobius
13:26:59.758 [mpv0][d] Set property: target-peak=1600
13:26:59.764 [mpv0][d] Set property: tone-mapping=mobius
```

IINA set `target-peak` based on the XDR display's capabilities and configured tone mapping as expected.

Playing a HDR video on a XDR display with tone mapping disabled:

```text
13:23:44.101 [mpv0][d] Set property: target-peak=auto
13:23:44.101 [mpv0][d] Set property: tone-mapping=
```

A bit odd that the mpv `tone-mapping` option is being set to the empty string.

Advanced IINA users might want to use a `mpv.conf` file with [conditional auto profiles](https://mpv.io/manual/stable/#conditional-auto-profiles) to set these options. As a test, the following `mpv.conf` file sets these options to different values:
<details>
<summary><b>mvp.conf:</b></summary>

```text
[HDR]
profile-desc=HDR
profile-cond=p["video-params/primaries"] == "bt.2020"
profile-restore=copy
target-peak=1000
tone-mapping=reinhard

[SDR]
profile-desc=SDR
profile-cond=p["video-params/primaries"] and p["video-params/primaries"] ~= "bt.2020"
profile-restore=copy
target-peak=500
tone-mapping=linear
```
</details>

This did not work. The profiles were properly applied by mpv and IINA then overwrote the option values.

### Analysis

The current  code only applies tone mapping when playing a HDR video on a display that supports HDR. The code needs to be changed to only consider if the video being played is HDR when deciding whether to apply tone mapping.

IINA's `Enable tone mapping` setting does not map to a mpv configuration. Tone mapping is always enabled (it may or may not be active). This IINA setting provided a way to disable the IINA code involved in tone mapping. However as can be seen above, IINA still sets the mpv options controlling tone mapping. To allow users to be able to control tone mapping using a `mpv.conf` file the code can be updated to not set any mpv tone mapping options if this IINA setting is not enabled.

### Fixing

The fix removes setting the mpv tone mapping options from `VideoView.requestEdrMode` and puts them in the new methods `setToneMappingForHDR` and `setToneMappingForSDR`. The method `refreshEdrMode` is changed to call the appropriate new method. This causes tone mapping to be applied when a HDR video is played on a SDR display, correcting #5119.

The code that sets the mpv options has been changed to do nothing if the IINA `Tone mapping` setting is not enabled. This allows advanced users to use a `mpv.conf` file to control these options.

### Additional Issues

This PR is focused on issue #5119. There are other issues with respects to tone mapping that should be discussed.

#### Alignment with mpv

From the mpv documentation for the [target-peak](https://mpv.io/manual/stable/#options-target-peak) option:
> in auto mode (the default), the chosen peak is an appropriate value based on the TRC in use. For SDR curves, it uses 203. For HDR curves, it uses 203 * the transfer function's nominal peak. If available, it will use the target display's peak brightness as reported by the display.

From a quick search, mpv lacks code to attempt to obtain the display's peak brightness under macOS. So IINA has its own code to obtain the display's brightness. This works for built in displays. Possibly in the further mpv can be enhanced to also do this. Where IINA is _not_ in alignment is that IINA _assumes_ the peak brightness of the display is 400, if it is unable to determine the display's brightness.

To properly obey the above description of `target-peak`, IINA should defer to `libmpv` if it is unable to determine the display's brightness and not assume 400 nits.

Should we make this change?

#### Inelegant User Interface

From the mpv documentation for the [target-peak](https://mpv.io/manual/stable/#options-target-peak) option:
> **--target-peak=<auto|nits>**
> Specifies the measured peak brightness of the output display, in cd/m^2 (AKA nits).

The problem for a user interface is that mpv has defined something whose value is a keyword _or_ a positive integer. This is not the only mpv option that has this kind of value. IINA currently treats `0` as meaning `auto`. That is far from ideal. It should be easy to look at the UI and tell that `target-peak` is being set automatically vs. a specific value has been manually set.

#### Missing User Interface

The video side panel does not contain tone mapping controls. The user may have multiple displays and may need to change target peak and possibly the tone mapping algorithm depending upon what monitor they are using.

#### Missing Validation and Error Reporting

As seen in this screenshot the old settings UI rejected invalid values:

error


The new settings UI allowed `Target peak` to be set to `-1`:
```text
19:30:13.841 [hdr1][d] Will enable tone mapping: target-peak=-1 algorithm=reinhard
19:30:13.841 [mpv1][d] Set property: target-peak=-1
19:30:13.841 [mpv1][d] Set property: tone-mapping=reinhard
```

**No error was reported.**

Checking the mpv log file shows mpv rejected the value: 
```text
[8455.345][v][cplayer] Set property: target-peak=-1 -> -4
```

The meaning of the `-4` return code:
```text
    /**
     * Generic catch-all error if a parameter is set to an invalid or
     * unsupported value. This is used if there is no better error code.
     */
    MPV_ERROR_INVALID_PARAMETER = -4,
```

IINA did not check the return code from `libmpv`. Not even to just log that the error occurred. This is true for a lot of the code that calls into `libmpv`. IINA should at least be logging such failures.

#### Additional Tone Mapping Options

The mpv [tone-mapping-param](https://mpv.io/manual/stable/#options-tone-mapping-param) option is a companion to the [tone-mapping](https://mpv.io/manual/stable/#options-tone-mapping) option. Some of the mpv documentation for that option:
> **--tone-mapping-param=\<value>**
>    Set tone mapping parameters. By default, this is set to the special string default, which maps to an algorithm-specific default value. Ignored if the tone mapping algorithm is not tunable. This affects the following tone mapping algorithms:
>
>    **clip**
        Specifies an extra linear coefficient to multiply into the signal before clipping. Defaults to 1.0.
>
> **mobius**
>        Specifies the transition point from linear to mobius transform. Every value below this point is guaranteed to be mapped 1:1. The higher the value, the more accurate the result will be, at the cost of losing bright details. Defaults to 0.3, which due to the steep initial slope still preserves in-range colors fairly accurately.
> …

Another mpv option where the value can be a keyword or a number.

Should IINA support setting this option?

iina/VideoView.swift
```diff
@@ -467,51 +472,84 @@ extension VideoView {
...
-      logHDR("Will enable tone mapping: target-peak=\(targetPeak) algorithm=\(algorithm)")
-      mpv.setInt(MPVOption.GPURendererOptions.targetPeak, targetPeak)
-      mpv.setString(MPVOption.GPURendererOptions.toneMapping, algorithm)
-    } else {
-      mpv.setString(MPVOption.GPURendererOptions.targetPeak, "auto")
-      mpv.setString(MPVOption.GPURendererOptions.toneMapping, "")
     }
-    return true
+    let algorithm = String(describing: Preference.enum(for: .toneMappingAlgorithm) as
+                           Preference.ToneMappingAlgorithmOption)
```
lhc70000:
```
Preference.ToneMappingAlgorithmOption(key: .toneMappingAlgorithm)
```
is slightly cleaner

low-batt: Using:
```swift
    let algorithm = Preference.ToneMappingAlgorithmOption(key: .toneMappingAlgorithm)
```

Does not compile. Two problems with it. That calls this:
```swift
    init?(key: Key) {
      self.init(rawValue: Preference.integer(for: key))
    }
```

So the expression returns a value of type `ToneMappingAlgorithmOption?`. The initializer returns an optional value because that is what the protocol `InitializingFromKey` requires. This generates a warning about a debug description for an optional value in this statement:
```swift
    logHDR("Will enable tone mapping: target-peak=\(targetPeak) algorithm=\(algorithm)")
```

And an error that `ToneMappingAlgorithmOption?` cannot be converted to `String` in this statement:
```swift
    mpv.setString(MPVOption.GPURendererOptions.toneMapping, algorithm)
```

The `enum` method returns a default value if the initializer returns `Nil`, so the caller does not need to deal with an optional value:
```swift
  static func `enum`<T: InitializingFromKey>(for key: Key) -> T {
    T.init(key: key) ?? T.defaultValue
  }
```

And `String(describing:)` is used to convert to a string. This same source pattern for accessing enum preference values can be seen in `Preference` and other classes

I agree it would be nice to simplify this. I'm not sure how to do that.

low-batt: How about this…

We add this method to `Preference` to obtain an enum value as a string:
```swift
  static func string<T: InitializingFromKey>(for key: Key, ofType t: T.Type) -> String {
    let value: T = Preference.enum(for: key)
    return String(describing: value)
  }
```

Then this code can be:
```swift
    let algorithm = Preference.string(for: .toneMappingAlgorithm,
                                      ofType: Preference.ToneMappingAlgorithmOption.self)
```

Slightly better. I suggest not making that change in this PR as this is a common pattern and instead changing this in its own PR that refactors all uses.

Thoughts?

low-batt: I've updated the PR with this change. If this is merged I will refactor similar code to use the new `Preference.string` function.

low-batt:
Rebased with `develop`.

PR #6214, opened 2026-07-23
Distinguish numpad keys from the number row in key bindings

### LINKED ISSUE #4898: Numeric Keypad Support Suggestion
**What you want IINA to do:**
In input config, fuzzily distinguish numpad keys from their corresponding number row keys. More specifically, if the input config only sets a binding for one of them, use both to do the same thing; however, if there are different bindings for both, distinguish the keys. For example: if the user sets the number row key "1" to do "A", and doesn't set a binding for the numpad key "KP1", pressing either key will do "A"; or if the user sets the numpad key "KP1"  to do "B", and doesn't set a binding for the number row key "1", pressing either key will do "B"; however, if the number row key "1" to do "A", and the numpad key "KP1" is set to do "B", pressing the two keys will do different things.

**What IINA does currently:**
Currently, IINA key mapping only supports number row keys 0 - 9. Setting key binding with the in-app key mapping feature will save numpad keys as number row keys instead. Editing the config file manually to set bings with numpad keys KP0 - KP9 will not work, the binds do show up in the IINA key binding settings, but as 0 - 9, and the app doesn't actually respond to them; if bindings for number row keys 0 - 9 have been set, pressing KP0 - KP9 will work the same as those instead.

To have more keys to use. And it won't affect existing input config files generated through IINA.

mpv fully supports numpad keys, but it distinguishes them from number row keys strictly, not interchangeable even if only one kind of them is set.

AbdelrahmanHafez:
- [x] Related issue: #4898
---
---

**Description:**

Numpad keys never trigger key bindings, because `mpvKeyCode(from:)` checks `isPrintable` before consulting `keyMap`, so numpad digits resolve to the same strings as the number row ("9" instead of "KP9"). `keyMap` already has the correct `KP0`-`KP9` entries for the keypad keycodes, they were just unreachable. As a result, `KP9` lines in a conf file parse fine and show up in the settings UI, but never match any incoming key event.

This PR:

- uses the mpv key names (`KP0`-`KP9`, `KP_DEC`) when the event has the `.numericPad` modifier flag, so numpad keys become distinguishable from the number row, same as mpv
- adds a fallback to the binding lookup: if a numpad key has no binding, the binding of its number-row twin is used (`KP9` -> `9`, `KP_DEC` -> `.`, modifiers preserved), so existing configs that only bind the plain keys keep working from the numpad. Binding both keys separately keeps them distinct, since an exact match always wins.
- stops assigning menu item key equivalents from `KP*` bindings. `NSMenuItem` cannot represent numpad keys, so the equivalent was degraded to the number-row char and the menu would swallow both physical keys and run the wrong action when the two keys had different bindings. This also makes the non-raw key bindings view show keypad keys as their mpv names (`KP5` instead of an ambiguous `5`).
- adds an `IINAApplication` subclass (set as `NSPrincipalClass`) whose `sendEvent` delivers numpad key presses that have a binding directly to the player window, ahead of menu key equivalent matching. Arrow keys (which also carry the `.numericPad` flag), non-player key windows, and windows with an attached sheet are left to normal dispatch, so dialogs and text fields behave as before.


low-batt:

Thank you very much for posting this pull request. I believe the lack of support for  numpad keys is something that needs correcting. Other developers handle this area of IINA. I will provide some feedback, but you should wait for the other developers to comment before making changes. They know this area of IINA and may disagree with my feedback. IINA is under heavy development right now and the other developers are overloaded, so it might be a while before we hear from them.

My own bias is for IINA to as much as possible match mpv's behavior. There are many advantages to matching mpv's behavior. Currently IINA is forced to differ from mpv in a couple of ways due to restrictions in the IINA UI code. That is not considered a good thing and one of those key restrictions (`keepaspect`) is being removed for the next release.

Thus if I was reviewing this PR I would **reject it** because it implements the "fuzzy" behavior suggested in the issue which intentionally deviates from the mpv behavior. Improvements upon mpv behavior should be taken up with the mpv project. Once mpv implements it IINA should then match up. Deviations from mpv behavior must be kept to a minimum.

There are other outstanding key binding PRs that have yet to be reviewed. I do not know if they also address the missing support for numpad keys.

Normally it would be good to include new tests with a PR that adds or changes functionality, however IINA currently has zero tests. Might be better to first establish testing in a separate PR. The outstanding PR #3787 does exactly that and includes tests for key bindings. The developers who review this PR will need to consider if PR #3787 should be reviewed and merged before this one.

One of the tasks for IINA when upgrading `libmpv` is to review changes to their key bindings found [here](https://github.com/mpv-player/mpv/blob/master/etc/input.conf) and update IINA's default key bindings found in [config](https://github.com/iina/iina/tree/develop/iina/config) appropriately. Adding numeric keypad support adds another reason for updating the bindings. Because keypad support is missing IINA's `mpv Default` key bindings do not contain the keypad bindings mpv has. This is just a reminder for the reviewers. Updating the default bindings should be done in a separate PR after `libmpv` has been upgraded.

@svobs Thoughts?

low-batt:
For a quick test I added these two bindings:

raw

I would have expected that pressing the `5` on the keyboard would cycle full screen mode and pressing `5` on the keypad would pause or resume playback. But both keys paused or resumed playback. Similar bindings worked as expected with mpv.

It is also not possible to distinguish that one key is bound to the keypad in the normal view:

normal

Remember, no need to address any feedback from me. See what the other developers have to say.

AbdelrahmanHafez:
Appreciate the detailed feedback and context. Makes sense. Let's wait for other reviews, and if they agree I'll take it to mpv, and then create a PR in IINA (if needed) to work with the new mpv behavior when/if they release a fix.

svobs:
> @svobs Thoughts?

To provide context for other readers: I've worked extensively in this area in IINA, but I am not a member of the core team and have no direct line of communication with them, so while I can provide technical insight, any opinions I provide are my own and do carry any weight with regard to what will or won't be accepted by them.

> My own bias is for IINA to as much as possible match mpv's behavior. 

I suspect the mpv maintainers will strongly resist adding this kind of fuzzy logic, as they have generally insisted on well-defined behavior in the past. Although it wouldn't hurt to ask.

But personally I think it would be a good fit for the IINA project to override mpv's behavior, exactly because IINA has done such overrides in the past to make things easier for casual users, sometimes to the detriment of power users.

In fact you could say the current behavior is exactly the way it is because it favors casual users.

Since probably the vast majority of users are using the Macbook built-in keyboard, you might define any user who is using a keyboard with a keypad as a power user. Whereas casual users are less likely to need or even understand that "number row 5" and "keypad 5" can do different things.

Unfortunately it's historically been difficult (in my experience) to get the maintainers of IINA to approve extra code to support power users. Though it does sometimes happen.

> Normally it would be good to include new tests with a PR that adds or changes functionality, however IINA currently has zero tests. Might be better to first establish testing in a separate PR. The outstanding PR https://github.com/iina/iina/pull/3787 does exactly that and includes tests for key bindings.

Thanks for the callout ;) I agree that tests are great, but removing them will improve the chances of this PR getting accepted.

Most importantly, looks like one technical hurdle still needs to be addressed: as I mentioned in [this post](https://github.com/iina/iina/issues/4898#issuecomment-2105571130):

> But menu item "key equivalents" are a different mechanism still. The public API for `NSMenuItem` provides only `keyEquivalentModifierMask` (which ignores the `numericPad` flag) and `keyEquivalent` which is a string. Perhaps there are some magic strings for the number pad chars? I couldn't find anything. I did find [this old link](https://stackoverflow.com/questions/4928158/nsmenuitem-custom-view-does-not-respond-to-keyequivalent?rq=3) which describes an awkward workaround which seems like more of a last resort.

What I described as an "awkward workaround" should probably be stated as a "necessary workaround" to address the case where a number key binding is mapped to a menu item (these APIs are so old that adding any custom behavior could be described as "awkward" by today's standards...) The application-wide event loop handles these before the player window and its default behavior does not distinguish between number row keys & their keypad key equivalents, so an application-wide override will be needed to change this behavior.

I'll bet this is probably why @low-batt experienced an unexpected result for `cycle fullscreen` & `cycle pause` [above](https://github.com/iina/iina/pull/6214#issuecomment-5063848153).

AbdelrahmanHafez:
Thanks for the additional feedback @svobs, I'll also wait for a member of the core team's feedback and then address all feedback in one go. If it's usually difficult to get things moving with mpv team, it's probably practical to just implement this in IINA, but the team might have different constraints. Looking forward to seeing this moving forward regardless of which direction. Do you know if there's an active team member we can mention here?

svobs:
> Do you know if there's an active team member we can mention here?

Maybe @low-batt will have more insight here as they seem to have some visibility into the core team's internal processes. To me it looks like no one on the core team who can sign off on significant backend changes is still active, other than @lhc70000 himself.

low-batt:
For the record, the opinions of @svobs carry a _lot_ of weight with me.

I too would not be surprised if the mpv project rejected proposals to change how key bindings work.

I have been trying to slowly remove some of the things IINA does that prevents you from doing some of the great things you can configure mpv to do. This is why deviating from mpv always worries me. It is sometimes hard to foresee the problems altering behavior might introduce. However the senior developers might like the fuzzy logic. We will have to wait for them to comment.

For key bindings we probably need @lhc70000. @uiryuu also has been active lately. My bad that I forgot to request a review. I will do that now.

Development of IINA 1.4.x basically ended with 1.4.2. Both 1.4.3 and 1.4.4 were tiny releases for critical problems. The team is now focused on what will be IINA 1.5.0 beta 1. One priority was to merge some major UI changes that had been in development on a branch to avoid the usual problem with needing to adapt and recode any changes to the `develop` branch to the new UI code. That merge has happened, but there is still more work to do on the UI. Likely the senior developers are very focused on that at the moment. Once the problems have been shaken out of the new UI we should be able to get the attention of the senior developers on the PRs have built up. If I get their attention I will mention this PR.

iina/PlayerCore.swift
```diff
@@ -320,6 +320,15 @@ class PlayerCore: NSObject {
...
 
+  /** Looks up the binding for a normalized keystroke. If none is bound, falls back to the binding of its
+   numpad / non-numpad twin (e.g. `KP9` matches a `9` binding and vice versa), so a config that only binds
+   one of the pair responds to both keys (#4898). */
+  static func keyBinding(for normalizedKeyCode: String) -> KeyMapping? {
+    if let kb = keyBindings[normalizedKeyCode] { return kb }
+    guard let alternative = KeyCodeHelper.numpadAlternative(for: normalizedKeyCode) else { return nil }
+    return keyBindings[alternative]
+  }
+
```
lhc70000, line 335: Since the sole purpose of `numPadTwins` and `numpadAlternative` is just string operation, I suggest using regular expression directly, like

```swift
    return keyBindings[
      kb ?? kb.replacingOccurrences(of: "KP(\\d)",
                                    with: "$1",
                                    options: .regularExpression)
    ]
```

lhc70000:
Also, please remove the unit test and test scheme. IINA currently doesn’t have tests; it's not a good practice, but we don't have time and energy to maintain and run tests.

AbdelrahmanHafez:
Done, removed the unit tests and the test scheme, and replaced the twin mapping with the regex approach. One difference from the snippet: I kept a `KP_DEC` -> `.` replacement next to `KP(\d)` -> `$1`, because numpad period resolved to `.` before this PR (it went through the printable-char path), so a config that binds `.` would have lost the numpad key otherwise. The fallback is also one-directional now, the number row no longer triggers `KP*`-only bindings.

lhc70000:
I think we can have this feature for now, especially because macOS doesn't distinguish numpad keys in the menu, and IINA usually sticks to macOS's behavior when there's a conflict with mpv.

For users with `KP` key bindings defined, this also will not cause any ambiguity. Or we can think this way: a user doesn't want this behavior only if they press a numpad key and expect nothing to happen - I can't think of a meaningful use case of it.

low-batt:
I see we decided the fuzzy behavior is desirable as it matches up with macOS behavior.

Where are we at with the two issues I brought up above? I pulled the latest commits and in the default view I cannot tell that one of my bindings differs from the other with respect to the key. I have to turn on `Display raw values` to see that one of them is bound to the keypad. Maybe the non-raw value should show `Keypad 5`?

Do we know if the behavior has changed? The behavior in that quick test certainly seemed wrong to me. I will test the behavior with the latest commits tomorrow.

low-batt (changes requested):
With the following key binding configuration file:
[Keypad.conf.zip](https://github.com/user-attachments/files/30391812/Keypad.conf.zip)

Which only contains these two bindings:
bindings


And using IINA built from the latest commit:
- Pressing the `5` key on the numeric keypad runs `cycle pause` as expected
- Pressing the `5/%` key in the number keys row also runs `cycle pause`, **not** `cycle fullscreen`

This does not match the behavior described in the issue:
> if the number row key "1" to do "A", and the numpad key "KP1" is set to do "B", pressing the two keys will do different things.

This behavior needs to be corrected before these changes can be merged.

Another problem is that you have to enable `Display raw values` in order to tell that `cycle pause` is bound to the `5` key on the keypad. I searched, but did not find any convention for indicating a key specific to the numeric keypad. Does anyone know of one?

As the convention is to use capital letters I'm thinking the non-raw view should display the key as `KEYPAD 5`, _in the case of the bindings used in the test where both keys are bound_. The inclusion of fuzzy logic raises the question of how the binding should be displayed in the non-raw view when there is only is only one binding and either key can be used. Should that display `5 or KEYPAD 5`? That at least would make it clear what the fuzzy logic is doing.

Thoughts everyone?

AbdelrahmanHafez:
Addressed: with your config, keypad 5 now runs `cycle pause` and row 5 runs `cycle fullscreen`. Menus no longer take key equivalents from `KP*` bindings, and an app-level `sendEvent` override (the workaround svobs described) hands bound numpad keys to the player window ahead of menu matching.

The non-raw view now shows the pause binding as `KP5`:

iina-6214-nonraw-fixed

On the display convention, I'd keep `KP5` rather than `KEYPAD 5` (matches the raw view and conf syntax), and for a single fuzzy binding I'd display it as written and keep the fuzziness behavioral. Let me know if I missed anything.

low-batt:
I'm totally confused as to why it is still not working for me. With the key binding configuration I posted both keys are executing `cycle fullscreen` for me. I double checked that I  am testing with the latest commit, 443f6daf919989e3fa26cf403f9f512e73dc4e37.

On this:
> On the display convention, I'd keep `KP5` rather than `KEYPAD 5` (matches the raw view and conf syntax)

But this is not the raw view. This is the view that is supposed to be easy for normal Mac users to understand. Looking through how other bindings are displayed in the "normal" view shows there is no requirement to match the raw view and conf file syntax. Looking closer at the conventions used in that view it probably should be `KEYPAD_5` with the underscore. If you search the net for `KP5` it is a pressure switch. The `KP` prefix is a mpv thing, not something normal Mac users will be familiar with. They might guess what it is. But why are we making users guess? See what the other developers think about this.

My suggestion for making the fuzzy binding explicit, so `5 or KEYPAD_5` when there is only one bindings, that is a little more controversial. I would not be surprised if the other devs reject that. But how else is the user to understand what IINA is doing?

iina.xcodeproj/project.pbxproj
```diff
@@ -2906,7 +2907,7 @@
 			mainGroup = 84EB1ECD1D2F51D3004FA5A1;
 			packageReferences = (
 				B4E446E625CB53920069F06E /* XCRemoteSwiftPackageReference "PromiseKit" */,
-				B4E446F525CB54EF0069F06E /* XCRemoteSwiftPackageReference "GRMustache" */,
+				B4E446F525CB54EF0069F06E /* XCRemoteSwiftPackageReference "GRMustache.swift" */,
```
low-batt: What is this change about?

iina/IINAApplication.swift
```diff
@@ -0,0 +1,51 @@
...
+      window.sendEvent(event)
+      return
+    }
+    super.sendEvent(event)
+  }
+
+  /// The player window a numpad key event should be delivered to, or `nil` to use normal
+  /// dispatch. When the app is inactive there is no key window; programmatically delivered
+  /// events (Accessibility, automation) should still reach the player window. A window with
+  /// an attached sheet is waiting for dialog input, so it must not receive playback commands.
```
low-batt, line 33: The convention for a function's documentation comment is to start it with a short one sentence summary of what it does. If a more detailed description is required then that should be given after a blank line. The comment for `routingWindow`,  `isBoundNumpadKeyEvent` and `keyBinding` are missing the summary. The documentation comment for `sendEvent` above correctly follows the convention. Xcode expects this convention as can be seen in this screenshot:
doc

svobs:
I did some basic testing too and see the same issue. Consider the following input config files I created.

### Case 1:

Configuration:
```
1 cycle pause
KP1 cycle mute
```

Test results:

| Key Pressed | Expected Result | Actual Result |
| ------------ | ---------------- | ----------------- |
| 1 (row) | Toggle Pause | ✅ Toggle Pause |
| 1 (keypad) | Toggle Mute | ❌ Toggle Pause |

### Case 2:

Configuration:
```
1 cycle pause
KP2 cycle mute
```

Test results:

| Key Pressed | Expected Result | Actual Result |
| ------------ | ---------------- | ----------------- |
| 1 (row) | Toggle Pause | ✅ Toggle Pause |
| 1 (keypad) | Toggle Pause | ✅ Toggle Pause |
| 2 (row) | Nothing | ✅ Nothing |
| 2 (keypad) | Toggle Mute | ✅ Toggle Mute |

> Looking closer at the conventions used in that view it probably should be KEYPAD_5 with the underscore

Agree that "keypad" should be spelled out.

AbdelrahmanHafez:
Thanks for the feedback @low-batt @svobs 
I'll take a look on this later this week/on the weekend.

AbdelrahmanHafez:
@low-batt @svobs fixed in 2defcc74. Tested and everything works as expected:

<details>
<summary>Screen recording showing all cases in action (⊞ is numpad)
</summary>

https://github.com/user-attachments/assets/63e1300b-fd9e-406a-a27b-304ec774ca95
</details>

<details>
<summary>Raw config screenshot</summary>
Screenshot 2026-08-10 at 05 01 41 AM
</details>
<details>
<summary>Non-raw config screenshot</summary>
Screenshot 2026-08-10 at 05 01 37 AM
</details>

<details>
<summary>Config text</summary>

```conf
1 cycle pause
KP1 cycle mute
5 cycle fullscreen
Alt+9 add volume 5
. frame-step
TAB cycle pause
ESC set fullscreen no
```

</details>


----------

AI generated 👇 

The remaining collision came from a second `performKeyEquivalent` call in both player window classes. The app-level override routed the keypad event correctly, then the window handed it back to the menu and it was interpreted as the number-row key again. Bound physical keypad events now skip that second menu lookup, while ordinary menu shortcuts and normal responder handling keep their existing path.

For the two cases @svobs posted:

- `1 cycle pause` + `KP1 cycle mute`: row 1 pauses, keypad 1 mutes.
- `1 cycle pause` + `KP2 cycle mute`: row 1 and keypad 1 pause, row 2 does nothing, keypad 2 mutes.

The non-raw view now spells these as `Keypad 1`, and a fuzzy binding is explicit, for example `5 or Keypad 5`. Raw view remains the exact config syntax. I went with `Keypad 5` rather than `KEYPAD_5` because this is the human-readable view, while raw mode already exposes `KP5`.

iina/IINAApplication.swift
AbdelrahmanHafez: Addressed.
Today I learned, thanks!

iina.xcodeproj/project.pbxproj
AbdelrahmanHafez: That was an accidental change, I reverted it in f7c67bfa. The remaining project-file changes only add `IINAApplication.swift` to the app target and set `IINAApplication` as `NSPrincipalClass`, both required by this PR.

PR #6220, opened 2026-07-25
Fix Inspector window flashing white when closed

### LINKED ISSUE #6143: Inspector window background turns white while closing
### IINA Version

1.4.3 Build 167

### Release Channel

Release

### macOS Version

Seqouia 15.7.7

### Steps to Reproduce

- Open the Inspector window while a video is playing (cmd-i)
- Close the Inspector window with cmd-w

### Expected Behavior

The Inspector window closes normally like any other window, e.g. the settings window.

### Actual Behavior

The background of the Inspector window turns white as it closes.

_No response_

### Logs

_No response_

### Sample Files

_No response_

Always

### Required Actions

### Optional Actions

hkarmoush:
## Summary

Closes the Inspector window (Cmd-I) with Cmd-W (or the red close button) and you'll see the whole panel flash white for a frame before it disappears.

The Inspector is the only HUD-style panel in IINA (`utility` + `HUD` style mask in `InspectorWindowController.xib`), but it never set an explicit `animationBehavior`, so it defaulted to `.default` — the standard document-window close animation. That animation renders a bitmap snapshot of the window to fade/shrink it out, and it doesn't account for HUD chrome, so the layer-backed content view's transparent backing shows through as white for that one frame.

`NSWindow.AnimationBehavior.utilityWindow` is the behavior Apple documents for inspector/HUD-style panels specifically, so this sets that in `windowDidLoad()` instead of leaving it at the document-window default.

## Known Limitations

- This only touches the close animation; the window's default opening/miniaturize animation was already fine.
- No other window in the app uses the HUD style, so this fix is scoped to `InspectorWindowController` only.

## Test Plan

- [x] Built with `xcodebuild -project iina.xcodeproj ONLY_ACTIVE_ARCH=NO -scheme iina -configuration Nightly` — succeeded
- [x] Opened Inspector (Cmd-I) during playback and closed it with Cmd-W repeatedly — no white flash
- [x] Confirmed the Inspector still opens, updates live, and resizes normally (no regression from the animation change)

Fixes #6143

### DIFF
--- a/iina/InspectorWindowController.swift
+++ b/iina/InspectorWindowController.swift
@@ -87,6 +87,11 @@ class InspectorWindowController: NSWindowController, NSWindowDelegate, NSTableVi
   override func windowDidLoad() {
     super.windowDidLoad()
 
+    // This is a HUD/utility panel, not a document window. The default close animation
+    // doesn't handle the HUD chrome correctly and flashes white; `.utilityWindow` is the
+    // animation Apple documents for exactly this panel style.
+    window?.animationBehavior = .utilityWindow
+
     watchProperties = Preference.array(for: .watchProperties) as! [String]
     watchTableView.delegate = self
     watchTableView.dataSource = self


low-batt:

Thank you for investigating the problem and posting this fix. I will let the senior developers review this PR as I have not been able to reproduce the problem.

uiryuu:
As I mentioned in https://github.com/iina/iina/issues/6143, I couldn't reproduced the issue either. @hkarmoush Can you share how did you managed to reproduced the issue? Also please remove the comments. Add simpler comments and maybe leave a link to the original issue is also fine.

@low-batt I guess the changes don't hurt and it can potentially fix the issue, even if we cannot reproduce it.

Issue #2595, opened 2019-09-30
Takes too long to open a very long playlist


MuntashirAkon:
IINA takes too long to open the playlist (sidebar) panel if it was loaded with a playlist containing huge media, and the spinner continues to spin (perhaps due to some heavy tasks running in the foreground which should be run on the background). Note: the playlist menu (Playback > Playlist) works just fine.

**System and IINA version:**

- macOS 10.14.6
- IINA 1.0.4

**Expected behavior:**
Playlist panel should open immediately regardless of the size of the playlist.

**Actual behavior:**
Playlist panel takes too long to open if the playlist is huge.

**Steps to reproduce:**
- Load a long playlist (for instance, [this m3u file](https://raw.githubusercontent.com/iptv-org/iptv/master/index.country.m3u)
- Try to open the playlist panel

**How often does this happen?**
Everytime.

PR #6221, opened 2026-07-25
Fix crash connecting an external display, #6215

### LINKED ISSUE #6215: 睡眠中重新插入外接显示器之后，iina崩溃
### IINA Version

1.4.4

### Release Channel

Release

### macOS Version

26.0.1 (25A362)

### Steps to Reproduce

现象：睡眠中重新插入外接显示器之后，iina崩溃

Answer chatgpt 5.6 sol:

The crash was caused by a race condition in IINA 1.4.4 while macOS was reconfiguring an external display.
The crash report shows an EXC_BREAKPOINT (SIGTRAP) on the main thread in Display.init(_:), called by DisplayController.addNewDisplays() after IINA received a display-configuration notification.
At the same time, the system log shows that external display ID 4 was being reprobed and temporarily reported an invalid 0 × 0 display mode. During this transient state, CGDisplayCopyAllDisplayModes() may return nil or an unexpected value. IINA force-casts the result to [CGDisplayMode], causing a Swift runtime trap and terminating the application.
This is an IINA display-management bug, not a problem with the video file, decoder, subtitles, or plugins. It was most likely triggered by the external display powering off, the Mac preparing to sleep, or a temporary dock/display reconnection.

### Expected Behavior

not crash

### Actual Behavior

IINA Quit Unexpectedly

[IINA-2026-07-24-090412.ips.zip](https://github.com/user-attachments/files/30329992/IINA-2026-07-24-090412.ips.zip)

### Logs

_No response_

### Sample Files

_No response_

Once a day

### Required Actions

### Optional Actions

low-batt:
This commit will:
- Change the `Display` initializer to support `CGDisplayCopyAllDisplayModes` returning `nil`
- Change the `displayModes` property to be optional
- Change the `Display.description` property to handle `displayModes` being `nil`
- Add a new `isVirtual` property to `Display` and change the initializer to set it
- Add support for the `isVirtual` property to the `description` property
- Change the initializer to handle product names that are the empty string
- Remove the check in `DisplayController` that requires the number of displays to change
- Add a new `displayId` property to the `NSScreen` extension
- Change `VideoView.updateDisplayLink` to use the new `displayId` property
- Change the `didChangeScreenParametersNotification` observer in `MainWindowController` to properly ignore screen changes IINA does not need to respond to
- Add a new `logAll` method to the `NSScreen` extension
- Change log methods in the `NSScreen` extension to not form messages if the log message will not be emitted
- Add the logger subsystem to calls to `NSScreen` logging methods where missing

The Apple documentation for the `CGDisplayCopyAllDisplayModes` method indicates it only returns `nil` if called with an invalid display ID. As the display ID was known to be valid the code was not expecting the results to be `nil`. According to the ChatGPT analysis posted in the issue macOS will return `nil` or an invalid display mode when macOS is re-probing the display. This commit will correct the crash by updating code to support `CGDisplayCopyAllDisplayModes` returning `nil`.

This commit will also correct problems where IINA was ignoring notifications about screen parameters changing that should have been processed.

This commit will also improve logging of displays and screens. This logging is needed to debug problems involving external displays.

- [x] Related issue: #6215
---
---

**Description:**

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -718,11 +718,9 @@ class MainWindowController: PlayerWindowController {
       // notification if changes require processing this notification.
       let screensChanged: Bool = {
         guard screens.count == cachedScreens.count else { return true }
-        for i in 0..<screens.count {
-          if screens[i].frame != cachedScreens[i].frame { return true }
-          if screens[i].displayId != cachedScreens[i].displayId { return true }
+        return zip(screens, cachedScreens).contains {
+          $0.frame != $1.frame || $0.displayId != $1.displayId
         }
-        return false
       }()
       guard screensChanged else { return }
       // Update the cached screens
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -64,8 +64,8 @@ class MainWindowController: PlayerWindowController {
 
   /** For blacking out other screens. */
   var screens: [NSScreen] = []
-  var cachedScreenCount = 0
   var blackWindows: [NSWindow] = []
+  var cachedScreens: [NSScreen] = []
 
   lazy var rotation: Int = {
     return player.mpv.getInt(MPVProperty.videoParamsRotate)
@@ -671,7 +671,7 @@ class MainWindowController: PlayerWindowController {
     fadeableViews.update()
 
     // other initialization
-    cachedScreenCount = NSScreen.screens.count
+    cachedScreens = NSScreen.screens
     [pipOverlayView].forEach {
       $0?.state = .active
     }
@@ -707,21 +707,42 @@ class MainWindowController: PlayerWindowController {
 
     addObserver(to: .default, forName: NSApplication.didChangeScreenParametersNotification) { [unowned self] _ in
       // This observer handles a situation that the user connected a new screen or removed a screen
-      let screenCount = NSScreen.screens.count
-      let countChanged = cachedScreenCount != screenCount
-      if fsState.isFullscreen && Preference.bool(for: .blackOutMonitor) && countChanged {
+      let screens = NSScreen.screens
+
+      // Activating extended dynamic range will cause notifications to be posted at a high rate
+      // because the screen's maximumExtendedDynamicRangeColorComponentValue property keeps changing
+      // as the display's brightness ramps up. This can continue to change if the display is
+      // configured to automatically adjust brightness if ambient lighting is changing. Instruments
+      // identified processing these notifications as a performance problem when they occur at a
+      // high rate. Compare the current list of screens to the cached list and only process this
+      // notification if changes require processing this notification.
+      let screensChanged: Bool = {
+        guard screens.count == cachedScreens.count else { return true }
+        for i in 0..<screens.count {
+          if screens[i].frame != cachedScreens[i].frame { return true }
+          if screens[i].displayId != cachedScreens[i].displayId { return true }
+        }
+        return false
+      }()
+      guard screensChanged else { return }
+      // Update the cached screens
+      cachedScreens = screens
+
+      log("Screen parameters have changed")
+      DisplayController.shared.addNewDisplays()
+      NSScreen.logAll(subsystem: subsystem)
+      NSScreen.log("Window is on screen", window.screen, details: false, subsystem: subsystem)
+
+      if fsState.isFullscreen && Preference.bool(for: .blackOutMonitor) {
         removeBlackWindow()
         blackOutOtherMonitors()
       }
-      // Update the cached value
-      cachedScreenCount = screenCount
       videoView.updateDisplayLink()
-      DisplayController.shared.addNewDisplays()
       // In normal full screen mode AppKit will automatically adjust the window frame if the window
       // is moved to a new screen such as when the window is on an external display and that display
       // is disconnected. In legacy full screen mode IINA is responsible for adjusting the window's
       // frame.
-      guard countChanged, fsState.isFullscreen, Preference.bool(for: .useLegacyFullScreen) else { return }
+      guard fsState.isFullscreen, Preference.bool(for: .useLegacyFullScreen) else { return }
       setWindowFrameForLegacyFullScreen()
     }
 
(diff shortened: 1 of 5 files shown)


low-batt:
The root cause of the crash has to do with this statement from the Apple documentation for [CGDisplayCopyAllDisplayModes](https://developer.apple.com/documentation/coregraphics/cgdisplaycopyalldisplaymodes(_:_:)#return-value):
> ### Return Value
> An array of display modes that the display supports, or `NULL` if the display is invalid. 

As the display was known to be valid, since it was just obtained using [CGGetOnlineDisplayList](https://developer.apple.com/documentation/coregraphics/cggetonlinedisplaylist(_:_:_:)), the code was not expecting the result to be `nil`. The ChatGPT analysis in the issue says:
> The crash was caused by a race condition in IINA 1.4.4 while macOS was reconfiguring an external display.
The crash report shows an EXC_BREAKPOINT (SIGTRAP) on the main thread in Display.init(\_:), called by DisplayController.addNewDisplays() after IINA received a display-configuration notification.
At the same time, the system log shows that external display ID 4 was being reprobed and temporarily reported an invalid 0 × 0 display mode. During this transient state, CGDisplayCopyAllDisplayModes() may return nil or an unexpected value. IINA force-casts the result to \[CGDisplayMode], causing a Swift runtime trap and terminating the application.
This is an IINA display-management bug, not a problem with the video file, decoder, subtitles, or plugins. It was most likely triggered by the external display powering off, the Mac preparing to sleep, or a temporary dock/display reconnection.

If indeed macOS is sometimes returning invalid information then that raises the question of how an app using these APIs is supposed to know if the information returned is valid? IINA uses display information to set mpv options such as [display-fps-override](https://mpv.io/manual/stable/#options-display-fps-override) and [target-peak](https://mpv.io/manual/stable/#options-target-peak). Rather concerning that the macOS APIs may not be absolutely trustworthy.

iina/MainWindowController.swift
```diff
@@ -634,21 +634,42 @@ class MainWindowController: PlayerWindowController {
...
+      // identified processing these notifications as a performance problem when they occur at a
+      // high rate. Compare the current list of screens to the cached list and only process this
+      // notification if changes require processing this notification.
+      let screensChanged: Bool = {
+        guard screens.count == cachedScreens.count else { return true }
+        for i in 0..<screens.count {
+          if screens[i].frame != cachedScreens[i].frame { return true }
+          if screens[i].displayId != cachedScreens[i].displayId { return true }
+        }
+        return false
```
uiryuu, line 725:
```suggestion
        return zip(screens, cachedScreens).contains {
          $0.frame != $1.frame || $0.displayId != $1.displayId
        }
```
for better readability

uiryuu (changes requested):

iina/MainWindowController.swift
low-batt: With this code readability is secondary to performance. Using [zip](https://developer.apple.com/documentation/swift/zip(_:_:)) is definitely easier to read, but does it come at a performance cost? If it is constructing tuples I would think it does. Thanks to Apple naming it zip, I was unable to find much discussion about this amongst all the discussion about zip files. The doc for [Zip2Sequence](https://developer.apple.com/documentation/swift/zip2sequence) says:
> A sequence of pairs built out of two underlying sequences.

The word "built" makes it sound like something is being constructed. I just cannot tell from Apple's descriptions how much the compiler will be able to optimize this.

This code is all about being able to as fast as possible discard notifications that do not need to be processed. We do not want to add any work that can be avoided.

low-batt:
Rebased with `develop` and corrected merge conflicts.

iina/MainWindowController.swift
low-batt: I finally had time to look closely into this. I took a look at the assembly code generated for the release build as well as the debug build.

I intentionally did not code it this like this as I expected the compiler optimizer would perform this optimization:
```swift
        for i in 0..<screens.count {
          let current = screens[i]
          let cached = cachedScreens[i]
          if current.frame != cached.frame { return true }
          if current.displayId != cached.displayId { return true }
        }
```

However when I tested coding it that way the resulting assembly code reduced building for release. I cannot think of a semantic reason why the compiler optimizer could not have performed the above transformation.

I checked the build settings and except for debug the build is configured for [Fastest, Smallest](https://developer.apple.com/documentation/xcode/doing-basic-optimization-to-reduce-your-app-s-size#Check-your-targets-build-settings-for-release-builds), which is the default for a release build:
> Fastest, Smallest: Optimize for size. This setting enables all Faster optimizations that do not typically increase code size. It also performs further optimizations designed to reduce code size. \[-Os]

Apple seems to be emphasizing small, maybe due to a focus on the iPhone?

There is also just `Fastest`:
> Fastest: Turns on all optimizations specified by the Faster setting and also turns on function inlining and register renaming options. This setting may result in a larger binary. \[-O3]

Possibly some of the optimizations I was expecting to see were placed under `Fastest, Aggressive Optimizations`:
> Fastest, Aggressive Optimizations: This setting enables Fastest but also enables aggressive optimizations that may break strict standards compliance but should work well on well-behaved code. \[-Ofast]

But that one sounds risky.

I asked some developer friends about this and their conclusion was: "You are a dinosaur"
:disappointed:

Apparently such optimizations are no longer the focus of the industry. 

Given this I will change this code to use [zip](https://developer.apple.com/documentation/swift/zip(_:_:)). 

In looking over the assembly generated when coded this way, I'm pretty sure Apple's use of the word "creates" is correct. This does seem to be allocating an array of tuples. The assembly code contains this:
```text
    0x106b4effc <+540>: bl     0x106b8a6ec               ; outlined destroy of Swift.Zip2Sequence<Swift.Array<__C.NSScreen>, Swift.Array<__C.NSScreen>> at <compiler-generated>
```

We should keep this behavior in mind when dealing with a large array. Might need to be careful about using `zip` in such a case. I've never looked into it, but there is an issue about playlist performance with some of the large playlist files found on the net. In a quick search I could not find that issue, but I did find issues #2595 and #3357. So playlists are one place where IINA needs to be careful about performance.

low-batt:
Rebased with `develop` again and changed screen notification observer to use zip.

Issue #1172, opened 2017-10-25
Video rotation in menu does not keep video size


pighair:
**System and IINA version:**

- macOS 10.13
- IINA 0.0.14.1

**Expected behavior:**

Window and video size are maintained as possible while rotating video.

If I rotate some video 90 degree of which width is bigger than desktop height, resizing it to maximum height is reasonable. But if I rotate it 180 or back to 0, the window width restored to the width before any rotation, keeping window size as possible.

**Actual behavior:**

Window and video size are maintained to possible maximum width from previous size.

**Steps to reproduce:**

### Reproduce example 1.
1. Prepare a landscape video of which half of width is smaller than desktop height. (I tried with 1920x1080 video on 2560x1600 monitor)
2. Press ⌘+0 to resize to half size. (Window size is now 960x540)
3. Rotate to 90. (Then window size changes to 846x1504, maintaining window width as possible. 1504 is maximum height of window. 1600-[menu height]-[dock height])
4. Press ⌘+0 to resize back to half size. (Window size is now 540x960.)
5. Rotate to 180 or rotate back to 0. (Then window size changes to 540x304, maintaining window width)

### Reproduce example 2.
1. Prepare a landscape video of which width is smaller than desktop height. (I tried with 720x480 video on 2560x1600 monitor)
2. Rotate to 90. (Then window size changes to 720x1080, maintaining window width)
3. Press ⌘+1 to resize back to original size. (Window size is now 500x750. _What??_)
4. Rotate to 180 or rotate back to 0. (Then window size changes to 500x334, maintaining window width)

### Reproduce example 3.
1. Prepare a landscape video of which width is smaller than desktop height. (I tried with 1280x720 video on 2560x1600 monitor)
2. Press ⌘+0 to resize to half size. (Window size is now 640x360)
3. Rotate to 90. (Then window size changes to 640x1138, maintaining window width)
4. Press ⌘+0 to resize back to original size. (Window size is now 500x890. It seems 500 px is minimum width, but I can resize window smaller by dragging corner to 198x300.)
5. Rotate to 180 or rotate back to 0. (Then window size changes to 534x300. It seems minimum height is 300 px. I can resize it smaller to 500x281)

[...]

PR #6222, opened 2026-07-26
Keep window size and position when rotating the video

### LINKED ISSUE #6224: Rotating video changes window size
### IINA Version

1.4.4

### Release Channel

Release

### macOS Version

macOS 15.7.4

### Steps to Reproduce

- Start IINA
- Open a high resolution video
- Grab the corner of the window and move inward to shrink the window size
- Move the window to the right lower corner of the desktop
- Click on `90°` in the submenu `Rotation` under the `Video` menu
- IINA rotates the video, but it also enlarges the window

### Expected Behavior

The video rotates without enlarging the window.

### Actual Behavior

Starting with the window reduced in size and located in the right lower corner of the desktop:

Image

Rotating the video enlarges the window:

Image

_No response_

### Logs

_No response_

### Sample Files

_No response_

So far every time.

### Required Actions

### Optional Actions

AbdelrahmanHafez:
- [x] Related issue: #1172
---
---

**Description:**

Rotating the video resets the player window: every 90° step goes down the same path as a video size change, so the window snaps back to the default scale for the video and gets re-centered, discarding whatever size and position the user had set. Rotating a video you're watching in a small corner window suddenly gives you a full-size window in the middle of the screen.

There is also a race in `onVideoReconfig`: it reads `info.rotation`, which is updated by the `video-rotate` property notification on a separate async hop to the main queue, so whether the reconfig sees the old or the new rotation depends on event timing.

This PR:

- reads `video-rotate` synchronously in `onVideoReconfig`, so the dwidth/dheight normalization no longer races with the property notification
- treats a reconfig whose new display size is the transpose of the previous one as a rotation, and routes it to the existing `handleVideoSizeChange(keepWindowSize: true)` path, so the window flips its aspect ratio in place (same area, same center) instead of being resized and re-centered

Not affected: initial sizing when opening a file (still driven by the resize preferences), playlist navigation, the mini player, and files with rotation metadata. Rotating a metadata-rotated file in place also works, e.g. a file with 270° metadata rotated by 90° flips back to landscape without moving the window.

CC @lhc70000 @low-batt

### DIFF
--- a/iina/PlayerCore.swift
+++ b/iina/PlayerCore.swift
@@ -2537,16 +2537,24 @@ class PlayerCore: NSObject {
     guard info.state.loaded else { return }
     var dwidth = mpv.getInt(MPVProperty.dwidth)
     var dheight = mpv.getInt(MPVProperty.dheight)
-    if info.rotation == 90 || info.rotation == 270 {
+    // Read the rotation synchronously. The video-rotate change notification updates
+    // info.rotation asynchronously, so it can still hold the previous value here.
+    let rotation = mpv.getInt(MPVOption.Video.videoRotate)
+    info.rotation = rotation
+    if rotation == 90 || rotation == 270 {
       swap(&dwidth, &dheight)
     }
     if dwidth != info.displayWidth! || dheight != info.displayHeight! {
       // filter the last video-reconfig event before quit
       if dwidth == 0 && dheight == 0 && mpv.getFlag(MPVProperty.coreIdle) { return }
+      // If the new size is the transpose of the old one, the video was rotated in 90°
+      // steps. Only update the window's aspect ratio, keeping its size and position,
+      // instead of resizing it as if a new video was opened (#1172).
+      let rotationChangedOnly = dwidth == info.displayHeight! && dheight == info.displayWidth!
       // video size changed
       info.displayWidth = dwidth
       info.displayHeight = dheight
-      notifyWindowVideoSizeChanged()
+      notifyWindowVideoSizeChanged(keepWindowSize: rotationChangedOnly)
     }
   }
 
@@ -2708,8 +2716,12 @@ class PlayerCore: NSObject {
     )
   }
 
-  func notifyWindowVideoSizeChanged() {
-    currentController.handleVideoSizeChange()
+  func notifyWindowVideoSizeChanged(keepWindowSize: Bool = false) {
+    if keepWindowSize && !isInMiniPlayer {
+      mainWindow.handleVideoSizeChange(keepWindowSize: true)
+    } else {
+      currentController.handleVideoSizeChange()
+    }
     if currentController.pendingShow {
       currentController.pendingShow = false
       currentController.showWindow(self)


low-batt:
I created a new issue (#6224) so we have some screenshots of the behavior.

low-batt:
Run the following test…

- Start IINA
- The settings panel appears
- On the left side of the panel click on `Key Bindings`
- Click on `IINA Defaults`
- Click on `Duplicate…`
- Enter `PR 6222` for the name of the new key bindings (or some other name)
- Click on `+`
- The `Key Mapping` pop-up appears
- Double click on the entry for `PLAY`
- Press the `5` key
- In `Or enter mpv command` enter `add video-rotate 15`
- Click `OK`
- On the left side of the panel click on `Video`
- Set the `Hardware decoder` setting to `Disabled`

With this key binding configuration and running with the fix in PR #6222, repeat the test described in the issue, but instead of using the video menu to rotate the video press the `5` key. This is the result:

broken


This fix is very helpful as it addresses the common case, but problems remain.

For completeness, with  `Lock Window Aspect Ratio` disabled this is the result:

tilt

iina/PlayerCore.swift
```diff
@@ -2537,16 +2537,24 @@ class PlayerCore: NSObject {
     guard info.state.loaded else { return }
     var dwidth = mpv.getInt(MPVProperty.dwidth)
     var dheight = mpv.getInt(MPVProperty.dheight)
-    if info.rotation == 90 || info.rotation == 270 {
+    // Read the rotation synchronously. The video-rotate change notification updates
+    // info.rotation asynchronously, so it can still hold the previous value here.
```
low-batt, line 2541: I love seeing this comment as without an explanation developers in the future might not understand why the `info` cache cannot be used in this case.

iina/PlayerCore.swift
```diff
@@ -2537,16 +2537,24 @@ class PlayerCore: NSObject {
...
+    let rotation = mpv.getInt(MPVOption.Video.videoRotate)
```
low-batt, line 2542: Not that it matters in this case, but just for the record, the mpv [video-rotate](https://mpv.io/manual/stable/#options-video-rotate) option is _not_ an integer, as can be seen in the following tests using mpv's IPC interface:
```text
low-batt@gag ~$ echo '{ "command": ["get_property", "video-rotate"] }' | socat - /tmp/mpv-socket | jq
{
  "data": 90,
  "request_id": 0,
  "error": "success"
}
```

That came back as an integer, but…
```text
low-batt@gag ~$ echo '{ "command": ["get_property", "video-rotate"] }' | socat - /tmp/mpv-socket | jq
{
  "data": false,
  "request_id": 0,
  "error": "success"
}
```

A boolean! Obtaining that same value as a string:
```text
low-batt@gag ~$ echo '{ "command": ["get_property_string", "video-rotate"] }' | socat - /tmp/mpv-socket | jq
{
  "data": "no",
  "request_id": 0,
  "error": "success"
}
```

This is one of those mpv options that accepts a keyword or an integer.

low-batt (approved):
I pulled the PR, built IINA and tested.

Similar problems remain, but this corrects the case most users will encounter. Changes look good to me.

PR #6227, opened 2026-07-27
Fix camera housing area no longer black in legacy full screen

### LINKED ISSUE #6201: Area around camera no longer black in legacy full screen
### IINA Version

develop at commit c987089

### Release Channel

Other

### macOS Version

macOS 15.7.4

### Steps to Reproduce

Follow these instructing using a Mac whose screen contains a camera housing:
- Build IINA from the `develop` branch
- Use the `defaults` command to set `enableNewSettings` to `true`
- Start IINA
- The _new_ settings panel appears
- Enable `Use legacy full screen`
- On the left side of the panel click on `UI`
- Set `Appearance` to `Light`
- Start playing a video 
- Click on `Enter Full Screen` under the `Video` menu
- The area around the camera housing is no longer blacked out as can be seen in the screenshot below

Image

### Expected Behavior

When using legacy full screen mode the area above the video is black matching the behavior when not using this custom full screen mode.

### Actual Behavior

In the current development version of IINA the area above the video is white, if IINA's appearance is set to `Light`.

This regression was introduced by commit 0df0243.

Previously [NSWindow.backgroundColor](https://developer.apple.com/documentation/appkit/nswindow/backgroundcolor) was set to black in `MainWindowController.windowDidLoad` as confirmed by the now erroneous comment:
https://github.com/iina/iina/blob/c987089efcced9e3088f907551fb8032e191ac32/iina/MainWindowController.swift#L452-L453

That commit also added this code to `PlayerWindowController.setMaterial`:
https://github.com/iina/iina/blob/c987089efcced9e3088f907551fb8032e191ac32/iina/PlayerWindowController.swift#L241

Assuming this change is needed then `MainWindowController` will need to set `backgroundColor` to `black` when entering legacy full screen and reset it when exiting.

_No response_

### Logs

_No response_

### Sample Files

_No response_

Every time.

### Required Actions

### Optional Actions

hkarmoush:
## Summary

On a Mac whose built-in screen has a camera housing, entering legacy full screen with a Light appearance leaves the area around the housing white instead of black (see screenshot in #6201).

Legacy full screen shortens the window's content view to avoid the camera housing (`setWindowFrameForLegacyFullScreen` in `MainWindowController.swift`), so the window's own `backgroundColor` shows through that gap. Commit 0df0243b changed `backgroundColor` from an unconditional `.black` to one that follows the current light/dark appearance (to make the windowed-mode sidebar chrome match the theme), which broke the "always black" assumption legacy full screen depends on for that gap.

This forces the background black specifically while the camera-housing gap is present, and restores the theme-following color when returning to windowed mode.

**Update:** a follow-up commit closes a second gap in the same regression family: `PlayerWindowController.setMaterial()` also unconditionally applies the appearance-following background color, and it fires live whenever the user changes the theme preference (including via the in-window Quick Settings sidebar) — with no regard for full screen state. That means changing the theme while already in legacy full screen on a camera-housing screen would silently re-whiten the gap the first commit had just forced black. `MainWindowController` now overrides `setMaterial` to re-force black in that case.

## Known Limitations

- This only affects Macs with a camera housing (`NSScreen.safeAreaInsets.top > 0`) using legacy full screen with a Light appearance — I don't have such a display, so this remains a code-level fix based on tracing the regression, not a visually confirmed repro. I attempted a manual visual sanity check by temporarily stubbing `cameraHousingHeight` to force the gap on this hardware, but couldn't capture a screenshot to verify pixel color (no Screen Recording permission in this environment) — reviewers with either real camera-housing hardware or that permission are best positioned to confirm visually.
- Scoped to `MainWindowController`; the equivalent codepath for other window controllers doesn't do this content-view shortening.
- Two related, lower-severity gaps in the same area were found but are left out of scope for this PR (happy to file follow-up issues if useful): `InteractiveMode.swift`'s enter/exit also sets an unconditional (non-appearance-aware) background color; and closing the window directly (e.g. Cmd-W) while still in legacy full screen bypasses the normal exit path that restores the theme-following color.

## Test Plan

- [x] Built with `xcodebuild -project iina.xcodeproj ONLY_ACTIVE_ARCH=NO -scheme iina -configuration Nightly` — succeeded
- [ ] Visual confirmation on a camera-housing display with Light appearance + legacy full screen (needs a reviewer with that hardware)

Fixes #6201

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -1751,6 +1751,16 @@ class MainWindowController: PlayerWindowController {
     view.setFrameSize(NSMakeSize(view.frame.width, screen.frame.height - unusable))
   }
 
+  /// `setMaterial` unconditionally sets the window's background to follow the current appearance, which would
+  /// undo the forced black background from `setWindowFrameForLegacyFullScreen` if the user changes the theme
+  /// while still in legacy full screen on a screen with a camera housing. Re-force it black in that case.
+  override func setMaterial(_ theme: Preference.Theme?) {
+    super.setMaterial(theme)
+    guard case .fullscreen(let legacy, _) = fsState, legacy,
+          window?.screen?.cameraHousingHeight != nil else { return }
+    window?.backgroundColor = .black
+  }
+
   private func legacyAnimateToFullscreen() {
     guard let window = self.window else { fatalError("make sure the window exists before animating") }
     // call delegate
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -1704,6 +1704,9 @@ class MainWindowController: PlayerWindowController {
     window.hasShadow = true
     (window as! MainWindow).forceKeyAndMain = false
     window.level = .normal
+    // Undo the forced black background applied for a camera housing gap (see
+    // setWindowFrameForLegacyFullScreen) now that the content view fills the window again.
+    window.backgroundColor = window.effectiveAppearance.isDark ? .black : .white
 
     restoreDockSettings()
     // restore window frame and aspect ratio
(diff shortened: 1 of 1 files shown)


lhc70000:
We can't change the window's background. We need to have a light background with a light Liquid Glass theme; otherwise, the docked sidebars will look very dimmed. A better approach is proposed in #6375, so I'm closing this one.

PR #6232, opened 2026-08-02
Add audio channel routing

LorenzoTTGT:
## Summary
- add an Audio settings control for stereo channel assignment and custom downmix gains
- apply routing through a labeled mpv lavfi/pan audio filter without replacing user audio filters
- update active players when routing preferences change and skip routing while SPDIF passthrough is enabled

## Validation
- git diff --check
- xcodebuild -project iina.xcodeproj -scheme iina -configuration Debug -destination "platform=macOS" build (local build reached CompileAssetCatalog, then failed because local asset tooling supports SVG template format 5.0 while repo assets use format 7.0; GitHub CI uses Xcode 26.5)

### DIFF
--- a/iina/AppData.swift
+++ b/iina/AppData.swift
@@ -129,6 +129,7 @@ struct Constants {
     static let flip = "iina_flip"
     static let mirror = "iina_mirror"
     static let audioEq = "iina_aeq"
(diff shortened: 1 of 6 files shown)


LorenzoTTGT:
Closing: this work is intended to stay on my fork for now, not as an upstream PR.

PR #6233, opened 2026-08-03
Fix RTL language window controls overlap sidebar, #6192

### LINKED ISSUE #6192: RTL language window controls overlap sidebar

low-batt:
This commit will:
- Add a `CommonWindow` class
- Change the following classes to use the new `CommonWindow` class:
  - `HistoryWindowController`
  - `LogWindowController`
  - `MainWindow`
  - `MiniPlayerWindow`
  - `SettingsWindow`
- Change the following XIBs to use the new `CommonWindow` class:
  - `AboutWindowController`
  - `FilterWindowController`
  - `GuideWindowController`
  - `InitialWindowController`
- Add a `NSUserInterfaceLayoutDirection` extension for debugging that provides a textual representation of the enum constants

The new `CommonWindow` class overrides the `windowTitlebarLayoutDirection` property with an implementation that returns the layout direction of the view containing the window close button, instead of the implementation provided by `NSWindow` that uses the primary system language. This will cause `NSWindow` to position the standard window buttons as they would appear if in the macOS system settings the language assigned to IINA matched the primary system language, correcting the overlap reported in issue #6192.

However the AppKit `NSWindow` implementation does not use the value of the `windowTitlebarLayoutDirection` property when laying out the title bar if the window uses a toolbar with the unified style where the toolbar appears next to the window title. This affects the `Inspector`, `Log Viewer` and `Playback History` windows. Due to this unfortunate AppKit behavior these windows will continue to have their standard window buttons positioned based on the primary system language. This also applies to windows created by plugins as they might use a unified toolbar.

- [x] Related issue: #6192
---
---

**Description:**

### DIFF
--- a/iina.xcodeproj/project.pbxproj
+++ b/iina.xcodeproj/project.pbxproj
@@ -128,6 +128,7 @@
 		519872FF26879B9B00F84BCC /* AccessibilityPreferences.swift in Sources */ = {isa = PBXBuildFile; fileRef = 519872FE26879B9B00F84BCC /* AccessibilityPreferences.swift */; };
 		519CCABD2BFFAEF10079DCAF /* HardwareDecodeCapabilities.swift in Sources */ = {isa = PBXBuildFile; fileRef = 519CCABC2BFFAEF10079DCAF /* HardwareDecodeCapabilities.swift */; };
 		51AC1CAB2A9FBF3700DF7079 /* OpenSubClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = 51AC1CAA2A9FBF3700DF7079 /* OpenSubClient.swift */; };
+		51BDDFB3301E97EE00E249CE /* CommonWindow.swift in Sources */ = {isa = PBXBuildFile; fileRef = 51BDDFB2301E97EE00E249CE /* CommonWindow.swift */; };
 		51C1BA3A291CA76700C1208A /* InfoDictionary.swift in Sources */ = {isa = PBXBuildFile; fileRef = 51C1BA39291CA76700C1208A /* InfoDictionary.swift */; };
 		51C889D52C42FF7F007B2584 /* PlayerState.swift in Sources */ = {isa = PBXBuildFile; fileRef = 51C889D42C42FF7F007B2584 /* PlayerState.swift */; };
 		51CA83E22E35AD6300A44B22 /* MPVAudioDevice.swift in Sources */ = {isa = PBXBuildFile; fileRef = 51CA83E12E35AD6300A44B22 /* MPVAudioDevice.swift */; };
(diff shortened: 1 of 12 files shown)


low-batt:
PR #6193 also fixes issue #6192, however the fix in that PR is intentionally limited to the main window. This PR is a  replacement for #6193 and provides a comprehensive fix. I will keep PR #6193 open as a draft in case we decide the limited fix is preferred.

### The Regression

As discussed in issue #6192, if the macOS primary language is set to a [left-to-right](https://en.wikipedia.org/wiki/Left-to-right_language#Directionality_and_orientation) language and IINA is set to a right-to-left language then window controls will overlap the top of the sidebar:

broken

This is a regression and must be fixed.

Notice the title bar layout left space on the right side for the standard window buttons.

### Intentional Apple Behavior?

From the Apple documentation for the [NSWindow.windowTitlebarLayoutDirection](https://developer.apple.com/documentation/appkit/nswindow/windowtitlebarlayoutdirection) property:
> ## windowTitlebarLayoutDirection
> **The direction the window’s title bar lays text out, either left to right or right to left.**
> 
> macOS 10.12+
> ```text
> var windowTitlebarLayoutDirection: > NSUserInterfaceLayoutDirection { get }
> ```
> 
> ## Discussion
>
> The layout direction of the window title bar includes the standard window buttons (close, minimize, maximize) and the title for the window. In general, this returns [NSUserInterfaceLayoutDirection.rightToLeft](https://developer.apple.com/documentation/appkit/nsuserinterfacelayoutdirection/righttoleft) if the primary system language is right to left. The layout direction may be right to left even in applications that don’t have a right to left language localization. Refer to this value if the application uses [titlebarAppearsTransparent](https://developer.apple.com/documentation/appkit/nswindow/titlebarappearstransparent) and places controls under the title bar.

Always questionable when a specification uses the word _may_. As a test, I added this extension:
<details>
<summary><b>NSUserInterfaceLayoutDirection extension:</b></summary>

```swift
#if DEBUG
extension NSUserInterfaceLayoutDirection: @retroactive CustomStringConvertible {
  public var description: String {
    switch self {
    case .leftToRight: return "leftToRight"
    case .rightToLeft: return "rightToLeft"
    @unknown default: return String(self.rawValue)
    }
  }
}
#endif
```
</details>

And this code to `MainWindowController.windowDidLoad`:
```swift
    log("window.windowTitlebarLayoutDirection: \(window.windowTitlebarLayoutDirection)")
    log("""
      standardWindowButtons[0].superview.userInterfaceLayoutDirection: \
      \(standardWindowButtons[0].superview!.userInterfaceLayoutDirection)
      """)
```

Which confirms `windowTitlebarLayoutDirection` does not match the direction of the app views:
```text
15:15:33.416 [window0][d] window.windowTitlebarLayoutDirection: leftToRight
15:15:33.416 [window0][d] standardWindowButtons[0].superview.userInterfaceLayoutDirection: rightToLeft
```

But the value of `windowTitlebarLayoutDirection` does not make sense. As seen here when the language is set to English the title is left justified:

english

But contrary to the `windowTitlebarLayoutDirection` value the title is right justified in the previous screenshot. It is only the standard window controls that are left justified. This AppKit behavior looks broken.

### Fixing

The fix overrides `windowTitlebarLayoutDirection` so it matches the view:
```swift
  override var windowTitlebarLayoutDirection: NSUserInterfaceLayoutDirection {
    guard !usesUnifiedToolbar, let button = standardWindowButton(.closeButton),
          let view = button.superview else {
      return super.windowTitlebarLayoutDirection
    }
    return view.userInterfaceLayoutDirection
  }
```

### Problem with the Fix

Apparently  the AppKit `NSWindow` implementation does not use the value of the `windowTitlebarLayoutDirection` property when laying out the title bar if the window uses a [toolbar](https://developer.apple.com/documentation/appkit/nstoolbar) with the [unified](https://developer.apple.com/documentation/appkit/nswindow/toolbarstyle-swift.enum/unified) style where the toolbar appears next to the window title:

fix-failure

It is as if the AppKit `NSWindow` implementation is by-passing `windowTitlebarLayoutDirection` and directly using the primary system language direction. So the fix can't be used for the `Inspector`, `Log Viewer` and `Playback History` windows. This is acceptable as nothing overlaps in these windows:

inspector

logger

history

Since we don't know if plugin windows will use a toolbar the fix can't be applied to those windows either.

And of course the [Sparkle](https://sparkle-project.org/) window cannot be fixed:

sparkle

### Fixed Windows

Music Mode:

mini

Settings:

settings

Welcome:

welcome

About:

about

Audio Filters:

audio-filters

Video Filters:

video-filters

low-batt:
Rebased with `develop`.

uiryuu:
I think https://github.com/iina/iina/issues/6192 can be deleted if this PR works well.

PR #6236, opened 2026-08-05
Fix video staying at PiP size after entering full screen from PiP

### LINKED ISSUE #6177: Entering full screen mode from PiP broken
### IINA Version

develop @ 3265b40

### Release Channel

Other

### macOS Version

macOS 15.7.4

### Steps to Reproduce

- Start IINA built from the current `develop` branch
- Open a video and allow it to start playing
- Click on the OSC button to enter PiP
- IINA enters PiP
- Double click on the PiP window to enter full screen mode
- IINA enters full screen mode
- The video retains the size of the PiP window and plays in the bottom left corner:

Image

### Expected Behavior

The video does not retain the size of the PiP window after entering full screen mode.

### Actual Behavior

As seen in the screenshot above the video fails to fill the full screen after exiting PiP.

This problem was introduced by commit 799c48a.

_No response_

### Logs

[iina.log](https://github.com/user-attachments/files/29665527/iina.log)

[mpv.log](https://github.com/user-attachments/files/29665529/mpv.log)

### Sample Files

I was playing a downloaded copy of the YouTube video [Big Buck Bunny 60fps 4K - Official Blender Foundation Short Film](https://www.youtube.com/watch?v=aqz-KE-bpKQ).

Every time.

### Required Actions

### Optional Actions

hkarmoush:
## Fixes #6177

Entering full screen mode directly from Picture-in-Picture left the video rendered at the PiP window's small size instead of resizing to fill the full screen window.

### Root cause

`addVideoViewToWindow()` re-adds `videoView` to `videoViewContainer` on every PiP-exit path (`doneExitingPIP()`, mini-player exit) but never re-established its Auto Layout constraints. Previously, constraints were only set up once in `windowDidLoad()` via a separate `setupVideoViewConstraints()` call, so any later call to `addVideoViewToWindow()` left the re-added view unconstrained and stuck at its last size.

### Fix

Move the `setupVideoViewConstraints()` call into `addVideoViewToWindow()` itself, right after the view is re-added, so constraints are (re-)established at every call site — including the two PiP-exit paths that previously skipped it — instead of only at initial window load.

### Testing

- Reproduced the bug on `develop`, confirmed the fix resolves it: PiP → double-click to enter full screen → video now correctly fills the full screen window.
- Verified the mini-player exit path (which also calls `addVideoViewToWindow()`) still behaves correctly.
- `xcodebuild -project iina.xcodeproj ONLY_ACTIVE_ARCH=NO -scheme iina -configuration Nightly` builds clean with no new warnings.


low-batt:
Thank you very much for investigating #6177 and posting this fix with the excellent analysis of what is going wrong. I pulled the PR, built IINA, tested and confirmed entering full screen mode from PiP works correctly with the changes.

As `addVideoViewToWindow` is small and `setupVideoViewConstraints` is only called by it, we might want to just move the  `setupVideoViewConstraints` code into `addVideoViewToWindow`.

What troubles me is that the code in `setupVideoViewConstraints` was intentionally kept separate. I need to hear from @lhc70000 whether there is some reason for that which I'm not spotting.

hkarmoush:
Thanks for testing and confirming this @low-batt!

Dug into the history a bit — `setupVideoViewConstraints` used to have a second caller back when it was split out (the `edgeToEdgeVideo` pref-change handler called it directly), but that responsibility has since moved to `setupVideoContainerConstraints`. So today `addVideoViewToWindow` really is the only caller. Happy to inline it if @lhc70000 confirms there's no other reason it's kept separate.

low-batt:
Thank you for looking into the history. Sounds to me like the contents of `setupVideoViewConstraints` should be moved into `addVideoViewToWindow`. I suggest going ahead and making that change. If you keep that as a separate commit to this PR it will be easy to back out should @lhc70000 for some reason want to retain `setupVideoViewConstraints`.

hkarmoush:
Done — pushed as a separate commit (846a366b) so it's easy to drop if @lhc70000 wants `setupVideoViewConstraints` kept around.

low-batt (approved):
Pulled latest commit, built IINA, confirmed problem is fixed.

Changes look good to me.

uiryuu:
The bug fix has already been included in a older branch (modular-window-4), so after resolve the conflicts, this PR reduced to a mere cosmetic change. But thanks for the PR and your effort anyway!

PR #6245, opened 2026-08-11
Fix plugin URL parsing without file-system permission

### LINKED ISSUE #6230: Plugin core.open(HTTPS URL) crashes IINA without file-system permission
### IINA Version

IINA 1.4.4 (168), ARM64

### Release Channel

Release

### macOS Version

26.5.2 (25F84)

### Steps to Reproduce

Use a plugin without `file-system` permission:

```json
{
"permissions": ["network-request"],
"allowedDomains": ["example.com"]
}
```

Then call:

```js
iina.core.open("https://example.com/video.mp4");
```

### Expected Behavior

The URL should open, or the API should return a JavaScript permission error.

### Actual Behavior

Crash.

[`core.open`](https://github.com/iina/iina/blob/v1.4.4/iina/JavascriptAPICore.swift#L53-L58) calls `parsePath(..., forceLocalPath: false)`, but [`parsePath`](https://github.com/iina/iina/blob/v1.4.4/iina/JavascriptAPI.swift#L58-L98) still requires `accessFileSystem` for ordinary URLs and force-unwraps the denied result.

_No response_

### Logs

_No response_

### Sample Files

_No response_

Always

### Required Actions

### Optional Actions

EhsanAzish80:
## Description

Fix `parsePath` so non-file URLs can be opened when `forceLocalPath` is false, without requiring the plugin's `file-system` permission.

Local paths and `file:` URLs remain permission-gated. Permission denial now returns a nil path instead of force-unwrapping the result and crashing.

Fixes #6230

## Testing

- Full Nightly configuration build succeeded.
- Verified HTTPS and RTSP URLs bypass file-system permission.
- Verified absolute paths, tilde paths, and `file:` URLs remain protected.
- Used a temporary macOS 12 deployment-target override because the installed Xcode 27 beta no longer accepts IINA's macOS 11 target.

AI tooling assisted with code analysis and build verification. I reviewed and validated the final change.

### DIFF
--- a/iina/JavascriptAPI.swift
+++ b/iina/JavascriptAPI.swift
@@ -74,6 +74,10 @@ class JavascriptAPI: NSObject {
       }
     }
 
+    if !forceLocalPath, let scheme = URL(string: path)?.scheme, scheme != "file" {
+      return (path, false)
+    }
+
     return whenPermitted(to: .accessFileSystem) {
       var absPath = path
       if let player = player, path.hasPrefix("@current/") {
@@ -95,7 +99,7 @@ class JavascriptAPI: NSObject {
         absPath.hasPrefix(pluginInstance.plugin.dataURL.path) ||
         absPath.hasPrefix(pluginInstance.plugin.tmpURL.path)
       )
-    }!
+    } ?? (nil, false)
   }
 
   private func trackPath(_ path: String, type: MPVTrack.TrackType) -> String? {


lhc70000:
Thanks for the PR. 

I don't think the scheme checking should be in `parsePath`, since it should only handle local paths and it's called from many other places with such an assumption. The check should be in `core.open` instead. But I also found several bigger issues when investigating this, so I ended up putting them all in #6372. I have added your credit in that PR, and let's move the conversation to there.

PR #6253, opened 2026-08-14
Fix settings window appearance not following selected theme

yorkyang2333:
- [ ] Related issue: #
---

**Description:**

Apply the saved appearance preference to the new settings window and observe changes to update it immediately. Selecting System clears the explicit appearance so the window follows macOS.

### DIFF
--- a/iina/SettingsWindow.swift
+++ b/iina/SettingsWindow.swift
@@ -62,6 +62,7 @@ class SettingsWindow: NSWindow {
   private var currentCompletionResults: [SettingsSearch.Entry] = []
 
   private var pendingHighlightItem: (id: Int, parentId: Int?)?
+  private let observedPrefKeys: [Preference.Key] = [.themeMaterial]
 
   init(_ pages: [SettingsPage]) {
     self.pages = pages
@@ -77,6 +78,7 @@ class SettingsWindow: NSWindow {
                backing: .buffered, defer: false)
 
     self.isReleasedWhenClosed = false
+    updateAppearance()
 
     let splitViewController = NSSplitViewController()
     self.contentViewController = splitViewController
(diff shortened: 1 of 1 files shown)


uiryuu:
The "Appearance" option in the settings meant to only change the appearance of the player (and the initial window), but not other windows (including settings window, log window, and inspector). We want those window to always follow the system setting, or just stay dark (the inspector).

Maybe we want to change the wording or change this behavior in the future. @lhc70000

yorkyang2333:
> The "Appearance" option in the settings meant to only change the appearance of the player (and the initial window), but not other windows (including settings window, log window, and inspector). We want those window to always follow the system setting, or just stay dark (the inspector).

For me this seems really misleading as the "Appearance" option in most other apps is generally acknowledged to **affect the entire app**, not just the player and initial window. Having the settings, log, and other windows behave differently makes the option feel unintuitive. I think this requires some reconsideration.

lhc70000:
Ideally a native application should always follow the system's appearance. When a user sets their system to, say, light, they will expect normal windows with normal controls to be light. This includes the about window, log viewer, filter lists, dialogs, etc. I don't see a reason why a user would expect seeing these windows in dark specifically in IINA. However, the player window has many overlay views on the video, so it's reasonable to prefer them to be dark in this case. It's an exception for the video window only (and extends to the initial window).

This setting, in IINA's early versions, controlled the material of the visual effect views in the player window and had more options such as ultra dark and medium light. We had many specific logic on how to change the materials for each of those views. Other windows were not affected simply because they don't use visual effect views. Now with those materials deprecated, only dark and light remained. On the other hand, in newer apps the appearance setting usually control the whole app because setting the `NSApp.appearance` is a one-liner, which doesn't look elegant to me in some cases.

yorkyang2333:
> Ideally a native application should always follow the system's appearance. When a user sets their system to, say, light, they will expect normal windows with normal controls to be light. This includes the about window, log viewer, filter lists, dialogs, etc. I don't see a reason why a user would expect seeing these windows in dark specifically in IINA. However, the player window has many overlay views on the video, so it's reasonable to prefer them to be dark in this case. It's an exception for the video window only (and extends to the initial window).
> 
> This setting, in IINA's early versions, controlled the material of the visual effect views in the player window and had more options such as ultra dark and medium light. We had many specific logic on how to change the materials for each of those views. Other windows were not affected simply because they don't use visual effect views. Now with those materials deprecated, only dark and light remained. On the other hand, in newer apps the appearance setting usually control the whole app because setting the `NSApp.appearance` is a one-liner, which doesn't look elegant to me in some cases.

@lhc70000 Based on your opinion, closing this PR for now.

PR #6254, opened 2026-08-14
Fixed manually installing plugins brings up the old settings page

yorkyang2333:
- [ ] Related issue: #
---
---

**Description:**

Fixed an issue where double clicking a `.iinaplgz` file to install a plugin brings up the old settings page.

### DIFF
--- a/iina/SettingsPagePlugin.swift
+++ b/iina/SettingsPagePlugin.swift
@@ -36,6 +36,7 @@ class SettingsPagePlugin: SettingsPage {
   fileprivate lazy var installView: PluginInstallView = .init(page: self)
   fileprivate lazy var listView: PluginListView = .init(page: self)
   fileprivate lazy var updateView: PluginUpdateView = .init(page: self)
+  fileprivate lazy var pluginManager = PluginManager(window: SettingsWindow.default)
 
   override func content() -> [SettingsSection] {
     return sections {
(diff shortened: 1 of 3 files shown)


iina/SettingsWindow.swift
```diff
@@ -651,6 +651,14 @@ extension SettingsWindow {
       sidebarList.selectRowIndexes(IndexSet(integer: selectIdx), byExtendingSelection: false)
     }
   }
+
+  func installPlugin(localPackageURL url: URL) {
```
uiryuu, line 652: I think this is a bad place to expose such an API. Consider moving it to `PluginManager`, and navigate the settings window to the plugin page from there.

iina/SettingsPagePlugin.swift
```diff
@@ -62,6 +62,14 @@ class SettingsPagePlugin: SettingsPage {
       }
     }
   }
+
+  func installPlugin(localPackageURL url: URL) {
```
uiryuu, line 67: There's another similar function in this file `@IBAction func installPluginFromLocalPackage(_ sender: Any) {`. It makes sense to call this function from there (to reduce code duplication).

iina/SettingsPagePlugin.swift
```diff
@@ -62,6 +62,14 @@ class SettingsPagePlugin: SettingsPage {
...
+    Task { @MainActor in
+      let pluginManager = PluginManager(window: SettingsWindow.default)
```
uiryuu: I don't think its appropriate to instantiate another `PluginManager`. Maybe consider moving the `PluginManager` from `PluginInstallView` to `SettingsPagePlugin` and reference the manager in settings page from the install view. Just my 2¢, your AI might have a better idea :)

uiryuu (changes requested):
Good catch! See my review below

iina/SettingsPagePlugin.swift
yorkyang2333: > Just my 2¢, your AI might have a better idea :)

Seems Gemini 3.7 Flash in Antigravity still didn't really liftoff :(

yorkyang2333:
> Good catch! See my review below

Should be all done now

uiryuu:
Thanks!

Issue #3153, opened 2020-10-14
Show frames instead of milliseconds


GHubbler:
For Audio it might be ok to have precise time display with milliseconds.
For movie files it would be more practicable to have the current frame displayed after the seconds, like in video editing programs usable: h:mm:ss:CurrentFrame.

theonlygusti:
How did you get it show milliseconds?

GHubbler:
Right mouse click one of the time-counters in the controls. 😉 
![MS_IINA](https://user-images.githubusercontent.com/10322831/104379054-32d87a00-5529-11eb-9dca-a90a21278e1c.png)

wantg:
WF........
How did you know that

patg75:
> For Audio it might be ok to have precise time display with milliseconds. For movie files it would be more practicable to have the current frame displayed after the seconds, like in video editing programs usable: h:mm:ss:CurrentFrame.

This! I'd love to have **both** frame count **AND** timecode h:mm:ss:ff as display options. So inconvenient to have to open clips in a second app to find out the TC or current frame.

To add to this: it would be perfect if saving a still frame could optionally include TC or frame as an appendix in the file name instead of the counter number.

ddofborg:
+1 for frame an option. Or can we do this already?

Globolo001:
Agreed. Please do this.

Preferably also with a seconds:framenumber

Milliseconds are practically meaningless with inherently quantized timeframes.

pindab0ter:
I would also find this very useful.

yorkyang2333:
I've opened PR #6256 in order to implement this, please check it out!

patg75:
Sorry, this is a long one:

I tested this with a NTSC file (24fps Big Buck Bunny re-encoded to `30000/1001` via ffmpeg) to check the concern about fractional frame rates over long durations.

If I understand this correctly, the frame calculation doesn't actually accumulate drift over time, because it's stateless:

```swift
frame = floor((second - floor(second)) * frameRate)
```

This recomputes the frame index from the absolute playback time on every call. There's no running frame counter being incremented across calls. So the classic NTSC drift problem (where a 30fps-assumption clock diverges from wall-clock time over hours) doesn't apply here. Each display update is an independent snapshot, not an accumulation. I verified this at various positions (e.g. seeking to `9:30.5` on a 29.97fps file correctly shows frame `15` = `round(0.5 * 30000/1001)`).

A few things worth pointing out though:

[...]

PR #6108, opened 2026-05-28
Added SMPTE

ednortheyvyse:
- [x] I have read [CONTRIBUTING.md](https://github.com/iina/iina/blob/develop/CONTRIBUTING.md)
- [x] This implements/fixes issue #3239 and #1003.

---

**Description:**

PR #6256, opened 2026-08-15
feat: add frame precision to time display

yorkyang2333:
- [x] Related issue: #3153
---
---

**Description:**

Add a “1 frame” precision option to the time display context menu. Video files display `h:mm:ss:CurrentFrame`, where `CurrentFrame` is the frame number within the current second, calculated from the active video track’s frame rate. If frame precision was selected for a video, audio playback temporarily falls back to 1 millisecond precision and restores frame precision when returning to video.

Screenshot 2026-08-15 at 19 27 15

### DIFF
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -48,6 +48,7 @@
 "osc.100ms" = "100 milliseconds";
 "osc.10ms" = "10 milliseconds";
 "osc.1ms" = "1 millisecond";
+"osc.1frame" = "1 frame";
 
 // QuickSettingViewController.swift
 "quicksetting.item_default" = "Default";
--- a/iina/DurationDisplayTextField.swift
+++ b/iina/DurationDisplayTextField.swift
@@ -9,6 +9,7 @@
 import Foundation
 
 class DurationDisplayTextField: NSTextField {
+  static let framePrecision: UInt = 4
   enum DisplayMode {
     case current
     case duration // displays the duration of the movie
@@ -20,6 +21,7 @@ class DurationDisplayTextField: NSTextField {
   var duration: VideoTime = .zero
   var current: VideoTime = .zero
   var remaining: VideoTime = .zero
+  var frameRate: Double?
 
   override init(frame frameRect: NSRect) {
     super.init(frame: .zero)
@@ -42,23 +44,24 @@ class DurationDisplayTextField: NSTextField {
   }
 
   func updateText(with duration: VideoTime, given current: VideoTime,
-                  and remaining: VideoTime) {
+                  and remaining: VideoTime, frameRate: Double? = nil) {
     self.duration = duration
     self.current = current
     self.remaining = remaining
+    self.frameRate = frameRate
     updateText()
   }
   
   private func updateText() {
-    let precision = DurationDisplayTextField.precision
+    let precision = displayPrecision
     let stringValue: String
     switch mode {
     case .current:
-      stringValue = current.stringRepresentationWithPrecision(precision)
+      stringValue = current.stringRepresentationWithPrecision(precision, frameRate: frameRate)
     case .duration:
-      stringValue = duration.stringRepresentationWithPrecision(precision)
+      stringValue = duration.stringRepresentationWithPrecision(precision, frameRate: frameRate)
     case .remaining:
-      stringValue = "-\(remaining.stringRepresentationWithPrecision(precision))"
+      stringValue = "-\(remaining.stringRepresentationWithPrecision(precision, frameRate: frameRate))"
     }
     self.stringValue = stringValue
   }
@@ -76,7 +79,7 @@ class DurationDisplayTextField: NSTextField {
   }
 
   override func rightMouseDown(with event: NSEvent) {
-    let precision = DurationDisplayTextField.precision
+    let precision = displayPrecision
     let menu = NSMenu(title: "Time label settings")
     menu.addItem(withTitle: NSLocalizedString("osc.precision", comment: "Precision"))
     ["1s", "100ms", "10ms", "1ms"].enumerated().forEach { (index, key) in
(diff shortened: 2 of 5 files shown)


Globolo001:
I got a local branch. And it's wayyy more complicated than one would expect. 

A 2h+ track at 19.956fps

How do you even attribute frames to a second?

yorkyang2333:
> I got a local branch. And it's wayyy more complicated than one would expect.
> 
> A 2h+ track at 19.956fps
> 
> How do you even attribute frames to a second?

I agree that accurately assigning frames to seconds becomes complicated for fractional frame rates. For this change I kept the scope limited: the frame field is calculated from the fractional part of the current playback time and the track’s nominal frame rate. It is meant as a simple frame-precision indicator, not a full SMPTE/drop-frame timecode implementation.

Globolo001:
Yap, that's exactly where I got to my dead-end haha. 

It works, but I have too little knowledge wether there are any edgecases at mkv container 3 hours in where a frame may be inaccurately assigned. 

Confusing topic. But been using it locally and works fine. 

But IINA really needs an overhaul at many edges rather than new features haha

yorkyang2333:
> But IINA really needs an overhaul at many edges rather than new features haha

Agreed. I've been on the nightly builds lately and hit quite a few glitches. The devs are so busy with upgrading dependancies and stuff that they don't really have time to address them :( Guess we need to wait till things stabilize.

low-batt:
Note that there is another open PR with changes in this area. See #6108.

Note that mpv provides the [estimated-frame-number](https://mpv.io/manual/stable/#command-interface-estimated-frame-number) property:
> **estimated-frame-number**
>  Number of current frame in current stream.
>
>  **Note**
>
>    This is only an estimate. (It's computed from two unreliable quantities: fps and possibly rounded timestamps.)

This brings up the question as to whether IINA should be doing its own calculation or obtaining this value from `libmpv`. If nothing else it would be interesting to see how IINA's calculation compares to mpv's.

Calls to the `libmpv` API  `mpv_get_property` can pend due to thread coordination / locking within `libmpv`. This is the reason why `PlayerCore` has a timer that updates `PlayerCore.info`:
https://github.com/iina/iina/blob/d1dca0a5101f31ad31e8aea854a15f963ea58f97/iina/PlayerCore.swift#L227

This cache allows the main thread to update the UI without the micro hangs that can occur when obtaining a property value from `libmpv`. See [refreshSyncUITimer](https://github.com/iina/iina/blob/d1dca0a5101f31ad31e8aea854a15f963ea58f97/iina/PlayerCore.swift#L2642) and [syncUI](https://github.com/iina/iina/blob/d1dca0a5101f31ad31e8aea854a15f963ea58f97/iina/PlayerCore.swift#L2787).

So if we were to decide to use the mpv property it would involve adding code to this system to obtain and cache the value. Do not take my comments on this as a need to make that change. Need to hear the opinion of the senior developers with regard to that.

On the state of `develop`…

I agree with the assessment that the Nightly build is currently filled with too many regressions. @yorkyang2333 your recent help with finding and fixing some of these is very much appreciated. For a long time the UI work was going on in a side branch. As usual with working on a branch, that was causing double work as changes added to `develop` required recoding for the branch. That was costing too much extra development time, so a lot of that has been merged into `develop`.  I believe there is still UI work going on in side branches.

The main goals for IINA 1.5.0 are:
- UI enhancements, including moving away from XIBs
- Upgrading dependencies, including moving away from [brew](https://brew.sh/)
- Stretch goal: [GPU Next](https://github.com/mpv-player/mpv/wiki/GPU-Next-vs-GPU) support
- Stretch goal: Metal support

As even with the latest `master` build mpv does not yet support clients of `libmpv` using [GPU Next](https://github.com/mpv-player/mpv/wiki/GPU-Next-vs-GPU) those stretch goals require IINA applying significant changes to sources from other projects. Not clear yet if these stretch goals will be reached or if it will be deemed that work on them is taking too long and has to be postponed.

Remember, I'm a _junior_ IINA developer and not always up to date on IINA's plans…

Hopefully after `develop` has been upgraded to mpv 0.41.0 and FFmpeg 9.0.1 the senior developers will have more time to review and merge PRs. Always good to have PRs outstanding and ready when that happens as usually there is a flurry of activity when their attention focuses on outstanding PRs.

patg75:
Stupidly enough, I posted my original response in the request / issues thread and not here 😂 so there's some duplication.
[**ORIGINAL POST**](https://github.com/iina/iina/issues/3153#issuecomment-5302814746)

As @low-batt  mentioned, there's the `estimated-frame-number` which I mentioned in my earlier post. But from my first tests in the last few hours, that turned out to be less reliable than I initially thought for this use case. As he already stated, mpv's docs call it

>  […] an estimate. (It's computed from two unreliable quantities: fps and possibly rounded timestamps.)

In practice it lags time-pos by 1-2 frames right after a manual seek (confirmed by testing: jump to 9:30.500 on a 29.97fps file showed frame 16 instead of 14. Using frame-by-frame frame moving to 09:30:14 showed me a jump to precision value of 09:30.437). IINA's sync timer also has no seek-debounce, so it can sample the property mid-transition.

So `estimated-frame-number` might still be a nice fit for an absolute "frame count" display (I'd say that bit of lag I found is tolerable there), but for the hh:mm:ss:FF timecode field specifically, deriving it from time-pos + frameRate (as @yorkyang2333's PR already does) seems to actually be the more reliable approach, since time-pos updates immediately on seek. 

But the ffmpeg SMPTE approach from [#6108](https://github.com/iina/iina/pull/6108) is very interesting, even though it seems they've broken their build. But there I found out about `video-frame-info/smpte-timecode` which might worth investigating.

I'll be happy to continue testing.

EDIT: Side note on mpv's `video-frame-info/smpte-timecode` (proper SMPTE support, no custom drop-frame math needed). It's only available since 0.40.0. But IINA currently bundles mpv 0.38.0 (tested the property in the inspector to be sure and it returns a value error). So sadly it's not usable yet without bumping the mpv dependency.

patg75:
@yorkyang2333 here's my current fork with the `estimated-frame-number` implementation and a context menu update.
I reverted the TC implementation back to your TC implementation because of the lag/drift I noticed when using `estimated-frame-number`. I guess for proper SMPTE we'll have to wait for mpv 0.40.0

https://github.com/patg75/iina/commit/5890657e0846b5ddb9f4ee0db676764fbabdaec0

yorkyang2333:
Thanks for pointing out that mpv already exposes estimated-frame-number. I took a closer look and compared it with the current time-pos + frame-rate calculation. For this specific hh:mm:ss:FF display, the current calculation appears to respond more reliably to manual seeks, while estimated-frame-number is better suited to an absolute frame-count display. I’ll keep this distinction in mind as the PR evolves.

Also, @low-batt thanks for the context on the current state of develop and the 1.5.0 roadmap. I’ll keep the PR open and see how things develop once the dependency upgrades settle down.

Issue #5419, opened 2025-03-19
Translate 3D videos to draggable 2D projections


Qard:
**What you want IINA to do:**

It would be nice to have an option to render 3D videos as 2D projections with a draggable viewport, similar to what web players generally do with 3D videos. Without 3D hardware to control the viewport by headset movement, it's a good alternative to at least get part of the experience.

**What IINA does currently:**

When playing a 3D movie it currently places the two full eye recordings side by side or one atop the other as the raw file itself is encoded that way.

**Why you think this should be added:**

It would make 3D content more accessible to those without VR headsets.

**Examples of other projects that have something similar:**

YouTube supports this. For example, in [this video](https://www.youtube.com/watch?v=eKumVFvGHFA) you can drag the video player area to move the viewport around.

Jeehut:
I would love to see this! Just to make it clear, the title seems a bit misleading since it's only talking about "3D" videos without mentioning that what's actually meant are "VR 180 3D" or "VR 360 3D" videos. "Just" 3D video support would mean simply showing one eye (without any draggable angle).

This requires mapping various kinds of videos to a 3D space (per eye) though and then placing a "virtual" camera in that space and calculating the pixels to be shown from the re-mapped video stream.

For example, common formats for 3D eye-mapping to support:
* Left-Right (left half is left eye, right half is right eye)
* Top-Bottom (top half is left eye, bottom half is right eye)
* Row-Interleaved (alternating rows for left/right eyes)
* Side-by-Side (squeezed/anamorphic L-R format)
* Separate streams (two separate video streams for each eye)

And for the VR projection formats:
* VR 180° (equirectangular half-sphere, popular on YouTube and Meta)
* VR 360° (full equirectangular sphere)
* VR 360° cubemap (six faces of a cube)
* Fisheye (curved lens distortion)
* Dual-fisheye (common in consumer 180° cameras)

Supporting these formats with a draggable viewport would make immersive content much more accessible to those without headsets while still preserving some of the spatial experience. It would also make reviewing recorded VR content easier without having to put on a headset for VR film makers (like me).

Issue #4272, opened 2023-03-17
Hope support for 3D videos


Darling-Lee:
I hope to increase support for 3D videos. Bino can be used as a reference(https://github.com/marlam/bino-mirror). However, Bino's support for MacOS is not friendly, and its functions are relatively lacking, such as the ability to read external subtitle files and cut the playback screen. Especially after version 2.0, it doesn't even have a playback control. But it is very good at playing 3D videos. In addition, VLC also supports interpreting other 3D videos as red-blue 3D, but VLC's hardware decoding capability is poor under MacOS, and it drops frames severely when playing high-resolution videos. I have been using IINA to play videos, but IINA has never supported 3D videos. I hope to add support for 3D videos in future versions, and be able to choose the input format type for 3D videos and render them into the desired output format type (such as rendering left-right 3D into red-blue 3D).

low-batt:
On this:
> IINA has never supported 3D videos

Similar to many open source projects IINA is layered on top of other projects. For audio/video playback IINA is using a library from the [mpv](https://mpv.io/) project. Many of the features of the mpv player are available in IINA. The mpv player uses libraries from [FFmpeg](https://ffmpeg.org/). FFmpeg video filters are available for use in IINA.

So does mpv support 3D? I found [this comment](https://github.com/mpv-player/mpv/issues/7155#issuecomment-1011278262) in an mpv issue about subtitles in 3D that says:
> I know you probably solved this already, but in case anyone else searches for this: using --vf=format:stereo-in will inform mpv of the format the 3D video uses, and subtitles will be adjusted accordingly (in your case, --vf=format:stereo-in=sbs2l; you can check all the available formats with --vf=format:stereo-in=help).

That is referring to the mpv [format](https://mpv.io/manual/stable/#video-filters-format) filter. The documentation for [stereo-in](https://mpv.io/manual/stable/#video-filters-%3Cstereo-in%3E) says:
> Set the stereo mode the video is assumed to be encoded in. Use `--vf=format:stereo-in=help` to list all available modes. Check with the stereo3d filter documentation to see what the names mean.

Running that help command:
<details>
<summary><b>Help Output:</b></summary>

```bash
low-batt@gag ~$ mpv --vf=format:stereo-in=help
Valid values for option stereo-in are:
    no
    mono
    sbs2l
    ab2r
    ab2l
    checkr
    checkl
    irr
    irl
    icr
    icl
    arcc
    sbs2r
    agmc
    al
    ar
[...]

Issue #1620, opened 2018-04-11
Is it able to play 3D(Half-OU, SBS...) videos as 2D?


Kroderia:
There are some 3D videos, in Half-OU, SBS and others. Is it able to make it play as 2D(cut down half of the screen and resize)? Or, would you add support for that?

yoavapi:
one+ year later - any support for 3d movies on 2d displays?

yangws:
Using a custom video filter can play 3D video file as 2D image:
1. create a custom ffmpeg video filter
2. set filter name as "stereo3d"
3. set filter value as valid stereo3d option which can found here: https://ffmpeg.org/ffmpeg-filters.html#stereo3d . eg: "sbsl:ml"

PR #6263, opened 2026-08-17
docs: add VR2D to the Community Plugins list

fetzu:
- [x] Related issues: #4272, #1620, #5419 
---
  - [x] The code is fully written by AI / ~~I am an AI agent~~.
---

**Description:**

Adds VR2D (`fetzu/iina-plugin-vr2d`) to the Community Plugins list.

It plays 3D VR video — 180°/360° equirectangular, equi-angular cubemap and 180–220° fisheye, side-by-side or over-under — as a flat rectilinear view you can pan and zoom, working out how the source is packed from file naming conventions, the container's stereo flag and the frame geometry. Reprojection is a `crop` + `v360` filter chain, so it uses only what mpv already ships with.

README only: one line, placed after Trakt Scrobbler to keep the section alphabetical.

**AI usage:** this one-line change and the plugin itself were written with Claude. The plugin lives in its own repository under the MIT licence and vendors no third-party code, so nothing here affects IINA's GPLv3 licensing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ IINA is always looking for contributions, whether it's through bug reports, code
 - **[recorder](https://github.com/5thDimensionalVader/recorder-iina)** (`5thDimensionalVader/recorder-iina`) - to clip a video using ffmpeg.
 - **[Skip Intro](https://github.com/pparanoiidd/iina-skip-intro)** (`pparanoiidd/iina-skip-intro`) - Detect and skip intros, recaps and credits.
 - **[Trakt Scrobbler](https://github.com/i3p9/iina-trakt-scrobbler)** (`i3p9/iina-trakt-scrobbler`) - Trakt.tv scrobbler plugin for IINA.
+- **[VR2D](https://github.com/fetzu/iina-plugin-vr2d)** (`fetzu/iina-plugin-vr2d`) - Watch 3D VR videos (180°/360°, side-by-side or over-under) flat, with pan, zoom and automatic detection.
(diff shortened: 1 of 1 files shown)


fetzu:
Note/Heads up: I have built the same thing into IINA itself — design proposal
#6264, and a draft PR at #6265.

Which raises the obvious question of whether this one still makes sense. I think
it does, at least for now: the plugin works on released IINA today, needs no
fork and no build, and the native version may well not land (or not for a
while). If/when it does land we can just drop this entry altogether.

low-batt (approved):
Pulled PR, inspected README, confirmed link works.

Changes look good to me.

low-batt:
@fetzu Hello from a _junior_ IINA developer.

On this:
> If/when it does land we can just drop this entry altogether.

Understood. Fine to remove the plugin later if/when a native version lands.

This caught us changing how community plugins are handled. For now updating the list in the README.md is needed because that is how the released version of IINA handles things. When someone updates the README we are now also updating a JSON file. The current development version of IINA reads this file from GitHub and will allow users to install a community plugin using IINA settings. I have added VR2D to the JSON file:
https://github.com/iina/iina/blob/4eb4ddaa18c6f5e02a40944467e09f0205c62f0b/plugins.json#L128-L133

Hopefully I got it right. This is what it looks like in the development version of IINA:
settings

fetzu:
Looks good, thank you very much for merging and for the JSON correction. Cheers !

PR #6269, opened 2026-08-19
Fix mini player volume slider dragging

yorkyang2333:
- [ ] Related issue: #
---

**Description:**

Restore the explicit `mouseDown` pass-through for `VolumeSlider`, so AppKit keeps native slider tracking active. This prevents the mini player window background from handling volume-slider drags when album art is hidden.

### DIFF
--- a/iina/VolumeSlider.swift
+++ b/iina/VolumeSlider.swift
@@ -22,6 +22,14 @@ class VolumeSlider: NSSlider {
   
   // MARK: - Mouse / Trackpad events
 
+  /// Keep AppKit's slider tracking path active on macOS versions affected by IINA issue #5768.
+  ///
+  /// Without this explicit override, AppKit may skip mouse tracking for NSSlider subclasses,
+  /// allowing a movable window background to handle the drag instead.
+  override func mouseDown(with event: NSEvent) {
+    super.mouseDown(with: event)
+  }
+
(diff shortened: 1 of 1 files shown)


uiryuu:
Can you explain how did you trigger this bug?

While trying to verify the fix works, I tried to scroll in this part of the mini player:

Untitled

and the volume (and progress slider) are still responsive to the trackpad scroll. Can you investigate the reason and also fix it in this PR?

yorkyang2333:
> Can you explain how did you trigger this bug?

In Music Mode (mini player), hide the album art and then try dragging the volume slider, and you'll find that you are dragging the window rather than the volume slider. (See video below)

https://github.com/user-attachments/assets/14b4bc16-d03c-4e83-9033-f77223fb7d6d

> and the volume (and progress slider) are still responsive to the trackpad scroll. Can you investigate the reason and also fix it in this PR?

Sorry, but can't reproduce this issue. This behaves normally on my side.

yorkyang2333:
I've found this isn't the right fix of this issue. I'll dive into it later and, if cannot identify the root cause, I'll turn it into an bug report in issues instead.

uiryuu:
Yeah, just wanted to tell you that this fix didn't solve the problem. I'll investigate the other one I mentioned myself.

PR #6277, opened 2026-08-21
Fix multiple problems with cache settings, #6188

### LINKED ISSUE #6188: Multiple problems with cache settings
### IINA Version

develop @ commit c5dccbc

### Release Channel

Other

### macOS Version

macOS 15.7.4

### Steps to Reproduce

- Build IINA from the `develop` branch
- Use the `defaults` command to set `enableNewSettings` to `true`
- Start IINA
- The _new_ settings panel appears
- On the left side of the panel click on `Network`
- In the `Cache` section look for the problems described below 

### Expected Behavior

The cache settings can be used to adjust IINA's behavior.

### Actual Behavior

Multiple problems are visible in this screenshot:

Image

---

On the `Enable cache` setting…

Looking over the macOS `System Settings` I don't see the word `Enable` anywhere. To match Apple's style this probably should be `Use cache`.

This is tied to the mpv [--cache](https://mpv.io/manual/stable/#options-cache) option which is not just on/off:
> **--cache=<yes|no|auto>**
> Decide whether to use network cache settings (default: auto).

Should we be supporting `auto`?

When you turn off `Enable cache` the other settings are not shown as disabled. Should some of these settings be subordinate settings and disabled when not using the cache?

----

On the `Seconds to prefetch` setting…

This is tied to the mpv [--cache-secs](https://mpv.io/manual/stable/#options-cache-secs) option:
> **--cache-secs=\<seconds>**
 >   How many seconds of audio/video to prefetch if the cache is active. This overrides the `--demuxer-readahead-secs` option if and only if the cache is enabled and the value is larger. The default value is set to something very high, so the actually achieved readahead will usually be limited by the value of the `--demuxer-max-bytes` option. Setting this option is usually only useful for limiting readahead.

As can be seen in the screenshot above the value is shown as `36,00`. It is not that the comma is in the wrong place, the last `0` is cut off. Increasing the value by one causes it to fit.

With values this large are the buttons that increment by one useful?

This: `Default 36000` should use a coma just like the value field does.

That default does not match up with mpv's default of 3,600,000:

```text
low-batt@gag ~$ mpv --list-options | grep cache-secs
 --cache-secs                     Double (0 to any) (default: 3600000)
```

The mpv [--demuxer-readahead-secs](https://mpv.io/manual/stable/#options-demuxer-readahead-secs) option that `--cache-secs` overrides:
> **--demuxer-readahead-secs=\<seconds>**
> If `--demuxer-thread` is enabled, this controls how much the demuxer should buffer ahead in seconds (default: 1). As long as no packet has a timestamp difference higher than the readahead amount relative to the last packet returned to the decoder, the demuxer keeps reading.
>
>    Note that enabling the cache (such as `--cache=yes`, or if the input is considered a network stream, and `--cache=auto` is used), this option is mostly ignored. (`--cache-secs` will override this. Technically, the maximum of both options is used.)
>
>    The main purpose of this option is to limit the readhead for local playback, since a large readahead value is not overly useful in this case.
>
>    (This value tends to be fuzzy, because many file formats don't store linear timestamps.)

The mpv default value matches the mpv documentation:
```text
low-batt@gag ~$ mpv --list-options | grep demuxer-read
 --demuxer-readahead-secs         Double (0 to any) (default: 1)
```

---

On the `Seconds to prefetch` setting…

In the new code this setting is bound to the IINA `cacheBufferSize` preference:
```swift
        SettingsItem.Input()
          .bindTo(.cacheBufferSize)
          .hasDescription()
```

Searching IINA sources, that preference is **NOT USED** by the backend.

In the old settings window that setting is bound to the IINA `defaultCacheSize` preference. That IINA preference is used by `MPVController`:
```swift
    setUserOption(PK.defaultCacheSize, type: .other, forName: MPVOption.Demuxer.demuxerMaxBytes,
                  verboseIfDefault: true) { key in
      return "\(Preference.integer(for: key))KiB"
    }
```

So this setting looks to be **broken** in the new settings window.

This is supposed to be tied to the mpv [--demuxer-max-bytes](https://mpv.io/manual/stable/#options-demuxer-max-bytes) option:
> **--demuxer-max-bytes=\<bytesize>**
>    This controls how much the demuxer is allowed to buffer ahead. The demuxer will normally try to read ahead as much as necessary, or as much is requested with `--demuxer-readahead-secs`. The option can be used to restrict the maximum readahead. This limits excessive readahead in case of broken files or desynced playback. The demuxer will stop reading additional packets as soon as one of the limits is reached. (The limits still can be slightly overstepped due to technical reasons.)
>
>    Set these limits higher if you get a packet queue overflow warning, and you think normal playback would be possible with a larger packet queue.
>
>  See `--list-options` for defaults and value range. \<bytesize> options accept suffixes such as KiB and MiB.

As seen in the code from `MPVController` and the mpv documentation, the units displayed should be `KiB` not `KB`.

The IINA default is close to the mpv, but doesn't match:
```text
low-batt@gag ~$ mpv --list-options | grep demuxer-max-bytes
 --demuxer-max-bytes              ByteSize (0 to 4.6116860184274e+18) (default: 150.000 MiB)
```

----

Shouldn't the `Show throbber when seeking` setting have an icon like the `Show throbber when paused waiting for the cache`?

----

**MISSING**

When I try and play YouTube videos I have trouble. The video starts to play and immediately pauses to fill the cache. Then it starts playing and again ends up pausing to fill the cache. I'm about to test these mpv options to see if they can be used to improve the behavior.

From the mpv manual entry for the [--cache-pause-initial](https://mpv.io/manual/stable/#options-cache-pause-initial) option:
> **--cache-pause-initial=<yes|no>**
> Enter "buffering" mode before starting playback (default: no). This can be used to ensure playback starts smoothly, in exchange for waiting some time to prefetch network data (as controlled by `--cache-pause-wait`). For example, some common behavior is that playback starts, but network caches immediately underrun when trying to decode more data as playback progresses.
>
>    Another thing that can happen is that the network prefetching is so CPU demanding (due to demuxing in the background) that playback drops frames at first. In these cases, it helps enabling this option, and setting `--cache-secs` and `--cache-pause-wait` to roughly the same value.
>
>    This option also triggers when playback is restarted after seeking.

That option sounds like it is intended to address exactly the problem I'm encountering.

From the mpv manual entry for the [--cache-pause-wait](https://mpv.io/manual/stable/#options-cache-pause-wait) option:
> **--cache-pause-wait=\<seconds>**
>    Number of seconds the packet cache should have buffered before starting playback again if "buffering" was entered (default: 1). This can be used to control how long the player rebuffers if `--cache-pause` is enabled, and the demuxer underruns. If the given time is higher than the maximum set with `--cache-secs or` `--demuxer-readahead-secs`, or prefetching ends before that for some other reason (like file end or maximum configured cache size reached), playback resumes earlier.

I don't know yet if `--cache-pause-initial` will be enough or if I will have to increase the `--cache-pause-wait` setting.

These seem like useful settings for IINA to support.

_No response_

### Logs

_No response_

### Sample Files

_No response_

Every time.

### Required Actions

### Optional Actions

low-batt:
This commit will:
- Change `SettingsPageNetwork.sectionCache` to use the `defaultCacheSize` key, which the old settings used instead of the duplicate key `cacheBufferSize` that is not supported by `MPVController`
- Remove the duplicate `Preference.cacheBufferSize` key
- Correct `MPVController.mpvInit` to set the mpv `cache` option to `yes` when the IINA `enableCache` setting is set to `true`
- Add support for the mpv `cache-pause-initial` and `cache-pause-wait` options
- Add help links for the cache settings to `SettingsPageNetwork.sectionCache`
- Add descriptions for the two new settings
- Change the cache settings to be subordinate to the `enableCache` setting

This commit will fix an existing problem where IINA will only set the mpv `cache` option to "no". Enabling the cache in the old settings or the new settings would not change the option value to "yes".

This commit will fix a new problem in the new settings window where changing the `Cache buffer size` setting did not change the value of the mpv `demuxer-max-bytes` option. This was because the new settings window used the `cacheBufferSize` key which was a duplicate of the `defaultCacheSize` used by the old settings window and supported by `MPVController`.

This commit will add support for the mpv `cache-pause-initial` and `cache-pause-wait` options. These options are very useful when streaming. With them you can configure IINA to not start playback when streaming until the configurable number of sections has been cached. This can be used to prevent entering and exiting buffering mode multiple times when first starting to play a stream.

- [x] Related issue: #6188
---
---

**Description:**


low-batt:
With the changes in this commit the cache settings will look like this:

updated

The four settings are subordinate to the setting that enables or disabled the cache. That setting is tied to the mpv [cache](https://mpv.io/manual/stable/#options-cache) setting. The subordinate settings are tied to these mpv options:
- [cache-pause-initial](https://mpv.io/manual/stable/#options-cache-pause-initial)
- [cache-pause-wait](https://mpv.io/manual/stable/#options-cache-pause-wait)
- [cache-secs](https://mpv.io/manual/stable/#options-cache-secs)
- [demuxer-max-bytes](https://mpv.io/manual/stable/#options-demuxer-max-bytes)

Information about these options not contained in the documentation can be seen by running this:
```text
mpv --list-options | grep 'cache-pause-initial\|cache-pause-wait\|cache-secs\|demuxer-max-bytes' 
```

That shows:
```text
low-batt@gag ~$ mpv --list-options | grep 'cache-pause-initial\|cache-pause-wait\|cache-secs\|demuxer-max-bytes' 
 --cache-pause-initial            Flag (default: no)
 --cache-pause-wait               Float (0 to 3.4028234663853e+38) (default: 1)
 --cache-secs                     Double (0 to any) (default: 3600000)
 --demuxer-max-bytes              ByteSize (0 to 4.6116860184274e+18) (default: 150.000 MiB)
low-batt@gag ~$ 
```

The new support for the `cache-pause-initial` and `cache-pause-wait` mpv options is very useful when streaming. I used to enable IINA's `Pause` setting under the `When media is opened` disclosure triangle so that playback would not start. I would then wait for a bit to allow the cache to fill and then start playback. These settings allow that to happen automatically. I did have to raise the `cache-pause-wait` setting from mpv's default value of one second to ten seconds.

I did not make the change suggested by @svobs to change the value of `Cache buffer size` to be in MiB instead of KiB. I can look into that, but wanted to get the fixes in this commit merged.

uiryuu (approved):
LGTM

PR #6284, opened 2026-08-23
Display screenshot formats in order

low-batt:
This commit will:
- Add a `bindToSorted` method to `SettingsItem`
- Add a `Preference.ScreenshotFormat` extension to `SettingsPageGeneral` that implements comparable
- Change `SettingsPageGeneral.sectionScreenshots` to use `bindToSorted` for the screenshot format setting
- Add the missing file extension to the PNG string in `VerbatimStrings`

- [ ] Related issue: #
---
---

**Description:**

### DIFF
--- a/iina/SettingsItem.swift
+++ b/iina/SettingsItem.swift
@@ -577,6 +577,16 @@ struct SettingsItem {
       return self
     }
 
+    func bindToSorted<T>(_ key: Preference.Key, ofType t: T.Type) -> Self
+    where T: RawRepresentable & CaseIterable & InitializingFromKey & Comparable, T.RawValue == Int
+    {
+      self.key = key
+      for c in t.allCases.sorted() {
+        valueTypes.append((c.rawValue, String(describing: c)))
+      }
+      return self
+    }
+
     func bindToCustom<T>(type t: T.Type, block: @escaping (NSPopUpButton) -> Void) -> Self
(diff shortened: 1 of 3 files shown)


low-batt:
Is there a reason IINA defaults to `png` instead of `jpg` like mpv?

Currently `develop` presents the list of screenshot formats to choose from like this:

before


No idea why all the entries show the file extension except for PNG. This PR also adds the missing "(.png)".

This is how the list of screenshot formats looks with the changes in this PR:

after


Currently the formats are ordered based on the order of the enum constants in Preference.ScreenshotFormat. The convention is that settings enum classes use integers for raw values with values implicitly assigned. Only two enum classes use explicitly assigned values, ScrollAction and Theme. These classes deviate because enumeration cases were removed, which would have changed the implicitly assigned values. It is critical that the raw values do not change because those values are stored in UserDefaults.

An easier way to change the sort order of screenshot formats would be to explicitly assign the raw values and then reorder the enumeration cases. A slight disadvantage of this method is that it is not obvious to a developer the importance of the order of the enumeration cases. Another disadvantage is that it would make the ScreenshotFormat deviate from the other enums. But the more important disadvantage of this approach is that it is only applicable for values displayed verbatim, as is the case for the screenshot formats. Enum classes that display localized strings cannot use this method. Sorting on the localized string is required. So this PR establishes that concept. Sorting is not always appropriate. Some settings need a specific order that is already provided by the enum class. But there are some settings where the list presented just looks random. A task for a future PR.

PR #6290, opened 2026-08-24
Fix missing validation in new settings window, #6289

### LINKED ISSUE #6289: Missing validation in new settings window
### IINA Version

Nightly @ commit fc5a458

### Release Channel

Nightly

### macOS Version

macOS 15.7.9

### Steps to Reproduce

- Start IINA Nightly @ commit fc5a458
- The settings panel appears
- On the left side of the panel click on `Interface`
- The right side of the panel displays the `Interface` tab
- On the left side of the panel click on `On Screen Controller`
- The right side of the panel displays the `On Screen Controller` section
- Enter `asdf` for the `Auto hide after` setting
- Press return to set the value
- IINA accepts the invalid value without reporting an error
- IINA stores the invalid value in [UserDefaults](https://developer.apple.com/documentation/foundation/userdefaults):
  > low-batt@gag ~$ defaults read com.colliderli.iina controlBarAutoHideTimeout
  >asdf

Image

### Expected Behavior

The new settings window does not allow settings to be set to invalid values.

In the case shown, IINA 1.4.4 refuses to accept `asdf` for the `Auto hide after` setting:

Image

### Actual Behavior

The new settings window is missing validation for many settings that take numeric values. As shown above this allows invalid values to be stored in [UserDefaults](https://developer.apple.com/documentation/foundation/userdefaults).

I will be labeling this as a release blocker due to invalid values being stored in [UserDefaults](https://developer.apple.com/documentation/foundation/userdefaults).

I will be posting a PR to address this issue shortly.

_No response_

### Logs

_No response_

### Sample Files

_No response_

Every time.

### Required Actions

### Optional Actions

low-batt:
This commit will:
- Add an `allowsFloats` parameter to the `SettingsItem.Input.range` function
- Add an `allowsFloats` property to `SettingsItem.Input` to store the new range parameter
- Change `SettingsItem.Input.getValueViews` to set the `allowsFloats` property of the `NumberFormatter` based on the parameter passed when establishing the range
- Change `SettingsItem.Input.getValueViews` to set the `NumberFormatter.numberStyle` property to decimal to permit the use of floating point numbers (when `allowsFloats` is true)
- Add a range function to `SettingsItem.SwitchWithInput` similar to the existing function in `SettingsItem.Input`
- Add range and `allowFloats` parameters to the `UIHelper.input` function
- Change `UIHelper.input` to construct a `NumberFormatter` when a range parameter is given and assign it to the `formatter` property of the `NSTextField` being returned
- Change all text fields in the settings window that accept numbers to reject invalid values

- [x] Related issue: #6289
---
---

**Description:**

### DIFF
--- a/iina/SettingsItem.swift
+++ b/iina/SettingsItem.swift
@@ -822,6 +822,7 @@ struct SettingsItem {
     private var customBindingBlock: ((NSTextField) -> Void)?
     private var isLongText = false
     private var range: ClosedRange<Double>?
+    private var allowsFloats = false
 
     private var cachedStepperValue: Double?
     private var stepper: NSStepper?
@@ -853,8 +854,9 @@ struct SettingsItem {
       return self
     }
 
-    func range(_ range: ClosedRange<Double>) -> Self {
+    func range(_ range: ClosedRange<Double>, allowsFloats: Bool = false) -> Self {
       self.range = range
+      self.allowsFloats = allowsFloats
       return self
     }
 
@@ -875,9 +877,11 @@ struct SettingsItem {
       setControlSize(textField)
       if let range {
         let formatter = NumberFormatter()
+        formatter.allowsFloats = allowsFloats
         formatter.usesGroupingSeparator = false
         formatter.minimum = NSNumber(value: range.lowerBound)
         formatter.maximum = NSNumber(value: range.upperBound)
+        formatter.numberStyle = .decimal
         textField.formatter = formatter
       }
       let stack = NSStackView(views: [textField])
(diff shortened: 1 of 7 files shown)


uiryuu (approved):

low-batt:
I wrote this up and then forgot to post it with the PR…

----

I'm thinking this is a poor solution, but I believe this must be fixed before a beta can be released, so maybe we go with this and refactor later.

I believe we should refactor Preference so that a settings (`Key`) does not just provide the [UserDefaults](https://developer.apple.com/documentation/foundation/userdefaults) key, but other attributes such as the type of value, for numeric values the permitted range and the default value. This would allow the settings page validation to be data driven. Thoughts?

---

Since it is desirable to prevent Nightly build users from adding invalid values that get persisted in `UserDefaults`, I'm going to go ahead and merge this now. But I do think we should consider moving to a data driven solution.

Issue #6288, opened 2026-08-24
very strange bug happening >1yr: vid scrubbing preview always being displayed even when mouse is away from on screen controller


comfortablemilk4454:
### IINA Version

1.4.4

### Release Channel

Release

### macOS Version

26.6.2

### Steps to Reproduce

vid scrubbing preview always being displayed even when mouse is away from on screen controller. this has been happening since i started using iina a year ago (mostly on release versions, and it happens on latest nightly too [1.5.0 Build 167 NIGHTLY fc5a458]). it'll only happen when your mouse is already moving when you launch a video at the exact split second the vid window starts to pop up. it's hard to reproduce intentionally because the timing is so precise, but i notice it happens ~5-10% of the time i launch a vid as im pretty fidgety with my mouse and usually move it after double clicking a vid file to launch it. if you time it right though its reproducible 100% of the time.

when it happens, it doesn't go away until you either quit the app or move your mouse off of the iina vid player window (although this sometimes doesn't work, quitting has been the only surefire temporary workaround). it persists through tabbing to other apps too with cmd tab. ive found that if you cmd w or click the red skittle it will only get rid of it until you launch the video again where it always comes back 100% of the time even if you're not moving the mouse on that 2nd time until you quit. 

i was able to get a log on 1.4.4, but i haven't been able to reproduce it with logging on yet on 1.5.0 nightly. the 1.4.4 vid is of the same instance as the log. i also am attaching a vid of it happening on 1.5.0, although i don't have a log for that, and i only started the recording after it had already started happening. as such, the 1.4.4 vid is the better reference as it shows the onset of the bug.

been happening on all versions of macos 15 & 26.

[vids.zip](https://github.com/user-attachments/files/31384370/vids.zip)

### Expected Behavior

preview goes away when your mouse strays from the scrubbing bar

### Actual Behavior

it doesn't

### Logs

[iina.log](https://github.com/user-attachments/files/31383771/iina.log)
[mpv.log](https://github.com/user-attachments/files/31383772/mpv.log)

### How often can you reproduce the issue

100%

### Required Actions

- [x] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [x] I tested with the Nightly build and confirmed the issue is not fixed.
- [ ] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

low-batt:
THANK YOU! For reporting this.

[...]

PR #6292, opened 2026-08-25
Fix seek time & thumbnail preview being shown when cursor outside OSC (issue #6288)

svobs:
…play slider

- [x] Related issue: #6288
---
---

**Description:**

* Replaces `isMouseInSlider` variable with real-time check for `!playSlider.isHidden && event.inAnyOf([playSlider])`.
* Also alters `refreshSeekTimeAndThumbnail` slightly so that it hides `thumbnailPeekView` if needed. (It was already showing it if needed, as well as hiding/showing `timePreviewView` if needed.) This makes this method easier for callers to use: just call it for `mouseEntered`, `mouseExited`, and `mouseMoved`, and it will know what to do.
* Removes already dead and now-obsolete code from from `mouseEntered`, `mouseExited`.
* After all the preceding, it seems that the previous code still relied  on `mouseExited` being called for the single case of closing the window, in order to hide the two views. A more robust solution is to just hide them at actual window close - so this adds that code.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -124,9 +124,7 @@ class MainWindowController: PlayerWindowController {
   var isWindowHidden: Bool = false
   var isWindowMiniaturizedDueToPip = false
 
-  // might use another obj to handle slider?
   var isMouseInWindow: Bool = false
-  var isMouseInSlider: Bool = false
   /** flag to ignore abrupt momentum scrolls */
   private var isMomentumScrollingAllowed = false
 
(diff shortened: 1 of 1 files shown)


uiryuu:
Can you generate a build so that @comfortablemilk4454 can verify the issue is fixed?

svobs:
> Can you generate a build so that @comfortablemilk4454 can verify the issue is fixed?

Here is a [test build](https://filebin.net/adp4klw688mycyc1) to try.

(@uiryuu I'm assuming this is what you mean, unless there is an easier way to do this)

uiryuu (approved):
The changes look good to me, and I also didn't find any bug from my testing. If @comfortablemilk4454 can confirm that this fixes the issue, then it's good to go.

comfortablemilk4454:
@uiryuu seems to work for me too, although this bug is hard to reproduce on purpose. will take some monitoring over the coming days, but fingers crossed!

uiryuu:
@comfortablemilk4454 Thanks for your input. Since this has been merged into develop, you can test out the latest nightly build.

comfortablemilk4454:
great, ty!

PR #6293, opened 2026-08-25
Use pencil instead of gear for key binding edit icon

svobs:
- [ ] Related issue: #
---
---

**Description:**
I finally tried out the new Settings UI and found the new Key Bindings UI to have very good design. But I was initially confused about how to edit a key binding, because I didn't associate the gear icon with a single-line edit. I really think using a pencil icon is more appropriate? I know this could have been a "discussion" post, but seems more precise to just give feedback in the form of a PR.

(One other point of unrequested feedback: I noticed the whole page scrolls. But would be better if all the elements stayed in position except for the the bindings list (i.e., everything below the Search field and above the separator which lies above the `Display raw values` checkbox.)

### DIFF
--- a/iina/SettingsPageKeyBindings.swift
+++ b/iina/SettingsPageKeyBindings.swift
@@ -539,7 +539,7 @@ fileprivate class KeyMappingCell: NSTableCellView {
       }
 
       self.editButton = createActionButton(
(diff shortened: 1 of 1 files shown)


uiryuu (commented):
I agree. @lhc70000 is the decider

PR #6306, opened 2026-08-30
Save playback position before playlist navigation

Hlinbit:
mpv only saves watch-later data on quit or loadfile replacement.

Switching playlist entries via `playlist-next`, `playlist-prev`, or direct selection bypasses those paths, so the current entry can lose its resume position.

Save the playback position immediately before replacing the active playlist item.


low-batt:

I would prefer this be implemented as a plugin. I'm not in favor of deviating from mpv's behavior by default. If this was added to IINA I would want a setting to control this.

This particular behavior of the watch later feature has been debated a lot in [mpv issues](https://github.com/mpv-player/mpv/issues). There are other cases where users request something similar. See the mpv issues for the arguments against some of the proposed changes.

PR #6307, opened 2026-08-30
miniplayer: show OSD and let mpv render the default album art in music mode

yorkyang2333:
- [ ] Related issue: #
---
---

**Description:**

Music mode never shows the OSD. The OSD view belongs to the main window, which is hidden while the mini player is on screen, so `sendOSD` bails out early on `!isInMiniPlayer` (#6050). That leaves music mode without any feedback for volume, seeking or pausing.

While looking into this I also noticed that when the media has no cover art, the mini player stacks a `defaultAlbumArt` NSView on top of the video view, which needs manual bookkeeping (track list, subtitle visibility) and leaves mpv rendering nothing behind it.

This PR addresses both:

1. The OSD state and the display/hide logic moved from `MainWindowController` down to `PlayerWindowController`, and the mini player creates its own `OSDView`. `sendOSD`/`hideOSD` now route to whichever window is active. Everything the main window OSD supports (progress bar, accessory views, auto-hide timeout, liquid glass style) keeps working.

2. The overlay view is gone. Instead, when entering music mode with a file that has no cover art, IINA adds its default album art as an external cover art track (`video-add <file> select <title> <lang> yes`) and removes the track when leaving music mode. mpv then renders it exactly like embedded cover art, so window sizing follows the image and embedded art still takes priority when present. Files with a real video track are never touched, and main window playback outside music mode is unchanged (the track is removed on switch back and does not survive file loads anyway).

A detail worth mentioning: the flags argument of `video-add` must be `select`. With `auto`, mpv adds the track but never selects it, so nothing is displayed and the track list logic would skip re-adding it.

One side effect I think is a plus: audio files without cover art now provide artwork to the system Now Playing panel, since the default art is a regular external album art track.

Tested with cover-less audio (art shows, single `video-add` in the log), audio with embedded art (no track added) and video files (main window untouched). Built with Xcode against the current develop.

The last commit is a drive-by one-character fix (`GrouopBy` → `GroupBy`) so the typos check passes; it has been failing on develop since 3e9b2813. Happy to move it to a separate PR if preferred.

### DIFF
--- a/iina/MainWindowController.swift
+++ b/iina/MainWindowController.swift
@@ -20,7 +20,6 @@ fileprivate let isMacOS11: Bool = {
 fileprivate let InteractiveModeBottomViewHeight: CGFloat = 60
 
 fileprivate let UIAnimationDuration = 0.25
-fileprivate let OSDAnimationDuration = 0.5
 fileprivate let CropAnimationDuration = 0.2
 
 fileprivate extension NSStackView.VisibilityPriority {
@@ -60,7 +59,6 @@ class MainWindowController: PlayerWindowController {
 
   /** For auto hiding UI after a timeout. */
   var hideControlTimer: Timer?
-  var hideOSDTimer: Timer?
 
   /** For blacking out other screens. */
   var screens: [NSScreen] = []
@@ -93,7 +91,6 @@ class MainWindowController: PlayerWindowController {
   var oscToolbarView: NSStackView!
   var oscSliderView: NSView!
 
-  var osdView: OSDView!
   var additionalInfoView: AdditionalInfoView!
   var bufferIndicatorView: BufferIndicatorView!
   var timePreviewView: TimePreviewView!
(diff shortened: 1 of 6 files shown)


low-batt:
On this:
>  The last commit is a drive-by one-character fix (GrouopBy → GroupBy) so the typos check passes; it has been failing on develop since 3e9b281. Happy to move it to a separate PR if preferred.

To be precise, it has been miss-spelled since 3e9b281, but is only now being detected by [typos](https://github.com/crate-ci/typos).

This behavior happens because we do not pin a version of `typos` and instead the CI build uses the latest version. The PR I just posted failed even though I installed `typos` on my Mac and ran it from Terminal before posting my PR. After seeing the failure I ran a `brew upgrade` which updated `typos` and sure enough now I get this error on my Mac.

Whenever this happens we fix such CI failures ASAP, so it would be helpful if you posted a PR specifically for this failure.

I don't think you will have to remove the change from this PR. Merging a fix for the miss-spelling should not cause a merge conflict as the sources will match.

yorkyang2333:
> Whenever this happens we fix such CI failures ASAP, so it would be helpful if you posted a PR specifically for this failure.

Cherry-picked to #6309.

Issue #6305, opened 2026-08-28
Many Subtitle Settings Don't Apply Correctly in 1.5.0 Nightlies


Uranosphaerite:
### IINA Version

1.5.0 (3e0f773b)

### Release Channel

Nightly

### macOS Version

27 (26A5421a)

### Steps to Reproduce

Open IINA 1.4.4 and IINA 1.5.0 Side by side. Apply the following settings and note the appearance. The override level is set to strip and the video uses ASS subtitles.

IINA 1.5.0 on top, IINA 1.4.4 on the bottom

Subtitle Shadow ON
<img width="1512" height="949" alt="Image" src="https://github.com/user-attachments/assets/81ce9715-88ae-43e0-82b2-d44c0e69e1e5" />

Subtitle Shadow ON, Background Color set
<img width="1512" height="949" alt="Image" src="https://github.com/user-attachments/assets/fc478e1a-3a3f-47b8-ab72-402124e158ab" />

Subtitle Shadow OFF, Background Color set
<img width="1512" height="949" alt="Image" src="https://github.com/user-attachments/assets/10667b09-73fd-42fd-b50c-32e77b88897a" />

Bold and italic don't change anything either, but this happens in both versions. There are many other weird behaviors with subtitles but this is the worst regression I saw. 

I believe that a lot of this behavior was caused by breaking changes in https://github.com/mpv-player/mpv/pull/14195, which was included in mpv 0.39.

### Expected Behavior

Subtitles render as intended by the settings

### Actual Behavior

Subtitles don't render in the way intended

### How often can you reproduce the issue

Every time

### Required Actions

- [x] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [x] I tested with the Nightly build and confirmed the issue is not fixed.
- [ ] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

low-batt:
Great writeup! Very glad you reported this.

On this:
> I believe that a lot of this behavior was caused by breaking changes in https://github.com/mpv-player/mpv/pull/14195, which was included in mpv 0.39.

Agree. Looks like mpv decided to make incompatible changes to align with the ASS specification.

From a quick look IINA will need to introduce some new subtitle settings and have to figure out how to migrate existing settings.

PR #6317, opened 2026-09-02
Added the new --sub-border-style option from mpv 0.39.0

uiryuu:
- [x] Related issue: #6305
---
---

**Description:**
Due to a change of interface from upstream https://github.com/mpv-player/mpv/pull/14195 (landed in 0.39.0), we have to specific `--sub-border-style` explicitly (instead of being inferred previously). 

Other important changes:
- `--sub-shadow-color` is an alias to `--sub-back-color` now. Therefore the `.subBgColorString` is removed in favor of `.subShadowColorString`.
- Since there's no background color for subs, the text label for sub color is dropped to prevent duplication.
- Updated the symbols and structure to make them more clear

@low-batt I didn't do the migration, since the migration logic is too messy and the gain is limited.

Before:
image

After:
image


iina/SettingsPageSubtitles.swift
```diff
@@ -89,17 +89,18 @@ class SettingsPageSubtitles: SettingsPage {
         SettingsItem.General(title: .text_Color)
           .image(name: "paintpalette")
           .withValueView(subtitlesColorView.view)
+        SettingsItem.PopupButton()
+          .image(name: ["character.textbox.badge.sparkles", "square.dashed.inset.filled", "square.dashed"])
+          .bindTo(.subBorderStyle, ofType: Preference.SubBorderStyle.self)
+          .withHelpLink(AppData.wikiLink.appending("/#options-sub-border-style"))
```
low-batt: This just takes me to the wiki home page?

low-batt:
No migration is fine.

It seems "shadow" is now "background", from [sub-back-color](https://mpv.io/manual/stable/#options-sub-back-color):
> **--sub-back-color=\<color>**
>    See `--sub-color`. Color used for sub text background.
>
>    `--sub-shadow-color` is an alias for `--sub-back-color`.

Except to make things confusing the default border style is `Outline and Shadow`.

And as seen in the screenshot below the subtitle side panel is using the term "Background", but the subtitles settings is still using "Shadow". Not sure what we should call it, but the settings and the side panel should use the same names.

Would be good if these settings were displayed in the same order in both places.

Shouldn't the side panel also provide the `Border Style` setting?

subtitles

uiryuu:
> And as seen in the screenshot below the subtitle side panel is using the term "Background", but the subtitles settings is still using "Shadow". Not sure what we should call it, but the settings and the side panel should use the same names.

I changed the background to shadow in the quicksettings to make things consistent.

> Would be good if these settings were displayed in the same order in both places.

Done

> Shouldn't the side panel also provide the Border Style setting?

I don't want to put too much content in the side panel. Otherwise the side panel will just be another settings. Also the border style might be too complicated, user can still explore it in the settings and follow the help link if needed.

uiryuu:
I'm merging for now, if you find anything you want to improve regarding this feature (wording, arrangement, etc..) you can propose another PR later.

PR #6347, opened 2026-09-10
Fix crash when bundled keybinding .conf is missing

Borisserz:
- [x] Related issue: #5561
---
---

**Description:**

`PrefKeyBindingViewController.defaultConfigs` (and the new Settings key-bindings page) force-unwrapped `Bundle.main.path(forResource:…)` for each bundled `.conf`. A missing resource crashed at launch while loading key bindings.

Resolve paths via a helper that skips/logs missing files, and stop force-unwrapping `"IINA Default"` in `PlayerCore.loadKeyBindings`. Also splits a `MainWindowController` expression that fails type-checking on Xcode 26.2 (CI uses 26.5).

**Test plan:**
- [x] `cd other/LogicTests && swift test` (3 tests)
- [x] Nightly `xcodebuild` build succeeded locally
- [ ] Launch with normal bundle → key bindings load as before
- [ ] Settings → Key Bindings still lists default configs when resources are present

### DIFF
--- a/iina.xcodeproj/project.pbxproj
+++ b/iina.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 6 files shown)


Borisserz:
Updated: removed the unrelated `MainWindowController` type-check workaround and the extra `other/LogicTests` package so this PR only contains the bug fix.

Issue #6255, opened 2026-08-15
Mouse natural scrolling setting does not affect volume scroll direction


rinchee07:
### IINA Version

1.4.4

### Release Channel

Release

### macOS Version

15.7.7

### Steps to Reproduce

1.Open a video in IINA.
2.Scroll to adjust the volume, and test scrolling in the IINA Settings window.
3.Enable/disable System Settings → Mouse → Natural scrolling.
3.Scroll to adjust the volume again, and test scrolling in the IINA Settings window again.

### Expected Behavior

The scrolling direction in the IINA Settings window and the volume scrolling direction in IINA should both reverse when System Settings → Mouse → Natural scrolling is enabled or disabled.

### Actual Behavior

Enabling or disabling System Settings → Mouse → Natural scrolling does not change the volume scrolling direction in IINA.
This also cannot be corrected with a third-party scroll-reversing tool（mos） while Natural scrolling is enabled, because reversing the mouse wheel direction fixes one behavior but makes the other one reversed.

I hope an Invert scroll direction option can also be added under IINA Settings → Control → Scroll, allowing the volume scroll direction to be reversed independently.

### How often can you reproduce the issue

everytime

### Required Actions

- [x] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [ ] I tested with the Nightly build and confirmed the issue is not fixed.
- [ ] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

Borisserz:
I'd like to take this one. Planned fix: stop re-inverting vertical scroll deltas when natural scrolling is on (that cancels the system setting for volume/seek).

PR #6353, opened 2026-09-11
Respect natural scrolling for volume and seek

Borisserz:
- [x] Related issue: #6255
---
---

**Description:**

When system Mouse natural scrolling is on, `scrollingDeltaY` is already inverted. IINA then negated `deltaY` again, so volume/seek direction did not follow the system setting (unlike AppKit scroll views in Settings).

Stop re-inverting vertical deltas; keep the existing horizontal flip when natural scrolling is off.

**Test plan:**
- [ ] Toggle System Settings → Mouse → Natural scrolling
- [ ] Vertical scroll over video for volume (or seek) reverses with the system setting
- [ ] Settings window scroll still matches


low-batt:
Sounds like we are changing scrolling behavior that maybe incorrect, but users are used to it? Doing such things without having a user setting to restore the old behavior typically generates loud complaints from some users. Wait and see if the senior developers agree with me.

Issue #5870, opened 2026-01-01
Delete current file on volume without Trash results in error (e.g. Network Share)


ahrib:
### IINA Version

1.4.1 build 160

### Release Channel

Release

### macOS Version

26.2 (25C56)

### Steps to Reproduce

two ways to reproduce issue

1. Using a SMB Network share

- use MacOS to mount a SMB share volume
- place test video on volume
- attempt to delete video using IINA gui

2. Force disable Trash on a volume (e.g. on a USB drive)
- cd /Volumes/myVolumeName
- sudo rm -Rf .Trashes
- sudo touch .Trashes
- place test video on volume
- attempt to delete video using IINA gui

### Expected Behavior

if no trash is available delete the file permanently, without error

### Actual Behavior

receive error below, and file remains on file system

Error deleting: “Test.mp4” couldn’t be moved to the Bin because the volume “ABC” doesn’t have one.

### How often can you reproduce the issue

always

### Required Actions

- [x] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [ ] I tested with the Nightly build and confirmed the issue is not fixed.
- [ ] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

anohren:
#2571 #3859 #4085

Pick your favorite thread

Borisserz:
I'd like to take this one. Planned fix: when `trashItem` fails because the volume has no Trash (`NSFeatureUnsupportedError`), fall back to permanent delete (same outcome as the existing hard-delete key command).

PR #6355, opened 2026-09-11
Delete file when volume has no Trash

Borisserz:
- [x] Related issue: #5870
---
---

**Description:**

`trashItem` fails with `NSFeatureUnsupportedError` on volumes without a Trash (e.g. some network shares / forced `.Trashes`). Delete Current File and playlist “delete file” then showed an alert and left the file.

Fall back to `removeItem` only for that error code.

**Test plan:**
- [ ] Local volume with Trash → Delete still moves to Bin
- [ ] Volume without Trash (SMB / no `.Trashes`) → Delete removes the file, no alert
- [ ] Other trash failures (permissions) still show the error alert

### DIFF
--- a/iina/Base.lproj/Localizable.strings
+++ b/iina/Base.lproj/Localizable.strings
@@ -331,6 +331,7 @@
 
 "alert.permanent_delete_no_trash.title" = "Trash Unavailable";
 "alert.permanent_delete_no_trash.message" = "This volume does not have a Trash. Permanently delete the file? This cannot be undone.";
+"alert.permanent_delete_no_trash.button" = "Delete Permanently";
 
 "alert.filter_hwdec.message" = "You are about to apply a video filter. However, your current hardware decoder may not support video filters well. You can either disable hardware decoding or switch to Auto(Copy), which will consume twice the memory.\nYou may need to repeat previous operation (like setting the crop) to make it work.";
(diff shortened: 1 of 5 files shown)


anohren:
To be clear, does this, in the UI, tell the user that their file will be Moved to Trash, but then switches on them behind the scenes and performs an unadvertised destructive operation instead?

Borisserz:
Yes — that was a fair concern. The playlist action is labeled “Move to Trash”, and the previous fallback permanently deleted without saying so.

Pushed an update: if `trashItem` fails with `NSFeatureUnsupportedError`, we now ask (“Trash Unavailable / Permanently delete…?”) and only delete if the user confirms. Cancel leaves the file alone. Also remove the playlist entry only after the file is actually gone.

anohren:
That sounds like an improvement. There is still one big concern though, and that is that users "don't read": If I've asked for something to be moved to trash and I'm presented with a button labeled "OK" or "Yes" I will assume that the rest of the message box is asking me to confirm the operation I asked for, and not that it has replaced the operation with another one. It's therefore advised to let the confirmation button itself inform the user of what will happen: "Delete permanently" (or whatever phrase macOS already uses for this, if any)

Borisserz:
Good call. Switched the confirm button from “OK” to **Delete Permanently** (Cancel still cancels), so the button itself states the destructive action.

PR #6363, opened 2026-09-13
Enlarge welcome window sidebar text

low-batt:
This commit will:
- Enlarge the size of the font used for the text in the sidebar of the welcome window
- Enlarge the build type indicator
- Remove the build number from the version shown as it is available in the about window
- For Nightly and Debug builds:
  - Add displaying the build date
  - Add a clickable button showing the branch and commit which links to the GitHub commit

- [ ] Related issue: #
---
---

**Description:**


low-batt:
Especially in release builds the `Welcome Window` sidebar looks a bit sparse with the pretty arrow looking a lot larger than the text:



The intent behind this PR is to spruce up the sidebar slightly. This is what it will look like for the 1.5.0 Release build:

release

For a clean look, Beta builds will no longer include the build number in the version. The user can always check the `About Window` for that:
beta

beta-clicked

Nightly and Debug builds now show the build date and a clickable button giving the branch and commit, similar to the `About Window`:
nightly

debug

uiryuu:
Not a graphics designer, just my 2¢ based on my first impressions:
- I feel like after the changes, "IINA" appears too close to the other stuff below it. I might want to put the version number a little bit lower.
- Now that the "IINA" is of larger font size, maybe consider making it not bold?
- The "Debug" and "Nightly" labels have different top and bottom spacings.
- What does it look like when the branch name is longer than the sidebar? Even if it doesn't clip, I still think the date and branch/commit info texts are too packed. Ok if we don't deal with that in this PR. Might be a good time to redesign this when converting this window to pure code.
- The beta popover doesn't seem to have border/shadow? Don't know if this is introduced in this PR or not

uiryuu:
image

In the about window, we have a similar display of the logo and version. Note that the app name is of a larger font size but not bold. Maybe consider unify them

low-batt:
On this:
> Not a graphics designer

Oh NO! I'm not either!

> I feel like after the changes, "IINA" appears too close to the other stuff below it. I might want to put the version number a little bit lower.

Ha! Not what I would have guessed. In earlier drafts I was keeping the version number close to IINA and had a noticeable gap between the version and the build type indicator. 

Focus on the screenshot of the release build which only has IINA and the version number. That is the important one. I kind of like the spacing on that one. Keeps the version attached to IINA. Are you sure you want me to move the version number down? If so, how much?

On this:
> Now that the "IINA" is of larger font size, maybe consider making it not bold?

And this from the second comment:
> In the about window, we have a similar display of the logo and version. Note that the app name is of a larger font size but not bold. Maybe consider unify them

I agree. Lets change the About Window. :smiley:

To me the Welcome Window  matches nicely with what you see in the IINA DMG file:

DMG

Look again at the About Window. Notice the smaller IINA on the top of the right side looks bold and in the same style. It is the IINA on the left side that differs from all the others.

I used the same font that is used in the About Window for the IINA on the left side of the panel,  `System Bold 30`. I was confused as to why the IINA in the About Window looked wimpy. I was suspecting maybe a constraint problem or some sort of defect. It looks the same as the IINA on the Welcome Window when viewed in the XIB. I just looked into why that is and traced it to this code that is overriding the XIB:
```swift
  @IBOutlet weak var iinaLabel: NSTextField! {
    didSet {
      iinaLabel.font = NSFont.systemFont(ofSize: 24, weight: .light)
    }
  }
```

I checked that code and it is from a long time ago, macOS 10.10-10.11 timeframe. If I remove that code the About Window looks like:

about-matching-font

I rather like that. This would unify the style across the About, Welcome and DMG.

Thoughts?

On this:
> The "Debug" and "Nightly" labels have different top and bottom spacings.

That was intentional. I viewed IINA and the version number as a unit. And the built type label and other info as a separate unit. So I was trying to leave a larger gap between the version and the build type. Look at the Beta screenshot and you can see the gap between the version and Beta label is larger than the gap between the IINA and version. In a draft version the gap between the IINA / version and build type + other info was greater. I didn't like the look of that and tightened it up a bit to what you see now.

On this:
> What does it look like when the branch name is longer than the sidebar? Even if it doesn't clip, I still think the date and branch/commit info texts are too packed. Ok if we don't deal with that in this PR. Might be a good time to redesign this when converting this window to pure code.

Bad, really bad. Both windows, as I brought this over from the About Window. As we have other eggs to fry I was trying to make this just a quick change and not open a can of worms as we will have a chance to do that when these windows are moved to code.

Initially I looked at converting the Welcome Window, but it is complex enough that I did not want to take that on at this time while we are trying to get 1.5.0 out.

Do you want me to look into fixing this in both windows?

On this:
> The beta popover doesn't seem to have border/shadow? Don't know if this is introduced in this PR or not

I did not make any changes to the popover. In running the IINA 1.5.0-beta1 release I see what looks to be a very thin shadow when I look at it live. But in this screenshot I can't see the shadow:

current-with-beta-showing

I was expecting that we would be adjusting these windows again when we switch away from XIBs.

Given the above feedback, what changes to you want me to make? As always, I defer to you on what we should do. Can also check in with @lhc70000 and he what he thinks.

low-batt:
With the latest commit the branch / commit button text will wrap if the branch name is long in both the `About Window` and the `Welcome Window`. The font style of the IINA label in the `About Window` has been changed to be bold like it is in the `Welcome Window` and the DMG background:
about-wrapped
welcome-wrapped

lhc70000:
I think it's better to put this into 1.5.1, together with a redesign and XIB removal.

PR #6372, opened 2026-09-15
Fix various permission issues in plugin API

lhc70000:
- [x] Related issue: #6230, also PR #6245
---
---

**Description:**

We currently have `network-request` and `file-system` permissions. While the plugin can open a file in various ways (e.g. `core.open`, by adding an external track, changing the playlist), the two later paths are not guarded by the permissions. Moreover, `core.open` also doesn't consider the case when the URL is not a local file (issue #6230).

The proposed logic is:

- Without permission: allow read/write/open in the plugin sandbox directory and `data://` URLs. Although opening a file is generally not dangerous, it still makes IINA read the file and scan its folder for subtitles, and the plugin can query the properties and read the opened file directly.
- Need `network-request` permission: simple HTTP/HTTPS/SMB URLs.
- Need `file-system` permission: File URLs outside plugin's sandbox, or other protocols (`edl`/`concat`, etc), since they can encode arbitrary local and remote files.

Several other related problems are also fixed:

- Currently the mpv APIs are not guarded by permissions, allowing arbitrary code execution such as `subprocess`. Added `safeMPVCommands` and moved `safeMPVOptions` to `AppData` so that any mpv API call not in the list requires the `file-system` permission.
- `loadExternalVideoFile()` and corresponding audio/subtitles functions: they assume the url is local and use `url.path` directly. If the url is not a file (enabled by the plugin API), there can be security issues (e.g. `https://example.com/local/path`)
- `MPVNode.free()` doesn't free the lists themselves, and a typo prevents freeing map elements.


uiryuu (approved):
LGTM

low-batt:
Still reviewing this PR, but here are a couple of initial comments…

Should not generate a warning if a plugin does not require any permissions:

jump-to-frame


To reproduce install the `Jump to Frame` community plugin.

I find giving a plugin write access to the file system much more disturbing that just giving it read access, so I'm surprised there aren't separate permissions instead one where I must grant both read and write access, along with _execute_ access!:

read-write

low-batt:
Had to restart IINA after installing `Jump to Frame` before it started to work. Seems it is expected that an IINA restart is required for some extensions to work. User has to click on `i` to know that.

All of these failed to install because IINA is trying to access documentation in the latest release and none of these publish a release:
- [Hold to Speed](https://github.com/Tommy12356F/iina-hold-to-speed)
- [Multiple Clips](https://github.com/karthisnk/multi-cutter-iina)
- [Pip Toggle for IINA](https://github.com/nastarandarjani/iina-pip-toggle)
- [recorder](https://github.com/5thDimensionalVader/recorder-iina)
- [Skip-Intro](https://github.com/pparanoiidd/iina-skip-intro)

[ListenBrainz Scrobbler](https://git.notfire.cc/notfire/iina-listenbrainz) is not hosted on GitHub, so I didn't test that one at all.

Many plugins require file system access. I did not try installing those.

I'll look closer at the code later today. Gotta run.

low-batt:
I was able to install [Hold to Speed](https://github.com/Tommy12356F/iina-hold-to-speed) using IINA 1.4.4, so the installation failure is a regression. This is what it looks like:
hold-to-speed-install-fail

low-batt (approved):
The problems I hit while testing this PR do not seem to be tied to these changes. The installation problem definitely needs to be fixed as it is a regression.

The changes in this PR look good to me.

lhc70000:
> there aren't separate permissions instead one where I must grant both read and write access, along with execute access

The permission system exists mainly because we need to let the user (a lot of them are non-technical) know the risk, rather than creating a sophisticated security system. We cannot monitor everything a plugin does. Setting watch_later to `.bashrc` can enable command injection, opening ffmpeg URL schemes can involve arbitrary actions, and reading `.ssh` can be equally dangerous. So r/w/x are not clearly separated. Therefore, in my opinion, the permission levels should only draw the boundary between a safe operation and a potentially dangerous operation. Without `file-system` permission, plugins can only call a small set of APIs that definitely don't have access to sensitive data, while all other operations can be abused and we cannot verify each of them.

Issue #6366, opened 2026-09-13
Window resize performance in 1.50 beta and nightly is very choppy


LeoNatan:
### IINA Version

1.50 beta

### Release Channel

Beta

### macOS Version

27.0

### Steps to Reproduce

Live resize of windows is very sluggish in 1.50 beta and nightly builds, regardless of the media open (tried 720p, 1080p, 4K, all hardware decoded). 1.4.4, while not completely hitch free, is much, much smoother.

https://github.com/user-attachments/assets/b5f40a0e-773a-4925-86b9-fc502e01de35

 
- Open a video
- Live resize the window

### Expected Behavior

Should be a smooth resize experience

### Actual Behavior

Very sluggish

### How often can you reproduce the issue

Consistent

### Required Actions

- [x] I tried my best to search from the existing issues (including closed issues) and didn't find duplicated issues.

### Optional Actions

- [x] I tested with the Nightly build and confirmed the issue is not fixed.
- [ ] I tested with the latest version of mpv and confirmed mpv doesn't have this issue.

LeoNatan:
If you need additional profiling, let me know.

LeoNatan:
Entering and exiting full screen is also affected, likely by the same issue.

LeoNatan:
I don't see anything suspicious in Activity Monitor, no CPU or GPU spikes. Both 1.4.4 and latest nightly rise to ~50% CPU usage when resizing the window, as does WindowServer.

uiryuu:
What mac are you using? Can you try to disable liquid glass in the settings and try again?

LeoNatan:
None of the Liquid Glass options are enabled. It’s an M3 Max.

josephoh:
Try going into System Settings --> Accessibility --> Display and enable "Reduce Transparency". Then go into System Settings --> Appearance and make sure the Liquid Glass slider is all the way to the right. This fixed it for me.

LeoNatan:
That’s not a fix.

low-batt:
For an issue like this lots of behind the scenes details matter. Would be good to collect logs. See the FAQ:  [How to collect log files when reporting an issue?](https://github.com/iina/iina/wiki/FAQ#how-to-collect-log-files-when-reporting-an-issue)

Usually hardware decoding is an advantage. But not always. For example, VP9 hardware decoding performs poorly. I do not suspect that as a cause since the title of the video suggests the HEVC codec was used. I mention this as an example of one of the details that matters. 

[...]

PR #6375, opened 2026-09-16
Keep camera housing black in legacy fullscreen

lhc70000:
- [x] Related issue: #6201
---
---

**Description:**

It seems that creating a new window (like the `blackWindows`) is the best solution.

Also removed the code setting the content view's frame since it's unnecessary now (and I think setting content view's frame is undefined behavior).


iina/MainWindowController.swift
```diff
@@ -67,6 +67,9 @@ class MainWindowController: PlayerWindowController {
   var blackWindows: [NSWindow] = []
   var cachedScreens: [NSScreen] = []
 
+  /** For hiding camera housing only in legacy full screen. */
```
low-batt: I've been hoping IINA would standardize on `///` and avoid the `/**`. Currently IINA has a mix of the two comment forms, but mostly the slash is used. I was going to ask you if you wanted me to cleanup the sources so only the slash form of comment is used throughout.

iina/MainWindowController.swift
```diff
@@ -1825,21 +1830,39 @@ class MainWindowController: PlayerWindowController {
...
+    let frame = NSRect(
+      x: screen.frame.minX,
+      y: screen.frame.minY,
+      width: screen.frame.width,
+      height: screen.frame.height - unusable,
+    )
+
     let useAnimation = {
       // Animation causes lagging under the macOS Tahoe beta, so don't allow it for now.
       guard #unavailable(macOS 26) else { return false }
```
low-batt, line 1846: Need to look into this. Maybe related to issue #6366?

iina/MainWindowController.swift
```diff
@@ -1825,21 +1830,39 @@ class MainWindowController: PlayerWindowController {
...
     window.setFrame(frame, display: true, animate: useAnimation)
-    guard let unusable = screen.cameraHousingHeight, let cv = window.contentView else { return }
-    // This screen contains an embedded camera. Shorten the height of the window's content view's
-    // frame and the video view container's frame to avoid having part of the window obscured by
-    // the camera housing.
-    let size = NSMakeSize(cv.frame.width, frame.height - unusable)
-    cv.setFrameSize(size)
-    videoViewContainer.setFrameSize(size)
+
+    // Force black background in camera housing
```
low-batt: Might want to explain the situation a bit more as a developer that uses the `Dark` `Appearance` default might not be thinking about the window having a white background when the `Light` `Appearance` is used.

iina/MainWindowController.swift
```diff
@@ -1825,21 +1830,39 @@ class MainWindowController: PlayerWindowController {
...
+    let housingWindow = NSWindow(contentRect: .zero, styleMask: [], backing: .buffered, defer: false, screen: screen)
```
low-batt: We could have continued to use the old method when `Appearance` is set to `Dark`. Probably best to keep it simple, but need to make sure a future developer understands this is about propertly handling the `Light` `Appearance` so they don't start thinking this code could be simplified to merely shrink the content view.

iina/MainWindowController.swift
```diff
@@ -1825,21 +1830,39 @@ class MainWindowController: PlayerWindowController {
   ///
   /// For screens that contain a camera housing views will be adjusted to not use that area of the screen.
   private func setWindowFrameForLegacyFullScreen() {
-    guard let window, let screen = window.screen ?? NSScreen.main else { return }
-    let frame = screen.frame
+    guard let window,
+          let screen = window.screen ?? NSScreen.main,
+          let unusable = screen.cameraHousingHeight else { return }
```
low-batt: NO GOOD! Including this in the initial `guard` means the frame is _not set_ for screens without a camera housing.

low-batt (changes requested):
A quick review as I'm busy today. This breaks legacy FS on screens without a camera housing.

There are other issues with using white in the main window background. I think they can be solved. I'll say more later. Gotta run…

low-batt (approved):
Pulled PR, tested, all working with latest commit.

